PR #5, opened 2022-09-07
Initial proposal for constructors and destructors

llvm-beanz:
This proposal adds user-defined constructors and destructors to HLSL and defines how they should interact with semantics and global initialization.


llvm-beanz:
Why does GitHub keep automatically closing this?

PR #6, opened 2022-09-07
Initial draft of uninitialized resource errors

llvm-beanz:
This proposal seeks to change how HLSL defers some data use analysis to post optimization. The proposal instead prefers early and consistent failures.

### DIFF
--- a/proposals/0008-uninitialized-resources.md
+++ b/proposals/0008-uninitialized-resources.md
@@ -58,12 +58,6 @@ code changes to impact code correctness.
 
 ## Proposed solution
 
-Describe your solution to the problem. Provide examples and describe how they
-work. Show how your solution is better than current workarounds: is it cleaner,
-safer, or more efficient?
-
-## Detailed design
-
 Building off clang's existing support for uninitialized variable analysis, we
 can move this analysis into the Clang control flow graph (CFG) instead of post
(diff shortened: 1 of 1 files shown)


proposals/0008-uninitialized-resources.md
```diff
@@ -0,0 +1,116 @@
...
+For example:
+
+```c++
+RWBuffer<int> In;
+RWBuffer<int> Out[2];
+void fn(bool Cond) {
+  RWBuffer<int> O = Out[0]; // warning: local resource not guaranteed to resolve to a unique global resource.
+  if (Cond) // note: if statement here is not compile-time resolved
+    O = Out[1];
+  O[0] = In[0]; // note: unresolved use occurs here
```
tex3d, line 109: Technically, since O is always assigned to a single resource range declaration, this is (can be made) legal.  But there is a problem if `Cond` is non-uniform, since NonUniformResourceIndex would need to be used at the indexing (create handle) site, and that indexing would be created by the compiler when replacing the conditional assignment of a resource with a conditional assignment of an index.

If instead of `Out[2]` you had separate `Out1` and `Out2` variables, then used those instead of `Out[0]` and `Out[1]`, then it would require `Cond` to be compile-time resolved.

PR #7, opened 2022-09-07
Initial proposal for math modes

llvm-beanz:
The new HLSL math modes aims to provide an alternative to the `precise` attribute which will both be less bug prone and provide a higher degree of control for developers.

### DIFF
--- a/proposals/0009-math-modes.md
+++ b/proposals/0009-math-modes.md
@@ -1,0 +1,85 @@
+# Math Modes
+
+* Proposal: [0009](0009-math-modes.md)
+* Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
+* Sponsor: [Chris Bieneman](https://github.com/llvm-beanz)
+* Status: **Under Consideration**
+* Planned Version: 202x
+
+## Introduction
+
+Shaders frequently need fine-grained control over the way the compiler optimizes
+(or doesn't) math optimizations. To enhance developer control in a more robust
+way than the existing `precice` qualifier, HLSL should gain a higher level
+concept of math modes.
+
+## Motivation
+
+The `precise` qualifier is particularly gnarly to implement and is wrought with
+bugs that are inscrutable to developers. The source of the implementation
+challenges come from the fact that the precise attribute is applied to a
+declaration, but does not impact the uses of the declaration. Instead it impacts
+reverse-propagates and applies to all of the math expressions that are used to
+compute the declared variable.
+
+The goal of this proposal is to create a new language feature which allows
+shader authors the control they need to replace precise, without the tooling
+complexity.
+
+## Proposed solution
+
+First and foremost, since the goal is to _replace_ the `precise` qualifier, this
+proposal should deprecate `precise` in the language. Allowing the new feature
+and `precise` to exist together for a single language version will ease
+adoption, however `precise` will be removed in the future.
+
+### Math Modes
+
+The core of this proposal is to introduce new math modes `strict` and `fast`.
+The `strict` math mode prohibits optimizations which may impact the precision of
+results. This prevents optimizations like fusing multiply and add instructions.
+The `fast` math mode prioritizes speed over precision guarantees. This does not
+mean that `fast` math is less precise. In some cases optimal fused operations
+are faster, but the results are less consistent across hardware architectures.
+
+As a simplified view, `strict` can be viewed as "execute the math as written in
+source", while `fast` is "let the optimizer do its thing".
+
+The new math modes are exposed through a variety of mechanisms described below.
+
+### Attributes
+
+The new `hlsl::math_mode()` attribute can be set to either `strict`, `fast`, or
+`default`. The default math mode is `fast`, however this can be changed with the
+compiler flag:`math-mode=precise|fast`.
+
+The math attribute can be applied to functions or expressions, but it does not
+propagate in the way that the `precise` keyword did.
+
+For example:
+
+```c++
+  [hlsl::math_mode(strict)]
+  float fma(float x, float y, float z) {
+    return x * y + z;
+  }
+
+  float fma2(float x, float y, float z) {
+    return [hlsl::math_mode(strict)] x * y + z; 
+  }
+```
+
+In both of the above functions optimization to an FMA instruction is prevented.
+
+### Namespaces
+
+New math namespaces are also added which provide simplified access to math
+operations with the expected performance characteristics.
+
+For example:
+
+```c++
+  float f = hlsl::strict::fma(1.0, 2.0, 3.0); // strict mode math
+  float f = hlsl::fast::fma(1.0, 2.0, 3.0);   // fast mode math
+  float f = hlsl::fma(1.0, 2.0, 3.0);         // default mode math
+```
--- a/proposals/0009-math-modes.md
+++ b/proposals/0009-math-modes.md
@@ -5,6 +5,7 @@
 * Sponsor: [Chris Bieneman](https://github.com/llvm-beanz)
 * Status: **Under Consideration**
 * Planned Version: 202x
+* Dependencies: [0002 C++ Attributes](0002-cxx-attributes.md)
 
 ## Introduction
 


tex3d (approved):
I like the dependencies line.

proposals/0009-math-modes.md
```diff
@@ -0,0 +1,86 @@
...
+## Introduction
+
+Shaders frequently need fine-grained control over the way the compiler optimizes
+(or doesn't) math optimizations. To enhance developer control in a more robust
+way than the existing `precice` qualifier, HLSL should gain a higher level
+concept of math modes.
+
+## Motivation
+
+The `precise` qualifier is particularly gnarly to implement and is wrought with
```
jeremyong, line 19: It'd be useful to have a little more insight into the motivation here. I can understand that _using_ a variable declared `precise` which doesn't carry the qualifier forward may result in some less-than-desirable behavior. However, from what I've seen, the overwhelmingly common use case for `precise` is for emitting data either to a render target (in particular, depth), or in an ISBE (i.e. vertex shader to pixel shader interpolant). In those scenarios, `precise` provides a nice guarantee that the results are bit exact across shader invocations, render passes, etc.

llvm-beanz: As shaders are growing in complexity we're moving in a direction where single-translation-unit shaders are unwieldy for both developers and compilers. In HLSL, we support linking DXIL through library shaders already, but there are a bunch of things that can get lost there when you look at `precise`.

For example if I compile a function `float fn(float)` into a library, then call it storing the value to a `precise float`, what is the expected behavior? Does the expected behavior change if I put the `[noinline]` attribute on `fn`?

If the code were all in one translation unit, the expectation is that after inlining, we propagate `precise` across the inlined portion of `fn`. That behavior puts some pretty hard restrictions on what a multi-source HLSL program would need to look like, and it more-or-less forces a link-reoptimize-fixup flow. It also gets _really_ mucky when you look at `[noinline]` (although admittedly a lot of things get mucky or just don't work with `[noinline]`).

The other factor here is the implementation of `precise` in DXC. We've found it to be fraught with errors. Because of the way it propagates across call boundaries DXC's implementation propagates `precise` on the IR instead of the AST. This causes other problems and is the source of a whole lot of bugs where the `precise` annotation gets dropped unexpectedly. To work around some bugs we actually disable some optimizations on `precise` values that would actually be beneficial to performance but it is not trivial to get them to work with `precise`.

All in, our goal here is to design a more powerful solution that is more clear, has more features, and is more straightforward from both the compiler and language perspective. The proposal to deprecate `precise` is mostly to get people to start thinking about updating their code so that we can have a path to eventually removing it.

Aishization: I can certainly appreciate the difficulty in propagating precise when dealing with multiple translation units. However I would be very weary about deprecating it.  The propagation is a vital feature when dealing with certain classes of rendering issue, such as ensuring that vertex positions match up between render passes. 

I have worked on platforms without an equivalent and it means either the developer has to manually propagate forcing the accuracy or having to disable optimizations entirely for the whole vertex shader. Either of these are rather undesirable. In my opinion, the propagation of precision needs to be done somewhere and I would rather it be done by the compiler stack than having to do it by hand as a developer.

jeremyong: I also sort of think `precise` is a sort of "killer feature" and fills a need I don't think `math_mode(strict)` will quite occupy. If a relaxed function is compiled in a TU and is later linked to emit data that needs the `precise` annotation, could we simply throw a linker error at this point?

Wouterdek: Given that this proposal is currently in the 'under consideration' state, I'd like to add that this feature would be very helpful for implementing [software extended-precision math](https://andrewthall.org/papers/df64_qf128.pdf) in HLSL. This is a case where you want specific operations to be executed in a strict fashion, _without affecting the math mode of inputs or subsequent operations_. Some other shader languages already offer this, I've seen it referred to as `invariant` mode.

Currently the Unreal Engine uses `precise` for [this feature](https://dev.epicgames.com/documentation/en-us/unreal-engine/large-world-coordinates-rendering-in-unreal-engine-5) due to a lack of a better alternative, but uses platform specific intrinsics where available to avoid the propagation behavior of `precise`. We would strongly prefer using `[hlsl::math_mode(strict)]` in that case if it was available to us.

llvm-beanz: @Wouterdek, thank you for the positive feedback. It is super useful for us to understand the use cases and potential benefit of features as we plan and allocate resources.

I'm unsure if this proposal will make HLSL 202x or 202y, my guess is the later due to the narrower scope of 202x (see: [Proposal 0020 - HLSL 202x and 202y](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0020-hlsl-202x-202y.md)).

I recently had a lengthy conversation recently about `invariant` and how it related to this proposal and the existing `precise` keyword. SPIR-V and MSL both apply the `invariant` property to the output of the shader which (in theory) allows an implementation to limit the scope of restricted optimizations. In practice, I think it hits a lot of the same issues that `precise` does because of the upward propagation, but maybe even more so. One example I pointed out was something like:

```hlsl
if (someMathFn(someFloat3Input, someOtherFloat3Input) > 0)
  Output = ...;
else
  Output = ...;
```

As `invariant` is specified the `someMathFn` in the condition needs to be "invariant" even though its result doesn't directly contribute to the output, whereas as `precise` is specified the `someMathFn` is unrestricted unless its result is stored to a `precise` variable declaration.

Wouterdek: @llvm-beanz Glad to hear this is still alive and in planning!

To be clear, I do prefer the `math_mode` attribute as it is proposed here. In fact, my reference to `invariant` may have caused some confusion, as I was referring to the implementation used in the proprietary shader compiler of a popular OEM. There, it does apply to expressions rather than variables, similar to this proposal. E.g.:

```
float a = someFn(...); // not strict
float b = invariant( someFn(a) ); // only the math operations in someFN(a) are performed strictly.
float c = someFn(b); // not strict
```

On the topic of `precise`, I do recognize the value of a variable-scoped property. The Unreal Engine does also use that for depth variables in certain cases. That said, given the choice between `precise` and better compilation scalability through well-defined DXIL linking, I feel the latter is/will become a greater necessity.

Nielsbishere: +1 on adding this to HLSL. Would like to add that there are more details of some of the use cases in Wouter's explanation at GPC (Solving numerical precision challenges for large worlds in UE5). This would be very useful for us too. The cascading effect of the precise keyword seem too unpredictable to be useful unless the developer is very aware of it.

Nielsbishere: I wanted to leave something here regarding a solution for precision as it's relevant to the GPC talk. It seems that fixed point can be a viable alternative if doubles are unavailable but perform very similarly (on NV, on amd/intel it probably performs worse) due to packing/unpacking. Though it seems to outperform doubles if it can be stored as fixed point and isn't packed tightly (on NV). Though for bandwidth fixed point is a lot better since you can pack a 4 fract, 37 integer, 1 sign x3 in 128-bit (+2 padding bits). It seems like it's suboptimal to the approach from the video, probably because integer math is less optimal on GPUs and there's a lot more operations involved. But here's some code if someone is interested (it would allow up to +-1.4M km precision 1/16th cm):
```hlsl
	uint64_t3 fixedPointUnpack(U32x4 packed) {
		U32x3 hi = (packed.www >> U32x3(0, 10, 20)) & ((1 << 10) - 1);
		U64x3 hiShift = (U64x3) hi << 32;
		return hiShift | packed.xyz;
	}

	U32x4 fixedPointPack(U64x3 unpacked) {
	
		U32x3 lo = (U32x3) unpacked;

		unpacked >>= 32;
		U32x3 hi = ((U32x3) unpacked) << U32x3(0, 10, 20);

		return U32x4(lo, dot(hi, 1.xxx));
	}

	U64x3 fixedPointSub(U64x3 a, U64x3 b) { return a - b; }
	U64x3 fixedPointAdd(U64x3 a, U64x3 b) { return a + b; }

	static const U32 fixedPointFraction = 4;		//1/16th cm
	static const U32 fixedPointShift = 41;			//42 bits (sign is excluded here)

	static const F32 fixedPointMultiplier = 1.f / (1 << fixedPointFraction);
	static const U64 fixedPointMask = ((U64)1 << (fixedPointShift + 1)) - 1;

	F32x3 fixedPointToFloat(U64x3 value) {

		Boolx3 sign = (Boolx3)(value >> fixedPointShift);
		U64x3 correctedForSign = value ^ fixedPointMask.xxx;
		++correctedForSign;

		value = select(sign, correctedForSign, value);
		return (F32x3)value * select(sign, -fixedPointMultiplier, fixedPointMultiplier);
	}

	U64x3 fixedPointFromFloat(F32x3 v) {

		Boolx3 sign = v < 0;
		v = abs(v);

		v *= (1 << fixedPointFraction).xxx;

		U64x3 res = (U64x3) v;
		U64x3 correctedForSign = (res - 1) ^ fixedPointMask.xxx;
		return select(sign, correctedForSign, res);
	}
```

PR #10, opened 2022-09-08
Initial HLSL Specifiction LaTeX stub

llvm-beanz:
This stubs out some LaTeX goop and a first stab at the Introduction for an HLSL Specification.

I'm sure LaTeX isn't everyone's preferred format, but it is git-friendly and supports all the formatting bells and whistles we might need.

### DIFF
--- a/specs/language/introduction.tex
+++ b/specs/language/introduction.tex
@@ -1,4 +1,4 @@
-\chapter{Introduction} \label{ch:Intro}
+\Ch{Introduction}{Intro}
 
 \p The \acrfull{hlsl} is the GPU programming language provided in conjunction
 with the \gls{dx} runtime. Over many years its use has expanded to cover every
@@ -26,7 +26,7 @@ where there is an intent to alter implementation behavior in the future. Since
 this document and the implementations are living sources, one or the other may
 be ahead in different regards at any point in time.
 
-\section{Scope} \label{sec:Scope}
+\Sec{Scope}{Intro.Scope}
 
 \p This document specifies the requirements for implementations of
 \acrshort{hlsl}. The \acrshort{hlsl} specification is based on and highly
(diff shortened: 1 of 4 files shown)


specs/language/introduction.tex
```diff
@@ -0,0 +1,165 @@
...
+\Sec{Terms and definitions}{Intro.Terms}
+
+\p This document aims to use terms consistent with their definitions in
+\gls{isoC} and \gls{isoCPP}. In cases where the definitions are unclear, or
+where this document diverges this section, the remaining sections in this
+chapter, and the attached \ref{glossaries}.
+
+\Sec{Runtime Targeting}{Intro.Runtime}
+
+\p \acrshort{hlsl} evolved out of \gls{dx} and gained popularity because it
```
pow2clk, line 59: The phrasing here makes it sound like HLSL replaced DirectX. I think something with "emerged" might communicate the intended message. "HLSL emerged from the evolution of DX to grant greater and greater control over the specifics of the GPU geometry and color processing. . . ". A bit wordy.

specs/language/introduction.tex
```diff
@@ -0,0 +1,165 @@
...
+\p Behavior of invalid program constructs or incorrect data which this standard
+imposes no requirements, or does not sufficiently detail.
+
+\Sub{Unspecified Behavior}{Intro.Defs.Unspecified}
+\p Behavior of a well formed program and correct data which may vary by the
+implementation, and the implementation is not expected to document the behavior.
+
+\Sub{Well-formed Program}{Intro.Defs.WellFormed}
+\p An HLSL program constructed according to the syntax rules, diagnosable
+semantic rules, and the One Definition Rule.
```
pow2clk, line 165: Just a note for future reference that we may want to come back to these and purge any we don't reference if there's any chance that might be the case.

pow2clk (approved):
I have a few pending comments, but they needn't hold this up regardless of how they are resolved.

specs/language/introduction.tex
llvm-beanz: Change incoming.

llvm-beanz:
I'm going to merge this and create a new PR to address the feedback. It seems like making the repo public stranded my fork so this branch can't be updated anymore 😭

PR #19, opened 2022-10-25
Start specifying lexical conventions

llvm-beanz:
This breaks down the unit of translation and the phases of translation. This breakdown follows C++ very closely.


llvm-beanz:
This change adds the "Lexical Conventions" chapter.

See: [hlsl.pdf](https://github.com/microsoft/hlsl-specs/files/11041464/hlsl.pdf)

specs/language/lex.tex
```diff
@@ -0,0 +1,48 @@
...
+minor alterations, specifically the removal of support for trigraph and digraph
+sequences. Below is a description of the phases.
+
+\begin{enumerate}
+  \item Source files are characters are mapped to the basic source character set
+  in an implementation-defined manner.
+  \item Any sequence of backslash (\texttt{\textbackslash}) immediately followed
+  by a new line is deleted, resulting in splicing lines together.
+  \item Tokenization occurs and comments are isolated. If a source file ends in
+  a partial comment or preprocessor token the program is ill-formed and a
```
pow2clk, line 28: I'm not sure I know what ending in a preprocessor token would mean. Can you give an example? Not necessarily for the text, but for my own understanding

specs/language/lex.tex
```diff
@@ -0,0 +1,48 @@
...
+  diagnostic shall be issued. Each comment block shall be treated as a single
+  white-space character.
+  \item Preprocessing directives are executed, macros are expanded,
```
pow2clk, line 31: We have a flag `flegacy-macro-expansion` that affects macro expansion that might bear mentioning?

pow2clk (commented):
Looks sensible to me. I just have a question and a possible suggestion

specs/language/lex.tex
llvm-beanz: There are a couple common cases of partial preprocessing tokens. One would be a `#if` without `#endif`.

A partial comment would be a file that had a starting `/*` but no ending `*/`.

specs/language/lex.tex
llvm-beanz: Given that is a legacy behavior, I don't think it describes the current HLSL language. My goal for this spec is to describe HLSL 2021 as we want the language to be rather than legacy support we might enable under flags or other compatibility modes.

specs/language/lex.tex
```diff
@@ -0,0 +1,48 @@
...
+  \texttt{pragma} and other unary operator expressions are executed. Processing
+  of \texttt{\#include} directives results in all preceding steps being executed
+  on the resolved file. Finally all preprocessing directives are removed from
```
bob80905, line 34: Is it worth mentioning that the process of processing include directives is recursive? I should also note that there isn't one final step to remove all preprocessing directives from the source, rather, (at least for includes), the included source gets preprocessed and injected into the parent source file, and the include directive gets removed in a step by step manner. Is this worth mentioning?

tex3d (approved):

specs/language/lex.tex
llvm-beanz: The compiler doesn't really inject the preprocessed source into the parent source file during normal compilation. Compilers generally support that for emitting preprocessed source, but during actual compilation the expansion occurs usually on the token iterators not by merging memory buffers to create a preprocessed file in memory.

The specification here is intended to be vague on this point so that the implementation is free to optimize as it sees fit.

bob80905 (approved):

PR #26, opened 2022-12-14
Proposal for non-member operator overloading

llvm-beanz:
This proposal enables users to work around one of the challenge points they've reported adopting HLSL 2021, while allowing the language to remain moving forward.

### DIFF
--- a/proposals/NNNN-non-member-operator-overloading.md
+++ b/proposals/NNNN-non-member-operator-overloading.md
@@ -1,0 +1,66 @@
+# Non-member Operator Overloading
+
+## Instructions
+
+* Proposal: [NNNN](NNNN-non-member-operator-overloading.md)
+* Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
+* Sponsor: [Chris Bieneman](https://github.com/llvm-beanz)
+* Status: **Under Consideration**
+* Planned Version: 202x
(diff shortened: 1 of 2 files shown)


proposals/NNNN-non-member-operator-overloading.md
```diff
@@ -0,0 +1,66 @@
...
+and supporting defining global and namespace scoped operator overloads. The
+limitations currently in place on operator overloading (disallowing overloading
+operators that idiomatically return references) will remain in effect on
+non-member operators as well unless the restrictions are lifted as proposed in
+[0006 Reference Types](0006-reference types.md).
+
+Because this solution should not break existing code, it could also be enabled
+under HLSL 2021 as an extension.
+
+## Alternatives considered
```
pow2clk, line 45: These are alternatives to the motivation at hand, but there are other benefits to non-member operators that these don’t represent an alternative to. I realize I’m being nitpicky, but I’m trying to imagine coming at this without the background we have

pow2clk (approved):
Is like to move this forward so my feedback can be addressed later as we refine the document. This is definitely something we want to do

proposals/NNNN-non-member-operator-overloading.md
llvm-beanz: That's a fair observation and I should note it. Since much of the motivation for prioritizing this feature is addressing an immediate developer need I felt having this context is important.

PR #27, opened 2023-01-02
Add GitHub pages build

llvm-beanz:
This change adds support for builing a GitHub pages static site reproducing the content of the GitHub markdown as well as built versions of the LaTeX specification. This change is a starting point, there is still a lot of work to improve the visual look and functionality of the site as well as the LaTeX to HTML generation.

A preview of the site is available off my fork of the repository here: https://llvm-beanz.github.io/hlsl-specs

Note: Some links that are hard-coded in the markdown files will not work until the page is generated and deployed to GitHub Pages.


proposals/0005-strict-initializer-lists.md
```diff
@@ -33,11 +33,13 @@ In HLSL the following code is valid:
     A a[2];
     int c;
   };
+  {% raw %}
   B b = {{1, 1.2}, {2, 2.2}, 3};   // Array elements specified as members
   B b2 = {1, 2, 3, 4, 5};          // each field initialized separately
   B b3 = {{1, {2, 3}}, {4, 5}};    // Completely random grouping of arguments
   int4 i4 = {1,2,3,4};             // valid int4 in C-syntax
   B b4 = {i4, 5};                  // int4 implicitly expanded to 4 arguments
+  {% endraw %}
```
tex3d, line 42: Are these because of the double braces `{{`?  We could also insert a space between them, but it's good to know you can escape a section like this too!

tex3d: Looks like we have another problem in [0002-cxx-attributes](https://llvm-beanz.github.io/hlsl-specs/proposals/0002-cxx-attributes.html) as well.  That's troubling...

llvm-beanz: Oof... yea, I'll fix 0002.

The horrible thing is that Jekyll pre-processes all the source files with Liquid and Liquid's syntax is silly conflictive with other actual programming languages.

tex3d: Is there any other way to solve this?  How do we capture the rules and make sure documents don't contain misinterpreted patterns?

tex3d (approved):
Looks good, but I wonder if there is a way to avoid random conflicts with liquid.

llvm-beanz:
> Looks good, but I wonder if there is a way to avoid random conflicts with liquid.

I wrapped the template in `raw`/`endraw` markers hoping that it will just make its way through everywhere. There is a way to disable liquid markup in a newer version of Jekyll than GitHub supports at the moment, so there might be something better in the future.

PR #37, opened 2023-04-04
Add vk::BufferPointer to HLSL

greg-lunarg:
Create a new first class type vk::BufferPointer which allows users to efficiently reference buffer device addresses and allows tools to analyze and report on usage in a logical rather than physical way i.e. names rather than numeric offsets.

### DIFF
--- a/proposals/0010-vk-buffer-ref.md
+++ b/proposals/0010-vk-buffer-ref.md
@@ -1,0 +1,280 @@
+<!-- {% raw %} -->
+# Buffer Pointers in HLSL With vk::BufferPointer
+
+* Author(s): [Greg Fischer](https://github.com/greg-lunarg)
+* Sponsor(s): [Chris Bieneman](https://github.com/llvm-beanz), [Steven Perron](https://github.com/s-perron), [Diego Novillo](https://github.com/dnovillo)
+* Status: **Under Consideration**
+* Planned Version: Retroactive addition to Vulkan X.X (requires SPIR-V X.X. Some language details require HLSL 202x
+
+## Introduction
+
+This proposal seeks to improve tool support for Vulkan shaders doing buffer device addressing by adding the vk::BufferPointer type to HLSL. 
+
+## Motivation
+
+vk::RawBufferLoad() is currently used to address physical storage buffer space. Unfortunately, use of this function has a number of shortcomings. One is that it generates low-level SPIR-V so that tools such as spirv-reflect, spirv-opt and renderdoc do not have the context to analyze and report on which members of a buffer are used in a logical manner. A bigger problem is that the HLSL programmer must compute the physical offsets of the members of a buffer which is error prone and difficult to maintain.
+
+For example, here is a shader using vk::RawBufferLoad(). Note the physical offset 16 hard-coded into the shader:
+
+```c++
+// struct GlobalsTest_t
+// {
+//       float4 g_vSomeConstantA;
+//       float4 g_vTestFloat4;
+//       float4 g_vSomeConstantB;
+// };
+
+struct TestPushConstant_t
+{
+      uint64_t m_nBufferDeviceAddress;  // GlobalsTest_t
+};
+
+[[vk::push_constant]] TestPushConstant_t g_PushConstants;
+
+float4 MainPs(void) : SV_Target0
+{
+    float4 vTest = vk::RawBufferLoad<float4>(g_PushConstants.m_nBufferDeviceAddress + 16);
+      
+      return vTest;
+}
+```
+
+The SPIR-V for this shader can be seen in Appendix A. Note the lack of logical context for the accessed buffer i.e. no declaration for the underlying structure GlobalsTest_t as is generated for other buffers.
+
+There is another way to use RawBufferLoad which does allow logical selection of the buffer fields, but it inefficiently loads the entire buffer to do it. See https://github.com/microsoft/DirectXShaderCompiler/issues/4986.
+
+The goal of this proposal is to have a solution that meets the following requirements:
+
+* Removes the need for having to manually or automatically generate offsets to load structured data with BufferDeviceAddress.
+* Get equivalent tooling functionality as is provided by the buffer reference feature in GLSL.  Namely, tools like RenderDoc are able to introspect the type information such that its buffer inspection and shader debugger are able to properly understand and represent the type of the data.
+* Make it possible through SPIR-V reflection to determine which members of a struct accessed by BufferDeviceAddress are statically referenced and at what offset.  This is already possible for other data like cbuffers in order for shader tooling to be able to identify which elements are used and where to put them.
+
+## Proposed solution
+
+Our solution is to add a new builtin type in the vk namespace that is a pointer to a buffer of a given type, `vk::BufferPointer<T,A>`. The template argument `T` must be a struct. `A` must be an integer and is the alignment in bytes of the pointer. If `A` is not specified, the alignment is assumed to be 16 bytes.
+
+This new type will have the following operations
+
+* Copy assignment and copy construction - These copy the value of the pointer from one variable to another.
+* Dereference Method - The get() method represents the struct rvalue pointed at by the pointer to which the get() is applied. Note that this does not necessarily imply the entire value is physically loaded at this point; it merely represents the rvalue that would be loaded, similar to the effect of the * operator when applied to a pointer in C++. As selection . operators are applied to the get(), the data that is physically loaded is reduced appropriately.
+* Null Pointer Method - The IsNull() method returns true if the pointer is 0, false if not.
+
+Note the operations that are not allowed:
+
+* There is no default construction. Every vk::BufferPointer<T> is either contained in a global resource (like a cbuffer, ubo, or ssbo), or it must be constructed using the copy constructor. 
+* There is no conversion from uint64_t to vk:BufferPointer.
+* There is no explicit pointer arithmetic. All addressing is implicitly done using the `.` pointer, or indexing an array in the struct T.
+* The comparison operators == and != are not supported for buffer pointers.
+
+Most of these restrictions are there for safety. They minimize the possibility of getting an invalid pointer. If the get() method is used on a null pointer, the behaviour is undefined.
+
+When used as a member in a buffer, vk::BufferPointer can be used to pass physical buffer addresses into a shader, and address and access buffer space with logical addressing, which allows tools such as spirv-opt, spirv-reflect and renderdoc to be able to better work with these shaders.
+
+For example, here is a shader using vk::BufferPointer to do the same thing as the shader above using vk::RawBufferLoad. Note the natural, logical syntax of the reference:
+
+```c++
+
+struct Globals_s
+{
+      float4 g_vSomeConstantA;
+      float4 g_vTestFloat4;
+      float4 g_vSomeConstantB;
+};
+
+typedef vk::BufferPointer<Globals_s> Globals_p;
+
+struct TestPushConstant_t
+{
+      Globals_p m_nBufferDeviceAddress;
+};
+
+[[vk::push_constant]] TestPushConstant_t g_PushConstants;
+
+float4 MainPs(void) : SV_Target0
+{
+      float4 vTest = g_PushConstants.m_nBufferDeviceAddress.g_vTestFloat4;
+      return vTest;
+}
+
+```
+
+In SPIR-V, Globals_p would be a pointer to the physical buffer storage class. The struct type of the push constant would contain one of those pointers. The SPIR-V for this shader can be seen in Appendix B. Note the logical context of the declaration and addressing of underlying struct Globals_s including Offset decorations all Globals_s members.
+
+## Linked Lists and Local Variables
+
+vk::BufferPointer can be used to program a linked list of identical buffers:
+
+```c++
+
+// Forward declaration
+typedef struct block_s block_t;
+typedef vk::BufferPointer<block_t> block_p;
+
+struct block_s
+{
+      float4 x;
+      block_p next;
+};
+
+struct TestPushConstant_t
+{
+      block_p root;
+};
+
+[[vk::push_constant]] TestPushConstant_t g_PushConstants;
+
+float4 MainPs(void) : SV_Target0
+{
+      block_p g_p = g_PushConstants.root;
+      g_p = g_p.next;
+      if (uint64_t(g_p) == 0) return float4(0.0,0.0,0.0,0.0);
+      return g_p.x
+}
+
+```
+
+Note also the ability to create local variables of type vk::BufferPointer such as g_p which can be read, written and dereferenced.
+
+## Design Details
+
+### Differences from C++ Pointers
+
+vk::BufferPointer is different from a C++ pointer in that a selection “.” operation can and must be applied to a buffer pointer to de-reference it. 
+
+### Buffer Pointer Target Alignment
+
+The target alignment `A` of `vk::BufferPointer(T,A)` must be at least as large as the largest component type in the buffer pointer's pointee struct type `T` or the compiler may issue an error.
+
+### Buffer Pointer Data Size and Alignment
+
+For the purpose of laying out a buffer containing a vk::BufferPointer, the data size and alignment is that of a uint64_t.
+
+### Buffer Pointer Pointee Buffer Layout
+
+The pointee of a vk::BufferPointer is considered to be a buffer and will be laid out as the user directs all buffers to be laid out through the dxc compiler. All layouts that are supported by dxc are supported for vk::BufferPointer pointee buffers.
+
+### Buffer Pointer Usage
+
+vk::BufferPointer cannot be used in Input and Output variables. It also cannot be used in Unions, when those finally appear in HLSL.
+
+A vk::BufferPointer can otherwise be used whereever the HLSL spec does not otherwise disallow it through listing of allowed types. Specifically, buffer members, local and static variables, function argument and return types can be vk::BufferPointer. Ray tracing payloads and shader buffer table records may also contain vk::BufferPointer.
+
+### Buffer Pointer and Semantic Annotations
+
+Applying HLSL semantic annotations to objects of type vk::BufferPointer is disallowed.
+
+## SPIR-V Appendices
+
+### Appendix A: SPIR-V for RawBufferLoad
+
+Note the lack of logical context for the accessed buffer i.e. no declaration for the underlying structure GlobalsTest_t as is generated for other buffers.
+
+```
+
+               OpCapability Shader
+               OpCapability Int64
+               OpCapability PhysicalStorageBufferAddresses
+               OpExtension "SPV_KHR_physical_storage_buffer"
+               OpMemoryModel PhysicalStorageBuffer64 GLSL450
+               OpEntryPoint Fragment %MainPs "MainPs" %out_var_SV_Target0 %g_PushConstants
+               OpExecutionMode %MainPs OriginUpperLeft
+               OpSource HLSL 600
+               OpName %type_PushConstant_TestPushConstant_t "type.PushConstant.TestPushConstant_t"
+               OpMemberName %type_PushConstant_TestPushConstant_t 0 "m_nBufferDeviceAddress"
+               OpName %g_PushConstants "g_PushConstants"
+               OpName %out_var_SV_Target0 "out.var.SV_Target0"
+               OpName %MainPs "MainPs"
+               OpDecorate %out_var_SV_Target0 Location 0
+               OpMemberDecorate %type_PushConstant_TestPushConstant_t 0 Offset 0
+               OpDecorate %type_PushConstant_TestPushConstant_t Block
+        %int = OpTypeInt 32 1
+      %int_0 = OpConstant %int 0
+      %ulong = OpTypeInt 64 0
+   %ulong_16 = OpConstant %ulong 16
+%type_PushConstant_TestPushConstant_t = OpTypeStruct %ulong
+%_ptr_PushConstant_type_PushConstant_TestPushConstant_t = OpTypePointer PushConstant %type_PushConstant_TestPushConstant_t
+      %float = OpTypeFloat 32
+    %v4float = OpTypeVector %float 4
+%_ptr_Output_v4float = OpTypePointer Output %v4float
+       %void = OpTypeVoid
+         %14 = OpTypeFunction %void
+         %15 = OpTypeFunction %v4float
+%_ptr_Function_v4float = OpTypePointer Function %v4float
+%_ptr_PushConstant_ulong = OpTypePointer PushConstant %ulong
+%_ptr_PhysicalStorageBuffer_v4float = OpTypePointer PhysicalStorageBuffer %v4float
+%g_PushConstants = OpVariable %_ptr_PushConstant_type_PushConstant_TestPushConstant_t PushConstant
+%out_var_SV_Target0 = OpVariable %_ptr_Output_v4float Output
+     %MainPs = OpFunction %void None %14
+         %19 = OpLabel
+         %20 = OpVariable %_ptr_Function_v4float Function
+         %21 = OpVariable %_ptr_Function_v4float Function
+         %22 = OpAccessChain %_ptr_PushConstant_ulong %g_PushConstants %int_0
+         %23 = OpLoad %ulong %22
+         %24 = OpIAdd %ulong %23 %ulong_16
+         %25 = OpBitcast %_ptr_PhysicalStorageBuffer_v4float %24
+         %26 = OpLoad %v4float %25 Aligned 4
+               OpStore %20 %26
+               OpStore %21 %26
+               OpStore %out_var_SV_Target0 %26
+               OpReturn
+               OpFunctionEnd
+
+```
+
+### Appendix B: SPIR-V for vk::buffer_ref
+
+Here is the SPIR-V for this shader. Note the logical context of the declaration and addressing of underlying struct Globals_s including Offset decorations all Globals_s members:
+
+```
+               OpCapability Shader
+               OpCapability PhysicalStorageBufferAddresses
+               OpExtension "SPV_KHR_physical_storage_buffer"
+               OpMemoryModel PhysicalStorageBuffer64 GLSL450
+               OpEntryPoint Fragment %MainPs "MainPs" %out_var_SV_Target0 %g_PushConstants
+               OpExecutionMode %MainPs OriginUpperLeft
+               OpSource HLSL 600
+               OpName %type_PushConstant_TestPushConstant_t "type.PushConstant.TestPushConstant_t"
+               OpMemberName %type_PushConstant_TestPushConstant_t 0 "m_nBufferDeviceAddress"
+               OpName %Globals_s "Globals_s"
+               OpMemberName %Globals_s 0 "g_vSomeConstantA"
+               OpMemberName %Globals_s 1 "g_vTestFloat4"
+               OpMemberName %Globals_s 2 "g_vSomeConstantB"
+               OpName %g_PushConstants "g_PushConstants"
+               OpName %out_var_SV_Target0 "out.var.SV_Target0"
+               OpName %MainPs "MainPs"
+               OpDecorate %out_var_SV_Target0 Location 0
+               OpMemberDecorate %Globals_s 0 Offset 0
+               OpMemberDecorate %Globals_s 1 Offset 16
+               OpMemberDecorate %Globals_s 2 Offset 32
+               OpDecorate %Globals_s Block
+               OpMemberDecorate %type_PushConstant_TestPushConstant_t 0 Offset 0
+               OpDecorate %type_PushConstant_TestPushConstant_t Block
+        %int = OpTypeInt 32 1
+      %int_0 = OpConstant %int 0
+      %int_1 = OpConstant %int 1
+      %float = OpTypeFloat 32
+    %v4float = OpTypeVector %float 4
+  %Globals_s = OpTypeStruct %v4float %v4float %v4float
+%_ptr_PhysicalStorageBuffer_Globals_s = OpTypePointer PhysicalStorageBuffer %Globals_s
+%type_PushConstant_TestPushConstant_t = OpTypeStruct %_ptr_PhysicalStorageBuffer_Globals_s
+%_ptr_PushConstant_type_PushConstant_TestPushConstant_t = OpTypePointer PushConstant %type_PushConstant_TestPushConstant_t
+%_ptr_Output_v4float = OpTypePointer Output %v4float
+       %void = OpTypeVoid
+         %15 = OpTypeFunction %void
+         %16 = OpTypeFunction %v4float
+%_ptr_Function_v4float = OpTypePointer Function %v4float
+%_ptr_PushConstant__ptr_PhysicalStorageBuffer_Globals_s = OpTypePointer PushConstant %_ptr_PhysicalStorageBuffer_Globals_s
+%_ptr_PhysicalStorageBuffer_v4float = OpTypePointer PhysicalStorageBuffer %v4float
+%g_PushConstants = OpVariable %_ptr_PushConstant_type_PushConstant_TestPushConstant_t PushConstant
+%out_var_SV_Target0 = OpVariable %_ptr_Output_v4float Output
+     %MainPs = OpFunction %void None %15
+         %20 = OpLabel
+         %23 = OpAccessChain %_ptr_PushConstant__ptr_PhysicalStorageBuffer_Globals_s %g_PushConstants %int_0
+         %24 = OpLoad %_ptr_PhysicalStorageBuffer_Globals_s %23
+         %25 = OpAccessChain %_ptr_PhysicalStorageBuffer_v4float %24 %int_1
+         %26 = OpLoad %v4float %25 Aligned 16
+               OpStore %out_var_SV_Target0 %26
+               OpReturn
+               OpFunctionEnd
+```
+<!-- {% endraw %} -->
--- a/proposals/0010-vk-buffer-ref.md
+++ b/proposals/0010-vk-buffer-ref.md
@@ -4,7 +4,7 @@
 * Author(s): [Greg Fischer](https://github.com/greg-lunarg)
 * Sponsor(s): [Chris Bieneman](https://github.com/llvm-beanz), [Steven Perron](https://github.com/s-perron), [Diego Novillo](https://github.com/dnovillo)
 * Status: **Under Consideration**
-* Planned Version: Retroactive addition to Vulkan X.X (requires SPIR-V X.X. Some language details require HLSL 202x
+* Planned Version: Retroactive addition to Vulkan 1.2 (requires SPIR-V 1.3. Some language details require HLSL 202x
 
 ## Introduction
 
@@ -12,7 +12,7 @@ This proposal seeks to improve tool support for Vulkan shaders doing buffer devi
 
 ## Motivation
 
-vk::RawBufferLoad() is currently used to address physical storage buffer space. Unfortunately, use of this function has a number of shortcomings. One is that it generates low-level SPIR-V so that tools such as spirv-reflect, spirv-opt and renderdoc do not have the context to analyze and report on which members of a buffer are used in a logical manner. A bigger problem is that the HLSL programmer must compute the physical offsets of the members of a buffer which is error prone and difficult to maintain.
+vk::RawBufferLoad() and vk::RawBufferStore are currently used to reference physical storage buffer space. Unfortunately, use of these functions has a number of shortcomings. One is that they generate low-level SPIR-V so that tools such as spirv-reflect, spirv-opt and renderdoc do not have the context to analyze and report on which members of a buffer are used in a logical manner. A bigger problem is that the HLSL programmer must compute the physical offsets of the members of a buffer which is error prone and difficult to maintain.
 
 For example, here is a shader using vk::RawBufferLoad(). Note the physical offset 16 hard-coded into the shader:
 
@@ -46,7 +46,7 @@ There is another way to use RawBufferLoad which does allow logical selection of 
 The goal of this proposal is to have a solution that meets the following requirements:
 
 * Removes the need for having to manually or automatically generate offsets to load structured data with BufferDeviceAddress.
-* Get equivalent tooling functionality as is provided by the buffer reference feature in GLSL.  Namely, tools like RenderDoc are able to introspect the type information such that its buffer inspection and shader debugger are able to properly understand and represent the type of the data.
+* Enables equivalent tooling functionality as is provided by the buffer reference feature in GLSL.  Namely, tools like RenderDoc are able to introspect the type information such that its buffer inspection and shader debugger are able to properly understand and represent the type of the data.
 * Make it possible through SPIR-V reflection to determine which members of a struct accessed by BufferDeviceAddress are statically referenced and at what offset.  This is already possible for other data like cbuffers in order for shader tooling to be able to identify which elements are used and where to put them.
 
 ## Proposed solution
@@ -56,7 +56,7 @@ Our solution is to add a new builtin type in the vk namespace that is a pointer 
 This new type will have the following operations
 
 * Copy assignment and copy construction - These copy the value of the pointer from one variable to another.
-* Dereference Method - The get() method represents the struct rvalue pointed at by the pointer to which the get() is applied. Note that this does not necessarily imply the entire value is physically loaded at this point; it merely represents the rvalue that would be loaded, similar to the effect of the * operator when applied to a pointer in C++. As selection . operators are applied to the get(), the data that is physically loaded is reduced appropriately.
+* Dereference Method - The Get() method represents the struct const lvalue reference of the pointer to which it is applied. The selection . operator can be applied to the Get() to further select a member from the referenced struct.
 * Null Pointer Method - The IsNull() method returns true if the pointer is 0, false if not.
 
 Note the operations that are not allowed:
@@ -66,7 +66,7 @@ Note the operations that are not allowed:
 * There is no explicit pointer arithmetic. All addressing is implicitly done using the `.` pointer, or indexing an array in the struct T.
 * The comparison operators == and != are not supported for buffer pointers.
 
-Most of these restrictions are there for safety. They minimize the possibility of getting an invalid pointer. If the get() method is used on a null pointer, the behaviour is undefined.
+Most of these restrictions are there for safety. They minimize the possibility of getting an invalid pointer. If the Get() method is used on a null pointer, the behaviour is undefined.
 
 When used as a member in a buffer, vk::BufferPointer can be used to pass physical buffer addresses into a shader, and address and access buffer space with logical addressing, which allows tools such as spirv-opt, spirv-reflect and renderdoc to be able to better work with these shaders.
 
@@ -92,7 +92,7 @@ struct TestPushConstant_t
 
 float4 MainPs(void) : SV_Target0
 {
-      float4 vTest = g_PushConstants.m_nBufferDeviceAddress.g_vTestFloat4;
+      float4 vTest = g_PushConstants.m_nBufferDeviceAddress.Get().g_vTestFloat4;
       return vTest;
 }
 
@@ -126,9 +126,9 @@ struct TestPushConstant_t
 float4 MainPs(void) : SV_Target0
 {
       block_p g_p = g_PushConstants.root;
-      g_p = g_p.next;
+      g_p = g_p.Get().next;
       if (uint64_t(g_p) == 0) return float4(0.0,0.0,0.0,0.0);
-      return g_p.x
+      return g_p.Get().x
 }
 
 ```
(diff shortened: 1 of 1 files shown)


proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,270 @@
...
+struct TestPushConstant_t
+{
+      Globals_p m_nBufferDeviceAddress;
+};
+
+[[vk::push_constant]] TestPushConstant_t g_PushConstants;
+
+float4 MainPs(void) : SV_Target0
+{
+      float4 vTest = g_PushConstants.m_nBufferDeviceAddress.g_vTestFloat4;
```
llvm-beanz, line 95: The accessing syntax is irregular here. A `vk::buffer_ptr<T>` can’t be treated as member accessing `T`. I think there are two options:

(1) Add a `Get` method to `vk::buffer_ptr` which returns a reference. This would be consistent with some of the existing buffer accessors, and provides regular syntax which can be supported for all HLSL versions.
(2) For HLSL 202x a dereference (`->`) operator could be added to indirect to the members.

proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,270 @@
...
+
+```
+
+Note also the ability to create local variables of type vk::buffer_ptr such as g_p which can be read, written and dereferenced.
+
+## Design Details
+
+### Differences from C++ Pointers
+
+vk::buffer_ptr is different from a C++ pointer in that a selection “.” operation can and must be applied to a buffer pointer to de-reference it.
```
llvm-beanz, line 142: I think this conflicts with the future directions of HLSL, so it would be better to add a more forward looking syntax.

proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Design Details
+
+### Differences from C++ Pointers
+
+vk::buffer_ptr is different from a C++ pointer in that a selection “.” operation can and must be applied to a buffer pointer to de-reference it. 
+
+### Buffer Pointer Target Alignment
+
+The target alignment `A` of `vk::buffer_ptr(T,A)` must be at least as large as the largest basic type in the buffer reference struct `T`.
```
llvm-beanz, line 146: I’m not sure this is correct. Is the intent here to match C alignment rules?

proposals/0010-vk-buffer-ref.md
greg-lunarg: > For HLSL 202x a dereference (->) operator could be added to indirect to the members.

Does that mean that "->" cannot be added today? If not, what approximate calendar could this be added and released?

greg-lunarg: Regarding get(), would the above line then be?:

`float4 vTest = g_PushConstants.m_nBufferDeviceAddress.get().g_vTestFloat4;

proposals/0010-vk-buffer-ref.md
greg-lunarg: OK

proposals/0010-vk-buffer-ref.md
greg-lunarg: This was just something I deduced based on my understanding of the overall feature.

When you say you think this is not correct, are you thinking:

1) `A` cannot be as small as the largest basic type alignment in the buffer, or

2) `A` can be smaller than the largest basic type alignment in the buffer?

proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,270 @@
...
+* Accessing members of T - This is done using the `.` operator.
+* Conversion to uint64_t - Return the integer representation of the address it points to. This can be used for null pointer checks.
+
+Note the operations that are not allowed:
+
+* There is no default construction. Every vk::buffer_ptr<T> is either contained in a global resource (like a cbuffer, ubo, or ssbo), or it must be constructed using the copy constructor. 
+* There is no equivalent to the `*` operator that will dereference the entire struct.
+* There is no conversion from uint64_t to vk:buffer_ptr.
+* There is no explicit pointer arithmetic. All addressing is implicitly done using the `.` pointer, or indexing an array in the struct T.
+* The comparison operators == and != are not supported for buffer pointers. Comparison to null pointer should be done by first converting to uint64_t.
```
llvm-beanz, line 67: Should we convert to `uint64_t` for comparison to null or should we support a boolean conversion for validity?

proposals/0010-vk-buffer-ref.md
llvm-beanz: The support for dereference operators is tied to adding references, which hasn’t yet been done. HLSL 202x doesn’t have a schedule yet, but we’re already starting to do some preliminary work on it.

Yes, that is how the `Get()` method should work.

proposals/0010-vk-buffer-ref.md
llvm-beanz: On many architectures alignments aren’t strictly tied to the sizes of basic types. For example, what is the required alignment for a float3?

If I have a float3 inside a struct, what is the alignment requirement of the struct?

In C a struct’s alignment is defined as the alignment of the most restrictive member (which is not the size of the member).

llvm-beanz:
As a general comment, please update the PR description with a meaningful description.

llvm-beanz:
I think we should also consider renaming the type to `BufferPointer`, which is more consistent with HLSL’s design. Curious if @s-perron or @Keenuts have thoughts on that.

Keenuts:
> I think we should also consider renaming the type to `BufferPointer`, which is more consistent with HLSL’s design. Curious if @s-perron or @Keenuts have thoughts on that.

You mean moving it outside of the `vk::` namespace or is is just about snake vs camel?
If it's moving it outside of the `vk::` namespace, yes, as long this is the kind of feature DX would/could support at some point in the future.

If it's only the case, sure, as all samplers/texture types are already camel-case.

proposals/0010-vk-buffer-ref.md
Keenuts: Is it required to have the full reference support to use `->`?
For now parser seems to ignore MemberExpr->isArrow() and allow `.` and `->` (isArrow is just always false), and then fails in 
`tools/clang/lib/Sema/SemaExprCXX.cpp`

Could we check the type here and allow arrow operators for this type without having the full reference implem?
This this feature can being used with the proper syntax, and then real reference support can roll-in without disturbances?

s-perron: I like the idea of a get method. I have mentioned that I do not like the idea that we cannot access the entire struct.

proposals/0010-vk-buffer-ref.md
s-perron: I think what Greg intended was correct, but is not worded clearly.

Correct me if I am wrong, but I think Greg meant to say:

> The target alignment `A` of `vk::buffer_ptr(T,A)` must be at least as large as the largest **alignment** of a basic type in the buffer reference struct `T`.

I believe this is the same as what Chris mentioned: 

> In C a struct’s alignment is defined as the alignment of the most restrictive member

s-perron (commented):
I think this generally looks good. It will be a useful feature. I don't have any suggestions other than those already made by others.

proposals/0010-vk-buffer-ref.md
llvm-beanz: We don’t need to have references fully implemented to add `->` and `*` dereference operators to built-in types, but those operators will be HLSL 202x features, not available in the older language versions.

There are some other changes that are planned for 202x which will require significant reworking of overload resolution (see: https://github.com/microsoft/hlsl-specs/pull/34), and I think that we should expect that adding new operator overloads will depend on that.

proposals/0010-vk-buffer-ref.md
llvm-beanz: I’m not sure how I feel about introducing a `nullptr` construct… I need to think on that.

llvm-beanz:
> If it's only the case, sure, as all samplers/texture types are already camel-case.

Ah yes, I just meant CamelCase instead of underscore_separated to be consistent with other HLSL data types.

proposals/0010-vk-buffer-ref.md
llvm-beanz: That’s definitely part of it.

Is the expected alignment based on the CPU or GPU data layout rules? The HLSL buffer packing rules don’t match C struct packing rules, so we probably need to be a bit explicit here about what is expected (see: https://github.com/microsoft/DirectXShaderCompiler/wiki/Buffer-Packing)

proposals/0010-vk-buffer-ref.md
s-perron: I personally like the boolean conversion idea or we could add a member function `IsNull()`. I find the member function more readable.

proposals/0010-vk-buffer-ref.md
greg-lunarg: Yes, Steven is right about my intent to refer to alignment.

I think part of the problem is my use of *basic* type. For example, for a float3, I consider the *basic* type is float. Is there a better term here? Elemental?

So, one implication of my statement is: if the buffer_ptr pointed to a struct with float3, the alignment A must be at least 4 bytes. Otherwise the compiler would give an error.

Practically, the way I use A is that I take alignment that the compiler gives me for the buffer member given the packing that it has applied to the buffer, then when I generate the SPIR-V OpLoad I use the minimum of that alignment and A.

Does this clear up all concerns?

proposals/0010-vk-buffer-ref.md
greg-lunarg: @llvm-beanz Can you please give an example of an existing HLSL method similar to your intent for get() here?

proposals/0010-vk-buffer-ref.md
greg-lunarg: The only reason for the convert to uint64_t is for the null check, so replacing the convert with an IsNull() method would be acceptable. It would also remove the dependency on the availability of uint64_t.

proposals/0010-vk-buffer-ref.md
llvm-beanz: The `RW` resource type subscript operators return references in the AST, which is roughly what you need to do here. The buffer `Load` methods also have some similarities.

greg-lunarg:
> I think we should also consider renaming the type to BufferPointer

So vk::BufferPointer? I am ok with that.

proposals/0010-vk-buffer-ref.md
llvm-beanz: I think what you're saying makes sense, but we need to get language into the spec that reflects the required alignment constraints.

greg-lunarg:
> As a general comment, please update the PR description with a meaningful description.

I have improved it. Let me know if you want more detail.

proposals/0010-vk-buffer-ref.md
llvm-beanz: Yea, I think `IsNull()` is much cleaner than supporting conversion to `uint64_t`. We really don't want/need to expose the address value directly.

proposals/0010-vk-buffer-ref.md
greg-lunarg: I am ok with a dereferencing method, but I have concerns about its spelling.

get() seems to imply a physical load of the entire buffer, which none of ".", "->" or "[]" do, at least in my mind.

For example, the following seems to imply that we would load the entire buffer, then select the member g_vTestFloat4:

`float4 vTest = g_PushConstants.m_nBufferDeviceAddress.get().g_vTestFloat4;`

Is there a better spelling we could come up with? One idea that is closer to what I would like is deref(). Would this be acceptable? Any other ideas?

To be clear, I don't have a problem with supporting a load of the entire buffer as @s-perron would like, but I don't want to imply it if we are not doing it, as in the case above.

llvm-beanz:
One area from the template that we really need to dig into and elaborate on here is:

> How does it work interact other HLSL features (semantics, buffers, etc)?

Some specific questions to consider:
* What would it mean to apply a semantic annotation to a `BufferPointer`?
* What is the expected behavior if a `BufferPointer` is inside a structure loaded from a resource?
* Can a `BufferPointer` be put in a Union?
* Can a `BufferPointer` be put inside a raytracing payload or attribute structure?

greg-lunarg:
> What would it mean to apply a semantic annotation to a BufferPointer?

Any annotations to a BufferPointer or a member contained directly or indirectly within a BufferPointer is disallowed.

greg-lunarg:
> What is the expected behavior if a BufferPointer is inside a structure loaded from a resource?

The BufferPointer is loaded along with all other values of the structure and has the same value as it did within the resource.

greg-lunarg:
> Can a BufferPointer be put in a Union?

Yes.

greg-lunarg:
> Can a BufferPointer be put inside a raytracing payload or attribute structure?

At this point, I am going to say no. The spec does say where the type can be used (buffer members, local variables, function args and returns). All other uses are disallowed.

greg-lunarg:
> I think what you're saying makes sense, but we need to get language into the spec that reflects the required alignment constraints.

The HLSL spec refers to *components*. So how about this?:

The target alignment `A` of `vk::BufferPointer(T,A)` must be at least as large as the largest alignment of all scalar component types contained in the BufferPointer's struct `T` and its structs.

proposals/0010-vk-buffer-ref.md
Keenuts: > We don’t need to have references fully implemented to add `->` and `*` dereference operators to built-in types, but those operators will be HLSL 202x features, not available in the older language versions.

Right, just unsure how the release of the HLSL 202x relates to this. Since we add a new type to the language, how would that be different from adding allowing the `->` operator for this type? Is it because you don't want to have the operator being tied to a specific type for now instead of a more generic semantic?

Agree with Greg, as the `get()` does seem to imply we load the full struct, and then hope for the optimizer to only load 1 field.

proposals/0010-vk-buffer-ref.md
Keenuts: Yes, fine with `IsNull()`. Just curious about any future divergence: cuda/opencl have pointers and `!= NULL`, so why not other shaders in a few years?  Using a bool operator with `!myBufferPointer` at least allows us be "compatible" with a more native pointer syntax, if we need one in the future.

llvm-beanz: I don't think we've ever made convergence with cuda/opencl a goal for HLSL. That said, you're probably right that we will want to support pointers in the future in HLSL.

The problem with a `bool` conversion operator is actually an existing issue in HLSL. Overloaded operators for user defined types cannot be used for Implicit conversions (there's a [bug](https://github.com/microsoft/DirectXShaderCompiler/issues/4096) tracking this). I posted a fly-by comment on that bug back in February because I think I spotted the problem, but I'm unsure if that can be safely changed in HLSL 2021.

If it can't be fixed safely in HLSL 2021, then a conversion operator would also need to be part of the HLSL 202x feature set where I have more extensive fixes for overload resolution planned.

llvm-beanz:
> Are there unions in HLSL? Not finding them. If there are, I think it should be ok. BufferPointers are defined to have size and alignment of a uint64_t. They are not opaque.

Unions were a planned feature in HLSL 2021 that didn't make it in time. They will almost certainly be part of HLSL 202x since much of the work is already done. See the proposal [here](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0004-unions.md).

> The HLSL spec refers to _components_. So how about this?:
> 
> The target alignment `A` of `vk::BufferPointer(T,A)` must be at least as large as the largest alignment of all scalar component types contained in the BufferPointer's struct `T` and its structs.

Since the intent here isn't to introduce new alignment requirements for data types maybe we can keep this simpler:

The address referenced by a `vk::BufferPointer<T>` must be correctly aligned for `T` following HLSL buffer storage alignment constraints.

> At this point, I am going to say no. The spec does say where the type can be used (buffer members, local variables, function args and returns). All other uses are disallowed.

I think this needs to be explored more. If they can be passed around through functions and stored in local variables... why not? (I don't have an actual answer or opinion here, I'm legitimately asking for a reasoning one way or the other).

greg-lunarg:
> Since the intent here isn't to introduce new alignment requirements for data types maybe we can keep this simpler:
>
> The address referenced by a vk::BufferPointer<T> must be correctly aligned for T following HLSL buffer storage alignment constraints.

How about?:

The alignment A of vk::BufferPointer<T,A> must be consistent with all elements within struct T.

proposals/0010-vk-buffer-ref.md
greg-lunarg: OK. I am fine with get().

greg-lunarg:
>>> Can a BufferPointer be put inside a raytracing payload or attribute structure?

>> At this point, I am going to say no. The spec does say where the type can be used (buffer members, local variables, function args and returns). All other uses are disallowed.

> I think this needs to be explored more. If they can be passed around through functions and stored in local variables... why not? (I don't have an actual answer or opinion here, I'm legitimately asking for a reasoning one way or the other).

As for payloads, I can't think of a reason for disallowing it, although I am not sure how useful it would be as payloads tend to be small.

It seems intersection attribute structures would be similar. For fixed function triangle intersection, it looks like those have a predetermined type (a struct containing a float2) so cannot use it. As for axis-aligned bounding box, it might be possible, but I again I am not sure how useful it would be since these are usually no more than 32 bytes in size.

natevm:
> > Can a BufferPointer be put inside a raytracing payload or attribute structure?
> 
> At this point, I am going to say no. The spec does say where the type can be used (buffer members, local variables, function args and returns). All other uses are disallowed.

It is critical that we allow these buffer pointer types to be embedded in shader binding table records, and arguably in ray payload structures as well. We depend on this for many advanced ray tracing workloads, and are currently using the RawBufferLoad/Store with 64 bit addresses to do this.

natevm:
> > > > Can a BufferPointer be put inside a raytracing payload or attribute structure?
> 
> > > At this point, I am going to say no. The spec does say where the type can be used (buffer members, local variables, function args and returns). All other uses are disallowed.
> 
> > I think this needs to be explored more. If they can be passed around through functions and stored in local variables... why not? (I don't have an actual answer or opinion here, I'm legitimately asking for a reasoning one way or the other).
> 
> As for payloads, I can't think of a reason for disallowing it, although I am not sure how useful it would be as payloads tend to be small.
> 
> It seems intersection attribute structures would be similar. For fixed function triangle intersection, it looks like those have a predetermined type (a struct containing a float2) so cannot use it. As for axis-aligned bounding box, it might be possible, but I again I am not sure how useful it would be since these are usually no more than 32 bytes in size.

In the past, we have passed pointers to stacks this way. An intersection program might increment a count and store an intersection into a buffer passed in the ray payload. Useful for anyhit based particle compositing.

natevm:
Is it possible to make a vk::BufferPointer<vk::BufferPointer> here? 

For example, a grid of points, where each grid cell contains a different number of points, one might want a multidimensional buffer pointer, where the first dimension is the cell in the grid, and the second dimension is the point in the cell. 

Another example would be direct sampling of emissive triangles from one of many emissive meshes. First dimension would be the mesh, and second would be the triangle in that mesh.

note, I’m asking about pointers to pointers here, since you might have one array of vertices referenced in multiple locations (eg the SBT hit record for one geometry, and in another separate buffer of buffers in a ray generation SBT record)

proposals/0010-vk-buffer-ref.md
Keenuts: Thanks for the thorough explanation!
for `get()`, missed the parallel with shared pointers. Good argument, fine with `get()`.

Thanks for the clarification on the frontier between HLSL version and shader model version. Those arguments make sense 😊

greg-lunarg:
@llvm-beanz @natevm I will remove language which says use of this type is disallowed beyond local variables and function args and returns. Still, I would rather not try enumerating ALL places they are allowed. Can I rather say?: BufferPointer is supported wherever the underlying runtime and API supports them. This includes local variables and function args and returns.

llvm-beanz:
> Still, I would rather not try enumerating ALL places they are allowed. Can I rather say?: BufferPointer is supported wherever the underlying runtime and API supports them. This includes local variables and function args and returns.

I don't think this is sufficient. When we move from specifying to implementing this change we need to implement the compiler to enforce the appropriate restrictions on where `BufferPointer` or data types containing `BufferPointer` members can be used, and we need to implement tests to verify that behavior. Leaving that as a gap in the proposal will only make it harder for us to verify a correct implementation.

natevm:
> > Is it possible to make a vk::BufferPointer<vk::BufferPointer<T,A>,A> here?
> 
> @natevm Does it work for you to have a vk::BufferPointer<T,A> where T is a struct containing a vk::BufferPointer? If so, I think this approach would solve all of your related requests above. Remember that a BufferPointer is a pointer to a buffer, and a buffer's underlying type is a struct. The struct is needed to support the layout that is needed. That is the reason for the restriction. But it should still allow you to do everything you want to do, I think.

That would work. Mostly just awkward for cases like buffers of vertices, since float3 is not a struct. At the same time, I get the concern over struct alignment, so that seems like a good compromise IMO.

natevm:
> > Still, I would rather not try enumerating ALL places they are allowed. Can I rather say?: BufferPointer is supported wherever the underlying runtime and API supports them. This includes local variables and function args and returns.
> 
> I don't think this is sufficient. When we move from specifying to implementing this change we need to implement the compiler to enforce the appropriate restrictions on where `BufferPointer` or data types containing `BufferPointer` members can be used, and we need to implement tests to verify that behavior. Leaving that as a gap in the proposal will only make it harder for us to verify a correct implementation.

Perhaps it would be better to identify cases where vk::BufferPointer wouldn’t be allowed? Are there any such cases that come to mind?

greg-lunarg:
@natevm You mentioned raytracing payloads as a useful place for BufferPointer to be used. Are there other raytracing structures they would be useful in? Please list.

greg-lunarg:
> I don't think this is sufficient. When we move from specifying to implementing this change we need to implement the compiler to enforce the appropriate restrictions on where BufferPointer or data types containing BufferPointer members can be used, and we need to implement tests to verify that behavior. Leaving that as a gap in the proposal will only make it harder for us to verify a correct implementation.

@llvm-beanz @natevm I think we want to disallow BufferPointer type from input and output variables. Any disagreement?

natevm:
> @natevm You mentioned raytracing payloads as a useful place for BufferPointer to be used. Are there other raytracing structures they would be useful in? Please list.

Shader binding table records are the other big one. Beyond that, none that I can think of. 🙂

greg-lunarg:
@llvm-beanz Should I say that vk::BufferPointer is disallowed unless targeting Vulkan? Is that just assumed for everything in the vk namespace?

Keenuts:
> @llvm-beanz Should I say that vk::BufferPointer is disallowed unless targeting Vulkan? Is that just assumed for everything in the vk namespace?

I would assume disallowed outside of vulkan, as the opposite would be weird.

Keenuts:
> > Can a BufferPointer be put in a Union?
> 
> Are there unions in HLSL? Not finding them. If there are, I think it should be ok. BufferPointers are defined to have size and alignment of a uint64_t. They are not opaque.

If we allow union on this type (since it's a uint64_t), isn't that going against the "we don't want to allow `uint64_t raw_addr = my_pointer`" ?

greg-lunarg:
> If we allow union on this type (since it's a uint64_t), isn't that going against the "we don't want to allow uint64_t raw_addr = my_pointer" ?

@Keenuts I don't think so.

Based on https://github.com/microsoft/hlsl-specs/pull/3, HLSL unions are like C++ unions. Using C++ unions to do casting is undefined behavior.

The C++ standard states the following in §9.5.1:

In a union, at most one of the non-static data members can be active at any time, that is, the value of at most one of the non-static data members can be stored in a union at any time. [...] The size of a union is sufficient to contain the largest of its non-static data members. Each non-static data member is allocated as if it were the sole member of a struct. All non-static data members of a union object have the same address.

natevm:
> > If we allow union on this type (since it's a uint64_t), isn't that going against the "we don't want to allow uint64_t raw_addr = my_pointer" ?
> 
> @Keenuts I don't think so.
> 
> Based on #3, HLSL unions are like C++ unions. Using C++ unions to do casting is undefined behavior.
> 
> The C++ standard states the following in §9.5.1:
> 
> In a union, at most one of the non-static data members can be active at any time, that is, the value of at most one of the non-static data members can be stored in a union at any time. [...] The size of a union is sufficient to contain the largest of its non-static data members. Each non-static data member is allocated as if it were the sole member of a struct. All non-static data members of a union object have the same address.

Why not allow us to construct a BufferPointer object from a raw address again? If it works, to be honest I probably would use this undefined union behavior just as a workaround to this seemingly odd limitation.

natevm:
> What raw address? A uint64_t? If this type is available, why do you need to set it from a "raw address"?

A uint64_t is currently used to represent raw device addresses in HLSL. I don't see why one shouldn't be allowed to construct a vk::BufferPointer from a uint64_t.

Keenuts:
> Using C++ unions to do casting is undefined behavior.

TIL type punning was only allowed in C, not C++..

proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,270 @@
...
+
+Most of these restrictions are there for safety. They minimize the possibility of getting an invalid pointer. If the `.` operator is used on a null pointer, the behaviour is undefined.
+
+When used as a member in a buffer, vk::buffer_ptr can be used to pass physical buffer addresses into a shader, and address and access buffer space with logical addressing, which allows tools such as spirv-opt, spirv-reflect and renderdoc to be able to better work with these shaders.
+
+For example, here is a shader using vk::buffer_ptr to do the same thing as the shader above using vk::RawBufferLoad. Note the natural, logical syntax of the reference:
```
devshgraphicsprogramming, line 73: to me all this describes a concept similar to a C++ reference, not a pointer. I suppose thats why its `GLSL_EXT_buffer_reference` and not `GLSL_EXT_buffer_pointer`.

Its not even about the the `.` being used over `->`

Not even the lack of pointer arithmetic.

The biggest thing that makes this a _reference_ and **not a pointer** is this
https://github.com/microsoft/DirectXShaderCompiler/pull/5089#issuecomment-1482072617
https://github.com/microsoft/DirectXShaderCompiler/pull/5089#issuecomment-1482087964
there seems to be no way to change the byte address in a struct, unless you somehow alias it with some other resource (which you obviously can't do for non-buffer variables because there's no `reinterpret_cast`) where the "pointer" is declared as a plain `uint64_t`.

Ergo my argument that this is a reference, and not a pointer.

llvm-beanz:
> The C++ standard states the following in §9.5.1:
> 
> In a union, at most one of the non-static data members can be active at any time, that is, the value of at most one of the non-static data members can be stored in a union at any time. [...] The size of a union is sufficient to contain the largest of its non-static data members. Each non-static data member is allocated as if it were the sole member of a struct. All non-static data members of a union object have the same address.

This is actually a really good point to bring up. While this is technically undefined behavior, I think every major C++ compiler implements C's rules for unions rather than anything else. I've filed an issue and assigned it to myself to revisit this in the Union's spec (see #38). I think we likely want to at least clarify the rules for HLSL there. My guess is that using unions for type punning is far too useful to not support explicitly.

proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,270 @@
+<!-- {% raw %} -->
+# Buffer Pointers in HLSL With vk::buffer_ptr
+
+* Author(s): [Greg Fischer](https://github.com/greg-lunarg)
+* Status: **Under Consideration**
+* Planned Version: ASAP?? SM6 required.
+
```
llvm-beanz, line 7: Please add myself, @s-perron, and @dnovillo as sponsors.

llvm-beanz (commented):
At this point I think we likely just need to see the document updated to reflect all the changes discussed throughout the review.

My inclination is that given an updated document we can accept this as an _under consideration_ proposal, file issues to track additional spec changes, and merge the change.

proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,270 @@
...
+This new type will have the following operations
+
+* Copy assignment and copy construction - These copy the value of the pointer from one variable to another.
+* Accessing members of T - This is done using the `.` operator.
+* Conversion to uint64_t - Return the integer representation of the address it points to. This can be used for null pointer checks.
+
+Note the operations that are not allowed:
+
+* There is no default construction. Every vk::buffer_ptr<T> is either contained in a global resource (like a cbuffer, ubo, or ssbo), or it must be constructed using the copy constructor.
```
llvm-beanz, line 63: I think this restriction actually means these can't be placed in Unions by C++ rules:

>If a union contains a non-static data member of reference type the program is ill-formed. At most one non-static data member of a union may have a brace-or-equal-initializer. [Note: If any non-static data member of a union has a non-trivial default constructor (12.1), copy constructor (12.8), move constructor (12.8), copy assignment operator (12.8), move assignment operator (12.8), or destructor (12.4), the corresponding member function of the union must be user-provided or it will be implicitly deleted (8.4.3) for the union. — end note ]

Since HLSL doesn't yet support constructors and destructors, this restriction would mean that `vk::BufferPointer` cannot be placed inside a union.

proposals/0010-vk-buffer-ref.md
llvm-beanz: I think the real challenge here is that references, by definition, can't be null, and this does explicitly say it can be null.

I do agree with you that in some contexts this does behave more like a reference.

I can imagine a future where `vk::BufferPointer` could be implemented as something like a `std::optional<T &__attribute__((address_space(...) ))`, but it could also be more like a C++ smart pointer abstraction... I think it could go either way.

s-perron: I also find it to be more like a pointer because an assignment from one to another will change the address it points to and not the memory that it points to:

```
vk:buffer_ptr a = g1;
vk:buffer_ptr b = g2;
a.get() = 0;
b.get() = 1;
vk:buffer_ptr c = b;
c = a; // For a C++ reference you would expect b.get() to become 0, but in this case b.get() does not change.
```

Then as Chris mentioned, the ability for the value to be null makes it lean more towards a pointer for me.

dnovillo: The example at the end with the linked list also makes me think pointer semantics.  In C terms, I'd think of it as a const pointer.

proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,270 @@
...
+
+float4 MainPs(void) : SV_Target0
+{
+      float4 vTest = g_PushConstants.m_nBufferDeviceAddress.g_vTestFloat4;
+      return vTest;
+}
+
+```
+
+In SPIR-V, Globals_t would be a pointer to the physical buffer storage class. The type of the push constant would contain one of those pointers. The SPIR-V for this shader can be seen in Appendix B. Note the logical context of the declaration and addressing of underlying struct Globals_t including Offset decorations all Globals_t members.
```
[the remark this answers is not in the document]
greg-lunarg: Yes. Thanks!

greg-lunarg:
I think I have now made all changes requested or approved by @llvm-beanz.

One request not yet added is for a method to convert uint64_t to vk::BufferPointer. This was removed at the request of @s-perron. There is explanatory text saying that this was done to increase safety and reduce the chances of creating an invalid pointer.

natevm:
> I think I have now made all changes requested or approved by @llvm-beanz.
> 
> One request not yet added is for a method to convert uint64_t to vk::BufferPointer. This was removed at the request of @s-perron. There is explanatory text saying that this was done to increase safety and reduce the chances of creating an invalid pointer.

This is a bad take, and we should add back in construction. Users can just as easily make invalid pointers by uploading incorrect data from the host to the device. Why should initializing a pointer on the device be any different? This stance on safety is inconsistent, and in my opinion impossible to truly achieve without a unified language across host and device.

I don’t want HLSL hand holding me and constraining the code I write. Especially when it’s high performance GPU code… Thats just frustrating. By preventing pointer initialization from a uint64_t, we force HLSL users to synchronize the GPU back to the CPU anytime a pointer needs initialization t from a 64 bit device address. 

Constructing from a 64 bit address also gives a clear mechanism for type casting (unless there’s another way to do this I’m unaware of?). And type casting is an important property of pointers in many high performance GPU algorithms, especially with raw and complex data structures. 

Ultimately I think this argument that it is about safety is inconsistent. Pointers will be unsafe, regardless of if they can be initialized by a 64 bit device address on the device or not. They should be flexible and allow me to write the code I need to write to achieve high performance applications. Without them, we place counterproductive constraints on developers without truly solving safety concerns.

s-perron:
> This is a bad take, and we should add back in initialization. Users can just as easily make invalid pointers by uploading incorrect data from the host to the device. Why should initializing a pointer on the device be any different? This stance on safety is inconsistent, and in my opinion impossible to truly achieve without a unified language across host and device.

Could we have this discussion in a follow up issue? There is more to this than just having invalid addresses. There are also aliasing rules for these pointers that we have not talked about at all. The sample spir-v does not have any "AliasedPointer" decorations, so the assumption might be that all buffer pointers have to be restrict pointers. If that is the case, it should be stated explicitly. We need to make sure we are  clear on the aliasing rules. Tracking aliasing in a compiler is much easier without arbitrary pointer arithmetic. I think all of these things need to be considered together, and this PR is already too busy that the conversation will be lost.

> Instead, we should be clearly warn that initialization is unsafe if the uint64_t is not a true device address.

We need something more restrictive than that. To have a valid spir-v module, we would have to add a requirement in the spec that address must be to memory on the appropriate storage class. We need to guarentee that it will not get the same address as a "logical address" in spir-v. This will break the spir-v aliasing rules, and will cause subtle bugs because of compiler optimizations.

> Ultimately I think this argument that it is about safety is inconsistent. Pointers will be unsafe, regardless of if they can be initialized by a 64 bit device address on the device or not.

You might be setting up a false dichotomy. You seem to be arguing that if we cannot make it safe in all cases, then there is no point in trying to reduce errors anywhere. One difference could be debugability. I'm guessing it would be much easier to debug the cpu than the gpu just based on the tooling that is available, so helping reduce bug in the shader could still be a productivity boost. We also have to consider how the compilers will optimize the code, and still keep it correct. When I use to work on C/C++ compilers, I spent too much time finding "compiler bugs" that were caused by the application violating the c++ ansi aliasing rules. This hurts everybody. I think we need to try to have a clear set of aliasing rules, and make it hard for people to break them, if not impossible. This will require more discussion.

llvm-beanz:
@greg-lunarg, in the future can you please not force-push updates? We can always squash the commits on merge, but force pushing during the review makes it more difficult to see what changed between versions.

greg-lunarg:
@natevm Once this PR is merged, can you please open a new issue against this document showing a use case where conversion from uint64_t is required to get reasonable performance. I am presuming you are somehow manufacturing new uint64_t addresses in the shader and sharing them. True?

@s-perron I will open up a new issue regarding the aliasing properties of vk::BufferPointer. TL;DR All different instances of type vk::BufferPointer are assumed to not alias each other. Any desired synchronization of writes and reads through different buffer pointer instances must be executed explicitly.

natevm:
@greg-lunarg I unfortunately suspect if we do not discuss device-side vk::BufferPointer initialization now, then it will never be prioritized, even if we make another issue. I also do not have the resources to really argue for this.

@s-perron a bit of a false dichotomy, but you also make a bit of a false dichotomy too. You are assuming preventing device initialization is the only way to improve safety, but there are likely additional, likely better options. For example, we might allow vk::BufferPointer initialization only if a special compiler flag is specified, which would dissuade its use by default but also allow the flexibility in special circumstances.

One other thought I had was that constructing vk::BufferPointers on the device might provide additional options for safety. DXC cannot predict host side behavior, but potentially down the road, pointer initialization could be error checked by emitting additional instructions in debug builds.

llvm-beanz:
> @greg-lunarg I unfortunately suspect if we do not discuss device-side vk::BufferPointer initialization now, then it will never be prioritized, even if we make another issue. I also do not have the resources to really argue for this. Don’t languages like CUDA expose pointer aliasing through the “volatile” keyword?

This gets really quickly outside the scope of what we can handle in the HLSL language. SPIR-V makes assumptions that pointers don't alias, we can't change SPIR-V, that's the Khronos Group's space.
 
> For example, we might allow vk::BufferPointer initialization only if a special compiler flag is specified, which would dissuade its use by default but also allow the flexibility in special circumstances.

I'm going to put a strong dislike on this idea... Compiler flags that change what is or isn't legal the language are bad, and we should strive to not introduce them (and remove the few we already have for legacy reasons).

natevm:
> 

Fair. Still, I think folks are going to expect to be able to cast a pointer from one type to another... That is a very basic feature.

proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Proposed solution
+
+Our solution is to add a new builtin type in the vk namespace that is a pointer to a buffer of a given type, `vk::BufferPointer<T,A>`. The template argument `T` must be a struct. `A` must be an integer and is the alignment in bytes of the pointer. If `A` is not specified, the alignment is assumed to be 16 bytes.
+
+This new type will have the following operations
+
+* Copy assignment and copy construction - These copy the value of the pointer from one variable to another.
+* Dereference Method - The get() method represents the struct rvalue pointed at by the pointer to which the get() is applied. Note that this does not necessarily imply the entire value is physically loaded at this point; it merely represents the rvalue that would be loaded, similar to the effect of the * operator when applied to a pointer in C++. As selection . operators are applied to the get(), the data that is physically loaded is reduced appropriately.
```
llvm-beanz, line 59: nit: `Get()` capital `G` to be consistent with HLSL style conventions.

I think it is probably more accurate that this is a `const` lvalue reference than an rvalue since I assume that's how this should likely be implemented.

proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,280 @@
...
+
+```
+
+Note also the ability to create local variables of type vk::BufferPointer such as g_p which can be read, written and dereferenced.
+
+## Design Details
+
+### Differences from C++ Pointers
+
+vk::BufferPointer is different from a C++ pointer in that a selection “.” operation can and must be applied to a buffer pointer to de-reference it.
```
llvm-beanz, line 142: This is out of date.

proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Design Details
+
+### Differences from C++ Pointers
+
+vk::BufferPointer is different from a C++ pointer in that a selection “.” operation can and must be applied to a buffer pointer to de-reference it. 
+
+### Buffer Pointer Target Alignment
+
+The target alignment `A` of `vk::BufferPointer(T,A)` must be at least as large as the largest component type in the buffer pointer's pointee struct type `T` or the compiler may issue an error.
```
llvm-beanz, line 146: nit: `vk::BufferPointer<T,A>`

proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,280 @@
...
+
+For the purpose of laying out a buffer containing a vk::BufferPointer, the data size and alignment is that of a uint64_t.
+
+### Buffer Pointer Pointee Buffer Layout
+
+The pointee of a vk::BufferPointer is considered to be a buffer and will be laid out as the user directs all buffers to be laid out through the dxc compiler. All layouts that are supported by dxc are supported for vk::BufferPointer pointee buffers.
+
+### Buffer Pointer Usage
+
+vk::BufferPointer cannot be used in Input and Output variables. It also cannot be used in Unions, when those finally appear in HLSL.
```
llvm-beanz, line 158: nit: strike the word `finally`.

natevm:
Ultimately, without pointer arithmetic or casting, I do not see a purpose for this proposal. In fact, it could be detrimental. These are the two key reasons why we are using vk::RawBufferLoad and Store. Without them, our library will cease to work, and we will be forced to either move away from HLSL, or to maintain a fork.

llvm-beanz:
> Ultimately, without pointer arithmetic or casting, I do not see a purpose for this proposal. In fact, it could be detrimental. These are the two key reasons why we are using vk::RawBufferLoad and Store. Without them, our library will cease to work, and we will be forced to either move away from HLSL, or to maintain a fork.

I think this is a bit hyperbolic. Your library isn't going to cease to work because this proposal doesn't remove any functionality that exists today. It may not be a fit for you to update to this feature, but that's not the same as pulling the rug out from under you and forcing you to "move away from HLSL, or to maintain a fork."

We hear your feedback. Casting is absolutely something we need to cover and describe in this proposal before we accept it as a feature for HLSL.

I think there are still some questions we need to sort out about arithmetic and aliasing.

This proposal is not yet done being designed. Even after we merge this document, it is **under consideration** not an accepted feature.

We'll iterate on the design until we get to something we're happy with, and our goal is to address meaningful needs by users, but it also almost certainly won't be everything that everyone wants.

Think of this PR as the first step in a process that has a few more steps to go. We're trying to move quickly to not roadblock and to get this into user hands as fast as we can while also ensuring a maintainable and high quality design.

natevm:
> I think this is a bit hyperbolic. Your library isn't going to cease to work because this proposal doesn't remove any functionality that exists today. It may not be a fit for you to update to this feature, but that's not the same as pulling the rug out from under you and forcing you to "move away from HLSL, or to maintain a fork."

Because this impacts my job and research, I am being a bit hyperbolic yes. But it is important. We really do need these features. So long as we don't remove the vk::RawBufferLoad and vk::RawBufferStore commands right away, at least until casting and aliasing are figured out wit vk::BufferPointers, then I'm ok to proceed.

> This proposal is not yet done being designed. Even after we merge this document, it is under consideration not an accepted feature. 

Ah, I misunderstood the process here. That makes sense. Yes, could be discussed in another PR on the same proposal.

natevm (commented):
The one small thing we might add to this proposal is that vk::RawBufferStore also is a thing that has issues, which would also be addressed with a vk::BufferPointer.

llvm-beanz:
>So long as we don't remove the vk::RawBufferLoad and vk::RawBufferStore commands right away, at least until casting and aliasing are figured out wit vk::BufferPointers, then I'm ok to proceed.

There is no proposal to remove `vk::RawBufferLoad` and `vk::RawBufferStore`. That is not in this proposal. Nobody is suggesting removing them. This proposal doesn't come close to covering all the uses for the existing features.

> Ah, I misunderstood the process here. That makes sense. Yes, could be discussed in another PR on the same proposal.

Our [process docs](https://github.com/microsoft/hlsl-specs/blob/main/docs/Process.md) aren't perfect, but they do cover the phases of the proposal process. This proposal is **under consideration** so there's still a bit to go before we start accepting changes to support it into DXC.

proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,280 @@
+<!-- {% raw %} -->
+# Buffer Pointers in HLSL With vk::BufferPointer
+
+* Author(s): [Greg Fischer](https://github.com/greg-lunarg)
+* Sponsor(s): [Chris Bieneman](https://github.com/llvm-beanz), [Steven Perron](https://github.com/s-perron), [Diego Novillo](https://github.com/dnovillo)
+* Status: **Under Consideration**
+* Planned Version: Retroactive addition to Vulkan X.X (requires SPIR-V X.X. Some language details require HLSL 202x
```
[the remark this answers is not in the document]
greg-lunarg: Vulkan 1.2 and SPIR-V 1.3. Done.

proposals/0010-vk-buffer-ref.md
greg-lunarg: Done.

proposals/0010-vk-buffer-ref.md
greg-lunarg: Changed to Get() to dereference.

proposals/0010-vk-buffer-ref.md
greg-lunarg: Done.

proposals/0010-vk-buffer-ref.md
greg-lunarg: Done.

proposals/0010-vk-buffer-ref.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Introduction
+
+This proposal seeks to improve tool support for Vulkan shaders doing buffer device addressing by adding the vk::BufferPointer type to HLSL. 
+
+## Motivation
+
+vk::RawBufferLoad() is currently used to address physical storage buffer space. Unfortunately, use of this function has a number of shortcomings. One is that it generates low-level SPIR-V so that tools such as spirv-reflect, spirv-opt and renderdoc do not have the context to analyze and report on which members of a buffer are used in a logical manner. A bigger problem is that the HLSL programmer must compute the physical offsets of the members of a buffer which is error prone and difficult to maintain.
```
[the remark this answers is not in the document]
greg-lunarg: Good catch. Thanks!

greg-lunarg:
I have pushed the latest round of review fixes.

s-perron (approved):

s-perron (approved):
I'll approve with a comment. There are still two open issue that need to be discussed in follow up issues. 

1. How will aliasing work? 
2. Will we allow some type of type punning, and, if so, how will we do it?

s-perron:
> SPIR-V makes assumptions that pointers don't alias, we can't change SPIR-V, that's the Khronos Group's space.

One of the issues we will need to discuss further is whether or now we need to expose SPIR-V's alias and restrict decorations. The points can alias, but they have to be labeled as aliasing.

https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#AliasingSection:~:text=For%20variables%20holding,alias%20each%20other.

llvm-beanz (approved):
I've filed some follow-up issues that reference this PR. This [search](https://github.com/microsoft/hlsl-specs/issues?q=is%3Aissue+is%3Aopen+%5B0010%5D+in%3Atitle) should collect them all.

I think this covers all the issues mentioned by @s-perron as well, but we will file more as needed.

For now, I think we can accept this feature as *under consideration* while we work through the remaining design issues.

Unless anyone objects I'll merge the PR before EOD.

PR #39, opened 2023-04-13
Add new issue templates

llvm-beanz:
These templates are for helping file spec issues and proposal follow-up issues so that we can better track proposals as they are being developed.

### DIFF
--- a/.github/ISSUE_TEMPLATE/proposal.md
+++ b/.github/ISSUE_TEMPLATE/proposal.md
@@ -8,7 +8,7 @@ assignees: ''
 ---
 
 **Which proposal does this relate to?**
-A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+Please link the file in the repo where the problem is documented.
 
 **Describe the issue or outstanding question.**
 This can be as short as a question that needs answering or as complex as needed
--- a/.github/ISSUE_TEMPLATE/proposal.md
+++ b/.github/ISSUE_TEMPLATE/proposal.md
@@ -1,0 +1,18 @@
+---
+name: Proposal Follow-up
+about: Follow-up issue for a proposal
+title: ''
+labels: 'active proposal'
+assignees: ''
+
+---
+
+**Which proposal does this relate to?**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the issue or outstanding question.**
+This can be as short as a question that needs answering or as complex as needed
+to convey the problem that needs solving.
+
+**Additional context**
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/proposal.md
```diff
@@ -0,0 +1,18 @@
...
+name: Proposal Follow-up
+about: Follow-up issue for a proposal
+title: ''
+labels: 'active proposal'
+assignees: ''
+
+---
+
+**Which proposal does this relate to?**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
```
dmpots, line 11: Is this a copy/paste error? Looks like the same text below. Shouldn't it say something like 

> Name the proposal this follow-up refers to using the XXXX-ProposalName format.

dmpots (approved):
Probably worth mentioning these in the process document as well. As it stands its not clear to me when these would/should be used.

.github/ISSUE_TEMPLATE/spec.md
```diff
@@ -0,0 +1,17 @@
+---
+name: Spec
+about: Highlight a problem with a published or adopted spec
+title: ''
+labels: 'spec'
+assignees: ''
+
+---
+
+**Which spec or proposal does this relate to?**
```
pow2clk, line 10: Under what circumstances would this be used for a proposal as opposed to using proposal.md?

tex3d: Given this, could the frontmatter line above be updated to clarify this?
```
about: Highlight a problem with a published or adopted spec
```

Maybe it could read something more like this?  What's the difference between a published and an adopted spec?
```
about: Highlight a problem with a published spec or accepted proposal
```

Or does this just complicate things?  Perhaps if proposal and spec documents are separate, then the first template would be for any proposal document no matter what the state, and this template would only be for spec documents.

Also, I see elsewhere in the Lifecycle->Accepted section of `Process.md`, it says: `Once a feature is accepted it becomes a planned feature for the release.`  Could `for the release` instead read `for a future release`?  That should eliminate the need to walk back from this, explaining that it may not "ship in the planned release" etc...  I'm not quite sure what "the release" refers to in the first place, unless each feature proposal will get a tagged for a specific release before being "accepted".

llvm-beanz: I intended `for the release` to refer to the `Planned Version` in the proposal doc. We can always defer a proposal either by putting it into the **deferred** state or by updating it to a different `Planned Version`.

tex3d (approved):

PR #46, opened 2023-04-14
Add design considerations and clarify process

llvm-beanz:
This update adds new documents on design considerations and contact information.

The design considerations document tries to codify some of the basic design language for HLSL and cover common topics that features will need to address in their specification.

This also updates some wording and details around the process docs and specifically about proposal sponsors. We think it is important that Vulkan-specific proposals have a sponsor from the Vulkan Working Group, so that is explicitly stated.

### DIFF
--- a/docs/Contact.md
+++ b/docs/Contact.md
@@ -4,15 +4,21 @@ This page lists some of the public spaces where you can find members of the HLSL
 
 ## Discord Channels
 
-* [DirectX](https://discord.com/invite/directx)
-* [LLVM](https://discord.gg/xS7Z362)
+* [DirectX](https://discord.com/invite/directx) - For D3D related
+  HLSL topics.
+* [LLVM](https://discord.gg/xS7Z362) - For HLSL Compiler Modernization and other
+  compiler topics.
 
 ## GitHub Issues
 
 * [DirectXShaderCompiler](https://github.com/microsoft/DirectXShaderCompiler/issues/new)
-* [hlsl-specs](https://github.com/microsoft/hlsl-specs/issues/new)
-* [LLVM](https://github.com/llvm/llvm-project/issues/new)
+  - For bugs and feature requests specific to the DXC compiler.
+* [hlsl-specs](https://github.com/microsoft/hlsl-specs/issues/new) - For feature
+  requests for the HLSL language or runtimes (Shader Models, Vulkan, etc).
+* [LLVM](https://github.com/llvm/llvm-project/issues/new) - For bugs and feature
+  requests specific to the Clang HLSL implementation.
 
 ## Email
 
-* [Ask HLSL](mailto:askhlsl@microsoft.com)
+* [Ask HLSL](mailto:askhlsl@microsoft.com) - For issues where privacy is
+  essential.
--- a/docs/Contact.md
+++ b/docs/Contact.md
@@ -1,0 +1,18 @@
+# Contacting the HLSL Team
+
+This page lists some of the public spaces where you can find members of the HLSL.
+
+## Discord Channels
+
+* [DirectX](https://discord.com/invite/directx)
+* [LLVM](https://discord.gg/xS7Z362)
+
+## GitHub Issues
+
+* [DirectXShaderCompiler](https://github.com/microsoft/DirectXShaderCompiler/issues/new)
+* [hlsl-specs](https://github.com/microsoft/hlsl-specs/issues/new)
+* [LLVM](https://github.com/llvm/llvm-project/issues/new)
+
+## Email
+
+* [Ask HLSL](mailto:askhlsl@microsoft.com)
--- a/docs/DesignConsiderations.md
+++ b/docs/DesignConsiderations.md
@@ -1,0 +1,84 @@
+# HLSL Design Considerations
+
+When designing and proposing a feature for HLSL there are some design
+considerations that should be taken into account.
+
+## Style Conventions
+
+HLSL's built-in types and methods should conform to a consistent coding style.
+
+* Data types, methods and built-in functions should all be `CamelCase`.
+* Namespaces and keywords are lowercase, `_` separated.
+* Vulkan-specific functionality should be added to the `vk` namespace.
+* Microsoft-style attributes interchangably use `CamelCase` and `_` separation,
+  but should prefer `CamelCase`
+* System Value semantics are case insensitive, should be specified `CamelCase`
+  and prefixed `SV_`.
+
+## Versioning
+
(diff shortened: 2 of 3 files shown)


tex3d (approved):
LGTM.

docs/Contact.md
```diff
@@ -0,0 +1,18 @@
...
+
+## GitHub Issues
+
+* [DirectXShaderCompiler](https://github.com/microsoft/DirectXShaderCompiler/issues/new)
+* [hlsl-specs](https://github.com/microsoft/hlsl-specs/issues/new)
+* [LLVM](https://github.com/llvm/llvm-project/issues/new)
+
+## Email
+
+* [Ask HLSL](mailto:askhlsl@microsoft.com)
```
pow2clk, line 18: It might be useful to suggest what each resource is most appropriate for.  My perfunctory suggestions:

DirectX Discord - D3D related HLSL topics
LLVM discord - Modernization project HLSL and other DX topics

DXC Github - bugs and feature requests specific to the DXC compiler
LLVM Github - bugs and feature requests specific to the modern clang HLSL/DX compiler
HLSL-Specs - feature requests for language and shader model versions

ask HLSL email - when privacy is essential or to prod a submission in any of the other contact methods.

docs/DesignConsiderations.md
```diff
@@ -0,0 +1,84 @@
...
+* Microsoft-style attributes interchangably use `CamelCase` and `_` separation,
+  but should prefer `CamelCase`
+* System Value semantics are case insensitive, should be specified `CamelCase`
+  and prefixed `SV_`.
+
+## Versioning
+
+All features should consider how users can adapt to the presence or absence of
+support. HLSL has two primary revisioning axis: language version and runtime
+version. How the two versions interact is not always a clean and easy line to
```
pow2clk, line 22: I can't decide if I mind that this mentions the guessable, but not immediately obvious connection between "runtime version" and "shader models". It's explained below and I see why a generic term was used, but perhaps something like "(e.g. shader models) might make it clearer sooner.

docs/DesignConsiderations.md
```diff
@@ -0,0 +1,84 @@
...
+* **`__HLSL_VERSION`** - Integer value for the HLSL language version. Unreleased
+  or experimental language versions are defined as a number larger than the
+  highest released version.
+* **`__SHADER_TARGET_STAGE`** - Integer value corresponding to the shader stage.
+  Shader stage. The shader stage values are exposed as
+  `__SHADER_STAGE_**STAGE**` (i.e. `__SHADER_STAGE_VERTEX`,
+  `__SHADER_STAGE_PIXEL`, ...)
+* **`__SHADER_TARGET_MAJOR`** - Major version for shader target (Vulkan or
+  Shader Model).
+* **`__SHADER_TARGET_MINOR`** - Minor version for shader target (Vulkan or
```
pow2clk, line 71: I don't think these are currently populated with the vulkan version, just the shader model.

docs/DesignConsiderations.md
```diff
@@ -0,0 +1,84 @@
...
+  Shader Model).
+
+If these macros are not sufficient for a given feature new macros or other
+mechanisms should be added as appropriate for the feature to enable developers
+to know if a given compiler release supports the required feature(s).
```
pow2clk, line 76: There are also DXC_VERSION_* macros that might serve as a way to determine if the compiler supports what is needed and the other macros fail.

docs/DesignConsiderations.md
```diff
@@ -0,0 +1,84 @@
...
+
+For features that are added retroactively to an already final runtime or
+language version, a `__has_feature` check should be added to allow the user to
+query for feature support in the preprocessor instead of forcing the user to
+check compiler versions explicitly. The Clang feature checking macros are
+documented
+[here](https://clang.llvm.org/docs/LanguageExtensions.html#feature-checking-macros).
+The `__has_feature` macro is known to work in DXC, and should be used.
```
pow2clk, line 84: That's interesting. Do we actually have it hooked up to any features?

docs/Process.md
```diff
@@ -31,15 +29,25 @@ further tweaked to align with the HLSL team's goals and priorities.
...
 
 By far the best way for an external contributor to propose a feature is through
-GitHub issues (See the section below on "Filing Issues"). If you can't be
-deterred from writing a proposal yourself you must find a member of the HLSL
-team to act as a _Sponsor_ for the change. The _Sponsor_ is responsible for
-tracking and helping change proposals through the proposal life cycle.
+GitHub issues (See the section below on "Filing Issues"). Issues in this
+repository will be used to publicly track feature requests for the HLSL language
+and HLSL runtime interfaces. Direct tooling feature requests to the
+[DirectXShaderCompiler](https://github.com/microsoft/DirectXShaderCompiler/issues/new).
```
pow2clk, line 35: "tooling feature requests" might benefit from some definition.

docs/Process.md
```diff
@@ -31,15 +29,25 @@ further tweaked to align with the HLSL team's goals and priorities.
...
+
+If you write a proposal yourself you must find a member of the HLSL team to act
+as a _Sponsor_ for the change. The _Sponsor_ is responsible for tracking and
+helping change proposals through the proposal life cycle. For Vulkan-specific
+features we require at least one _Sponsor_ from the HLSL team and one _Sponsor_
+from the Khronos Vulkan Working Group. If you need assistance finding a
+_Sponsor_ for a proposal reach out to the [HLSL Team](Contact.md).
```
pow2clk, line 42: That won't help to find a sponsor from Khronos. I don't know how that would be accomplished really.

docs/DesignConsiderations.md
llvm-beanz: I specifically chose not to use the term "shader models" because it is DX-specific, and doesn't capture the Vulkan version angle.

docs/DesignConsiderations.md
llvm-beanz: Eeek... that's unfortunate.

docs/DesignConsiderations.md
llvm-beanz: I left those off because I think we really shouldn't rely on the compiler version. Especially given how many places we have trouble bouncing between versions for the Preview. This year a feature is available in "2206", but not in "2208", and fixes in "1.7.2208" aren't present in "1.8.2206". So the version mapping is just all a mess.

docs/Process.md
llvm-beanz: The point of that is to reach out to us and we can help find someone (likely going through the Google team).

PR #53, opened 2023-05-05
Specify aliasing and address space for vk::BufferPointer

### LINKED ISSUE #42: [0010] What pointer annotations do we need to expose? (address space, aliasing)

Do we need to expose pointer annotations for address spaces and aliasing? If not what are the inferred address space and aliasing behaviors.

**Additional context**

greg-lunarg:
Fixes #42.

### DIFF
--- a/proposals/0010-vk-buffer-ref.md
+++ b/proposals/0010-vk-buffer-ref.md
@@ -171,7 +171,7 @@ An attribute vk::aliased_pointer can be attached to a variable, function paramet
 
 ### Buffer Pointers and Address Space
 
-All buffer pointers are presumed to point into the same address space. No new address space attributes are proposed.
+All buffer pointers are presumed to point into the host memory address space. No new address space attributes are proposed.
 
 ## SPIR-V Appendices
 
--- a/proposals/0010-vk-buffer-ref.md
+++ b/proposals/0010-vk-buffer-ref.md
@@ -165,7 +165,7 @@ Applying HLSL semantic annotations to objects of type vk::BufferPointer is disal
 
 ### Buffer Pointers and Aliasing
 
-All buffer pointers are considered restricted pointers. In other words, given two different objects of buffer pointer type, it is assumed that their pointees do not overlap by default.
+By default, buffer pointers are assumed to be restrict pointers as defined by the C99 standard.
 
 An attribute vk::aliased_pointer can be attached to a variable, function parameter or a block member of buffer pointer type. It is assumed that the pointee of an object with this attribute can overlap with the pointee of any other object with this attribute.
 
--- a/proposals/0010-vk-buffer-ref.md
+++ b/proposals/0010-vk-buffer-ref.md
@@ -163,6 +163,16 @@ A vk::BufferPointer can otherwise be used whereever the HLSL spec does not other
 
 Applying HLSL semantic annotations to objects of type vk::BufferPointer is disallowed.
 
+### Buffer Pointers and Aliasing
+
+All buffer pointers are considered restricted pointers. In other words, given two different objects of buffer pointer type, it is assumed that their pointees do not overlap by default.
+
+An attribute vk::aliased_pointer can be attached to a variable, function parameter or a block member of buffer pointer type. It is assumed that the pointee of an object with this attribute can overlap with the pointee of any other object with this attribute.
+
+### Buffer Pointers and Address Space
+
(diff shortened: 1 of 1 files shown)


proposals/0010-vk-buffer-ref.md
```diff
@@ -163,6 +163,16 @@ A vk::BufferPointer can otherwise be used whereever the HLSL spec does not other
 
 Applying HLSL semantic annotations to objects of type vk::BufferPointer is disallowed.
 
+### Buffer Pointers and Aliasing
+
+All buffer pointers are considered restricted pointers. In other words, given two different objects of buffer pointer type, it is assumed that their pointees do not overlap by default.
```
s-perron, line 168: > In other words, given two different objects of buffer pointer type, it is assumed that their pointees do not overlap by default.

This definition is too general. I think you need to define something like the "base", as you mentioned in https://github.com/microsoft/hlsl-specs/issues/42#issuecomment-1516924360. 

The [WGSL spec](https://gpuweb.github.io/gpuweb/wgsl/#alias-analysis) does this well. The call it the "root identifier" instead.

Don't try to stick too close to the spir-v aliasing rules. They need to be updated: https://gitlab.khronos.org/spirv/SPIR-V/-/issues/731

s-perron (changes requested):
I think this needs to be expanded a little. As stated, it would be nearly impossible to write a vaild program because any copy of a buffer pointer breaks the rules.

proposals/0010-vk-buffer-ref.md
s-perron: You might also simply reference the C standard for the definition of restrict to make it more precise.

greg-lunarg: Firstly, sorry for the delayed response. I will try to respond in a more timely fashion going forward.

> I think this needs to be expanded a little. As stated, it would be nearly impossible to write a valid program because any copy of a buffer pointer breaks the rules.

It is not having copies of pointers that breaks the rules. It is dereferencing the copies in a way that depends on them overlapping that breaks the rules.

So my question is: is there a useful graphics shader that cannot be written when disallowing aliasing between "different" pointer values?

greg-lunarg: If we decide that allowing aliasing between "different" pointer values is necessary, then I would allow aliasing only in the following cases:

1) Dereferences from the same named pointer variable or member of a named buffer.

2) Dereferences from a cast of the same named uint64_t variable or member of a named buffer to the same buffer pointer type.

2) Create an attribute vk::aliased which can be applied to a variable or member. Any dereferences from such pointers are assumed to be aliased.

s-perron: I think we are talking past each other. In general, I think we agree on what should happen, we just disagree that the current wording captures what you want. I find the single sentence

> In other words, given two different objects of buffer pointer type, it is assumed that their pointees do not overlap by default.

to simplistic to capture a concept that the C standard felt needed to be must more involved. See https://en.cppreference.com/w/c/language/restrict.

 I would suggest you ignore my first comment in this thread, and do the second one. Reference the C99 standard of the restrict type qualifier to define restrict.

Something like

> By default, buffer pointers are assumed to be restrict pointers as defined by the C99 standard.

should be good enough.

greg-lunarg:
@llvm-beanz Anything else needed here?

proposals/0010-vk-buffer-ref.md
```diff
@@ -163,6 +163,16 @@ A vk::BufferPointer can otherwise be used whereever the HLSL spec does not other
...
 
+### Buffer Pointers and Aliasing
+
+By default, buffer pointers are assumed to be restrict pointers as defined by the C99 standard.
+
+An attribute vk::aliased_pointer can be attached to a variable, function parameter or a block member of buffer pointer type. It is assumed that the pointee of an object with this attribute can overlap with the pointee of any other object with this attribute.
+
+### Buffer Pointers and Address Space
+
+All buffer pointers are presumed to point into the same address space. No new address space attributes are proposed.
```
llvm-beanz, line 174: Which address space? The default address space in HLSL is thread-local memory, which I assume isn't where these are expected to refer. I assume these are host memory, so we should state that explicitly.

greg-lunarg: Done.

llvm-beanz (approved):

PR #55, opened 2023-05-08
Add type punning rules to vk::BufferPointer

### LINKED ISSUE #43: [0010] Is type punning allowed?

With the planned introduction of [unions](proposals/0004-unions.md) to HLSL we need to know if `vk::BufferPointer<T,A>` objects are allowed in unions and if type punning is allowed.

**Additional context**
Related [unions issue](https://github.com/microsoft/hlsl-specs/issues/38).

greg-lunarg:
Fixes #43

### DIFF
--- a/proposals/0010-vk-buffer-ref.md
+++ b/proposals/0010-vk-buffer-ref.md
@@ -165,7 +165,7 @@ Applying HLSL semantic annotations to objects of type vk::BufferPointer is disal
 
 ### Buffer Pointers and Type Punning Through Unions
 
-While buffer pointer types are allowed in unions, type punning with buffer pointer types is disallowed.
+While buffer pointer types are allowed in unions, type punning with buffer pointer types is disallowed as it is with all other types in HLSL. Specifically, when a member of a union is defined, all other members become undefined, no matter the types.
 
 ## SPIR-V Appendices
 
--- a/proposals/0010-vk-buffer-ref.md
+++ b/proposals/0010-vk-buffer-ref.md
@@ -163,6 +163,10 @@ A vk::BufferPointer can otherwise be used whereever the HLSL spec does not other
 
 Applying HLSL semantic annotations to objects of type vk::BufferPointer is disallowed.
 
+### Buffer Pointers and Type Punning Through Unions
+
+While buffer pointer types are allowed in unions, type punning with buffer pointer types is disallowed.
+
 ## SPIR-V Appendices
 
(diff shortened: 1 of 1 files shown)


evaspitslines:
Is there currently already a PR that implements these functions? Or is it being worked on? 👀

greg-lunarg:
> Is there currently already a PR that implements these functions? Or is it being worked on?

There is a PR implementing something similar, but it was rejected because it did not follow the process being followed here.

An implementation of this proposal will likely be based on the previous PR.

s-perron (commented):
LGTM

proposals/0010-vk-buffer-ref.md
```diff
@@ -163,6 +163,10 @@ A vk::BufferPointer can otherwise be used whereever the HLSL spec does not other
 
 Applying HLSL semantic annotations to objects of type vk::BufferPointer is disallowed.
 
+### Buffer Pointers and Type Punning Through Unions
+
+While buffer pointer types are allowed in unions, type punning with buffer pointer types is disallowed.
```
llvm-beanz, line 168: I'm not sure this wording is sufficient.

Take this case:
```c++
Buffer<Int> I;

union Pain {
  vk::BufferPointer<SomeStruct> Ptr;
  uint64_t Int;
  double FP;
};

void Init(inout Pain P) {
  P.Int = I[0];
}

SomeStruct Load(Pain P) {
  return P.Ptr.Get();
}

void Fn() {
  Pain P;
  Init(P);
  SomeStruct S = Load(P);
}
```

The union `P` in `Fn` is initialized in `Init` to a `uint64_t`), then used as a `vk::BufferPointer`. What is the expected behavior here?

It can't be a compile error because `Init`, `Load` and `Fn` don't need to be defined in the same translation unit so we can't analyze that usage pattern. Is it undefined?

greg-lunarg: Yes, my intention is that this would be undefined behavior.

In general, type punning across members of a union is considered undefined behavior in C++. I don't see anything in the HLSL union proposal that changes this. So it seems that this should be true no matter the types of the members, including buffer pointers.

I will augment the text of this section to be more explicit about this.

llvm-beanz (approved):
I'm going to call this good enough. I don't think this is the right language or describes the behavior we want, but I think we can defer getting it correct until the Unions proposal actually defines how type punning works.

s-perron (approved):

PR #56, opened 2023-05-09
Add pseudo-class definition to vk::BufferPointer spec

### LINKED ISSUE #45: [0010] Include a pseudo-hlsl/c++ class declaration for `vk::BufferPointer<T,A>` and any associated built-in functions/types

For clarity and understanding of the spec it would be helpful to include a pseudo HLSL/C++ class declaration for `vk::BufferPointer<T,A>` in the spec and any reasonable pseudo-code for implementation. This both will help in the readability and understanding of the new objects and APIs and helps us move toward our goal of being able to define HLSL types in HLSL.

**Additional context**

greg-lunarg:
Fixes #45

### DIFF
--- a/proposals/0010-vk-buffer-ref.md
+++ b/proposals/0010-vk-buffer-ref.md
@@ -63,7 +63,7 @@ class vk::BufferPointer {
 }
 ```
 
-This class represents a pointer to a buffer of type struct `S`. `align` is the alignment in bytes of the pointer. If `align` is not specified, the alignment is assumed to be 16 bytes.
+This class represents a pointer to a buffer of type struct `S`. `align` is the alignment in bytes of the pointer. If `align` is not specified, the alignment is assumed to be alignof(S).
 
 This new type will have the following operations
 
--- a/proposals/0010-vk-buffer-ref.md
+++ b/proposals/0010-vk-buffer-ref.md
@@ -51,7 +51,19 @@ The goal of this proposal is to have a solution that meets the following require
 
 ## Proposed solution
 
-Our solution is to add a new builtin type in the vk namespace that is a pointer to a buffer of a given type, `vk::BufferPointer<T,A>`. The template argument `T` must be a struct. `A` must be an integer and is the alignment in bytes of the pointer. If `A` is not specified, the alignment is assumed to be 16 bytes.
+Our solution is to add a new builtin type in the vk namespace that is a pointer to a buffer of a given type:
+
+```c++
+template <struct S, int align>
+class vk::BufferPointer {
+    vk::BufferPointer(const vk::BufferPointer&);
+    vk::BufferPointer& operator=(const vk::BufferPointer&);
+    vk::BufferPointer(const uint64_t);
+    S& Get() const;
+}
+```
+
+This class represents a pointer to a buffer of type struct `S`. `align` is the alignment in bytes of the pointer. If `align` is not specified, the alignment is assumed to be 16 bytes.
 
 This new type will have the following operations
 


greg-lunarg:
IsNull() is not included as it was removed in #51.

s-perron:
@llvm-beanz Is this okay with you?

proposals/0010-vk-buffer-ref.md
```diff
@@ -51,7 +51,19 @@ The goal of this proposal is to have a solution that meets the following require
...
+template <struct S, int align>
+class vk::BufferPointer {
+    vk::BufferPointer(const vk::BufferPointer&);
+    vk::BufferPointer& operator=(const vk::BufferPointer&);
+    vk::BufferPointer(const uint64_t);
+    S& Get() const;
+}
+```
+
+This class represents a pointer to a buffer of type struct `S`. `align` is the alignment in bytes of the pointer. If `align` is not specified, the alignment is assumed to be 16 bytes.
```
llvm-beanz, line 66: Is the default alignment `16` or `max(16, alignof(S))`? Can we add that as the default value for the template parameter above?

greg-lunarg: Can a struct have an alignment greater than 16? I had assumed not, hence the default of 16.

proposals/0010-vk-buffer-ref.md
```diff
@@ -51,7 +51,19 @@ The goal of this proposal is to have a solution that meets the following require
 
 ## Proposed solution
 
-Our solution is to add a new builtin type in the vk namespace that is a pointer to a buffer of a given type, `vk::BufferPointer<T,A>`. The template argument `T` must be a struct. `A` must be an integer and is the alignment in bytes of the pointer. If `A` is not specified, the alignment is assumed to be 16 bytes.
+Our solution is to add a new builtin type in the vk namespace that is a pointer to a buffer of a given type:
+
+```c++
+template <struct S, int align>
+class vk::BufferPointer {
```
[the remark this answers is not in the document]
greg-lunarg: OK.

proposals/0010-vk-buffer-ref.md
llvm-beanz: I don't know that we enforce larger than 16-byte alignments on data types in HLSL right now, but certainly in C++ there are data types that require more restrictive alignments and there's no language-defined limit on alignment in HLSL.

For some C++ examples see: https://godbolt.org/z/YPh3Pfvr9

greg-lunarg: OK, will use max formula.

greg-lunarg: On further consideration, I am thinking that the default alignment should just be alignof(S). If you disagree, let me know.

proposals/0010-vk-buffer-ref.md
```diff
@@ -51,7 +51,19 @@ The goal of this proposal is to have a solution that meets the following require
...
+    vk::BufferPointer(const vk::BufferPointer&);
+    vk::BufferPointer& operator=(const vk::BufferPointer&);
+    vk::BufferPointer(const uint64_t);
```
devshgraphicsprogramming, line 61: I'm going to rain on your parade and tell you about Intel Gen12 and ARC GPUs

None of these GPUs report `shaderInt64`, which means you cant use `uint64_t` type in shaders.
Its a real "Intel" moment as according to @nanokatze the HW even supports 64bit atomics.

Its really weird that VK BDA nor GLSL Buffer Reference did not require `uint64` as a dependency, but anyway as per the GLSL_buffer_reference_spec
>     If neither 1GL_ARB_gpu_shader_int641 nor 1GL_EXT_shader_explicit_arithmetic_types_int641 is supported, then remove the constructors to convert to and from uint64_t.

In order to get this running on Arc GPUs you need a `uint2` version for each method that deals with `uint64_t`s.

llvm-beanz: I'm disinclined to say we should add a `uint2` variation of the methods. HLSL requires 64-bit integer data types for a bunch of features and we have for years expected that vendor backends either support 64-bit integer types natively or emulate them.

If we wanted to holistically support all of HLSL on targets where the vendor compiler can't support 64-bit types (through hardware or emulation), I think this would be one in a long list of things we'd need to fix.

devshgraphicsprogramming: You do realize this is Intel and its Windows drivers we're talking about?

devshgraphicsprogramming: You can see that Intel Gen11 (architecture released in 2019) and Gen12 don't report support for `shaderInt64` in Vulkan on Windows as recently as last month, but Gen9 did since forever and ARC has just started to.

Also half of Gen9 is officially designated End-of-Life since last december, so Gen11 will likely follow it into the grave before getting any 64bit integer support on Vulkan.

I'm just annoyed that for my graphics programming career spanning 14+ years, I've always had to deal with Intel's Windows OpenGL and Vulkan drivers not getting anywhere near the care and attention its Linux drivers get (let alone the Windows DirectX).

This is also the case for AMD and Raster-Ordered-Views done with fragment shader interlock in GL and VK, but thats just an extension without any dependents, not a semi-core feature you expect to rely on to implement other stuff.

Obviously the hilarious part of this all is that Intel knows how to emulate/implement `double` floating point because it has to in order to release an OpenGL 4.5 driver as GL 4.0 has the double floating point arithmetic extension in core.

Its not that Intel can't, its that Intel chooses not to and its not like people in the Vulkan ecosystem have Microsoft's power and can simply twist Intel's arm into implementing it, or simply do it for them by having the WDDM stack intervene in the middle and emulate it for them.

nanokatze: Actually I just checked, Intel implements shaderInt64 on newest HW as well since 101.4577 driver (implements vk 1.3.250)

llvm-beanz (approved):

s-perron (approved):

PR #59, opened 2023-06-13
Update to inline SPIR-V

s-perron:
Some features in the inline SPIR-V are hard to use, and some features are
missing. This proposal suggestes some changes that we can consider.

### DIFF
--- a/proposals/0011-inline-spirv.md
+++ b/proposals/0011-inline-spirv.md
@@ -85,9 +85,58 @@ Most SPIR-V extension so only certain types of changes to the SPIR-V sped:
 1. add a new storage class, or
 1. add a new builtin input.
 
-Most extension also define [capabilities](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Capability) that are used to enable the new features, but the capabilities does nothing on its own.
+Most extension also define [capabilities](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Capability) that are used to enable the new features, but the capabilities do nothing on their own.
+
+The proposed solution will add syntax to define each of these features aliased to a meaningful name, so that they can be defined in a header file. Then users will be able to use the header file, and use the feature by using the name. They will not have to see consider the inline SPIR-V. The proposed solution will also expect that the SPIR-V referenced is defined in the version of spirv-headers that was used to build DXC.
 
 ### Execution modes
+
+The existing `vk::ext_execution_mode` is a builtin function. I'm guessing it was defined as a builtin function because the original designers want to be able to add an execution mode that was not yet defined in spirv-headers. Since we are not considering that workflow anymore, I propose that we define a new attribute `vk::spvexecutionmode(ExecutionMode)`.
+
+Then suppose we wanted to implement the [SPV_KHR_post_depth_coverage](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_post_depth_coverage.html) extension in a header file.
+
+The header file could be something like 
+
+```
+// spv_khr_post_depth_coverage.h
+
+namespace spv {
+  namespace khr {
+    const uint32_t PostDepthCoverageExecutionMode = 4446;
+  }
+}
+```
+
+The a user could use the extension:
+
+```
+#include "spv_khr_post_depth_coverage.h"
+
+[[vk::spvexecutionmode(spv::khr::PostDepthCoverageExecutionMode)]]
+float4 PSMain(...) : SV_TARGET
+{ ...
+}
+```
+
+Another way of using it would be:
+
+```
+// spv_khr_post_depth_coverage.h
+
+#define PostDepthCoverageExecutionMode vk::spvexecutionmode(spv::khr::PostDepthCoverageExecutionMode)
+```
+
+Then the user could write:
+
+```
+#include "spv_khr_post_depth_coverage.h"
+
+[[PostDepthCoverageExecutionMode]]
+float4 PSMain(...) : SV_TARGET
+{ ...
+}
+```
+
 ### Instructions and extended instructions
 ### Types
 ### Decorations
--- a/proposals/0011-inline-spirv.md
+++ b/proposals/0011-inline-spirv.md
@@ -62,13 +62,6 @@ For example, `vk::ext_execution_mode` is a builtin function that
 [needs to be called from the entry point](https://github.com/microsoft/DirectXShaderCompiler/blob/adc0363539ef423ca3f6e9d0211a665756b81080/tools/clang/test/CodeGenSPIRV/spv.intrinsicExecutionMode.hlsl#L13)
 to which it applies. It would be much better as an attribute.
 
-The call to the execution mode builtin requires the enabling extension and
-capabilities to be added as attributes on the call. This seems unnecessary since
-that information should be available to the compiler through the spirv-headers
-JSON. This would be needed only if the spirv-header have not been updated to
-include the extension. We will need to decide if we want to support that
-workflow.
-
 The inline type definitions are also very involved. You can see the
 [sample](https://github.com/microsoft/DirectXShaderCompiler/blob/adc0363539ef423ca3f6e9d0211a665756b81080/tools/clang/test/CodeGenSPIRV/spv.intrinsicTypeInteger.hlsl)
 to get an idea of how hard it is to define a type.
@@ -78,8 +71,8 @@ shader stages. There are examples that show how to
 [declare a builtin variable](https://github.com/microsoft/DirectXShaderCompiler/blob/adc0363539ef423ca3f6e9d0211a665756b81080/tools/clang/test/CodeGenSPIRV/spv.intrinsicDecorate.hlsl)
 in a pixel shader. However, it
 [does not work in compute shaders](https://github.com/microsoft/DirectXShaderCompiler/issues/4217),
-or any other shader stage where there cannot be arbitrary inputs. This makes it
-impossible to implement many extensions.
+or any other shader stage where there cannot be arbitrary inputs. This makes
+many extensions impossible to implement.
 
 Common feedback on the feature is that is is
 ["a bit patchy, edge cases that don't work and that sort of thing"](https://github.com/microsoft/DirectXShaderCompiler/issues/5181#issuecomment-1537757720).
@@ -91,7 +84,8 @@ use cases that were considered at the time. Now that we have more experience, we
 should be able to improve on the implementation, while keeping the style
 similar.
 
-Most SPIR-V extension so only certain types of changes to the SPIR-V sped:
+Most SPIR-V extensions only introduce certain types of changes to the SPIR-V
+spec:
 
 1.  add a new
     [execution mode](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Execution_Mode),
@@ -99,12 +93,12 @@ Most SPIR-V extension so only certain types of changes to the SPIR-V sped:
 1.  add a new type,
 1.  add a new decoration,
 1.  add a new storage class, or
-1.  add a new builtin input.
+1.  add a new builtin.
 
-Most extension also define
+Most extensions also define
 [capabilities](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Capability)
-that are used to enable the new features, but the capabilities do nothing on
-their own.
+that are used to enable the new features, but the capabilities usually do
+nothing on their own.
 
 The proposed solution will add syntax to define each of these features aliased
 to a meaningful name, so that they can be defined in a header file. Then users
@@ -130,31 +124,10 @@ The header file could be something like
 ```
 // spv_khr_post_depth_coverage.h
 
-namespace spv {
-  namespace khr {
-    const uint32_t PostDepthCoverageExecutionMode = 4446;
-  }
-}
-```
-
-The a user could use the extension:
-
-```
-#include "spv_khr_post_depth_coverage.h"
-
-[[vk::spvexecutionmode(spv::khr::PostDepthCoverageExecutionMode)]]
-float4 PSMain(...) : SV_TARGET
-{ ...
-}
-```
-
-Another way of using it would be:
-
-```
-// spv_khr_post_depth_coverage.h
-
+// It would be nice to have this live in a namespace spv::khr::, but not possible with attributes.
+const uint32_t SampleMaskPostDepthCoverageCapabilityId = 4447;
 const uint32_t PostDepthCoverageExecutionModeId = 4446;
-#define PostDepthCoverageExecutionMode vk::spvexecutionmode(spv::khr::PostDepthCoverageExecutionModeId)
+#define SPV_KHR_PostDepthCoverageExecutionMode vk::ext_extension("SPV_KHR_post_depth_coverage"), vk::ext_capability(SampleMaskPostDepthCoverageCapabilityId), vk::spvexecutionmode(spv::khr::PostDepthCoverageExecutionModeId)
 ```
 
 Then the user could write:
@@ -162,7 +135,7 @@ Then the user could write:
 ```
 #include "spv_khr_post_depth_coverage.h"
 
-[[PostDepthCoverageExecutionMode]]
+[[SPV_KHR_PostDepthCoverageExecutionMode]]
 float4 PSMain(...) : SV_TARGET
 { ...
 }
@@ -192,7 +165,7 @@ Some instructions expect a pointer as an operand. This is why the
 would need this. I suggestion we deprecate it in
 [proposal 0006](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0006-reference-types.md).
 
-Other instruction require that a type be a literal. This is fairly common. The
+Other instructions require that a type be a literal. This is fairly common. The
 attribute `vk::ext_literal` was added to tell the compiler that is needs to add
 a literal value on the instruction instead of the result id of a load. This type
 of attribute would be meaningless in HLSL generally, and I will have to keep the
@@ -243,8 +216,8 @@ VmeImageINTEL<Texture2D> image;
 AvcMcePayloadINTEL payload;
 ```
 
-This would have the largest implementation cost, and for possibly little value.
-Not many extensions (TODO: Fill in actual numbers) create new types.
+We should consider the cost versus the value of adding this type. Just 3 out of
+111 spir-v extensions define a new type.
 
 ### Decorations
 
@@ -321,6 +294,10 @@ add it to the OpEntryPoint for the entry points from which it is reachable.
 The developer can set the builtin output by simply calling the function.
 
 ### Capability and extension instructions
+
+Existing inline spir-v has attributes to indicate that a capability or extension
+is needed by a particular part of the code. There are examples above, and there
+is no reason to change it.
 
 ## Detailed design
 
--- a/proposals/0011-inline-spirv.md
+++ b/proposals/0011-inline-spirv.md
@@ -11,31 +11,31 @@
 > When filling out the template below for a new feature proposal, please do the
 > following first:
 
-> 1. exclude the "Planned Version", "PRs" and "Issues" from the header.
-> 2. Do not spend time writing the "Detailed design" until the feature has been
->    merged in the "Under Consideration" phase.
-> 3. Delete this Instructions section including the line below.
+> 1.  exclude the "Planned Version", "PRs" and "Issues" from the header.
+> 2.  Do not spend time writing the "Detailed design" until the feature has been
+>     merged in the "Under Consideration" phase.
+> 3.  Delete this Instructions section including the line below.
 
----
+--------------------------------------------------------------------------------
 
-* Proposal: [0011](0011-inline-spirv.md)
-* Author(s): [Steven Perron](https://github.com/s-perron)
-* Sponsor: TBD
-* Status: **Under Consideration**
+*   Proposal: [0011](0011-inline-spirv.md)
+*   Author(s): [Steven Perron](https://github.com/s-perron)
+*   Sponsor: TBD
+*   Status: **Under Consideration**
 
 *During the review process, add the following fields as needed:*
 
-* Planned Version: 20YY
-* PRs: [#NNNN](https://github.com/microsoft/DirectXShaderCompiler/pull/NNNN)
-* Issues:
-  [#NNNN](https://github.com/microsoft/DirectXShaderCompiler/issues/NNNN)
+*   Planned Version: 20YY
+*   PRs: [#NNNN](https://github.com/microsoft/DirectXShaderCompiler/pull/NNNN)
+*   Issues:
+    [#NNNN](https://github.com/microsoft/DirectXShaderCompiler/issues/NNNN)
 
 ## Introduction
 
 Define a set of Vulkan specific builtin functions, types, and attributes that
-will allow vendors to implement their extensions. The vendors should be able
-to write a self-contained header file that users can use without
-knowing about inline SPIR-V.
+will allow vendors to implement their extensions. The vendors should be able to
+write a self-contained header file that users can use without knowing about
+inline SPIR-V.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+Define a set of Vulkan specific builtin functions, types, and attributes that
+will allow vendors to implement their extensions. The vendors should be able to
+write a self-contained header file that users can use without knowing about
+inline SPIR-V.
+
+## Motivation
+
+In Vulkan, there is a culture of allowing a vendor to define extensions that are
+only interesting to them. From the perspective of a driver this is acceptable
+because company B does not have to support the extension put out by vendor B, if
```
danginsburg, line 44: "company A does not have to support the extension put out by vendor B"?

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+generally requires a code change to the compiler that must be maintained.
+
+If those extensions could be defined as a header file that DXC consumes, then a
+vendor would not have to modify the compiler other than possibly updating the
+submodules.
+
+The existing
+[inline SPIR-V](https://github.com/microsoft/DirectXShaderCompiler/wiki/GL_EXT_spirv_intrinsics-for-SPIR-V-code-gen)
+was the first attempt at solving this problem. However, the features are not
+easy to use, the documentation is incorrect, and is does not interact with the
```
danginsburg, line 58: "and it"?

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+{ ...
+}
+```
+
+### Instructions and extended instructions
+
+The existing `vk::ext_instruction` attribute is used to define a function as a
+specific spir-v instruction. Suppose we wanted to implement the
+[SPV_INTEL_subgroups](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/INTEL/SPV_INTEL_subgroups.html)
+extension in a header file. It contains a 8 new instructions. Each one could be
```
danginsburg, line 149: "contains 8 new instructions"

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+```
+
+Then the user could simply use the types:
+
+```
+VmeImageINTEL<Texture2D> image;
+AvcMcePayloadINTEL payload;
+```
+
+We should consider the cost versus the value of adding this type. Just 3 out of
```
danginsburg, line 219: I am in favor of adding the types with the syntax you proposed.

danginsburg (approved):
I strongly support this proposal, thank you for putting it together.

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+a literal value on the instruction instead of the result id of a load. This type
+of attribute would be meaningless in HLSL generally, and I will have to keep the
+attribute.
+
+I believe this part of the existing inline SPIR-V works well, and should be
+kept. If we feel strongly about naming conventions, we could change the names.
+
+### Types
+
+Some extensions introduce new types. Very few extension add new types. It is
```
dgkoch, line 179:
```suggestion
Some extensions introduce new types. Very few extensions add new types. It is
```

proposals/0011-inline-spirv.md
dgkoch:
```suggestion
extension in a header file. It contains 8 new instructions. Each one could be
```

proposals/0011-inline-spirv.md
dgkoch:
```suggestion
because company A does not have to support the extension put out by vendor B, if
```

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+variable. Conceptually this is similar to setting an address space. I would not
+want to add anything new, but I suggest we deprecate this when address spaces
+are added to HLSL more generally.
+
+### Builtin input
+
+The existing inline spir-v has limited support for adding a builtin. There is a
+sample that shows how to
+[add a builtin input to a pixel shader](https://github.com/microsoft/DirectXShaderCompiler/blob/128b6fd16b449df696a5c9f9405982903a4f88c4/tools/clang/test/CodeGenSPIRV/spv.intrinsicDecorate.hlsl).
+This work for pixel shaders because the parameters to a pixel shader can have
```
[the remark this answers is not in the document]
dgkoch:
```suggestion
This works for pixel shaders because the parameters to a pixel shader can have
```

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+### Storage classes
+
+The existing inline spir-v allows the developer to set the storage class for a
+variable. Conceptually this is similar to setting an address space. I would not
+want to add anything new, but I suggest we deprecate this when address spaces
+are added to HLSL more generally.
+
+### Builtin input
+
+The existing inline spir-v has limited support for adding a builtin. There is a
```
dgkoch, line 237:
```suggestion
The existing inline spir-v has limited support for adding a builtin input. There is a
```

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+Then the compiler will be able to add a variable to in the input storage class,
+with the builtin decoration, with a type that is the same as the return type of
+the function, and add it to the OpEntryPoint for the entry points from which it
+is reachable.
+
+The developer can use the builtin input by simply calling the function.
+
+### Builtin output
+
+The existing inline spir-v has limited support for adding a builtin. This would
```
dgkoch, line 272:
```suggestion
The existing inline spir-v has limited support for adding a builtin output. This would
```

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+*   Does this change require DXIL changes?
+*   Can it be CodeGen'd to SPIR-V?
+
+## Alternatives considered (Optional)
+
+### Auto-generating builtin functions and attributes from [spirv.core.grammar.json](https://github.com/KhronosGroup/SPIRV-Headers/blob/8e2ad27488ed2f87c068c01a8f5e8979f7086405/include/spirv/unified1/spirv.core.grammar.json)
+
+When we considered auto-generating, we noticed that not enough information is
+available. For example, it is not possible to tell if a builtin should be an
+input or output, and we cannot know it types. The other problem is that it will
```
dgkoch, line 323:
```suggestion
input or output, and we cannot know it's types. The other problem is that it will
```

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+```
+template<typename ImageType>
+typedef VmeImageINTEL vk::SpirvType</* OpTypeVmeImageINTEL */ 5700, Imagetype>
+typedef AvcMcePayloadINTEL vk::SprivType</* OpTypeAvcMcePayloadINTEL */ 5704>
+```
+
+Then the user could simply use the types:
+
+```
+VmeImageINTEL<Texture2D> image;
```
Keenuts, line 215: Are you planning on allowing those to interact with current annotations?
(If we have some vendor-specific `Texture4D` type by ex, which needs some `vk::image_format`)
Do we say "if the type is using inline spir-v, then then cannot mix with other attributes? (seems the simplest way, but might need to be specified?)

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+definition for the function. When the user calls the function, it will set the
+value of the output to `val`.
+
+For example, we could declare the `gl_FragStencilRefARB` builtin in a header
+file like this:
+
+```
+[[vk::ext_extension("SPV_EXT_shader_stencil_export")]]
+[[vk::ext_builtin_input(/* FragStencilRefEXT */ 5014)]]
+void gl_FragStencilRefARB(int);
```
Keenuts, line 287: Why is that needed for built-ins to be used like functions?
Would it be possible to declare the builtin-type like the other types (if needed),
and have a global variable with the buildin annotation?
Or is declaring such global an issue for implementation?

Keenuts (commented):
Some high-level questions, there are probably technical reasons for those decisions, just curious 😊

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+
+Finally, adding decorations to declare builtin inputs does not work in all
+shader stages. There are examples that show how to
+[declare a builtin variable](https://github.com/microsoft/DirectXShaderCompiler/blob/adc0363539ef423ca3f6e9d0211a665756b81080/tools/clang/test/CodeGenSPIRV/spv.intrinsicDecorate.hlsl)
+in a pixel shader. However, it
+[does not work in compute shaders](https://github.com/microsoft/DirectXShaderCompiler/issues/4217),
+or any other shader stage where there cannot be arbitrary inputs. This makes
+many extensions impossible to implement.
+
+Common feedback on the feature is that is is
```
gnl21, line 77:
```suggestion
Common feedback on the feature is that it is
```

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+
+The existing inline spir-v has limited support for adding a builtin. There is a
+sample that shows how to
+[add a builtin input to a pixel shader](https://github.com/microsoft/DirectXShaderCompiler/blob/128b6fd16b449df696a5c9f9405982903a4f88c4/tools/clang/test/CodeGenSPIRV/spv.intrinsicDecorate.hlsl).
+This work for pixel shaders because the parameters to a pixel shader can have
+arbitrary semantics. However, when we quickly ran into problems when
+[trying to declare a spir-v specific builtin in a compute shader](https://github.com/microsoft/DirectXShaderCompiler/issues/4217).
+
+The reason that a builtin input requires more than just a decoration on the
+variable. It must be in the input storage class, and get added to the
```
gnl21, line 245: This first sentence seems to be missing something. Maybe it should be just:
```suggestion
A builtin input requires more than just a decoration on the
variable. It must be in the input storage class, and get added to the
```

proposals/0011-inline-spirv.md
gnl21:
```suggestion
input or output, and we cannot know its types. The other problem is that it will
```

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+const uint32_t PostDepthCoverageExecutionModeId = 4446;
+#define SPV_KHR_PostDepthCoverageExecutionMode vk::ext_extension("SPV_KHR_post_depth_coverage"), vk::ext_capability(SampleMaskPostDepthCoverageCapabilityId), vk::spvexecutionmode(spv::khr::PostDepthCoverageExecutionModeId)
+```
+
+Then the user could write:
+
+```
+#include "spv_khr_post_depth_coverage.h"
+
+[[SPV_KHR_PostDepthCoverageExecutionMode]]
```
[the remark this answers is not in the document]
s-perron: This is a attribute that is connected to the function on the next line, so it will not apply to any other entry point.

proposals/0011-inline-spirv.md
s-perron: If they are using inline spir-v, then have the ability to set all of the parameters like image format in the template of `vk::SpirvType` and they do not need `vk::image_format`. However, if the attribute is a decoration that should apply to the type, then it could still apply. I do not see any good example now, but I don't want to rule it out.

proposals/0011-inline-spirv.md
s-perron: That is a good question. That idea could work. The syntax would be:

```
[[vk::ext_extension("SPV_EXT_shader_stencil_export")]]
[[vk::ext_builtin_output(/* FragStencilRefEXT */ 5014)]]
int gl_FragStencilRefARB;
```

Then `gl_FragStencilRefARB` could be used like any other global variable. I think that could work. I'll wait for more feedback before I update the proposal.

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+
+Most extensions also define
+[capabilities](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Capability)
+that are used to enable the new features, but the capabilities usually do
+nothing on their own.
+
+The proposed solution will add syntax to define each of these features aliased
+to a meaningful name, so that they can be defined in a header file. Then users
+will be able to use the header file, and use the feature by using the name. They
+will not have to see consider the inline SPIR-V. The proposed solution will also
```
janharaldfredriksen-arm, line 106:
```suggestion
will not have to see the inline SPIR-V. The proposed solution will also
```
(or leave "consider")

llvm-beanz:
@s-perron, sorry, we’ve had a couple busy weeks and the holiday last week had lots of people out. I’ve assigned this to myself, @tex3d, and @pow2clk so that it shows up on our work queues.

I had some comments that I had started to put together (nothing big), and I’ll get those posted today or tomorrow and we can aim to get this merged soon.

llvm-beanz (commented):
Mostly nit-picky language comments.

I should add some documentation about this but I strongly recommend using a language linter to help avoid passive language, and to avoid first-person or wishy-washy language.

I use [write-good](https://github.com/btford/write-good) on all my technical writing as a good starting point. It has plugins for lots of common text editors.

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+extension in a header file. It contains 8 new instructions. Each one could be
+defined in the header file as a function, and then the function can be called by
+the users. For example,
+
+```
+[[vk::ext_capability(5568)]]
+[[vk::ext_extension("SPV_INTEL_subgroups")]]
+[[vk::ext_instruction(/* OpSubgroupShuffleINTEL */ 5571)]]
+template<typename T>
+T SubgroupShuffleINTEL(T data, uint32 invocationId);
```
mklefrancois, line 158: Allowing 'template' would be fantastic; as another example, it would allow the implementation of [hitObjectTraceRayNV](https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/NV/SPV_NV_shader_invocation_reorder.html) for Shader-Execution-Reorder in a header.  Currently, if we attempt to do the following, we receive the error "G53006855: an attribute list cannot appear here," and the payload's explicit structure needs to be defined.

```
[[vk::ext_instruction(OpHitObjectTraceRayNV)]]
template<typename T>
void hitObjectTraceRayNV(
    [[vk::ext_reference]] HitObjectNV hitObject,
    RaytracingAccelerationStructure as,
    uint RayFlags,
    uint CullMask,
    uint SBTOffset,
    uint SBTStride,
    uint MissIndex,
    float3 RayOrigin,
    float RayTmin,
    float3 RayDirection,
    float RayTMax,
    [[vk::ext_reference]] [[vk::ext_storage_class(RayPayloadKHR)]] T payload
  );
```

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,347 @@
...
+### Builtin output
+
+The existing inline spir-v has limited support for adding a builtin output. This would
+have most of the same problems as declaring a builtin input.
+
+Nine of the 111 extensions add new builtin outputs. It would be good to have a
+clean way to do this. I'm proposing a new function attribute
+`vk::ext_builtin_ouput(val)` that, like `vk::ext_instruction`, will imply a
+definition for the function. When the user calls the function, it will set the
+value of the output to `val`.
```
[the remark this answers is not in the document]
s-perron: Much better.

proposals/0011-inline-spirv.md
s-perron: I will open a follow up issue to consider this. I'll look at implementation issue to determine which is easiest. I'm conderned that there is a lot of code that handled variables based on their representation in the AST, and I'll have to special case it to look for the attribute.

s-perron:
@llvm-beanz Ping.

llvm-beanz (commented):
Two last little things from me:
- [x] File an issue on exploring better representation of types in the source.
- [x] Put yourself and myself as "Sponsors".

s-perron:
Both are done. I opened https://github.com/microsoft/hlsl-specs/issues/76 to follow up on Nathan's idea.

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,293 @@
...
+
+The header file could create a partial instantiation with a more meaningful
+name. For example, if you wanted to declare the types from the
+[SPV_INTEL_device_side_avc_motion_estimation](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/INTEL/SPV_INTEL_device_side_avc_motion_estimation.html)
+you could have
+
+```
+template<typename ImageType>
+typedef VmeImageINTEL vk::SpirvType</* OpTypeVmeImageINTEL */ 5700, Imagetype>
+typedef AvcMcePayloadINTEL vk::SprivType</* OpTypeAvcMcePayloadINTEL */ 5704>
```
bogner, line 173: This syntax doesn't really make sense. Probably better to use type alias syntax, given that it's available in HLSL2021, ie
```c++
template <typename ImageType>
using VmeImageINTEL = vk::SpirvType</* OpTypeVmeImageINTEL */ 5700, ImageType>;
```

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,293 @@
...
+adding a capability and annotation.
+
+If we can hide all of the cases that could be implicitly added in a header file,
+then there is very little burden on users. This is why we chose to not have
+implicit inclusion of the capabilities and extensions.
+
+## Acknowledgments (Optional)
+
+Take a moment to acknowledge the contributions of people other than the author
+and sponsor.
```
bogner, line 291: I think this section can be removed instead of leaving the boilerplate around if there aren't any specific acknowledgements

proposals/0011-inline-spirv.md
devshgraphicsprogramming: +1 on this, would be amazing, also the templating is very neat

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,293 @@
...
+typedef VmeImageINTEL vk::SpirvType</* OpTypeVmeImageINTEL */ 5700, Imagetype>
+typedef AvcMcePayloadINTEL vk::SprivType</* OpTypeAvcMcePayloadINTEL */ 5704>
+```
+
+Then the user could simply use the types:
+
+```
+VmeImageINTEL<Texture2D> image;
+AvcMcePayloadINTEL payload;
+```
```
devshgraphicsprogramming, line 181: also make sure that `vk::result_id<vk::SpirvType<...> >` is allowed, its currently breaking my `OpConvertUToPtr`

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,293 @@
...
+    declare an object of that type. This means the same function cannot be used
+    to declare two different types. Also, if there are multiple header files,
+    they cannot use the same id for two different types. This can become hard to
+    manage.
+
+This proposal deprecates the old mechanism, and replaces it with a new type
+`vk::SpirvType<int OpCode, ...>`. The template on the type contains the opcode
+and all of the parameters necessary for that opcode. The difficulty with this is
+that the operands are not just literal integer values. Sometimes they are
+another type.
```
devshgraphicsprogramming, line 163: As it stands its with the old mechanism impossible to for example, delare a pointer to a struct defined in HLSL. So I could in-fact do this?
```cpp
template<typename T>
using ptr_t = vk::SpirvType</*OpTypePointer*/ 32, /*PhysicalStorageBuffer*/ 5349, T>;
```
instead of what I have right now, which is
```cpp
// 15 just happens to guess typeid(MyStruct) by accident, also its only ever defined if its used somewhere else outside BDA
#define typeid_MyStruct 15

// I need to call it myself in a way reachable from the entry point which is mildly annoying
[[vk::ext_type_def(60,/*SpvOpTypePointer*/ 32)]]
void createPtrType([[vk::ext_literal]] int storageClass, [[vk::ext_literal]] int type);

// Shouldn't I be doing the rest with?
typedef /*vk::ext_result_id<*/vk::ext_type<60> /*>*/ imm_p_MyStruct;

[[vk::ext_capability(/*SpvPhysicalStorageBufferAddresses*/ 5347)]]
[[vk::ext_instruction(/*SpvOpConvertUToPtr*/ 120)]]
imm_p_MyStruct ConvertUToPtr(uint64_t src);

...

void main()
{
    nbl::hlsl::spirv::impl::createPtrType(5349,typeid_MyStruct);

...
}
```

devshgraphicsprogramming:
let me show you a vaguely "nuts" usage of SPIR-V intrinsics
https://godbolt.org/z/EEoebnTPK

As you can see I'm a bit impatient for proposals 0006 and 0010 to get ironed out and decided, so I want to make a bit of my own throw away code.

```cpp
struct MyStruct{
    uint32_t a;
    uint32_t b;
    uint32_t c;
};
// 15 just happens to guess typeid(MyStruct) by accident, also its only ever defined if its used somewhere else outside BDA
#define typeid_MyStruct 15

namespace nbl
{
namespace hlsl
{

namespace spirv
{

namespace impl
{

// I need to call it myself in a way reachable from the entry point which is mildly annoying
[[vk::ext_type_def(60,/*SpvOpTypePointer*/ 32)]]
void createPtrType([[vk::ext_literal]] int storageClass, [[vk::ext_literal]] int type);

// Shouldn't I be doing the rest with?
typedef /*vk::ext_result_id<*/vk::ext_type<60> /*>*/ imm_p_MyStruct;

[[vk::ext_capability(/*SpvPhysicalStorageBufferAddresses*/ 5347)]]
[[vk::ext_instruction(/*SpvOpConvertUToPtr*/ 120)]]
imm_p_MyStruct ConvertUToPtr(uint64_t src);

template<typename T>
[[vk::ext_instruction(/*SpvOpLoad*/ 61)]]
T bda_load(imm_p_MyStruct ptr, [[vk::ext_literal]] int memoryOperands);

template<typename T>
[[vk::ext_instruction(/*SpvOpStore*/ 62)]]
void bda_store([[vk::ext_reference]] T val, imm_p_MyStruct ptr, [[vk::ext_literal]] int memoryOperands);

}

template<typename T>
T bda_load(uint64_t addr)
{
    nbl::hlsl::spirv::impl::createPtrType(5349,typeid_MyStruct);

    return impl::bda_load<T>(impl::ConvertUToPtr(addr),/*None*/0x1);
}

template<typename T>
void bda_store(inout T val, uint64_t addr)
{
    nbl::hlsl::spirv::impl::createPtrType(5349,typeid_MyStruct);

    impl::bda_store<T>(val,impl::ConvertUToPtr(addr),/*None*/0x1);
}

}

template<typename T>
struct ssbo_pointer_t
{
    T deref()
    {
        return spirv::bda_load(value);
    }
    ssbo_pointer_t<T> operator+(const int64_t incr)
    {
        ssbo_pointer_t<T> retval = {value+sizeof(T)*incr};
        return retval;
    }

    uint64_t value;
};

template<typename T, typename U>
ssbo_pointer_t<U> _reinterpret_cast(ssbo_pointer_t<T> ptr)
{
    ssbo_pointer_t<U> retval = {ptr.value};
    return retval;
}

//what can you do when you can't overload operator-> (again not really appropriate, cause its not deref)
//#define MEMBER(t,m) nbl::hlsl::_reinterpret_cast<decltype(decltype(t)::m)>(t.value+offsetof(decltype(t),m))
// Another way to do it is operator-> an `ssbo_reference_t` struct and more intrinsics to construct a whole OpAccessChain via spawning `ssbo_reference_t`

}
}

struct PushConsts {
  nbl::hlsl::ssbo_pointer_t<MyStruct> addr;
  MyStruct dummy;
};
[[vk::push_constant]] PushConsts pushConsts;

[shader("compute")]
[numthreads(1, 1, 1)]
void test( uint3 DTid : SV_DispatchThreadID )
{    
    MyStruct mystruct = nbl::hlsl::spirv::bda_load<MyStruct>(pushConsts.addr.value+4);
    mystruct.a = 6;
    nbl::hlsl::spirv::bda_store(mystruct,pushConsts.addr.value);

//    nbl::hlsl::ssbo_pointer_t<uint> p_a = MEMBER(pushConsts.addr.value,a);
}
```

if you want to make SPIR-V intrinsics more useful, try to get it to a place where things like this are possible.

I believe everything I need for this (Except for `offsetof` and `decltype` I only need for member access) is already throught about (templated custom SPIR-V types from HLSL user-types, no weird function calls to declare types) in this PR. But sharing anyway in-case I'm wrong.

proposals/0011-inline-spirv.md
s-perron: vk::SpirvType<...> will be a type like any other, it just has no public members. You will be able to use it anywhere a type can be used.

proposals/0011-inline-spirv.md
```diff
@@ -0,0 +1,293 @@
...
+
+```
+[[vk::ext_builtin_input(/* NumWorkgroups */ 24)]]
+uint32 gl_NumWorkGroups();
+```
+
+Then the compiler will be able to add a variable to in the input storage class,
+with the builtin decoration, with a type that is the same as the return type of
+the function, and add it to the OpEntryPoint for the entry points from which it
+is reachable.
```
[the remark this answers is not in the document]
bogner: Yes, that language makes it clearer. Thanks for the explanation.

bogner (approved):

s-perron:
@llvm-beanz Anything left?

PR #61, opened 2023-06-22
[SM68] Initial proposal for Wave Matrix

llvm-beanz:
This proposal adds new WaveMatrix data types that support lane cooperative matrix multiplication.

A preview implementation of this proposal is available as part of the DXC 1.8.2306 preview release.


proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+the `detail` namespace are not exposed in the HLSL runtime. They are provided
+here to make the specification more concise to consume. Objects in no namespace
+are exposed as public interfaces.
+
+### WaveMatrix Fill
+
+All WaveMatrix objects have a `Fill` method of the form `void Fill(ElTy Value)`
+where `ElTy` is the element type.
+
+The `Fill` method fills the matrix or matrix fragment with the provided value.
```
pow2clk, line 86: I think a more technical description would be "Assigns the given `Value` to every element in the matrix or matrix fragment".

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+All wave threads must provide the same value or the result is undefined. All
+WaveMatrix objects have the same `Fill` method with the same behavior.
+
+### WaveMatrix Matrix Objects
```
pow2clk, line 90: I feel like the explanation of `Fill` above would be a bit more logical after this section

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+
+The code below approximately declares the base interface that WaveMatrix matrix
+objects implement.
```
pow2clk, line 93: Maybe add "the following sections will explain in detail what these methods do and what the parameters represent."

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+The `Alignment` must be at least `sizeof(ElTy)` otherwise the load behavior is
+undefined.
+
+Loading from or storing to a `groupshared` array takes the starting index, and
+the row stride as a number of elements.
+
+##### Orientation
+
+When loading and storing WaveMatrix matrices a boolean parameter is provided to
+indicate if the matrix being loaded or stored is column major.
```
pow2clk, line 134: Maybe a brief explanation of what column major means is in order. I think we should say that, when false, row-major orientation is indicated

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+
+Matrices may be stored in row or column layout. Matrices are always loaded into
+row-major orientation in memory for WaveMatrix objects. The `Load` and `Store`
+methods perform matrix transposition when loading or storing column major
+matrices.
+
+##### Stride
+
+When loading and storing from `groupshared` arrays, the stride is expressed in
+number of elements.
```
pow2clk, line 144: As above, perhaps a quick definition of what the stride means in this context in addition to the units for the matrix types

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+
+When loading and storing from `[RW]ByteAddressBuffer` types, the stride is
+expressed in bytes. The row stride must be a multiple of the size of the
+element, and greater than or equal to the size of the element multiplied by the
```
pow2clk, line 148: I don't think a comma is needed here since there are just two operands to "and"

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+
+template <typename ElTy, int NRows, int NCols>
+class WaveMatrixRight : detail::WaveMatrixBase<ElTy, NRows, NCols> {
+  uint MatrixDepth();
+};
+```
+
+`ElTy` must be either a 32 or 16 bit floating point type or an 8 bit packed
+signed or unsigned integer type. `NRows` and `NCols` must be compile-time
+constant expressions.
```
pow2clk, line 169: It is fairly obvious, but perhaps stating that NRows is the number of rows and NCols the number of columns.

I think we should explain what Left and Right matrices are and how they can be used

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+number of elements.
+
+When loading and storing from `[RW]ByteAddressBuffer` types, the stride is
+expressed in bytes. The row stride must be a multiple of the size of the
+element, and greater than or equal to the size of the element multiplied by the
+number of elements per row. Any value below the minimum legal values are
+ignored. The behavior of row stride values that are not a multiple of element
+stride is undefined.
+
+#### WaveMatrix Left & Right
```
pow2clk, line 153: Text similar to the above "The code below approximately declares the base interface that WaveMatrix matrix
objects implement." would make clear the intent of this section

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+```
+
+`ElTy` must be either a 32 or 16 bit floating point type or an 8 bit packed
+signed or unsigned integer type. `NRows` and `NCols` must be compile-time
+constant expressions.
+ 
+##### WaveMatrix(Left|Right) MatrixDepth
+
+The `MatrixDepth` method returns the hardware-dependent depth for the matrix
+multiplication unit. The resulting value must be an even multiple of 16.
```
pow2clk, line 174: I think you're using "even" here to indicate that there is no remainder. I'm not sure it's necessary and it might be interpreted as the number itself will not be odd. Obviously a multiple of 16 won't be, but still it's a bit ambiguous initially.

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+
+### WaveMatrix Matrix Fragment Objects
+
+The code below approximately declares the base interface that all WaveMatrix
+fragment objects implement. A WaveMatrix fragment object stores a single row or
+column of a WaveMatrix matrix.
```
pow2clk, line 180: Does what it represents depend on whether the matrix or row or column major?

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+
+```c++
+namespace detail {
+template <typename ElTy, int NRows, int NCols>
+class WaveMatrixFragmentBase : WaveMatrixBase<ElTy, NRows, NCols> {
```
pow2clk, line 185: The inheritance here would seem to suggest that redeclaring these methods isn't strictly necessary. I agree that repeating it makes things clearer, but then I don't know if the inheritance adds anything and it makes the reader check above to see if there are any differences. I did exactly that briefly and saw none. Are there any?

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+
+Loading from or storing to a `groupshared` array takes the starting index, and
+the element stride as a number of elements.
+
+##### Stride
+
+When loading and storing from `groupshared` arrays, the stride is expressed in
+number of elements.
+
+When loading and storing from `[RW]ByteAddressBuffer` types the stride must be
```
pow2clk, line 220: This lacks the clarification that these are specified in bytes

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+greater than or equal to the size of the element type. Any value below the
+minimum legal values are ignored. The behavior of row stride values that are not
+a multiple of element stride is undefined.
+
+### WaveMatrix Accumulator Objects
+
+The WaveMatrix Accumulator objects come in three forms which are represented by
+two categories: matrix accumulators and fragment accumulators. All accumulators
```
pow2clk, line 228: I found "which are represented by" to be confusing because there are 3 of one and 2 of another. Would "fall into two categories" be equally accurate?

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+                     std::true_type>;
+
+template <typename T> struct is_8bit_packed_int_type = std::false_type;
+
+template <typename T>
+struct is_32bit_int_type = std::enable_if_t<(std::is_same<T, int32_t>::value ||
+                                             std::is_same<T, uint32_t>::value),
+                                            std::true_type>;
+
+template <typename T> struct is_32bit_int_type = std::false_type;
```
pow2clk, line 265: I'm afraid I'm not familiar with this feature of C++ and it looks like you're redeclaring variables. Am I exceptionally ignorant of this feature (possible) or is it going to confuse others?

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+  MultiplyAccumulate(WaveMatrixLeft<MatElTy1, NRows, NCols> LHS,
+                     WaveMatrixRight<MatElTy2, NRows, NCols> RHS);
+};
+
+```
+
+#### Broadcast Add
+
+The `WaveMatrixAccumulator::Add` methods perform an element-wise addition of an
+accumulator matrix and the provided matrix or fragment accumulator.
```
pow2clk, line 302: The last "accumulator" seems out of place to me. Is the intent to say that it adds an accumulator matrix (which I think I know what is) to a matrix accumulator or a [matrix] fragment accumulator? I'm not sure it's clear what those are and the terms are kind of ambiguous with the accumulator matrix term

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+
+If the argument is a fragment object, the `Add` method broadcasts the argument
+accumulator up from an Mx1 or 1xN matrix to an MxN matrix then performs
+element-wise addition.
```
pow2clk, line 306: I think this is what I tend to call a "splat". Is it a conversion to a full matrix that copies the values of the fragment into a temporary full matrix that gets accumulated? How do the values get selected? I think I can guess, but I think it should be explicit.

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+
+#### WaveMatrixAccumulator::Multiply
+
+The `WaveMatrixAccumulator::Multiply` method performs multiplication of the left
+and right arguments and stores the result back into the `WaveMatrixAccumulator`.
```
pow2clk, line 311: I know you specify that it "stores" the result, but given that this is the only one that doesn't accumulate (right?) maybe we should call out explicitly that it overwrites the contents of the accumulator

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+This is a wave-level operation and cannot be used inside divergent control flow.
```
pow2clk, line 312: If it is, is the result undefined? 

Longer term goal, but perhaps we should crib from other specs' formal definitions of "should", "must" etc.

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+type unless the accumulator is a signed or unsigned 32-bit integer type. For
+signed or unsigned 32-bit integer accumulators, signed or unsigned 8-bit packed
+integers are also supported and can be mixed interchangeably.
+
+#### WaveMatrixAccumulator::MultiplyAccumulate
+
+The `WaveMatrixAccumulator::MultiplyAccumulate` method performs multiplication
+of the left and right arguments and adds the result back into the
+`WaveMatrixAccumulator`. This is a wave-level operation and cannot be used
+inside divergent control flow.
```
pow2clk, line 324: as above, what if I do?

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+accumulator type unless the accumulator is a signed or unsigned 32-bit integer
+type. For signed or unsigned 32-bit integer accumulators, signed or unsigned
+8-bit packed integers are also supported and can be mixed interchangeably.
+
+### WaveMatrix Fragment Accumulators
+
+WaveMatrix intrinsics are defined to support quantization calculations.
+Including calculating a sum for the rows of the left matrix and a sum of the
+columns of the right matrix. The `WaveMatrixRightRowAcc` and
+`WaveMatrixLeftColAcc` fragment accumulators perform this operation.
```
pow2clk, line 336: Should they have "Fragment" in the name? I was initially confused why they took full matrices as parameters, but I see through the inheritance that they are fragment accumulators

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+    : WaveMatrixAccumulatorBase<ElTy, NRows, NCols,
+                                detail::WaveMatrixFragmentBase> {
+  void SumAccumulate(WaveMatrixRight<MatElTy, NRows, NCols> RightMatrix);
+};
+```
+
+#### Zero Point
+
+The following is the equation for matrix multiplication with zero point
+adjustment included:
```
pow2clk, line 358: This section feels out of place in between the definition of WaveMatrix Fragment Acuumulators and the description of their SumAccumulate method. I recognize that it is referenced in the section just after it, but if this is defining the multiply operations, perhaps it can go after the multiply method description and the next section can link back to it?

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+
+$- Z_a * (\sum_{i=0}^{K} B_{[i,y]})$ is the zero point adjustment for matrix $A$
+
+$- Z_b * (\sum_{i=0}^{K} A_{[x,i]})$ is the zero point adjustment for matrix $B$
+
+$+ Z_a * Z_b * K$ is the static zero point adjustment for both matrix $A$ and $B$
+
+$Z_*$ are constant zero points values
+
+#### Wave Matrix SumAccumulate
```
pow2clk, line 372: maybe "WaveMatrix Fragment SumAccumulate" to be consistent with the title of that section?

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+
+The `SumAccumulate` methods accumulate the values of the argument matrix into
+the WaveMatrix fragment accumulator. The fragment WaveMatrix must have the same
```
pow2clk, line 375: Reading this and the description above the class, it's not clear to me what actually gets placed into the accumulator. Is it the sum of all elements in each row of a row-major matrix and the sum of all elements in each column of a column-major matrix?

pow2clk (commented):
Sorry for the slew of comments 😕

proposals/xxxx-wave-matrix.md
```diff
@@ -0,0 +1,396 @@
...
+of many machine learning and image processing workloads. Adding native support
+to HLSL will enable high-performance matrix operations across all supported
+hardware with Shader Model 6.8 drivers.
+
+## Proposed solution
+
+WaveMatrix introduces new matrix templates to facilitate wave cooperative
+operations:
+
+```c++
```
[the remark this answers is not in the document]
jeremyong: FYI if you check the 3rd party grammars supported by Github's Linguist integration [here](https://github.com/github-linguist/linguist/blob/master/vendor/README.md), HLSL is listed as being supported by Tim's textmate grammar (so `hlsl` is a valid fenced code block and we just need to update that grammar to update it as the language changes). I would actually suggest having an "official" Microsoft-maintained grammar file that you can update as the compiler updates if it isn't too much trouble. It'd be super helpful for HLSL tool maintainers to have a more "official" syntax highlighter.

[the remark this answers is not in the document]
llvm-beanz: In this code block the HLSL highlighting is probably okay, but GitHub’s HLSL syntax highlighting doesn’t handle HLSL 2021 particularly well. The C++ mode syntax highlighting does much better IMO.

As an example here’s HLSL highlighted:

```hlsl
namespace detail {
template <typename ElTy, int NRows, int NCols>
class WaveMatrixBase {
};
} // namespace detail
```

And the same code C++ highlighted:

```c++
namespace detail {
template <typename ElTy, int NRows, int NCols>
class WaveMatrixBase {
};
} // namespace detail
```

I’m fine changing the simpler code samples to be HLSL highlighted, but I’d greatly prefer to keep the ones that have templates C++-mode. I had used C++ everywhere for consistency, but that isn’t strictly necessary.

proposals/xxxx-wave-matrix.md
llvm-beanz: I want to be a little careful about how deep this document goes in terms of usage. This isn't intended to be user documentation, this is a specification of the language feature.

proposals/xxxx-wave-matrix.md
llvm-beanz: No, it depends on which specific fragment type you use, either `WaveMatrixLeftColAcc` or `WaveMatrixRightRowAcc`, which are a column or row respectively.

proposals/xxxx-wave-matrix.md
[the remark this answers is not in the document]
jeremyong: Right, I think the thing to do is to fork the C++ tmLanguage file as a base, adding HLSL specific constructs/keywords, and then swap it as the official highlighting grammar for github (instructions for doing that are [here](https://github.com/github-linguist/linguist/blob/master/CONTRIBUTING.md#changing-the-source-of-a-syntax-highlighting-grammar).

proposals/xxxx-wave-matrix.md
llvm-beanz: The inheritance should not be there. The fragment classes don't support orientation specification.

proposals/xxxx-wave-matrix.md
llvm-beanz: That's because I mixed syntaxes and made this invalid. That should be a type alias:
```c++
using is_32bit_int_type = std::false_type;
```

proposals/xxxx-wave-matrix.md
llvm-beanz: Yes, it adds one accumulator to another. It doesn't work with matrix types, just the accumulator types on both sides.

proposals/xxxx-wave-matrix.md
llvm-beanz: I think this is explicitly not specified. The hardware can decide if it needs to copy the fragment to a full matrix or if it can more optimally perform a row/column-wise operation.

proposals/xxxx-wave-matrix.md
llvm-beanz: I don't love the names of those types, but I think if we add "Fragment" to the name we're going to be getting dangerously close to a type name that can't fit on one line of code with reasonable line wrapping rules.

Degerz:
https://github.com/microsoft/DirectXShaderCompiler/pull/6807

I assume that this proposal has been denied ?

llvm-beanz:
> [microsoft/DirectXShaderCompiler#6807](https://github.com/microsoft/DirectXShaderCompiler/pull/6807)
> 
> I assume that this proposal has been denied ?

Denied is probably the wrong phrasing... In need of significant reworking.

This is definitely a feature we want, but the design proposed here has some pretty big limitations that we haven't had time to address. That resulted in us pausing development on it and not shipping it with the Shader Model 6.8 release.

We hope to refine this feature (in public) and include an updated version of it in the future.

llvm-beanz:
I'm going to close this PR. We are committed to bringing a wave matrix feature to HLSL, and the current plan is to do so in SM 6.9. This proposal will not form the basis of that feature so it is probably best to close this PR.

The feedback on this PR is extremely useful and we'll work to incorporate it into the revised feature.

PR #62, opened 2023-06-22
[SM68] Add proposal for Work Graphs

### LINKED ISSUE #101: [Work Graphs] Remove SYNC_FLAG from Barrier intrinsics
https://github.com/microsoft/hlsl-specs/pull/62
https://github.com/microsoft/hlsl-specs/pull/77

The sync_flags parameter has been removed from the barrier intrinsic, we need to update the DXIL and WG specs.

**Additional context**

https://microsoft.github.io/DirectX-Specs/d3d/WorkGraphs.html#barrier

llvm-beanz:
The Work Graphs feature is a new approach to GPU work creation for HLSL on Shader Model 6.8. The feature extends library shaders to support node shaders which create work graphs.

A preview implementation of this feature is available in the DXC 1.8.2306 preview release.

Fixes #101


proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,455 @@
...
+A broadcasting launch mode node entry may also be a compute entry (annotated with `[shader("compute")]`). In this case the node's input is an empty input record.
+
+#### Function Attributes
+
+All node shaders except _thread launch_ nodes must specify the thread group size using the `[numthreads(<x>, <y>, <z>)]` attribute.
+
+Node shaders support all compute shader function annotations, but also have new annotations unique to node shaders.
+
+##### **`[NodeLaunch("<mode>")]`**
+Valid values for node launch mode are `broadcasting`, `coalescing`, or `thread`. If the `NodeLaunch` attribute is not specified on a node entry the default launch mode is `broadcasting`.
```
llvm-beanz, line 58: Note: in historically the attribute name (i.e. `NodeLaunch`) is case-insensitive, but the arguments are case-sensitive. We should follow that with Work Graph attributes.

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,545 @@
...
+uninitialized when created.
+
+## Detailed design
+
+### Node Entry Functions
+
+Node shaders are compute shaders built into library targets. Shader entries
+annotated as `[shader("node")]` are usable as work graph nodes.
+
+A broadcasting launch mode node entry may also be a compute entry (annotated
```
llvm-beanz, line 73: @pow2clk & @tex3d, I wonder whether or not we should change this. Do we get much benefit from having compute shaders work as node shaders? It wouldn't really be hard to wrap entry points to share shader bodies, and this seems to introduce some unnecessary complication in the language and compiler.

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,534 @@
...
+  /// Any violation of these requirements is undefined behavior.
+  ///
+  /// This method returns `false` for thread groups that are not the last to
+  /// finish and these thread groups are not allowed to read or write to the
+  /// input. Reading or writing the input after this call on a thread that
+  /// returned `false` is undefined behavior.
+  ///
+  /// This method returns `true` for the last thread group to finish. That
+  /// thread group can continue reading and writing to the input.
+  bool FinishedCrossGroupSharing();
```
pow2clk, line 235: That's a lot of text in comments in a pseudo-code block. You'd have more formatting options if you put it outside in metadata. The backticks aren't going to do anything as-is.

pow2clk (commented):
Sorry to cut out half way, but I wanted to give you something to think about and I have to go just now.

proposals/xxxx-work-graphs.md
llvm-beanz: The intent for this was that the comments are doxygen formatted (hence the
@<tag> tags). Doxygen does support markdown formatting. These header burbs are
actually complete enough that we could probably generate a doxygen header doc
from them.

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,534 @@
...
+  /// reference to the underlying data.
+  std::enable_if_t<IsRW, RecordTy>::type &Get();
+};
+
+/// @brief Interface for GroupNodeInputRecord and RWGroupNodeInputRecord
+template <typename RecordTy, bool IsRW = false> class GroupNodeInputRecordBase {
+  /// @brief Returns the number of records that have been coalesced into the
+  /// current thread group.
+  ///
+  /// Returns 1..._MaxCount_, where _MaxCount_ is specified by the
```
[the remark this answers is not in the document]
llvm-beanz: There is no guarantee that it a `GroupNodeInputRecord` has `MaxCount` records
when an entry function is called. The guarantee is that there is at least one
record (otherwise the node wouldn't be called), and that there are no more than
defined by the `MaxRecord` attribute.

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,534 @@
...
+##### **`[NodeLaunch("<mode>")]`**
+Valid values for node launch mode are `broadcasting`, `coalescing`, or `thread`.
+If the `NodeLaunch` attribute is not specified on a node entry the default
+launch mode is `broadcasting`.
+
+##### **`[NodeIsProgramEntry]`**
+
+Shaders that can receive input records from both inside and outside the work
+graph (i.e. from a command list) require this attribute. Graphs can have more
+than one entry shaders that receive inputs from outside the graph. Nodes which
```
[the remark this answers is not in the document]
llvm-beanz: This attribute is a little weird and the description of it in the D3D spec I didn't find any less confusing. My understanding is that we need to know which nodes are "entry" nodes (meaning they receive records from outside the graph). We make two implicit assumptions:

1) Any node shader that isn't explicitly targeted by an output from another node _is implicitly_ a graph entry node.
2) Any node shader that is explicitly targeted by an output from another node _is not_ a graph entry node.

This attribute is only required for the case where a node is explicitly targeted by an output from another node _and_ the user wants to directly invoke it with records from the command list.

The other sentence in here is that a Work Graph can have more than one entry shader, so there is no limit to the number of implicit or explicit graph entry nodes.

Based on my description here do you have a suggestion for how I can clarify the spec?

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,535 @@
...
+  bool IsValid() const;
+};
+
+class EmptyNodeOutput {
+  /// @brief Adds `Count` empty output records to the thread node outputs.
+  ///
+  /// Must be called in thread group uniform control flow.
+  void ThreadIncrementOutputCount(uint Count);
+
+  /// @brief Adds `Count` empty output records to the group node outputs.
```
tex3d, line 308:
```suggestion
  /// @brief Adds `Count` empty output records to the node output, once for the
  /// group, instead of summing the value across threads.
```

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,535 @@
...
+```
+
+##### Entry Parameter Attributes
+
+> Consistent with HLSL grammar attribute names are case-insensitive. Any
+> attributes that take string arguments, the string argument values are
+> case-sensitive.
+
+###### **`[MaxRecords(<count>)]`**
+Applies to node inputs in coalescing launch nodes or outputs for any node mode.
```
tex3d, line 379:
```suggestion
Applies to node inputs in coalescing launch or outputs for any launch mode.
```

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,535 @@
...
+Node outputs require either the `MaxRecords` or `MaxRecordsSharedWith`
+attribute.
+
+###### **`[MaxRecordsSharedWith("<name>")]`**
+This attribute applies to node outputs. The named node must have the
+`MaxRecords` attribute applied to its entry. This attribute and the `MaxRecords`
+attribute are mutually exclusive.
+
+The node that this attribute is applied to shares a maximum record allocation
+with the named node.
```
tex3d, line 398: You specify a parameter here, not a node name.  No quotes.
```suggestion
###### **`[MaxRecordsSharedWith(<parameter>)]`**
This attribute applies to node outputs. The named parameter must have the
`MaxRecords` attribute. This attribute and the `MaxRecords` attribute are
mutually exclusive.

The node output that this attribute is applied to shares a maximum record
allocation with the named node output parameter.
``````

File: proposals/xxxx-work-graphs.md [hunk 9]
```diff
@@ -0,0 +1,535 @@
...
+###### **`[NodeID("<name>", <index> = 0)]`**
+This attribute applies to output nodes and defines the name and index of the
+output node. If not provided, the default index is 0.
+
+If this attribute is not present on an output, the default node ID for the
+output is the name of the parameter, and the index is the default index (0).
+
+###### **`[AllowSparseNodes]`**
+This attribute applies to outputs and allows the work graph to be created even
+if there is not a node defined for the specified output.
```
tex3d, line 412:
```suggestion
This attribute applies to outputs and allows the work graph to be created even
if there is not a node defined for the specified output.  If the output is an
array, each element may or may not have a downstream node defined in the graph.
`IsValid()` can be used to determine whether an output node is defined in the
graph.
```

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,535 @@
...
+```
+
+For nodes that recurse, the `GetRemainingRecursionLevels()` function returns the
+number of remaining recursion levels before reaching the node's maximum
+recursion depth.
+
+#### Barrier
+
+```c++
+enum class MEMORY_TYPE_FLAG : uint {
```
tex3d, line 437: In HLSL, `MEMORY_TYPE_FLAG` and `SEMANTIC_FLAG` are not `enum class`, they are like `enum` instead.  They do not require `MEMORY_TYPE_FLAG::` for use.

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,535 @@
...
+  GROUP_SCOPE = 0x00000002,
+  DEVICE_SCOPE = 0x00000004,
+};
+
+/// @brief Request a barrier for a set of memory types and/or thread group
+/// execution sync.
+/// @param MemoryTypeFlags Flag bits as defined by MEMORY_TYPE_FLAG.
+/// @param SemanticFlags Flag bits as defined by SEMANTIC_FLAG.
+///
+/// `Barrier` must be called from thread group uniform control flow.
```
tex3d, line 456: Barrier only requires group uniform control flow when `GROUP_SYNC` is used.
```suggestion
/// `Barrier` must be called from thread group uniform control flow when
/// `SemanticFlags` includes `GROUP_SYNC`.
```

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,535 @@
...
+
+/// @brief Request a barrier for just the memory used by an object.
+/// @param TargetObject The object or resource which owns the memory to apply
+/// the barrier to.
+/// @param SemanticFlags Flag bits as defined by SEMANTIC_FLAG.
+///
+/// The TargetObject parameter can be a particular node input/output record
+/// object or UAV resource. Groupshared variables are not currently supported.
+///
+/// `Barrier` must be called from thread group uniform control flow.
```
tex3d, line 467: Barrier only requires group uniform control flow when `GROUP_SYNC` is used.
```suggestion
/// `Barrier` must be called from thread group uniform control flow when
/// `SemanticFlags` includes `GROUP_SYNC`.
```

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,535 @@
...
+
+In the context of a node shader, `Barrier` enables requesting a memory barrier on
+input and/or output record memory specifically, while the implementation is free
+to store the data in any memory region.
+
+The pseudo-code below shows implementing the existing HLSL memory barrier
+functions using the new `Barrier` function.
+
+```C++
+void AllMemoryBarrier() { Barrier(ALL_MEMORY, DEVICE_SCOPE, 0); }
```
tex3d, line 487:
```suggestion
void AllMemoryBarrier() { Barrier(ALL_MEMORY, DEVICE_SCOPE); }
```

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,535 @@
...
+
+void AllMemoryBarrierWithGroupSync() {
+  Barrier(ALL_MEMORY, DEVICE_SCOPE | GROUP_SYNC);
+}
+
+void DeviceMemoryBarrier() {
+  Barrier(UAV_MEMORY | GROUP_SHARED_MEMORY, DEVICE_SCOPE, 0);
```
tex3d, line 494:
```suggestion
  Barrier(UAV_MEMORY, DEVICE_SCOPE);
```

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,535 @@
...
+}
+
+void DeviceMemoryBarrierWithGroupSync() {
+  Barrier(UAV_MEMORY | GROUP_SHARED_MEMORY, DEVICE_SCOPE | GROUP_SYNC, 0);
```
tex3d, line 498:
```suggestion
  Barrier(UAV_MEMORY, DEVICE_SCOPE | GROUP_SYNC);
```

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,535 @@
...
+}
+
+void GroupMemoryBarrier() { Barrier(GROUP_SHARED_MEMORY, GROUP_SCOPE, 0); }
```
tex3d, line 501:
```suggestion
void GroupMemoryBarrier() { Barrier(GROUP_SHARED_MEMORY, GROUP_SCOPE); }
```

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,551 @@
...
+#### Entry Parameters
+
+Specific types for input parameters depend on the node launch mode, but all node
+shaders support two categories of parameters:
+
+* Zero or one [Node Input Parameter](#node-input-objects).
+* Zero or more [Node Output Parameter(s)](#node-output-objects).
+
+[Broadcast and Coalescing Launch](launch-modes) shaders also support optional
+system value parameters.
```
tex3d, line 149: I think you misunderstood my previous comment, which I'll resolve since it's pointing to the old text anyway.  I meant that this spec is missing a section describing the allowed system value parameters.

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,551 @@
...
+  /// input. Reading or writing the input after this call on a thread that
+  /// returned `false` is undefined behavior.
+  ///
+  /// This method returns `true` for the last thread group to finish. That
+  /// thread group can continue reading and writing to the input.
+  bool FinishedCrossGroupSharing();
+};
+
+template <typename RecordTy>
+using GroupNodeInputRecord = detail::GroupNodeInputRecordsBase<RecordTy, false>;
```
tex3d, line 247:
```suggestion
using GroupNodeInputRecords = detail::GroupNodeInputRecordsBase<RecordTy, false>;
```

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,551 @@
...
+
+template <typename RecordTy>
+using RWGroupNodeInputRecord = detail::GroupNodeInputRecordsBase<RecordTy, true>;
```
tex3d, line 250:
```suggestion
using RWGroupNodeInputRecords = detail::GroupNodeInputRecordsBase<RecordTy, true>;
```

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,535 @@
+<!-- {% raw %} -->
+
+# Work Graphs
+
+* Proposal: [NNNN](NNNN-work-graphs.md)
+* Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
+* Sponsor: [Greg Roth](https://github.com/pow2clk), [Tex
+  Riddell](https://github.com/tex3d)
+* Status: **Under Consideration**
+* Planned Version: Shader Model 6.8
```
tex3d, line 10: Shouldn't we update these fields, since really aren't going through approval process for an already-shipped feature.
Also, can we assign a number to this now?  I'd like for the mesh nodes spec to have a number that's higher than this one, since it extends it 😄.

proposals/xxxx-work-graphs.md
```diff
@@ -0,0 +1,535 @@
...
+### Launch Modes
+
+_Node_ shaders have one of three _launch modes_:
+
+* Thread
+* Broadcasting
+* Coalescing
+
+_Thread launch_ nodes represent an individual thread of work that processes a
+single _input record_. Thread launch nodes use a thread group size of `(1,1,1)`.
```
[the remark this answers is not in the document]
pow2clk: Just FYI, I'm going to reword this a bit. The compiler accepts numthreads on thread nodes if the dimensions are `(1,1,1)`, but it doesn't require it.

proposals/xxxx-work-graphs.md
pow2clk: I've lightly modified the corresponding section from the API spec and moved it to its own section after the inputs documentation.

proposals/xxxx-work-graphs.md
pow2clk: Renumbered and classified.

tex3d (approved):
Looks good!

pow2clk (approved):

PR #77, opened 2023-08-17
[SM6.8] Add proposal for DXIL 1.8

llvm-beanz:
This change proposes an abbreviated DXIL 1.8 specification. We should fill this out with more detail as appropriate, but this should document the basic bits that we need to verify the compiler implementation.


llvm-beanz:
> While a new shader model is introduced, do you think it might be possible to also deprecate constant buffer layout usage (with a flag) at the same time? Or would that introduce too much risk to the shader model release? I'm thinking the later that gets pushed out, the further into the future we'd have to maintain all that legacy cbuffer layout code :)

Unfortunately, the non-legacy cbuffer layout has lots of bugs and fixing those is out of scope for SM 6.8 (we're too late in the process). It is on our mind as a change we want to get resolved for the future, but it will miss SM 6.8

proposals/NNNN-dxil18.md
```diff
@@ -0,0 +1,282 @@
...
+                              (uint32_t)NodeIOFlags::Input |
+                              (uint32_t)NodeIOFlags::ThreadRecord,
+
+    GroupNodeOutputRecords = (uint32_t)NodeIOFlags::ReadWrite |
+                             (uint32_t)NodeIOFlags::Output |
+                             (uint32_t)NodeIOFlags::GroupRecord,
+    ThreadNodeOutputRecords = (uint32_t)NodeIOFlags::ReadWrite |
+                              (uint32_t)NodeIOFlags::Output |
+                              (uint32_t)NodeIOFlags::ThreadRecord,
+  };
```
pow2clk, line 259: I think this might be more easily consumable as a table either with a single field that lists the flags enabled or maybe even with a column for each flag.

proposals/NNNN-dxil18.md
```diff
@@ -0,0 +1,282 @@
...
+```
+
+### Reading and Writing Node Records
+
+Unlike memory operations in earlier versions of DXIL, reading and writing node
+record memory uses LLVM's load and store instructions. DXIL 1.8 reserves address
+space 6 for node memory. The `getNodeRecordPtr` DXIL operation returns a pointer
+to node record memory in address space 6.
```
pow2clk, line 267: I see now this addresses the space. I feel like maybe this would be more useful higher anyway and it would make the appearance of address space 6 in that table less surprising as well as the absence of some expected intrinsics.

pow2clk (approved):
I don't know if anything is missing, but whatevber might be can be added later.

proposals/NNNN-dxil18.md
```diff
@@ -0,0 +1,262 @@
...
+│kDxilNodeMaxRecursionDepthTag            │    19    │MDList: (i32)             │
+├─────────────────────────────────────────┼──────────┼──────────────────────────┤
+│kDxilNodeInputsTag                       │    20    │MDList: (MDList[])        │
+├─────────────────────────────────────────┼──────────┼──────────────────────────┤
+│kDxilNodeOutputsTag                      │    21    │MDList: (MDList[])        │
+├─────────────────────────────────────────┼──────────┼──────────────────────────┤
+│kDxilNodeMaxDispatchGridTag              │    22    │MD list: (i32, i32, i32)  │
+├─────────────────────────────────────────┼──────────┼──────────────────────────┤
+│kDxilRangedWaveSize                      │    23    │MD list: (i32, i32, i32)  │
+└─────────────────────────────────────────┴──────────┴──────────────────────────┘
```
tex3d, line 130: The Value column doesn't look right.  Each value seems to imply that it points to another MDList containing a value or values.  This isn't the case.  The single values here are operands in-line after the associated tag that identifies the next operand.  The multiple values *are* another MDList, containing that set of values.

The node inputs/outputs is confusing, as it appears to me like an MDList containing a single MDList operand with some set of entries (which might imply that this set of entries are the inputs or outputs).

The meaning/encoding of most of the metadata values is not described.

```suggestion
┌─────────────────────────────────────────┬──────────┬──────────────────────────────┐
│                   Tag                   │ Constant │          Value               │
├─────────────────────────────────────────┼──────────┼──────────────────────────────┤
│kDxilNumThreadsTag                       │    4     │MDList: (i32 x, i32 y, i32 z) │
├─────────────────────────────────────────┼──────────┼──────────────────────────────┤
│kDxilNodeLaunchTypeTag                   │    13    │i32                           │
├─────────────────────────────────────────┼──────────┼──────────────────────────────┤
│kDxilNodeIsProgramEntryTag               │    14    │i1                            │
├─────────────────────────────────────────┼──────────┼──────────────────────────────┤
│kDxilNodeIdTag                           │    15    │MDList: (MDString, i32)       │
├─────────────────────────────────────────┼──────────┼──────────────────────────────┤
│kDxilNodeLocalRootArgumentsTableIndexTag │    16    │i32                           │
├─────────────────────────────────────────┼──────────┼──────────────────────────────┤
│kDxilShareInputOfTag                     │    17    │MDList: (MDString, i32)       │
├─────────────────────────────────────────┼──────────┼──────────────────────────────┤
│kDxilNodeDispatchGridTag                 │    18    │MDList: (i32 x, i32 y, i32 z) │
├─────────────────────────────────────────┼──────────┼──────────────────────────────┤
│kDxilNodeMaxRecursionDepthTag            │    19    │i32                           │
├─────────────────────────────────────────┼──────────┼──────────────────────────────┤
│kDxilNodeInputsTag                       │    20    │MDList: (MDList[], ...)       │
├─────────────────────────────────────────┼──────────┼──────────────────────────────┤
│kDxilNodeOutputsTag                      │    21    │MDList: (MDList[], ...)       │
├─────────────────────────────────────────┼──────────┼──────────────────────────────┤
│kDxilNodeMaxDispatchGridTag              │    22    │MDList: (i32 x, i32 y, i32 z) │
├─────────────────────────────────────────┼──────────┼──────────────────────────────┤
│kDxilRangedWaveSize                      │    23    │MDList: (i32 x, i32 y, i32 z) │
└─────────────────────────────────────────┴──────────┴──────────────────────────────┘
```

proposals/NNNN-dxil18.md
```diff
@@ -0,0 +1,262 @@
...
+│kDxilNodeRecordTypeTag                   │    2     │MDList: (MDList[])        │
+├─────────────────────────────────────────┼──────────┼──────────────────────────┤
+│kDxilNodeMaxRecordsTag                   │    3     │MDList: (i32)             │
+├─────────────────────────────────────────┼──────────┼──────────────────────────┤
+│kDxilNodeMaxRecordsSharedWithTag         │    4     │MDList: (MDString, i32)   │
+├─────────────────────────────────────────┼──────────┼──────────────────────────┤
+│kDxilNodeOutputArraySizeTag              │    5     │MDList: (i32)             │
+├─────────────────────────────────────────┼──────────┼──────────────────────────┤
+│kDxilNodeAllowSparseNodesTag             │    6     │MDList: (i1)              │
+└─────────────────────────────────────────┴──────────┴──────────────────────────┘
```
tex3d, line 157: Similar issues with value as described in prior comment.

```suggestion
┌─────────────────────────────────────────┬──────────┬──────────────────────────┐
│                   Tag                   │  Index   │          Value           │
├─────────────────────────────────────────┼──────────┼──────────────────────────┤
│kDxilNodeOutputIDTag                     │    0     │MDList: (MDString, i32)   │
├─────────────────────────────────────────┼──────────┼──────────────────────────┤
│kDxilNodeIOFlagsTag                      │    1     │i32                       │
├─────────────────────────────────────────┼──────────┼──────────────────────────┤
│kDxilNodeRecordTypeTag                   │    2     │MDList: (<tag-value list>)│
├─────────────────────────────────────────┼──────────┼──────────────────────────┤
│kDxilNodeMaxRecordsTag                   │    3     │i32                       │
├─────────────────────────────────────────┼──────────┼──────────────────────────┤
│kDxilNodeMaxRecordsSharedWithTag         │    4     │MDList: (MDString, i32)   │
├─────────────────────────────────────────┼──────────┼──────────────────────────┤
│kDxilNodeOutputArraySizeTag              │    5     │i32                       │
├─────────────────────────────────────────┼──────────┼──────────────────────────┤
│kDxilNodeAllowSparseNodesTag             │    6     │i1                        │
└─────────────────────────────────────────┴──────────┴──────────────────────────┘
```

proposals/NNNN-dxil18.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+`kDxilNodeRecordType` is a tagged metadata list based on the following tags:
+
+```
+┌─────────────────────────────────────────┬──────────┬──────────────────────────┐
+│                   Tag                   │ Constant │          Value           │
+├─────────────────────────────────────────┼──────────┼──────────────────────────┤
+│kDxilNodeRecordSizeTag                   │    0     │MDList: (i32)             │
```
tex3d, line 166:
```suggestion
│kDxilNodeRecordSizeTag                   │    0     │i32                       │
```

tex3d (approved):
I'm going to approve so we may add the document.  However, there are a lot of deficiencies to address after we add the document, which should be captured as follow-up issues/tasks.

We should probably update the status to reflect the fact that we have already shipped this as well.

Besides the review feedback so far, there is also:
- Missing feature details for: expanded-comparison-sampling and extended-command-info
- Lacking descriptions for each new DXIL intrinsic and their parameters.
- Lacking descriptions for new metadata
- Lacking new feature flags and descriptions

There may be more missing that I'm not thinking of right now, but we can address that as we update the document and these things come up.

PR #94, opened 2023-09-07
Fleshing out Execution and Memory models

llvm-beanz:
This update Intro.Defs earlier in the introduction and puts more detail into the SPMD programming model and memory model.

The PDF draft of this change is available [here](https://github.com/microsoft/hlsl-specs/files/12552783/hlsl.pdf).

### DIFF
--- a/specs/language/introduction.tex
+++ b/specs/language/introduction.tex
@@ -254,6 +254,10 @@ an implementation-defined way. A runtime implementation may sort the
 implementation-defined way. During execution no guarantees are made that all
 \gls{lane}s in a \gls{wave} are actively executing.
 
+\p \gls{wave}, \gls{quad}, and \gls{threadgroup} operations require execution
+synchronization of applicable active and helper \gls{lane}s as defined by the
+individual operation.
+
 \Sub{Optimization Restrictions}{Intro.Model.Restrictions}
 
 \p An optimizing compiler may not optimize code generation such that it changes
--- a/specs/language/introduction.tex
+++ b/specs/language/introduction.tex
@@ -54,6 +54,56 @@ document and should be used in conjunction with interpreting this standard.
 where this document diverges this section, the remaining sections in this
 chapter, and the attached \ref{glossaries}.
 
+\Sec{Common Definitions}{Intro.Defs}
+
+\p The following definitions are consistent between \acrshort{hlsl} and the
+\gls{isoC} and \gls{isoCPP} specifications, however they are included here for
+reader convenience.
+
+\Sub{Correct Data}{Intro.Defs.CorrectData}
+\p Data is correct if it represents values that have specified or unspecified
+but not undefined behavior for all the operations in which it is used. Data that
+is the result of undefined behavior is not correct, and may be treated as
+undefined.
+
+\Sub{Diagnostic Message}{Intro.Defs.Diags}
+\p An implementation defined message belonging to a subset of the
+implementation's output messages which communicates diagnostic information to
+the user.
+
+\Sub{Ill-formed Program}{Intro.Defs.IllFormed}
+\p A program that is not well-formed, for which the implementation is expected
+to return unsuccessfully and produce one or more diagnostic messages.
+
+\Sub{Implementation-defined Behavior}{Intro.Defs.ImpDef}
+\p Behavior of a well-formed program and correct data which may vary by the
+implementation, and the implementation is expected to document the behavior.
+
+\Sub{Implementation Limits}{Intro.Defs.ImpLimits}
+\p Restrictions imposed upon programs by the implementation of either the
+compiler or runtime environment. The compiler may seek to surface
+runtime-imposed limits to the user for improved user experience.
+
+\Sub{Undefined Behavior}{Intro.Defs.Undefined}
+\p Behavior of invalid program constructs or incorrect data for which this
+standard imposes no requirements, or does not sufficiently detail.
+
+\Sub{Unspecified Behavior}{Intro.Defs.Unspecified}
+\p Behavior of a well-formed program and correct data which may vary by the
+implementation, and the implementation is not expected to document the behavior.
+
+\Sub{Well-formed Program}{Intro.Defs.WellFormed}
+\p An \acrshort{hlsl} program constructed according to the syntax rules,
+diagnosable semantic rules, and the One Definition Rule.
+
+\Sub{Runtime Implementation}{Intro.Defs.Runtime}
+\p A runtime implementation
+refers to a full-stack implementation of a software runtime that can facilitate
+the execution of \acrshort{hlsl} programs. This broad definition includes
+libraries and device driver implementations. The \acrshort{hlsl} specification
+does not distinguish between the user-facing programming interfaces and the
+vendor-specific backing implementation.
+
 \Sec{Runtime Targeting}{Intro.Runtime}
 
 \p \acrshort{hlsl} emerged from the evolution of \gls{dx} to grant greater
@@ -64,7 +114,7 @@ integral part of the description for \acrshort{hlsl} . Some \acrshort{hlsl}
 features require specific \gls{sm} features, and are only supported by compilers
 when targeting those \gls{sm} versions or later.
 
-\Sec{\acrfull{spmd} Programming Model}{Intro.Model}
+\Sec{\acrlong{spmd} Programming Model}{Intro.Model}
 
 \p \acrshort{hlsl} uses a \acrfull{spmd} programming model where a program
 describes operations on a single element of data, but when the program executes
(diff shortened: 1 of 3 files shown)


specs/language/introduction.tex
```diff
@@ -54,6 +54,47 @@
...
+\p An implementation defined message belonging to a subset of the
+implementation's output messages which communicates diagnostic information to
+the user.
+
+\Sub{Ill-formed Program}{Intro.Defs.IllFormed}
+\p A program that is not well formed, for which the implementation is expected
+to return unsuccessfully and produce one or more diagnostic messages.
+
+\Sub{Implementation-defined Behavior}{Intro.Defs.ImpDef}
+\p Behavior of a well formed program and correct data which may vary by the
```
jeremyong, line 73: Nit: I think "well-formed" is a more accepted spelling, similar to your use of "ill-formed" elsewhere.

specs/language/introduction.tex
```diff
@@ -54,6 +54,47 @@
...
+implementation, and the implementation is expected to document the behavior.
+
+\Sub{Implementation Limits}{Intro.Defs.ImpLimits}
+\p Restrictions imposed upon programs by the implementation.
```
jeremyong, line 77: I wonder if it's worth clarifying what "the implemention" refers to, since we have both DXC/Clang, but also downstream consumers of the DXIL output.

specs/language/introduction.tex
```diff
@@ -54,6 +54,47 @@
...
+
+\Sub{Undefined Behavior}{Intro.Defs.Undefined}
+\p Behavior of invalid program constructs or incorrect data which this standard
```
jeremyong, line 80: wording: "... or incorrect data _for which_ this standard ..."

specs/language/introduction.tex
```diff
@@ -54,6 +54,47 @@
...
+imposes no requirements, or does not sufficiently detail.
+
+\Sub{Unspecified Behavior}{Intro.Defs.Unspecified}
+\p Behavior of a well formed program and correct data which may vary by the
```
jeremyong, line 84: What does "correct data" refer to here? I would consider rephrasing to something like "Undocumented implementation-dependent behavior of a well-formed program" (a bit more compact)

specs/language/introduction.tex
```diff
@@ -54,6 +54,47 @@
...
+implementation, and the implementation is not expected to document the behavior.
+
+\Sub{Well-formed Program}{Intro.Defs.WellFormed} \p An \acrshort{hlsl} program
+constructed according to the syntax rules, diagnosable semantic rules, and the
+One Definition Rule.
```
jeremyong, line 89: Interesting that ODR is called out as a separate rule, are there no other rules in the same classification (link-time) that need to be considered for a program to be considered "well-formed"?

specs/language/introduction.tex
```diff
@@ -77,46 +118,67 @@
...
+specification is useful for defining the execution and memory model as well as
+specific semantics of language constructs.
+
+\Par{\gls{lane}}{Intro.Model.Terms.Lane}
+A \gls{lane} represents a single computed element in an \acrshort{spmd}
 program. In a traditional programming model it would be analogous to a thread of
 execution, however it differs in one key way. In multi-threaded programming
 threads advance independent of each other. In \acrshort{spmd} programs, a group
 of \gls{lane}s execute instructions in lock step because each instruction is a
 \acrshort{simd} instruction computing the results for multiple \gls{lane}s
```
jeremyong, line 139: wording suggestion:

> In \acrshort{spmd} programs, a group of \gls{lane}s share a common instruction stream, which they execute in lockstep.

Note: I believe "lockstep" is commonly used as a single word now, although it is occasionally hyphenated also. This wording also includes instructions that don't "compute results" (e.g. barrier instructions).

specs/language/introduction.tex
```diff
@@ -77,46 +118,67 @@
...
-requirement that be the case. \acrshort{hlsl} is explicitly designed to run on
-hardware with arbitrary \gls{wave} sizes.
-
-\Sub{\gls{quad}}{Intro.Model.Quad}
-
-\p A \gls{quad} is a subdivision of four \gls{lane}s in a \gls{wave} which are
+\Par{\gls{wave}}{Intro.Model.Terms.Wave} A grouping of \gls{lane}s for execution
+is called a \gls{wave}. \gls{wave} sizes vary by hardware architecture. Some
+hardware implementations support multiple wave sizes. Generally wave sizes are
+powers of two, but there is no requirement that be the case. \acrshort{hlsl} is
```
jeremyong, line 145: Is there a "multiple-of-four" requirement?

specs/language/introduction.tex
```diff
@@ -77,46 +118,67 @@
...
+A \gls{quad} is a subdivision of four \gls{lane}s in a \gls{wave} which are
 computing adjacent values. In pixel shaders a \gls{quad} may represent four
 adjacent pixels and \gls{quad} operations allow passing data between adjacent
 lanes. In compute shaders quads may be one or two dimensional depending on the
 workload dimensionality described in the \texttt{numthreads} attribute on the
 entry function. (FIXME: Add reference to attribute)
 
-\Sub{\gls{threadgroup}}{Intro.Model.Group}
+\Par{\gls{threadgroup}}{Intro.Model.Terms.Group}
+A grouping of \gls{lane}s executing the same shader to produce a combined
```
jeremyong, line 161: It may be worth explicitly clarifying the execution model expected when a thread group size is less than the wave size, and/or not a multiple of the wave size.

specs/language/introduction.tex
```diff
@@ -77,46 +118,67 @@
...
+result is called a \gls{threadgroup}. \gls{threadgroup}s are independent of
+\acrshort{simd} hardware specifications. The dimensions of a \gls{threadgroup}
+are defined in three dimensions. The maximum dimensions of a \gls{threadgroup}
```
jeremyong, line 164: wording: I suggest "the maximum extent along each dimension of a ..." because "maximum dimensions" here is just 3.

specs/language/introduction.tex
```diff
@@ -77,46 +118,67 @@
...
-\p A grouping of \gls{threadgroup}s which represents the full execution of a
+\Par{\gls{dispatch}}{Intro.Model.Terms.Dispatch}
+A grouping of \gls{threadgroup}s which represents the full execution of a
 \acrshort{hlsl} program and results in a completed result for all input data
 elements.
 
+\Sub{\acrshort{spmd} Execution Model}{Intro.Model.Exec}
+
+\p A runtime implementation will provide an implementation-defined mechanism for
+defining a \gls{dispatch}. In an implementation-defined way a runtime will
```
jeremyong, line 175: I might rephrase this a bit, although I'm not sure what the best phrasing is. I think I would naturally write something like:

> Implementations are permitted to vary with respect to hardware resource management and execution scheduling provided that the observed behavior conforms to the behavior described in this specification. For example, a runtime implementation may choose to...

specs/language/introduction.tex
```diff
@@ -130,37 +192,38 @@
...
+\texttt{groupshared} keyword. The underlying memory for any declaration
+annotated with \texttt{groupshared} is shared across an entire
+\gls{threadgroup}. Reads and writes to \gls{threadgroup} Memory, may occur in
+any order except as restricted by synchronization intrinsics.
+
+\Par{Device Memory}{Intro.Memory.Spaces.Device}
+Device memory is memory available to all \gls{lane}s executing on the device.
+This memory may be read or written to by multiple \gls{threadgroup}s that are
+executing concurrently. Reads and writes to device memory may occur in any order
+except as restricted by synchronization intrinsics. Some device memory may be
```
jeremyong, line 216: "synchronization intrinsics" might preclude specific annotations like `globallycoherent` which I don't personally consider to be an "intrinsic"

specs/language/introduction.tex
```diff
@@ -130,37 +192,38 @@
...
+visible to the host. Device memory that is visible to the host may have
+additional synchronization concerns.
```
jeremyong, line 218: Probably worth elaborating what those concerns are specifically

specs/language/introduction.tex
```diff
@@ -130,37 +192,38 @@
...
+
+\Par{Constant Memory}{Intro.Memory.Spaces.Constant}
+Constant memory is similar to device memory in that it is available to all
+\gls{lane}s executing on the device. Constant memory is read-only, and an
+implementation can assume that constant memory is immutable and cannot change
+during execution.
+
+\Par{Host Memory}{Intro.Memory.Spaces.Host}
+Host memory is memory that is directly addressable by the host. Reads and writes
```
jeremyong, line 227: What about resizable bar and host-addressable VRAM?

jeremyong (commented):
Thanks for working on this doc! I left some wording suggestions and comments that are hopefully helpful.

specs/language/introduction.tex
llvm-beanz: It's possible in the future we may need to incorporate more post-compile rules. I actually expect that we may accrue a few to represent the DXIL and SPIR-V validators, but I want to hold off referencing that explicitly until I get a little further along.

llvm-beanz:
I've put in a bunch of word smithing and updates.

I think this is mostly in sane state now.

[Updated PDF](https://github.com/microsoft/hlsl-specs/files/13529743/hlsl.pdf)

dmpots (approved):

specs/language/introduction.tex
```diff
@@ -130,37 +241,38 @@
...
 elaborate further.
 
-\Sec{Common Definitions}{Intro.Defs}
+\Sub{Memory Spaces}{Intro.Memory.Spaces}
 
-\p The following definitions are consistent between \acrshort{hlsl} and the
-\gls{isoC} and \gls{isoCPP} specifications, however they are included here for
-reader convenience.
+\p \acrshort{hlsl} programs manipulate data stored in five distinct memory
+spaces: thread, threadgroup, device, constant and host.
```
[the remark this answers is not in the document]
llvm-beanz: I had stripped `host` out and missed that spot. We define a host address space in the DXIL spec, but we don't use it, so I was thinking I'd simplify this until we really need it.

llvm-beanz:
@Keenuts, the last change I added here is in response to the discussion we had in email this week. Do you think that wording correctly addresses the concerns?

specs/language/introduction.tex
```diff
@@ -78,46 +128,144 @@
...
+\p A \textit{helper} \gls{lane} is a lane which would not be executed by the
+initial launch conditions except that its computations are required for adjacent
+pixel operations in pixel fragment shaders. A \textit{helper} \gls{lane} will
+execute all computations but will not perform writes to buffers, and any outputs
+it produces are discarded. \textit{Helper} lanes may be required for
+\gls{lane}-cooperative operations to execute correctly.
+
+\p A \textit{inactive} \gls{lane} is a lane that is not executed by the initial
+launch conditions. This occurs when there is not sufficient inputs to fill the
+all \gls{lane}s in the \gls{wave}.
```
Keenuts, line 221: Maybe not worth to mention, but isn't another cause is register pressure? If the register pressure is too high, AFAIK lanes might be disabled and the wave size be only have a fraction of the theoretical wave-size.

Keenuts:
> @Keenuts, the last change I added here is in response to the discussion we had in email this week. Do you think that wording correctly addresses the concerns?

Sure, this block as-is makes sense: no optimization should change the program behavior (except optimizations like fast-math? But maybe that's considered to be a codegen switch, not an optimization switch?)

What is IMO the important missing bit (but it's a whole subject) is to define what is expected for actual execution: the spec mentions lock-step is not mandatory, hence the spec as-is is allows the same CFG/shader to yield 2 different results. Example: running on Maxwell and Volta architecture, with the latter allowing not reconverging directly after a divergence.

llvm-beanz:
> Sure, this block as-is makes sense: no optimization should change the program behavior (except optimizations like fast-math? But maybe that's considered to be a codegen switch, not an optimization switch?)

Within the spec you can't change the program behavior outside what is specified. HLSL has fast-math by default, so that will be the specified behavior for floating point math operations.

> What is IMO the important missing bit (but it's a whole subject) is to define what is expected for actual execution: the spec mentions lock-step is not mandatory, hence the spec as-is is allows the same CFG/shader to yield 2 different results. Example: running on Maxwell and Volta architecture, with the latter allowing not reconverging directly after a divergence.

Lockstep isn't strictly required by the spec for anything. Even wave operations can be emulated without lockstep execution if you have appropriate synchronization primitives.

For Volta, the expectation is that wave operations will effectively require warp synchronization. I think that's a driver problem for Nvidia more than anything else. Maybe that's the missing explicit wording.

I can add something like:

> Wave, Quad, and Thread Group operations require execution synchronization of applicable active and helper Lanes as defined by the individual operation.

Keenuts:
> For Volta, the expectation is that wave operations will effectively require warp synchronization. I think that's a driver problem for Nvidia more than anything else. Maybe that's the missing explicit wording.
> 
> I can add something like:
> 
> > Wave, Quad, and Thread Group operations require execution synchronization of applicable active and helper Lanes as defined by the individual operation.

Yes, and where those barriers are put should probably be described in the control-flow structures (here using Cuda's syntax mixed with HLSL)

```
switch (value) {
    case 0:
        WaveActiveSum(1);
   default:
       __syncwrap(MASK_FOR_ALL_THREADS);
       WaveActiveSum(10);
}
```

llvm-beanz:
> ```
> switch (value) {
>     case 0:
>         WaveActiveSum(1);
>    default:
>        __syncwrap(MASK_FOR_ALL_THREADS);
>        WaveActiveSum(10);
> }
> ```

I don't think this is correct. Every `Wave*` intrinsic needs to have a `__syncwarp` associated with it. In your example it may have only been required for the `default:` case, but in a more complicated example they're effectively needed for all Wave and Quad operations with a mask that represents all the participating lanes.

Keenuts:
> > ```
> > switch (value) {
> >     case 0:
> >         WaveActiveSum(1);
> >    default:
> >        __syncwrap(MASK_FOR_ALL_THREADS);
> >        WaveActiveSum(10);
> > }
> > ```
> 
> I don't think this is correct. Every `Wave*` intrinsic needs to have a `__syncwarp` associated with it. In your example it may have only been required for the `default:` case, but in a more complicated example they're effectively needed for all Wave and Quad operations with a mask that represents all the participating lanes.

Yes, of course. If we add a `case 1` which doesn't branch, then the mask has to be different, what I meant with this is the control-flow structure itself needs to define where synchronization is expected, not just the intrinsic:
 - we both agree the we must synchronize as soon as we can, when we fall-through. But this has to be made explicit IMO, and it's related to the control flow construct.
 (Same for loops breaks, etc).
 
 But It seems this goes beyond this initial PR, so probably not the best place to dive into this.

llvm-beanz:
Yea, I think some of this will come down to the wording as we write the behavior of switch statements and case labels, but at the same time I don't think we need to define synchronization for control flow structures.

It isn't illegal to reshape control flow during optimization if it doesn't change program behavior, so synchronization of execution isn't really a property of the control flow structures. Even within the definitions of C, it is illegal to merge and hoist control flow if it changes the resulting execution. C doesn't have the SIMT/SPMD considerations that HLSL does, but we can lean on similar rules for HLSL and get the expected behavior.

The C spec defined case statements from ANSI C through until C17 (the wording is slightly different in the latest draft of C23) with the following statement:

> Labels in themselves do not alter the flow of control, which continues unimpeded across them.

If that's the basis for how `case` labels are defined, and `Wave*` operations are defined to have wave sync behavior, I think that defines the expected behavior completely.

I don't want our spec to make statements about expectations for control flow structure that aren't required because we do want conforming compilers to be allowed to optimize in cases where they can.

specs/language/introduction.tex
```diff
@@ -78,46 +128,148 @@
...
+of \gls{lane}s may execute instructions in lockstep because each instruction may
+be a \acrshort{simd} instruction computing the results for multiple \gls{lane}s
+simultaneously, or synchronizing execution across multiple \gls{lane}s or
+\gls{wave}s. A \gls{lane} has an associated \textit{lane state} which denotes
+the execution status of the lane (\ref{Intro.Model.Terms.LaneState}).
+
+\SubSub{\gls{wave}}{Intro.Model.Terms.Wave}
 
-\Sub{\gls{wave}}{Intro.Model.Wave}
+\p A grouping of \gls{lane}s for execution is called a \gls{wave}. The size of a
```
damyanp, line 158: All lanes in a wave will execute in lockstep (right?) Wonder if that can be mentioned here.  Text above suggests that lanes are only running in lockstep when there's some kind of synchronization going on.

damyanp (approved):

specs/language/introduction.tex
llvm-beanz: They aren't required to be lock-step except at synchronization points. While it has been true with a lot of hardware that lock-step is guaranteed, that's less true with newer hardware. NV's latest GPUs have per-thread program counters which can allow much more creative divergence in execution.

llvm-beanz:
I've marked the outstanding conversations as resolved. I think that I've addressed all the feedback. I'm happy to revisit anything in subsequent PRs. I'm sure this will all need some more work as we get further along.

PR #95, opened 2023-09-08
[0010] Clarify behavior of vk::aliased_pointer on vk::BufferPointer

### LINKED ISSUE #87: [0010] Define behavior of casting between aliased & non-aliased pointers
[Proposal 0010 - vk::BufferPointer](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0010-vk-buffer-ref.md)

We should define conversion rules for aliased and non-aliased pointers.

**Additional context**
[See comment here](https://github.com/microsoft/hlsl-specs/issues/84#issuecomment-1708072582)

greg-lunarg:
Specifically, clarify interaction with default (restrict) BufferPointer.

Fixes #87.

### DIFF
--- a/proposals/0010-vk-buffer-ref.md
+++ b/proposals/0010-vk-buffer-ref.md
@@ -184,7 +184,11 @@ Applying HLSL semantic annotations to objects of type vk::BufferPointer is disal
 
 By default, buffer pointers are assumed to be restrict pointers as defined by the C99 standard.
 
-An attribute vk::aliased_pointer can be attached to a variable, function parameter or a block member of BufferPointer type. It ie assumed that the pointee of a BufferPointer with this attribute can overlap with the pointee of any other BufferPointer with this attribute. This also means that the pointee of a BufferPointer with this attribute does not overlap with the pointee of a default (restrict) BufferPointer.
+An attribute vk::aliased_pointer can be attached to a variable, function parameter or a struct member of BufferPointer type. It is assumed that the pointee of a BufferPointer with this attribute can overlap with the pointee of any other BufferPointer with this attribute if they have the same pointee type and their scopes intersect. This also means that the pointee of a BufferPointer with this attribute does not overlap with the pointee of a default (restrict) BufferPointer.
+
+The result of vk::static_pointer_cast and vk::reinterpret_pointer_cast as well as all constructors is restrict.
+
+A pointer value can be assigned to a variable, function parameter or struct member entity, even if the aliasing disagrees. Such an assignment is an implicit cast of this property.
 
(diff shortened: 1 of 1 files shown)


proposals/0010-vk-buffer-ref.md
```diff
@@ -184,7 +184,7 @@ Applying HLSL semantic annotations to objects of type vk::BufferPointer is disal
 
 By default, buffer pointers are assumed to be restrict pointers as defined by the C99 standard.
 
-An attribute vk::aliased_pointer can be attached to a variable, function parameter or a block member of buffer pointer type. It is assumed that the pointee of an object with this attribute can overlap with the pointee of any other object with this attribute.
+An attribute vk::aliased_pointer can be attached to a variable, function parameter or a block member of BufferPointer type. It ie assumed that the pointee of a BufferPointer with this attribute can overlap with the pointee of any other BufferPointer with this attribute. This also means that the pointee of a BufferPointer with this attribute does not overlap with the pointee of a default (restrict) BufferPointer.
```
s-perron, line 187: Looks like you introduced a typo "It ie assumed ...".

s-perron (approved):

proposals/0010-vk-buffer-ref.md
s-perron, line 187: The meaning of the restrict attribute only applies to a scope.

For example the following is allowed in C:

```
int foo(restrict int* p) {
  return *p;
}

int func(int * p) {
  return foo(p); // implicit cast from "aliased" to "restrict".
}
```

The extra sentence you added is not true. An aliased pointer and restrict pointer can point to overlapping storage as long as the aliased pointer is not used during the lifetime of the restrict pointer.

s-perron (changes requested):
How does this fix #87?

proposals/0010-vk-buffer-ref.md
greg-lunarg: I agree this PR does not necessarily address #87. I will attempt to redress that.

I am not sure I agree that my "extra" statement above is false, although perhaps underspecified. My statement above assumes that the two pointers in question have overlapping scopes. If they don't have overlapping scopes, than their overlapping  memory is moot, it seems to me. But I will add that the pointers have overlapping scopes to clarify.

greg-lunarg:
I have added clarifying text for alias casting. I have also added an example.

greg-lunarg:
@s-perron Please review

s-perron:
@llvm-beanz Can you review this for the second approval.

llvm-beanz (approved):

Issue #96, opened 2023-09-11
[0011] Begin Review Period


llvm-beanz:
**Which proposal does this relate to?**
[Proposal 0011 - Inline SPIR-V](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0011-inline-spirv.md)

**Review Period**
With all outstanding issues resolved, the Inline SPIR-V spec is now entering a final review period. This period is a final call for feedback on the proposal. All feedback is due 9/22. If there is no outstanding feedback at EOD 9/22, the spec will be accepted on 9/25.

(cc @s-perron)

devshgraphicsprogramming:
How can I build a type corresponding to a BDA / GLSL buffer_reference for some struct `T` using ` vk::SpirvType`?

How could I declare the equivalent of
```glsl
layout() buffer coherent read_only
{
   uint count;
   uint timestamps[];
};
```

PR #97, opened 2023-09-11
[0011] Begin Review Period

llvm-beanz:
This begins the review period as described in #96.

### DIFF
--- a/proposals/0011-inline-spirv.md
+++ b/proposals/0011-inline-spirv.md
@@ -6,7 +6,7 @@
 *   Author(s): [Steven Perron](https://github.com/s-perron)
 *   Sponsor: [Chris B](https://github.com/llvm-beanz) and
     [Steven Perron](https://github.com/s-perron)
-*   Status: **Under Consideration**
+*   Status: **Under Review**
 
 ## Introduction
 


devshgraphicsprogramming:
Ok so there's one "challenge" I have.

Suppose I want to declare a type from a SPIR-V intrinsic, like `uint8_t` or `int8_t` to solve:
- https://github.com/microsoft/DirectXShaderCompiler/issues/4637
- https://github.com/microsoft/DirectXShaderCompiler/issues/4778

Basically the problem I see is that a plain
```cpp
namespace vk
{
template<uint32_t OpCode>
using SpirvType;
}
```

is enough to declare a SPIR-V **non** "opaque" type that can be read from UBO or stored to SSBO (or anything else really).

Also it will likely make the `sizeof()` operator error or ICE the compiler in weird ways.

IMHO that template should also have parameters for size and alignment, and either  a bool or enum bitfield for what kinds of storage of this type is allowed (UBO/SSBO/InputOutput/PushConstant/etc.).

s-perron (approved):

cassiebeckley:
@devshgraphicsprogramming as I understand it, `sizeof` already does not work correctly. As it stands, fixing all of the problems with it is likely to be significantly more complex than implementing this proposal as written.

devshgraphicsprogramming:
> @devshgraphicsprogramming as I understand it, `sizeof` already does not work correctly. As it stands, fixing all of the problems with it is likely to be significantly more complex than implementing this proposal as written.

well `sizeof` works correctly on all "builtin" types and aggregates if you compile with `-fvk-use-scalar-layout`

I'm not a Clang expert, what exactly is the problem in making the parameter list of `SpirvType`, `<OpCode,Size,Alignment>` and then have `sizeof` report the correct size?

Surely that has to be simpler than letting `sizeof(SpirvType<OpCode>)` return 0, garbage, or an error diagnostic?

s-perron:
> > @devshgraphicsprogramming as I understand it, `sizeof` already does not work correctly. As it stands, fixing all of the problems with it is likely to be significantly more complex than implementing this proposal as written.
> 
> well `sizeof` works correctly on all "builtin" types and aggregates if you compile with `-fvk-use-scalar-layout` In fact I make regular and heavy usage of it.
> 
> I'm not a Clang expert, what exactly is the problem in making the parameter list of `SpirvType`, `<OpCode,Size,Alignment>` and then have `sizeof` report the correct size?
> 
> Surely that has to be simpler than letting `sizeof(SpirvType<OpCode>)` return 0, garbage, or an error diagnostic?

We are in the middle of implementing, and we will probably have to make some changes that relate to the size and alignment. We will get back once we have something implemented and have worked out our uses cases.

In general the `sizeof` feature is broken for Vulkan. Note that you had to use a particular layout. We are less concerned about that. We are more concerned with being able to use the type inside structs and arrays.

devshgraphicsprogramming:
> We will get back once we have something implemented and have worked out our uses cases.

Look forward to it.

> In general the sizeof feature is broken for Vulkan. Note that you had to use a particular layout. We are less concerned about that.

I use that layout because all desktop Vulkan 1.3 devices that still get driver updates support it, also it "makes sense" when you want heavy code&data-structure sharing with C++

The non-scalar layouts don't make sense from C++ side of things because you can't have `(sizeof(T)%alignof(T)) != 0` which is the case for any 3 dimensional vector or matrix.

bogner (approved):

PR #110, opened 2023-10-26
[SM 6.8] Introduce symbol and string tables

llvm-beanz:
This proposal introduces new container parts to adopt a file-level symbol table and string table.

### DIFF
--- a/proposals/xxxx-symbol-and-string-table.md
+++ b/proposals/xxxx-symbol-and-string-table.md
@@ -1,0 +1,204 @@
+<!-- {% raw %} -->
+
+# DirectX Container Symbol and String Table
+
+## Instruction* Proposal: [XXXX](xxxx-symbol-and-string-table.md)
+* Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
+* Sponsor: [Chris Bieneman](https://github.com/llvm-beanz)
+* Status: **Under Consideration**
+* Planned Version: SM 6.8+
+
+## Introduction
+
+This proposal introduces two new parts to the DXIL container format to capture a
+symbol table and file-level string table.
+
+## Motivation
+
+As the usage of DXIL libraries expands and we seek to make DXIL linking better
+supported having a symbol table for DXIL libraries will allow the linker to more
+efficiently choose whether or not a library needs to be loaded during linking.
+Additionally this change adds a separate DXIL part to contain a file-wide string
+table. Today DXIL container parts each contain their own string tables. Those
+string tables can (and do) have duplicated strings.
+
+## Proposed solution
+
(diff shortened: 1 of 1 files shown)


llvm-beanz:
An experimental implementation of this proposal is available here:

https://github.com/llvm-beanz/DirectXShaderCompiler/commit/966586f71b04dbfc7c04ab143414acbdb668b86b

proposals/xxxx-symbol-and-string-table.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Proposed solution
+
+Add two new parts to the DXIL container format with magic IDs `SYM0` and `STR0`
+which will represent the symbol table and string table respectively.
+
+### Symbol Table
+
+The new symbol table is designed to match the Windows PE/COFF symbol table and
+uses the Windows PE/COFF bigobj symbol structures and enumeration values. There
```
tex3d, line 33: I can't find documentation that matches the structure used below.  [This documentation](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#symbol-name-representation) does not match this structure, probably because they are different for "bigobj", but why can't I find any documentation for that structure?  Can you link documentation that matches the structure here if you know of any?

I see hints that this should be `IMAGE_SYMBOL_EX`, but can't find that structure when I do a web search or MS documentation search, other than a [rust mapping](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/SystemServices/struct.IMAGE_SYMBOL_EX.html) which also doesn't match the structure exactly (for Type).

I did find `IMAGE_SYMBOL_EX` in `ntimage.h`, and again `Type` doesn't match, though it occupies the same space as `BaseType` and `ComplexType` combined.  And `SectionIndex` is signed `LONG SectionNumber`.

tex3d (commented):
I don't see how this proposal provides us with any benefits at this time.
- "Motivation" does not define a clear problem to solve, or adequately describe the current situation.  Instead, it presents a desire for something we already have in another form, and a gripe where this proposal will only make things worse (string duplication), since this proposal can't replace the existing data structures.
- "Proposed solution" does not present a solution to a problem, it presents things you want to add.
- I have problems with various details in the proposal, but I think those are moot, since ...
- I don't think we should consider the proposal, especially for SM 6.8, but even for the next shader model, until we have a more comprehensive road map to a cleaner, more unified destination, and this proposal becomes something that looks like a clear step along that path without introducing more complex intermediate states that we have to support.

llvm-beanz:
I'm going to close this PR since we're going to instead go with #112 for SM 6.8. We'll revisit string and symbol tables at a later date.

PR #112, opened 2023-11-01
Proposal for adding shader function name to PSV0 part

tex3d:
This proposes we add a new EntryFunctionName field for PSV0 to encode the entry name for D3D12 runtime use.

### DIFF
--- a/proposals/NNNN-psv0-entry-name.md
+++ b/proposals/NNNN-psv0-entry-name.md
@@ -38,10 +38,10 @@ signatures, and a `PSV0` (`PipelineStateValidation`) part that encodes all
 additional information for the runtime.
 
 None of these parts included for non-library shaders currently capture the name
-of the entry function inside the DxilModule.  This name is available through
-metadata in the DxilModule, but that is encoded in llvm bitcode, so it is not
-accessible to the runtime without adding significant dependencies, and parsing
-through a lot of data it otherwise does not need to parse.
+of the entry function inside the DXIL part data. This name is available in
+the encoded llvm bitcode, but the runtime is unable to parse bitcode.
+Enabling the runtime to parse LLVM bitcode would add dependencies,
+and require parsing a large amount of unnecessary data.
 
 ## Proposed solution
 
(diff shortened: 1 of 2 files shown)


proposals/NNNN-psv0-entry-name.md
```diff
@@ -0,0 +1,213 @@
...
+`DxilContainerAssembler`, when a new version is selected.
+
+In this case this consists of:
+- Adding one accessor that returns a null-terminated utf-8 string pointer, by
+  looking it up in the string table at the offset provided by the new field in
+  the new version of the `PSVRuntimeInfoN` structure.
+- In `DxilContainerAssembler`, when writing the new version of `PSV0`, add the
+  entry function name to the string table, and write the offset a new
+  `EntryFunctionName` field in the new `PSVRuntimeInfoN` structure.
+
```
llvm-beanz, line 116: This "Detailed Design" section does not actually describe the concrete changes you're proposing. The design should provide enough information that I can implement this change in LLVM without needing to look at DXC's codebase to reverse engineer the new structures.

tex3d: > The design should provide enough information that I can implement this change in LLVM without needing to look at DXC's codebase to reverse engineer the new structures.

In order to implement "this change" (one added field), you must have implemented the rest of PSV0.  "concrete changes" require a concrete context to reference.  However, I have nothing concrete to refer to other than the DXC codebase, since that's where PSV0 is currently defined.  If I cannot reference that, I must specify all of PSV0 independently from DXC.  Yet it was made clear that this was not required for this proposal.

I linked to the comment that describes the data layout for PSV0 above.  I described the elements this spec interacts with, how the versioning works, and the concrete changes necessary to add this field to a derived structure, as well as other implementation details for DXC.  I linked to a PR implementing this proposal.  Yes, these refer to the DXC codebase.  That's because the DXC codebase is the only definition I can reference.

For the purpose of evaluating this proposal, what I've included should be more than sufficient.  If we want an independent specification for PSV0, we can spend time writing that.  I do not think that is required to evaluate this proposal.

By the way: it's not called "reverse engineering" when you have the full source code and comments describing layouts and intentions in a working project you can build.

I can think of things I could do to localize more of the specifics referred to here, but none of that will satisfy the "enough information" requirement above unless I end up fully specifying PSV0 here.

Can you be more specific about what you think this proposal should include and/or exclude based on what's here so far?

llvm-beanz (approved):
Mostly looks good. Some nits and a few suggestions around wording to remove narrative language and focus on the relevant details.

python3kgae (approved):

PR #114, opened 2023-11-03
[0010] Fix buffer address space to be Vulkan device address space.

### LINKED ISSUE #98: [0010] Fix the address space for BufferPointers
[Proposal 0010: vk::BufferPointer](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0010-vk-buffer-ref.md)

The spec says 

> All buffer pointers are presumed to point into the host memory address space.

This is not correct. The pointers will contain the address of a buffer that can be used by the device to access the buffer. ~in device memory. That is, the address that is used by the device to access the buffer.~ **EDIT: @jeremyong pointed out my careless wording.**  The memory could be located on the device or simply accessible to the device. The shader does not know or care.

**Additional context**

See https://github.com/microsoft/hlsl-specs/issues/84#issuecomment-1713976696 and the follow up comments by @natevm and me.

The values for these pointers come from calls to [vkGetBufferDeviceAddress](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetBufferDeviceAddressEXT.html) on the host. If we were to keep the language the same as vulkan, we would say that it is in the "device address space". I am open to other suggestions.

greg-lunarg:
Fixes #98

### DIFF
--- a/proposals/0010-vk-buffer-ref.md
+++ b/proposals/0010-vk-buffer-ref.md
@@ -188,7 +188,7 @@ An attribute vk::aliased_pointer can be attached to a variable, function paramet
 
 ### Buffer Pointers and Address Space
 
-All buffer pointers are presumed to point into the host memory address space. No new address space attributes are proposed.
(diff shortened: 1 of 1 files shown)


greg-lunarg:
@llvm-beanz @s-perron  I think this fixes #98. PTAL.

s-perron (approved):

llvm-beanz (approved):

PR #116, opened 2023-11-08
Adding lexing specification for vector literals

llvm-beanz:
This change stubbs out some secitons in the lexer chapter with commented out section definitions and fleshes out the specific setion for the vector literal grammar.

[PDF of Changes](https://github.com/microsoft/hlsl-specs/files/13298698/hlsl.pdf)

### DIFF
--- a/specs/language/lex.tex
+++ b/specs/language/lex.tex
@@ -184,3 +184,95 @@ which terminates at the next new line.
 
 \p Character sequences in header names are mapped to header files or external
 source file names in an implementation defined way.
+
+\Sec{Preprocessing numbers}{Lex.PPNumber}
+
+\begin{grammar}
+  \define{pp-number}\br
+  digit\br
+  \texttt{.} digit\br
+  pp-number \texttt{'} digit\br
+  pp-number \texttt{'} non-digit\br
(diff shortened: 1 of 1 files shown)


bogner (approved):

specs/language/lex.tex
```diff
@@ -184,3 +184,95 @@
...
+  integer-literal \texttt{.} scalar-element-sequence\br
+  floating-point-literal \texttt{.} scalar-element-sequence
+
+  \define{scalar-element-sequence}\br
+  scalar-element-sequence-x\br
+  scalar-element-sequence-r
+
+  \define{scalar-element-sequence-x}\br
+  \texttt{x}\br
+  scalar-element-sequence-x \texttt{x}
```
tex3d, line 263: Should these be limiting length to a maximum of 4 somehow, or is that something handled elsewhere?

tex3d (approved):

specs/language/lex.tex
llvm-beanz: The grammar doesn’t technically need to limit, it can repeat infinitely but the semantic rules will restrict the valid sizes of vectors to the appropriate limited range, and that will come when vector types actually get defined.

PR #129, opened 2023-11-21
[0011] Use variable syntax instead of function syntax for SPIR-V builtin

cassiebeckley:
In order to better match the semantics of builtin variables, use static variables instead of functions in order to get and set builtins.

### DIFF
--- a/proposals/0011-inline-spirv.md
+++ b/proposals/0011-inline-spirv.md
@@ -214,14 +214,14 @@ with the entry point.
 To support adding builtin inputs from source, this proposal adds a new
 `vk::ext_builtin_input` attribute which takes a `builtInId` parameter and
 applies to a variable declaration. This attribute must be applied to a static
-variable.
+const variable.
 
 For example, the `gl_NumWorkGroups` builtin could be declared in a header file
 like this:
 
 ```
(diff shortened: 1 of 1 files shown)


cassiebeckley:
@s-perron for your perusal

proposals/0011-inline-spirv.md
```diff
@@ -211,49 +211,54 @@ With the existing inline SPIR-V, the variable can be decorated and assigned to
...
+applies to a variable declaration. This attribute must be applied to a static
+variable.
 
 For example, the `gl_NumWorkGroups` builtin could be declared in a header file
 like this:
 
 ```
 [[vk::ext_builtin_input(/* NumWorkgroups */ 24)]]
-uint3 gl_NumWorkGroups();
+static uint3 gl_NumWorkGroups;
```
s-perron, line 224: Since we cannot write to the builtin input, do we want to make the variable const as well? Then we can get all of the checks that it cannot be written to for free.

cassiebeckley: Sure, that seems like a good idea. Updated.

s-perron (approved):

s-perron:
@llvm-beanz @pow2clk  FYI

llvm-beanz (approved):

PR #130, opened 2023-12-01
Add Acronym and Glossary to TOC; fix Glossary reference in Introduction.

bharadwajy:
Fix an apparent typo.

### DIFF
--- a/specs/language/introduction.tex
+++ b/specs/language/introduction.tex
@@ -51,8 +51,8 @@ document and should be used in conjunction with interpreting this standard.
 
 \p This document aims to use terms consistent with their definitions in
 \gls{isoC} and \gls{isoCPP}. In cases where the definitions are unclear, or
-where this document diverges this section, the remaining sections in this
-chapter, and the attached \ref{glossaries}.
+where this document diverges this section, refer to the remaining sections in this
+chapter and the attached \nameref{main}.
 
 \Sec{Runtime Targeting}{Intro.Runtime}
 
--- a/specs/language/hlsl.tex
+++ b/specs/language/hlsl.tex
@@ -1,7 +1,8 @@
 \documentclass[oneside]{book}
 \usepackage[utf8]{inputenc}
 \usepackage[english]{babel}
-\usepackage[acronym]{glossaries}
+\usepackage{nameref}
+\usepackage[acronym,numberedsection=nameref,toc]{glossaries}
 \usepackage{multicol}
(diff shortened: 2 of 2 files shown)


llvm-beanz (approved):
Thank you! Nice fixes.

specs/language/introduction.tex
```diff
@@ -51,8 +51,8 @@
 
 \p This document aims to use terms consistent with their definitions in
 \gls{isoC} and \gls{isoCPP}. In cases where the definitions are unclear, or
-where this document diverges this section, the remaining sections in this
-chapter, and the attached \ref{glossaries}.
+where this document diverges this section, refer to the remaining sections in this
```
dmpots, line 54: This still sounds a bit confusing to me

> where this document diverges this section

Should it maybe be "where this document diverges **from** this section"? I guess its not clear to me what its diverging from. Is it supposed to be the `iscC` and `isoCPP` referenced above?

dmpots (approved):

specs/language/introduction.tex
bharadwajy: > This still sounds a bit confusing to me
> 
> > where this document diverges this section
> 
> Should it maybe be "where this document diverges **from** this section"? I guess its not clear to me what its diverging from. Is it supposed to be the `iscC` and `isoCPP` referenced above?

Yes, that makes sense. I missed it. Tried to make minimal changes while ensuring the intent is keep (as I understood it). @llvm-beanz, suggestions?

llvm-beanz: The intent is for this section and the glossaries to cover all places where our terms diverge from the ISO C/C++ standards. So probably something like:

> In cases where the definitions are unclear, or where this document diverges from \gls{isoC} and \gls{isoCPP}, the definitions in this section, the remaining sections in this chapter, and the attached \ref{glossaries} supersede other sources.

bharadwajy: > The intent is for this section and the glossaries to cover all places where our terms diverge from the ISO C/C++ standards. So probably something like:
> 
> > In cases where the definitions are unclear, or where this document diverges from \gls{isoC} and \gls{isoCPP}, the definitions in this section, the remaining sections in this chapter, and the attached \ref{glossaries} supersede other sources.

OK. I already merged this PR with changes with lesser detail than suggested. Maybe you can make your suggested clarification part of #94?? Thanks!

Issue #93, opened 2023-09-07
[0010] Method to obtain `uint64_t` from `BufferPointer`


devshgraphicsprogramming:
**Which proposal does this relate to?**

0010 - Buffer Pointers

**Describe the issue or outstanding question.**

In my opinion we need a function/method to convert `vk::BufferPointer<T,A>` to `uint64_t`.

This is because the proposal explicitly disallows arithmetic operators on `BufferPointer` and doesn't mention there being `operator[]`.

Furthermore indexing past the end of an array in a struct is Invalid/UB in SPIR-V, and `OpTypeRuntimeArray` cannot be slapped onto any odd struct member outside of an StorageBuffer block.
https://github.com/KhronosGroup/SPIRV-Tools/issues/1458

This means that in order to implement the following design pattern from C/C++
```cpp
struct DiscreteCumulativeProbability
{
   uint count;
   // actual array length determined by `count`
   float values[1];
};
```
we cannot simply declare in HLSL202x
```cpp
struct DiscreteCumulativeProbability
{
   uint count;
   // actual array length determined by `count`
   float values[];
};
```

Even if I didn't need to work around #57 and could get the address of `ptr.Get().value[1]` without helper structs/code, I'd still need to be able to perform some form of arithmetic/indexing.

Also I guess `vk::BufferPointer<T[]>` is not the syntax you want to deal with.

Finally being able to just get the `uint64_t` address value is a matter of control, we currently have this level of control with `RawBufferLoad` and `RawBufferStore` and its not nice for it to go away.

**Additional context**

See this comment: https://github.com/microsoft/hlsl-specs/issues/84#issuecomment-1707973062

devshgraphicsprogramming:
btw in case I wasn't clear, this is not about having any **implicit** casts.

I'd find `reintepret_cast<uint64_t>(ptr)` acceptable, as well as explicit `uint64_t()` construction, or just any method added to `vk::BufferPointer` acceptable.

devshgraphicsprogramming:
Another acceptable solution would be for the SPIR-V intrinsics to interact with `vk::BufferPointer` such that the pointer result-id is passed directly to an intrinsic which requires a `[[vk::ext_reference]]`, so I can implement my own cast
```cpp
template<typename T, uint16_t A>
[[vk::ext_capability(/*PhysicalStorageBufferAddresses*/ 5347)]]
[[vk::instruction(/*OpConvertPtrToU*/117)]]
[...]

PR #134, opened 2023-12-01
[0010] Allow casting to uint64_t

s-perron:
We have decided to allow casting to uint64_t. It will make create
more opportunities for developers to make mistakes, but they do not
have to use it. The implementation cost does not seem too high.

Fixes https://github.com/microsoft/hlsl-specs/issues/93

### DIFF
--- a/proposals/0010-vk-buffer-ref.md
+++ b/proposals/0010-vk-buffer-ref.md
@@ -109,6 +109,8 @@ This new type will have the following operations
     syntax vk::BufferPointer<T,A>(u).
 *   A buffer pointer can be cast to a bool. If so, it returns FALSE if the
     pointer is null, TRUE otherwise.
+*   A buffer pointer can be cast to a uint64_t. The cast will return the 64-bit
+    address that the pointer points to.
 
 Note the operations that are not allowed:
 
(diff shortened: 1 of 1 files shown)


s-perron:
@greg-lunarg FYI

proposals/0010-vk-buffer-ref.md
```diff
@@ -120,8 +122,12 @@ Note the operations that are not allowed:
 *   The comparison operators == and != are not supported for buffer pointers.
 
 Most of these restrictions are there for safety. They minimize the possibility
```
Keenuts, line 124: The section above is not true anymore no?
Since we can now build a `BufferPointer` from a uint64_t, and case it back to uint64_t, all the operations above (arithmetic, comparison, etc) are "allowed".

Keenuts: It's more I don't see why arithmetic/comparison is not allowed now, since we allow a much less safe option which is to freely convert any uint64_t to a pointer and back.

s-perron: This forces the developer to be explicit that they are using integers. The idea it to make the dangerous behaviour hard, ugly, and annoying. It also matches what is done in SPIR-V, where the pointer must be cast to an integer before these operations are done.

s-perron: > The section above is not true anymore no? Since we can now build a `BufferPointer` from a uint64_t, and case it back to uint64_t, all the operations above (arithmetic, comparison, etc) are "allowed".

They are "supported" indirectly. You are still not allowed to use "==" with two buffer pointers, so this statement is still true.

greg-lunarg:
I am ok with adding this to the current design.

Do we want to eliminate the cast to bool to test for null pointer as this new capability lets us do this with a comparison to zero? This feels a little cleaner to me and removes the ambiguity around what might constitute a null pointer.

s-perron:
I was considering removing the cast to bool too. I can do that.

proposals/0010-vk-buffer-ref.md
Keenuts: Right, thanks for the precision. This makes sense.

s-perron:
I rebased, and updated the linked list example now that https://github.com/microsoft/hlsl-specs/pull/108 has been merged.

PR #138, opened 2023-12-12
Add stub grammar definitions and sections

llvm-beanz:
These are just empty placeholder sections with some grammar definitions to fill in the ID expression grammars. Properly documenting this will require going back and defining scopes and other lexical constructs.

Since all of this is more or less inherited from C/C++ I'm going to leave them as stubs for now and focus on the HLSL-specific sections.

This change also introduces numbered sub-subsections and formatting for optional grammar elements.

### DIFF
--- a/specs/language/expressions.tex
+++ b/specs/language/expressions.tex
@@ -96,3 +96,40 @@ non-static member function.
 and value category as \textit{E} without the enclosing parenthesis. A
 parenthesized expression may be used in the same contexts with the same meaning
 as the same non-parenthesized expression.
+
+\Sub{Names}{Expr.Primary.ID}
+
+\begin{note}
+  The grammar and behaviors of this section are almost identical to C/C++ with
+  some subtractions (notably lambdas and destructors).
+\end{note}
+
+\begin{grammar}
+  \define{id-expression}\br
+  unqualified-id\br
+  qualified-id
+\end{grammar}
+
+\SubSub{Unqualified Identifiers}{Expr.Primary.ID.Unqual}
+
+\begin{grammar}
+  \define{unqualified-id}\br
+  identifier\br
+  operator-function-id\br
+  conversion-function-id\br
+  template-id\br
+\end{grammar}
+
+\SubSub{Qualified Identifiers}{Expr.Primary.ID.Qual}
+
+\begin{grammar}
+  \define{qualified-id}\br
+  nested-name-specifier \opt{\keyword{template}} unqualified-id\br
+  \define{nested-name-specifier}\br
+  \terminal{::}\br
+  type-name \terminal{::}\br
+  namespace-name \terminal{::}\br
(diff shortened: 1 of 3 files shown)


specs/language/expressions.tex
```diff
@@ -96,3 +96,40 @@
 and value category as \textit{E} without the enclosing parenthesis. A
 parenthesized expression may be used in the same contexts with the same meaning
 as the same non-parenthesized expression.
+
+\Sub{Names}{Expr.Primary.ID}
+
+\begin{note}
+  The grammar and behaviors of this section are almost identical to C/C++ with
+  some subtractions (notably lambdas and destructors).
```
bharadwajy, line 104: Lambdas are primary expressions [per this](https://timsong-cpp.github.io/cppwp/expr#prim). So, the note of their exclusion may be more appropriate in `\Sec{Primary Expressions}{Expr.Primary}`, than here. 

On a related note are `fold-expression` and `requires-expression` planned to be supported? If not, does that need to be called out  in `\Sec{Primary Expressions}{Expr.Primary}` as is being done for `destructors` (in this section) and  `lambda-expressions`? 

A general question: do the differences/omissions from C++ spec need to call out at all as it is evident that they are not listed as supported grammar tokens in *this* spec?

llvm-beanz: > Lambdas are primary expressions [per this](https://timsong-cpp.github.io/cppwp/expr#prim). So, the note of their exclusion may be more appropriate in `\Sec{Primary Expressions}{Expr.Primary}`, than here.

I mentioned lambdas here because there are intricacies about how unqualified ids are resolved in lambdas. It isn't really critical one way or the other. This particular note is intended just as a placeholder because other HLSL-specific parts of the spec are higher priority to flesh out. We should remove this note when we fill out these sections properly.

> On a related note are `fold-expression` and `requires-expression` planned to be supported? If not, does that need to be called out in `\Sec{Primary Expressions}{Expr.Primary}` as is being done for `destructors` (in this section) and `lambda-expressions`?

HLSL is also C++98-based (with a few leaks from C++11), so I've tried to constrain away from mentioning features of newer versions of C++. I don't really want to exhaustively note all the ways HLSL is different from C++, the goal is really just to make the draft and final specifications consumable.

> A general question: do the differences/omissions from C++ spec need to be called out at all as it is evident that they are not listed as supported grammar tokens in _this_ spec?

I'm hopefully containing direct comparisons to C++ to just `note` and `footnote` blocks. Those are really provided to add additional clarification for readers, but not as the body of the specification. The C++ standard does this kind of thing as "editor notes", which are present in drafts but strips them from the final specification.

bharadwajy (approved):
LGTM!

damyanp (approved):

tex3d (approved):
LGTM.

Issue #135, opened 2023-12-06
Deprecate existing `uniform` keyword usage


llvm-beanz:
**Is your feature request related to a problem? Please describe.**
DXC parses the `uniform` keyword in several contexts and propagates it to the AST, but it doesn't _do_ anything with the keyword.

Someday it would be nice to teach the type system about uniformity, but in the current mode where we parse the keyword then ignore it, changing the behavior of the keyword is hard. We should deprecate the keyword in the next HLSL language version and migrate people off it so that we can consider reclaiming it for future use.

**Describe the solution you'd like**
Deprecate the `uniform` keyword in HLSL 202x, and remove it in HLSL 202x + 1.

**Describe alternatives you've considered**
Any significant change to the `uniform` keyword would potentially break existing code since the keyword has no semantic analysis applied (other than verifying it is applied to a variable declaration).

PR #140, opened 2023-12-13
[Expr] Add Expr.Post.Call

llvm-beanz:
This section specifies how HLSL function calls work as well as how argument and paramters are represented, their lifetimes and semantics.

### DIFF
--- a/specs/language/expressions.tex
+++ b/specs/language/expressions.tex
@@ -196,45 +196,45 @@ lvalue directly decays via array-to-pointer decay. \footnote{This results in
 
 \p Arguments to \textit{output} and \textit{input/output parameters} must be
 lvalues. \textit{Output parameters} are not initialized prior to the call; they
-are passed as an uninitialized cx-value (\ref{Basic.lval}). An \textit{output
+are passed as an uninitialized cxvalue (\ref{Basic.lval}). An \textit{output
 parameter} is only initialized explicitly inside the called function. It is
 undefined behavior to not explicitly initialize an \textit{output parameter}
-before returning from the function in which it is defined. The cx-value created
+before returning from the function in which it is defined. The cxvalue created
 from an argument to an \textit{input/output parameter} is initialized through
 copy-initialization from the lvalue argument expression. In both cases, the
-cx-value shall have the type of the parameter and the argument can be converted
+cxvalue shall have the type of the parameter and the argument can be converted
 to that type through implicit or explicit conversion.
 
 \p If an argument to an \textit{output} or \textit{input/output parameter} is a
-constant sized array, the array is copied to a temporary cx-value following the
+constant sized array, the array is copied to a temporary cxvalue following the
 same rules for any other data type. If an argument to an \textit{output} or
 \textit{input/output parameter} is an unsized array type, the array lvalue
 directly decays via array-to-pointer decay. An argument of a constant sized
-array of type \texttt{T[N]} can be converted to a cx-value of an unsized array
+array of type \texttt{T[N]} can be converted to a cxvalue of an unsized array
 of type \texttt{T[]} through array to pointer decay. An unsized array of type
 \texttt{T[]}, cannot be implicitly converted to a a constant sized array of type
 \texttt{T[N]}.
 
-\p On expiration of the cx-value, the value is assigned back to the argument
+\p On expiration of the cxvalue, the value is assigned back to the argument
 lvalue expression following an inverted conversion if applicable. The argument
 expression must be of a type or able to convert to a type that has defined
-copy-initialization to and from the parameter type. The lifetime of the cx-value
+copy-initialization to and from the parameter type. The lifetime of the cxvalue
 begins at argument expression evaluation, and ends after the function returns. A
-cx-value argument is passed by-address to the caller.
+cxvalue argument is passed by-address to the caller.
 
 \p If the lvalue passed to an \textit{output} or \textit{input/output parameter}
 does not alias any other parameter passed to that function, an implementation
 may avoid the creation of excess temporaries by passing the address of the
-lvalue instead of creating the cx-value.
+lvalue instead of creating the cxvalue.
 
 \p When a function is called, any parameter of object type must have completely
 defined type, and any parameter of array of object type must have completely
 defined element type.\footnote{HLSL \textit{output} and \textit{input/output
 parameters} are passed by value, so they must also have complete type.} The
 lifetime of a parameter ends on return of the function in which it is
-defined.\footnote{As stated above cx-value parameters are passed-by-address, so
-the expiring parameter is the reference to the address, not the cx-value. The
-cx-value expires in the caller.} Initialization and destruction of each
+defined.\footnote{As stated above cxvalue parameters are passed-by-address, so
+the expiring parameter is the reference to the address, not the cxvalue. The
+cxvalue expires in the caller.} Initialization and destruction of each
 parameter occurs within the context of the calling function.
 
 \p The value of a function call is the value returned by the called function.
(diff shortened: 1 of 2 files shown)


specs/language/expressions.tex
```diff
@@ -160,3 +160,86 @@
...
+\p When a function is called, each parameter shall be initialized with its
+corresponding argument. The order in which parameters are initialized is
+undefined. \footnote{Today in DXC targeting DXIL matches the Microsoft C++ ABI
+and evaluates argument expressions right-to-left, while SPIR-V generation
+matches the Itanium ABI evaluating parameters left-to-right. There are good
+arguments for unifying these behaviors, and arguments for keeping them
+different.}
+
+\p If the function is a non-static member function the \texttt{this} argument
+shall be initialized to a reference to the object of the call as if casted by an
```
damyanp, line 182: > as if casted

Casted to which type?

specs/language/expressions.tex
```diff
@@ -160,3 +160,86 @@
...
+\p Arguments to \textit{output} and \textit{input/output parameters} must be
+lvalues. \textit{Output parameters} are not initialized prior to the call; they
+are passed as an uninitialized cx-value (\ref{Basic.lval}). An \textit{output
+parameter} is only initialized explicitly inside the called function. It is
+undefined behavior to not explicitly initialize an \textit{output parameter}
+before returning from the function in which it is defined. The cx-value created
+from an argument to an \textit{input/output parameter} is initialized through
+copy-initialization from the lvalue argument expression. In both cases, the
+cx-value shall have the type of the parameter and the argument can be converted
+to that type through implicit or explicit conversion.
```
damyanp, line 205: > the argument can be converted to that type through implicit or explicit conversion

Is this saying that the implementation has a choice of how to do it?  What's the difference between implicit or explicit conversion here?  (At this point, isn't everything implicit?)

specs/language/expressions.tex
```diff
@@ -160,3 +160,86 @@
...
+copy-initialization to and from the parameter type. The lifetime of the cx-value
+begins at argument expression evaluation, and ends after the function returns. A
+cx-value argument is passed by-address to the caller.
+
+\p If the lvalue passed to an \textit{output} or \textit{input/output parameter}
+does not alias any other parameter passed to that function, an implementation
+may avoid the creation of excess temporaries by passing the address of the
+lvalue instead of creating the cx-value.
+
+\p When a function is called parameters with object type must have completely
```
damyanp, line 229: missing comma after "called"?

specs/language/expressions.tex
```diff
@@ -160,3 +160,86 @@
...
+defined type, and parameters of array of object must have completely defined
+element type.\footnote{HLSL \textit{output} and \textit{input/output parameters}
```
damyanp, line 231: Since we're talking about "parameters", should both of the "type"s in this sentence also be plural?

specs/language/expressions.tex
```diff
@@ -160,3 +160,86 @@
...
+parameter is the reference to the address, not the cx-value. The cx-value
+expires in the caller.} Initialization and destruction of each parameter occurs
+within the context of the calling function.
+
+\p The value of a function call is the value returned by the called function.
+
+\p A function call is an lvalue if the result type is an lvalue reference type;
+otherwise it is a prvalue.
+
+\p If a function call is a prvalue of object type, the type of the prvalue shall
```
damyanp, line 244: "shall" -> "must"?  (it's must on line 230).

damyanp (approved):

specs/language/expressions.tex
llvm-beanz: A user can explicitly convert argument types. Some casts in HLSL are allowed to produce lvalues (which is a bit odd). An example of this in practice is:

```c++
struct Doggo {
  int Legs;
  bool Tail;
};

struct Kitten {
  int Legs;
  bool Teeth;
};

void Pet(inout Doggo D) {}

void main() {
  Kitten K = {4, true};
  Pet((Doggo)K);
}
```
[Compiler Explorer](https://godbolt.org/z/5jM63voa1)

In HLSL 2018, the case above was allowed implicitly or explicitly, in HLSL 2021+ it is only allowed explicitly. An example that HLSL 2021 allows implicitly would be:

```c++
void Trunc(int3 I) {}

void main() {
  float3 V = {1.5, 2.5, 3.5};
  Trunc(V);
}
```
[Compiler Explorer](https://godbolt.org/z/7o5e5hWzT)

specs/language/expressions.tex
```diff
@@ -160,3 +160,87 @@
...
+\p A function call may be an \textit{ordinary function}, or a \textit{member
+function}. In a function call to an \textit{ordinary function}, the
+\textit{postfix-expression} must be an lvalue that refers to a function. In a
+function call to a \textit{member function}, the \textit{postfix-expression}
+will be an implicit or explicit class member access whose \textit{id-expression}
+is a member function name.
+
+\p When a function is called, each parameter shall be initialized with its
+corresponding argument. The order in which parameters are initialized is
+undefined. \footnote{Today in DXC targeting DXIL matches the Microsoft C++ ABI
```
bharadwajy, line 175:
```suggestion
unspecified. \footnote{Today in DXC targeting DXIL matches the Microsoft C++ ABI
```

Consider the term "unspecified" - given the [C++ spec definition](https://timsong-cpp.github.io/cppwp/defns.unspecified#def:behavior,unspecified) would apply to such situations and is introduced/inherited in `\Sub{Unspecified Behavior}{Intro.Defs.Unspecified}` (introduction.tex).

specs/language/expressions.tex
```diff
@@ -160,3 +160,87 @@
...
+
+\p If the function is a non-static member function the \texttt{this} argument
+shall be initialized to a reference to the object of the call as if casted by an
+explicit cast expression to an lvalue reference of the type that the function is
+declared as a member of.
+
+\p Parameters are either \textit{input parameters}, \textit{output parameters},
+or \textit{input/output parameters} as denoted in the called function's
+declaration (\ref{Decl.Function}).
+
```
bharadwajy, line 189: HLSL Function Arguments [description](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-function-parameters) lists `uniform` in addition to the above as parameter annotations. Is this spec a forward-looking one with an intent to deprecate it (#135) and hence it was not listed?

bharadwajy: Found out that DXC does not support uniform parameters.

specs/language/expressions.tex
```diff
@@ -160,3 +160,87 @@
...
+\p \textit{Input parameters} are passed by-value into a function. If an argument
+to an \textit{input parameter} is of constant-sized array type, the array is
+copied to a temporary and the temporary value is converted to an address via
+array-to-pointer decay. If an argument is an unsized array type, the array
+lvalue directly decays via array-to-pointer decay. \footnote{This results in
+\textit{input} parameters of unsized arrays being modifiable by a function.}
+
+\p Arguments to \textit{output} and \textit{input/output parameters} must be
+lvalues. \textit{Output parameters} are not initialized prior to the call; they
+are passed as an uninitialized cx-value (\ref{Basic.lval}). An \textit{output
```
bharadwajy, line 199: Nit: Replace `cx-value` with `cxvalue` as defined in `\ref{Basic.lval}`, here and in various places in the document.

bharadwajy (approved):

specs/language/expressions.tex
bogner: aside: we should probably implement a warning for stuff like this use of `Trunc`... I doubt anyone expects it to work how it does.

bogner (approved):

specs/language/expressions.tex
llvm-beanz: Agreed. DXC has some warnings for implicit conversions, but not nearly enough.

specs/language/expressions.tex
llvm-beanz: Done.

PR #142, opened 2023-12-14
Fix failing GitHub action to deploy docs

llvm-beanz:
Our doc build started failing due to a change in the azure images. The failure is the same as what is reported in this issue: https://github.com/actions/runner-images/issues/7603.

The fix described there is to call `sudo apt-get update`, which is the change here.


llvm-beanz:
I filed an issue assigned to myself to create an action to build the docs as a pre-merge check:
https://github.com/microsoft/hlsl-specs/issues/143

bharadwajy (approved):

PR #144, opened 2023-12-14
Fix Pandoc build

llvm-beanz:
For some reason pandoc doesn't like this formatting.


dmpots (approved):

bogner (approved):
This is unfortunate, but fixing it now and revisiting later seems like the best thing to do

PR #147, opened 2024-01-03
Expanded Comparison Sampling spec

pow2clk:
This documents how expanded comparison sampling should be implemented and what interfaces it should present to the shader author, the driver developer, and the runtime developer. In addition it outlines what new behaviors the compiler will have and how they should be tested.


proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,432 @@
...
+    %dx.types.Handle,     ; sampler handle
+    float,                ; coordinate c0
+    float,                ; coordinate c1
+    float,                ; coordinate c2
+    float,                ; coordinate c3
+    i32,                  ; offset o0
+    i32,                  ; offset o1
+    i32,                  ; offset o2
+    float,                ; compare value
+    float,                ; bias: in [-16.f,15.99f]
```
python3kgae, line 263: Is the bias required to be an immediate value or the range [-16.f,15.99f] is only be validated when the bias is immediate value?

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,432 @@
...
+ for non-comparison samplers:
+
+* `CalculateLevelOfDetailUnclamped(SamplerState S, float<num> Location)`
+  returns the raw LOD value determined implicitly using gradients representing
+  how much the _Location_ changes across neighboring pixels regardless of
+  whether it falls outside the valid range.
+* `CalculateLevelOfDetail(SamplerState S, float<num> Location)`
+  returns the implicitly-determined LOD value clamped to the valid range.
+
+There are no `CalculateLevelOfDetail` variants for comparison samplers.
```
tex3d, line 64: While all this detail is good, it doesn't seem to follow the advice in the newly introduced template for "Introduction".  Perhaps these details belong under "Motivation" instead?  Then "Introduction" can better follow the template's guidance: "10,000 ft view of the change being proposed. Try to keep to one paragraph and less than 10 sentences."

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,432 @@
...
+### HLSL additions
+
+These new HLSL texture methods will be added:
+
+```c++
+<Format> <TexObject>::SampleCmpGrad(
+    SamplerComparisonState S,
+    float<num> Location,
+    float<num> DDX,
+    float<num> DDY,
```
tex3d, line 129: `num` for these arguments are not all the same.  For the `Location` argument, it may be one dimension higher for `Texture*Array` types.  Therefore, there needs to be two different `num` variables: one that is the count of texture dimensions without the array dimension, and one with the array dimension.

Dimension with array is used for `Location`, while dimension without array is used for `DDX`, `DDY`, and `Offset`.

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,432 @@
...
+    float CompareValue,
+    [float Clamp,]
+    [out uint Status]
+);
+```
+
+Where `TexObject` represents any texture object type except `Texture2DMS*`,
+ `Texture3D`, and `TextureCube*` objects.
```
tex3d, line 147: Perhaps we should list the allowed texture types, instead of the exclusions?

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,432 @@
...
+ when used by the non-comparison builtin methods:
+
+* `SampleCmpBias(SamplerComparisonState S, float<num> Location, float Bias)`
+* `SampleCmpBias(SamplerComparisonState S, float<num> Location, float Bias, int<num> Offset)`
+* `SampleCmpBias(SamplerComparisonState S, float<num> Location, float Bias, int<num> Offset, float Clamp)`
+* `SampleCmpBias(SamplerComparisonState S, float<num> Location, float Bias, int<num> Offset, float Clamp, uint Status)`
+* `SampleCmpGrad(SamplerState S, float<num> Location, float DDX, float DDY)`
+* `SampleCmpGrad(SamplerState S, float<num> Location, float DDX, float DDY, int<num> Offset)`
+* `SampleCmpGrad(SamplerState S, float<num> Location, float DDX, float DDY, int<num> Offset, float Clamp)`
+* `SampleCmpGrad(SamplerState S, float<num> Location, float DDX, float DDY, int<num> Offset, float Clamp, uint Status)`
```
tex3d, line 105: `DDX` and `DDY` require dimensionality of texture surface (minus array dim - see other comment regarding the `<num>` used in the prototypes).

proposals/0013-expanded-comparison-sampling.md
pow2clk: Here is the origin of the restriction: https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#inst_SAMPLE_B  

" Restrictions:   ...
                2) the range of srcLODBias is ([-16.0f](https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#D3D11_MIP_LOD_BIAS_MIN) to [15.99f](https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#D3D11_MIP_LOD_BIAS_MAX));
                   values outside of this range will produce undefined results"

This is reflected in the sampleBias documentation in DXIL.rst.

proposals/0013-expanded-comparison-sampling.md
pow2clk: Right. Careless copy and pasting.

After I wrote this, I ended up editing the table in DXIL.rst because it had some inaccuracies for Cube map texture sample methods: https://github.com/microsoft/DirectXShaderCompiler/pull/6131

proposals/0013-expanded-comparison-sampling.md
pow2clk: will do

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,432 @@
...
+
+### Diagnostic Changes
+
+#### Removed errors
+
+* Using `CalculateLevelOfDetail[Unclamped]` with a `SamplerComparisonState`
+  sampler will no longer produce an error when compiling with appropriate shader
+  model versions.
+* Invoking new methods `SampleCmpBias` and `SampleCmpGrad` with valid texture
+  objects and parameters will not produce an unrecognized function error.
```
tex3d, line 293: I'm not sure how useful/necessary it is to list newly introduced intrinsic functions as if they are removed errors in this way.

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,432 @@
...
+  (See Instr.SampleCompType, Instr.SamplerModeForSampleC,
+   Instr.ResourceKindForCalcLOD, Instr.ResourceKindForSample, and
+   Instr.ResourceClassForSamplerGather)
+
+### Runtime Additions
+
+#### Runtime information
+
+A shader flag indicating usage of `SampleCmpBias` or `SampleCmpGrad` to indicate
+ that the shader requires the presence of a capability bit indicating support.
```
tex3d, line 320: I think we should suggest a name for the new feature flag here, shouldn't we?  Perhaps `ExtendedComparisonSampling`.

We should also be clear where this flag is added - a new flag in the `DFCC_FeatureInfo` or `SFI0` part.

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,432 @@
...
+
+```C++
+typedef enum D3D12_FEATURE {
+    ...
+    D3D12_FEATURE_D3D12_OPTIONS22
+} D3D12_FEATURE;
+
+typedef struct D3D12_FEATURE_DATA_D3D12_OPTIONS22 {
+    ...
+    BOOL ExtendedComparisonSamplesSupported;
```
tex3d, line 340: Could `Samples` be replaced with `Sampling` to match the spec?  Also, for some reason "Sampling" seems less awkward to me.

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,432 @@
...
+
+## Testing
+
+### Correct Behavior Testing
+
+Verify that `SampleCmpGrad` and the new `CalculateLevelOfDetail*` overloads
+ generate appropriate code including the new DXIL and SPIRV intrinsics for all
+ shader stages.
+Verify that `SampleCmpBias` produces appropriate code including the new DXIL and
+ SPIRV intrinsics for pixel and compute shader stages.
```
tex3d, line 356: Shouldn't this include node shaders as well?

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,432 @@
+# Expanded Comparison Sampling
+
+* Proposal: [0013](0013-expanded-comparison-sampling.md)
+* Author(s): [Greg Roth](https://github.com/pow2clk),
+             [Jesse Natalie](https://github.com/jenatali)
+* Sponsor: [Greg Roth](https://github.com/pow2clk)
+* Status: **Under Consideration**
+* Planned Version: Shader Model 6.8, validator version 1.8
```
llvm-beanz, line 8: Is the validator version really relevant here? This feature is part of 6.8, the fact that the support lands in 1.8 validator or 1.8.xxx DXC seems like an implementation detail.

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,432 @@
...
+HLSL has sampling operations which sample from a texture according to the
+ settings in the provided sampler object.
+The most useful forms of sampling involve the calculation of a MIP level
+ to determine which level(s) of detail(LOD) of the sampled texture should be
+ used and how they should be weighted.
+
+Different non-comparison sampler operations calculate the MIP level in different
+ ways:
+
+* `Sample(SamplerState S, float<num> Location)`
```
llvm-beanz, line 31: Maybe instead of `float<num>` which isn't valid syntax and is a bit odd, `vector<float, X>` which is valid-ish?

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,432 @@
...
+  returns the implicitly-determined LOD value clamped to the valid range.
+
+## Detailed Design
+
+### HLSL additions
+
+These new HLSL texture methods will be added:
+
+```c++
+<Format> <TexObject>::SampleCmpGrad(
```
llvm-beanz, line 125: Can we write these as more complete pseudo-c++? Something like:

```c++
template<typename ElementType, int ElementSize>
vector<ElementType, ElementSize> TextureBase::SampleCmpGrad(
    SamplerComparisonState S,
    vector<float,ElementSize> Location,
    ...);
```

In the Work Graphs spec I tried to be explicit about adding full class hierarchies in pseudo code to make it easier to see exactly what interfaces are being added on which types, what the return types are, and how they relate to the templated parameters.

proposals/0013-expanded-comparison-sampling.md
llvm-beanz: Yea, I would prefer that we explicitly state "we're adding method X to type Y" for each type we are changing rather than tying to have shorter but less explicit language.

In the future if we add a new `Texture4D` type, this language would be confusing to read, but an explicit list of the types that are changed here will stay more accurate.

pow2clk: done

proposals/0013-expanded-comparison-sampling.md
pow2clk: We've occasionally toyed with the idea of divorcing the validator version from the shader model version. I did it mainly because Tex did in #112, but I'm not going to insist on it.

tex3d: Oh, I included it in that PR because availability of this new info in PSV0 for the runtime is only dependent on validator version.  In fact, I should probably remove "SM 6.8" because it's not dependent on shader model.

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,441 @@
...
+### HLSL additions
+
+These new HLSL texture methods will be added:
+
+```c++
+template<typename ElementType, int CoordsCt, int DimsCt>
+vector<ElementType, ElementSize> TextureBase::SampleCmpGrad(
+    SamplerComparisonState S,
+    vector<float,CoordsCt> Location,
+    vector<float,DimsCt> DDX,
```
tex3d, line 137: Why switch from using `crds` and `dims` above to `CoordsCt` and `DimsCt` here?  The other ones had a description for what they meant at least.

tex3d: I bet it was due to search pattern being broken, where the `float<crds>` form is used elsewhere, instead of `vector<float,CoordsCt>`.

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,441 @@
...
+
+Since there is no defined equation to determine which mip level an
+ implementation should use, the best approach is to compare the results of
+ different operations involving mip levels to ensure that they are consistent.
+
+To test `CalculateLevelOfDetail*` methods,
+use a texture with sentinel values in each mip level and ensure that the mip
+  level returned by CalculateLevelOfDetail matches the behavior of the original
+  SampleCmp operation by providing the sentinel value of the mip level
+  CalculateLevelOfDetail* made you expect and making sure that it matches.
```
tex3d, line 412: It might be just me, but I found the wording here confusing.  I feel like a simpler step-by-step breakdown might have been easier to follow.  Here's how I interpreted this (eventually):
- Fill each mip level of a texture with a single value matching the mip level, such as `0` for mip level 0, `1` for mip level 1, and so on.
- In a shader, perform a CalculateLevelOfDetail operation to retrieve the expected mip level for a sample operation being tested.
- With comparison mode set to equal on the comparison sampler, perform a SampleCmp operation, providing the expected mip level as the comparison value.
- Return the comparison result, which is expected to be 1 for success.

proposals/0013-expanded-comparison-sampling.md
```diff
@@ -0,0 +1,441 @@
...
+This will ensure that CalculateLevelOfDetail correctly returns the mip value
+that sample would have used.
+Ensure as well that `CalculateLevelOfDetail` clamps the mip result and
+ `CalculateLevelOfDetailUnclamped` does not.
+
+Given the nature of SampleCmp operations, the tests will have to be performed
+ with comparison values and success determined based on whether it matches what
+ is expected.
```
tex3d, line 420: Considering the parallel nature of shaders, couldn't this be performed without iteration in a single pass in a pixel shader?

For instance, you could output several values from each pixel in the pixel shader, then check that the values are within expected tolerances.  The values could consist of 1. a reference value (such as the result of CalculateLOD, no bias), 2. the comparison result with bias, 3. the bias value, which could be an input interpolated on one RT coordinate dimension (like X).  On the other RT coordinate dimension (like Y), you could vary something that impacts the scale of the input coordinate gradient to provide a range of derivative scales, and thus LODs, across that RT dimension.  Additionally, you could include a sampled result with non-comparison sampler as a reference.

proposals/0013-expanded-comparison-sampling.md
pow2clk: It was to try to make things clearer. Since we're using templates, I felt somehow I could be more verbose without being confusing. I should define them anyway, but some of this needs to be spread to other locations too, I'm just waiting to see if this addresses Chris' comments before I apply it elsewhere.

proposals/0013-expanded-comparison-sampling.md
pow2clk: I'll take another run at it with your comments

llvm-beanz (approved):

proposals/0013-expanded-comparison-sampling.md
pow2clk: I've changed them to use template-style type indicators, which are correct and potentially less confusing.

proposals/0013-expanded-comparison-sampling.md
pow2clk: Removed

proposals/0013-expanded-comparison-sampling.md
pow2clk: `SampleCmpGradientOrBias` was the chosen name. It's better documented now.

proposals/0013-expanded-comparison-sampling.md
pow2clk: Looks like D3D went with `SampleCmpGradientAndBiasSupported` and OPTIONS21. So I'll change that.

proposals/0013-expanded-comparison-sampling.md
pow2clk: Done

proposals/0013-expanded-comparison-sampling.md
pow2clk: I've since propagated all the template usage everywhere

proposals/0013-expanded-comparison-sampling.md
pow2clk: At this stage, the test is written and I'm unsure what it does. As such, I'm inclined to leave the language which I feel doesn't preclude a parallel test, but also doesn't prescribe it.

proposals/0013-expanded-comparison-sampling.md
pow2clk: Done

proposals/0013-expanded-comparison-sampling.md
pow2clk: Done

tex3d (approved):
Looks good!

PR #148, opened 2024-01-03
 Create shader model spec template

pow2clk:
The existing basic template has a lot of useful information, but lacks
some things that are of interest when documenting shader model features.
This alters the format of the versioning, adds details about language,
interchange format, diagnostics, and runtime/driver interfaces to the
implementation details section including questions to prompt details
that might be important to document. In addition, testing and migration
sections are included to go into how the feature can best be vetted and
adopted.


proposals/templates/shader-model-template.md
```diff
@@ -0,0 +1,140 @@
...
+* How does it interact with different shader stages?
+* How does it work interact other HLSL features (semantics, buffers, etc)?
+* How does this interact with C++ features that aren't already in HLSL?
+* Does this have implications for existing HLSL source code compatibility?
+
+### Interchange Format Additions
+
+* What DXIL changes does this change require?
+* What Metadata changes does this require?
+* How will SPIRV be supported?
```
tex3d, line 72: Also: What information is required for runtime validation?

tex3d: Oh, I guess you addressed this separately.

proposals/templates/shader-model-template.md
```diff
@@ -0,0 +1,140 @@
...
+specifications directory once complete. Each feature will need different levels
+of detail here, but some common things to think through are:
+
+### HLSL Additions
+
+* How is this feature represented in the grammar?
+* How does it interact with different shader stages?
+* How does it work interact other HLSL features (semantics, buffers, etc)?
+* How does this interact with C++ features that aren't already in HLSL?
+* Does this have implications for existing HLSL source code compatibility?
```
tex3d, line 66: Should HLSL language changes/additions proposed here be described in a separate HLSL language proposal and referenced here?

tex3d (approved):
Looks good.

proposals/templates/shader-model-template.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Runtime Additions
+
+#### Runtime information
+
+* What information does the compiler need to provide for the runtime and how?
+
+#### Device Capability
+
+* How does it intereact with other Shader Model options?
```
damyanp, line 92: typoe: interact

proposals/templates/shader-model-template.md
```diff
@@ -0,0 +1,140 @@
...
+* What shader model and/or optional feature is required for the bulk of this feature to be present?
```
damyanp, line 93: It seems odd to ask what shader model is required for a shader model proposal?  Or maybe I'm misunderstanding something?

proposals/templates/shader-model-template.md
```diff
@@ -0,0 +1,140 @@
+<!-- {% raw %} -->
+
+# Feature name
+
+## Instructions
```
damyanp, line 5: I wonder if there's some value in something explaining what type of proposal this template is intended for (so, eg, a proposal for a new language feature is redirected to the appropriate template)?

damyanp (approved):

proposals/templates/shader-model-template.md
llvm-beanz: Ideally the language impacting changes in shader models are just additions of new types, attributes and functions. I think in that common case a single spec makes the most sense.

In rare cases we may have situations where language changes are required to expose new shader model functionality. Ideally we should avoid those cases and phase changes, but I also don't think we should strictly prohibit it.

proposals/templates/shader-model-template.md
```diff
@@ -0,0 +1,140 @@
...
+
+This section should grow into a feature specification that will live in the
+specifications directory once complete. Each feature will need different levels
+of detail here, but some common things to think through are:
+
+### HLSL Additions
+
+* How is this feature represented in the grammar?
+* How does it interact with different shader stages?
+* How does it work interact other HLSL features (semantics, buffers, etc)?
```
bharadwajy, line 64:
```suggestion
* How does it interact other HLSL features (semantics, buffers, etc)?
```

Typo?

proposals/templates/shader-model-template.md
```diff
@@ -0,0 +1,140 @@
...
+  what transition suport will we provide?
+  * New compilation failures should have a clear error message and ideally a FIXIT
+  * Changes in codegen should include a warning and possibly a rewriter
+  * Errors that are produced for previously valid shader code whould give ample
+    notice to developers that the change is coming and might involve rollout stages
+
+* Note that changes that allow shaders that failed to compile before to compile
+  require testing that the code produced is appropriate, but they do not require
+  any special transition support. In these cases, this section might be skipped.
+
```
bharadwajy, line 126: Should there be an optional(?) section for features that would be considered for deprecation starting this shader model version as part of the transition strategy section?

proposals/templates/shader-model-template.md
tex3d: Perhaps a better name for the template would be `shader-feature-template.md`, since it should be a proposal for a feature, not necessarily an entire shader model.

proposals/templates/shader-model-template.md
pow2clk: I think that's outside the scope for what this was intended to be, which was not a document for an entire shader model version, but a single feature that would contribute to that version. We should have a way of documenting deprecations, but I don't think it has enough in common with the intent here to be part of this document.

pow2clk:
Hate to force push, but github got weird on me and wouldn't even let me push an unaltered branch onto that same upstream branch saying I was missing commits.

PR #149, opened 2024-01-04
Wave size range proposal

pow2clk:
Adds the ability to specify a range of wave sizes that a shader can use instead of being limited to just one value

### DIFF
--- a/proposals/0014-wave-size-range.md
+++ b/proposals/0014-wave-size-range.md
@@ -4,18 +4,22 @@
 * Author(s): [Greg Roth](https://github.com/pow2clk)
 * Sponsor: [Greg Roth](https://github.com/pow2clk)
 * Status: **Under Consideration**
-* Planned Version: Shader Model 6.8, validator version 1.8
+* Planned Version: Shader Model 6.8
 
 ## Introduction
 
 Shader Model 6.6 included the ability to specify a wave size on a shader entry
- point in order to indicate that this shader depends on a specific wave size.
+ point in order to indicate either that a shader depends on or strongly prefers
+ a specific wave size.
 If that size of wave isn't available to the current hardware,
  shader will fail to load.
-Existing queries allow the developer to determine if this will happen in
- advance.
+If that wave size is among those supported by the platform,
+ it must be the one used when the shader is executed.
 This enables creating custom shaders optimized for the wave size of a
  particularly interesting platform and loading them when possible.
+Existing queries provide the developer with all the information necessary
+ to determine what a platform supports,
+ which should let them choose shaders accordingly.
 
 ## Motivation
 
@@ -47,13 +51,17 @@ This allows the same shader to be used in the event of the availability of a
 
 Some platforms that support a range of wave sizes might overlap with the range
  specified as supported by the shader in more than one value.
-In these cases, the runtime has a choice of wave sizes that it probably doesn't
- have the information needed to choose the best wave size.
-To give the runtime the needed information, it is useful to specify the
+In these cases, the runtime driver has a choice of wave sizes,
+ but it probably doesn't have the information needed to choose the best wave size.
+
+To provide the needed information, the shader author could specify the
  preferred wave size in addition to the full range of acceptable values.
 By adding an optional third `WaveSize` parameter, shader authors can specify the
  optimal wave size for a given shader in addition to the range of acceptable
  sizes.
+This also preserves the ability to force the driver to choose your preferred
+ wave size when multiples are available just as was done with the single value
+ `WaveSize` attribute.
 
 ## Detailed Design
 
@@ -75,17 +83,25 @@ Where `minWaveSize` is the minimum wave size supported by the shader
 
 ### DXIL Additions
 
-The existing `WaveSize` value is stored as a metadata scalar constant.
+The existing `WaveSize` value is stored as a metadata 32-bit integer constant.
 To store the additional values, an additional metadata tag indicating a tuple of
- three values representing the parameters to `WaveSize` given by the user.
+ three 32-bit integers representing the given parameters to `WaveSize`.
 In the case where no preferred wave size is specified, the value zero will
  indicate that nothing was specified.
 
+|         Tag             | Constant |         Value           |
+|-------------------------|----------|-------------------------|
+|kDxilRangedWaveSizeTag   |    23    |MD list: (i32, i32, i32) |
+
 ### SPIR-V Additions
 
-To make use of this attribute, SPIRV would have to represent it in new SPIRV
- operations that drivers can read to determine what the values are an how to
- use to them to choose the best wave size for a given shader.
+To represent the wave size range and preferred value in SPIR-V,
+ a new `ExecutionMode` value would need to be defined for the `OpExecutionMode`
+ instruction to allow specifying `SubgroupSize` with three operands instead of
+ one.
+Like the other formats, the location of the operands would indicate
+ what each value represents, the first two being the min and max wave sizes and
+ the third being the preferred wave size.
 
 ### Diagnostic Changes
 
@@ -100,9 +116,9 @@ These are where new or slightly altered errors are produced:
  attribute.
 * If the preferred wave size value indicated by the third parameter of
  `WaveSize` is not in the range specified by the first two parameters.
-* If multiple `WaveSize` attributes are applied to the same entry point,
- regardless of what overload they are, the existing error indicating an
- attribute conflict is produced.
+* If multiple `WaveSize` attributes are applied to the same entry point
+ with different numbers or values of parameters,
+ the existing error indicating an attribute conflict is produced.
 * If more than three or fewer than one parameter is applied to `WaveSize`.
 * If negative values are provided for any of the `WaveSize` parameters.
 * If float values are provided for any of the `WaveSize` parameters.
@@ -114,11 +130,15 @@ These are where new or slightly altered errors are produced:
 
 Validation should confirm:
 
-* The tuple that the wave size range tag points to has exactly three elements.
 * Each element in that is a power of two integer between 4 and 128 inclusive.
   The third parameter may also be zero.
 * The minimum wave size is less than the maximum wave size.
-* The preferred wave size lies between the minimum and maximum.
+* The preferred wave size is zero or lies between the minimum and maximum.
+* Where validator version is less than 1.6, fail on shaders that use `WaveSize`.
+* Where validator version is less than 1.8, fail when the `WaveSize` metadata
+ tag has more than one parameter.
+* Where validator version is greater than or equal to 1.8, fail when The tuple
+ that the wave size range tag points to does nto have exactly three elements.
 
 ### Runtime Additions
 
@@ -127,8 +147,6 @@ Validation should confirm:
 The PSV0 runtime data structure already contains both
  `MinimumExpectedWaveLaneCount` and `MaximumExpectedWaveLaneCount` members that
  can be used to transmit the minimum and maximum values to the runtime.
-A new `PreferredWaveLaneCount` value will need to be added to the PSV3 revision
- of that struct to accommodate the third preferred value.
 
 #### Device Capability
 
@@ -228,10 +246,10 @@ For each wave size in the platform range,
  ensure that the following shader range, preferred value, and platform range
  combinations are accepted and use the preferred value:
 
-* The the preferred value and shader maximum is equal to the current platform
+* The  preferred value and shader maximum is equal to the current platform
   range value and the shader minimum is one power of two less than the platform
   minimum.
-* The the preferred value and shader minimum is equal to the current platform
+* The preferred value and shader minimum is equal to the current platform
   range value and the shader maximum is one power of two more than the platform
   maximum.
 * The shader range is the full 4-128 and the preferred value is the current
@@ -241,10 +259,10 @@ For each wave size in the platform range,
  ensure that the following shader range, preferred value, and platform range
  combinations are accepted:
 
-* The the shader maximum is equal to the current platform range value and the
+* The shader maximum is equal to the current platform range value and the
   preferred value and shader minimum is one power of two less than the platform
   minimum.
-* The the shader minimum is equal to the current platform range value and the
+* The shader minimum is equal to the current platform range value and the
   preferred value and shader maximum is one power of two more than the platform
   maximum.
 * The shader range is the full 4-128 and the preferred value is one power of two
@@ -254,9 +272,14 @@ For each wave size in the platform range,
 
 ## Alternatives considered
 
-The preferred wave size is seen as useful, but unlike the simple range,
- it requires adding new runtime data.
-That effort should be factored into the consideration of this alternative.
+Useful as it is, the preferred wave size parameter adds some level of testing, diagnostic, and other implementation complexity.
+It wasn't part of the original discussions that motivated this feature,
+ but it is necessary to maintain one aspect of the original `WaveSize` behavior.
+In addition to allowing the platform to reject wave sizes that are unsupported
+ at all, the single-value `WaveSize` attribute tells the platform which wave
+ size to choose among however many options the platform offers.
+Without the preferred wave size, there wouldn't be a way to specify this
+ preference and the platform might choose arbitrarily among the supported values.
 
 Instead of modifying the existing attribute, an additional range attribute taking
  only two parameters might be provided to specify the range while keeping the
@@ -281,3 +304,17 @@ The only issue here is that it complicates correctness checking a bit where
 The simple solution would be to allow only one of either `WaveSize` or
  `WaveSizeRange`.
 Mild aesthetic preference ultimately opted to reuse the existing attribute.
+
+## Acknowledgements
+
+This document received invaluable contributions and feedback from various Microsoft
+ team members and our partners.
+
+Special thanks:
+
+* Alan Baker
+* Chris Bieneman
+* Martin Fuller
+* Amar Patel
+* Damyan Pepper
+* Tex Riddell
--- a/proposals/0014-wave-size-range.md
+++ b/proposals/0014-wave-size-range.md
@@ -27,7 +27,7 @@ Shader Model 6.6 provides no mechanism for specifying multiple wave sizes that a
  single shader might be able to support.
 Consequently, if more than one target wave size is of interest, separate entry
  points will have to be created for each wave size.
-These will each have to be compiled and shipped separately and, at runtime,
+These will each have to be compiled and shipped separately and, at run time,
  the appropriate shaders will have to be selected.
 This increases the size of the shipped product and potentially slows down
  runtime shader loading.
@@ -86,7 +86,7 @@ This allows the same shader to be used in the event of the availability of a
 
 Some platforms that support a range of wave sizes might overlap with the range
  specified as supported by the shader in more than one value.
-In these cases, the runtime driver has a choice of wave sizes,
+In these cases, the graphics driver has a choice of wave sizes,
  but it probably doesn't have the information needed to choose the best wave size.
 
 To provide the needed information, the shader author could specify the
@@ -159,7 +159,6 @@ These are where new or slightly altered errors are produced:
 * If float values are provided for any of the `WaveSize` parameters.
 * If non-numerical values are provided for any of the `WaveSize` parameters.
 * If non-literal variables are provided for any of the `WaveSize` parameters.
-* If any of the parameters to `WaveSize` are less than 4 or greater than 128.
 
 #### Validation Changes
 
@@ -179,6 +178,7 @@ Validation should confirm:
 
 #### Runtime information
 
+No additions are needed here.
 The PSV0 runtime data structure already contains both
  `MinimumExpectedWaveLaneCount` and `MaximumExpectedWaveLaneCount` members that
  can be used to transmit the minimum and maximum values to the runtime.
@@ -202,8 +202,6 @@ Verify this compiler output:
   the third in the same tuple.
 * That the PSV0 `MinimumExpectedWaveLaneCount` and `MaximumExpectedWaveLaneCount`
   values reflect those provided for the wave size range.
-* That the PSV3 `PreferredWaveLaneCount` value reflects the third preferred
-  value when provided and zero when omitted.
 
 #### Diagnostics Testing
 
@@ -307,7 +305,8 @@ For each wave size in the platform range,
 
 ## Alternatives considered
 
-Useful as it is, the preferred wave size parameter adds some level of testing, diagnostic, and other implementation complexity.
+Useful as it is, the preferred wave size parameter adds some level of testing,
+ diagnostic, and other implementation complexity.
 It wasn't part of the original discussions that motivated this feature,
  but it is necessary to maintain one aspect of the original `WaveSize` behavior.
 In addition to allowing the platform to reject wave sizes that are unsupported
--- a/proposals/0014-wave-size-range.md
+++ b/proposals/0014-wave-size-range.md
@@ -37,6 +37,41 @@ The query to determine what wave sizes are supported takes a range and similar
  information is transmitted from the shader to the runtime, though not the
  driver.
 
+### Motivational Example
+
+As an example of how this might be useful, consider a tile culling shader
+For each tile, the shader has to compute the maximum and minimum depth values.
+
+To do this without a preferred wave size involves expensive atomic an barrier
+ operations.
+To keep track of the current minimum and maximum values for each tile requires
+ a storage buffer.
+This buffer needs to be pre-initialized to unachievably low values for the
+ maximum and unachievalby high values for the minimum to ensure that whatever
+ the first value encountered is, it will replace the initialization.
+This pre-clearing requires an expensive resource barrier.
+After that setup, each thread in the tile will have to
+ use atomic operations to compare this thread's depth value
+ and determine the overall mininum and maximum depth for the whole tile,
+ storing it in the tile-specific location in the buffer.
+
+Shader Model 6.6's `WaveSize` already introduces some improvement opportunities
+ on this approach.
+A shader could specify a wave size of 64 and each thread could load 4 depth
+ values and perform a the minimum and maximum calculations locally.
+These local results are then used in a wave minimum/maximum calculation and
+ writes the final result.
+This removes the need for atomics and the advance buffer clearing.
+
+With the addition of a 
+
+Lets say that we are running instead with a range, which is say [4,64]. But we can indicate that 64 is optimal. (Perhaps [4,64,64] to say which is our preferred wave size)
+
+So if we get wave64, SV_GroupWaveCount == 1, we run the console version on PC, great! This was the hint we provided and we get the best version of the shader.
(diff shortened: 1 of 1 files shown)


proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+* Proposal: [0014](0014-wave-size-range.md)
+* Author(s): [Greg Roth](https://github.com/pow2clk)
+* Sponsor: [Greg Roth](https://github.com/pow2clk)
+* Status: **Under Consideration**
+* Planned Version: Shader Model 6.8, validator version 1.8
+
+## Introduction
+
+Shader Model 6.6 included the ability to specify a wave size on a shader entry
+ point in order to indicate that this shader depends on a specific wave size.
```
damyanp, line 12: nit: how come some of these lines are indented by one character?

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+If that size of wave isn't available to the current hardware,
```
damyanp, line 13: I think a key motivation of the SM 6.6 feature was that some hardware supports multiple wave sizes.  The attribute is required to force the driver/hardware to run that shader using the specified wave size.  Without it there's no way for the developer to tell the driver what the required wave size is.

So the feature wasn't really about allowing developers to choose which shader to load, it was about allowing developers to influence decisions that the driver was making when executing the shader.

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+  range value and the shader maximum is one power of two more than the platform
+  maximum.
+* The shader range is the full 4-128 and the preferred value is the current
+  platform range value.
+
+For each wave size in the platform range,
+ ensure that the following shader range, preferred value, and platform range
+ combinations are accepted:
+
+* The the shader maximum is equal to the current platform range value and the
```
damyanp, line 244: There's some repeated `the`s around here.

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+  preferred value and shader maximum is one power of two more than the platform
+  maximum.
+* The shader range is the full 4-128 and the preferred value is one power of two
+  less than the current platform range value.
+* The shader range is the full 4-128 and the preferred value is one power of two
+  greater than the current platform range value.
+
+## Alternatives considered
+
+The preferred wave size is seen as useful, but unlike the simple range,
```
damyanp, line 257: Has someone presented a concrete use case for the preferred wave size?

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+
+* The wave size range tag pointing to anything but a tuple of 3
+* A tuple value is not an integer
+* A range tuple value is -4, 0, 1, 2, 3, 127, 129, or 256
+* A preferred tuple value is -4, 1, 2, 3, 127, 129, or 256
+* The minimum wave size value is equal to the maximum
+* The minimum wave size value is greater than the maximum
+* The preferred wave size is outside the specified range, but otherwise valid
+
+### Execution Testing
```
damyanp, line 191: Some of these validations sound a lot like HLKs.  Would we normally call out new HLKs as part of this proposal?

damyanp (changes requested):

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+
+By adding an optional second `WaveSize` parameter, shader authors can provide
+ two values that represent the range of wave sizes that the shader can support.
+This allows the same shader to be used in the event of the availability of a
+ wave size that the shader has optimized for specifically as well as other sizes
+ that are supported, but might not be so optimal.
+
+Some platforms that support a range of wave sizes might overlap with the range
+ specified as supported by the shader in more than one value.
+In these cases, the runtime has a choice of wave sizes that it probably doesn't
```
amarpMSFT, line 50: nit: it wouldn't be the runtime choosing but rather the driver.

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+Verify this compiler output:
+
+* The two parameter overload of `WaveSize` correctly transmits those values to a
+  metadata tuple with a zero third value that is pointed to by the correct tag
+  in the entry point attribute list.
+* The three parameter overload of `WaveSize` transmits those values as well as
+  the third in the same tuple.
+* That the PSV0 `MinimumExpectedWaveLaneCount` and `MaximumExpectedWaveLaneCount`
+  values reflect those provided for the wave size range.
+* That the PSV3 `PreferredWaveLaneCount` value reflects the third preferred
```
amarpMSFT, line 152: The runtime doesn't need to know the PreferredWaveLaneCount - nothing actionable for it.  That said, it doesn't hurt to have the information in case somehow it does become useful in the future.  Assuming the cost of the work isn't a big deal.

proposals/0014-wave-size-range.md
amarpMSFT: Yes, Martin Fuller motivated it well in an email thread you're on 10/26/2023.

proposals/0014-wave-size-range.md
tex3d: It was also about being able to write a shader that requires a particular wave size in an explicit way, such that it won't be allowed to run with a different wave size, since doing so can result in undefined behavior.

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+
+### Diagnostic Changes
+
+#### New Errors
+
+These are where new or slightly altered errors are produced:
+
+* If any of the parameters of `WaveSize` are not literal power of two integers
+  between 4 and 128 inclusive.
+* If the minimum wave size value is greater than or equal to the max wave size
```
tex3d, line 98: Shouldn't it be legal for min to be equal to max?  That would be the degenerate case, equivalent to a single value (max==min==preferred).

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+ as respectively specified by the first and second parameters in the `WaveSize`
+ attribute.
+* If the preferred wave size value indicated by the third parameter of
+ `WaveSize` is not in the range specified by the first two parameters.
+* If multiple `WaveSize` attributes are applied to the same entry point,
+ regardless of what overload they are, the existing error indicating an
+ attribute conflict is produced.
```
tex3d, line 105: Can it be legal if the attribute specifies the same parameters?  This is our usual behavior for such attributes, I believe.

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+* If more than three or fewer than one parameter is applied to `WaveSize`.
+* If negative values are provided for any of the `WaveSize` parameters.
+* If float values are provided for any of the `WaveSize` parameters.
+* If non-numerical values are provided for any of the `WaveSize` parameters.
+* If non-literal variables are provided for any of the `WaveSize` parameters.
+* If any of the parameters to `WaveSize` are less than 4 or greater than 128.
```
tex3d, line 111: Aren't these points already covered by the first one:
```
* If any of the parameters of `WaveSize` are not literal power of two integers
  between 4 and 128 inclusive.
```

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+ use to them to choose the best wave size for a given shader.
+
+### Diagnostic Changes
+
+#### New Errors
+
+These are where new or slightly altered errors are produced:
+
+* If any of the parameters of `WaveSize` are not literal power of two integers
+  between 4 and 128 inclusive.
```
tex3d, line 97: Can we allow for explicit `0` for preferred to indicate no preference?  Isn't that what we encode when not included?

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+Shader Model 6.6 included the ability to specify a wave size on a shader entry
+ point in order to indicate that this shader depends on a specific wave size.
+If that size of wave isn't available to the current hardware,
+ shader will fail to load.
+Existing queries allow the developer to determine if this will happen in
+ advance.
+This enables creating custom shaders optimized for the wave size of a
+ particularly interesting platform and loading them when possible.
+
+## Motivation
```
damyanp, line 20: I think we have a couple of example use cases we know about that we could describe in the motivation here, and I think we're missing discussion about the driver having to choose a wave size on hardware that supports that.

proposals/0014-wave-size-range.md
damyanp: I'm wondering why 1 and 2 aren't valid wave sizes?  (I suspect I'm missing something obvious)

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+* If float values are provided for any of the `WaveSize` parameters.
+* If non-numerical values are provided for any of the `WaveSize` parameters.
+* If non-literal variables are provided for any of the `WaveSize` parameters.
+* If any of the parameters to `WaveSize` are less than 4 or greater than 128.
+
+#### Validation Changes
+
+Validation should confirm:
+
+* The tuple that the wave size range tag points to has exactly three elements.
```
tex3d, line 117: We actually need to validate different rules depending on the DXIL version, I think:
- < 1.6: not allowed
- < 1.8: one element
- otherwise, three elements

Some of this should be covered by the metadata loading with error or extra metadata indicator.

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+* Each element in that is a power of two integer between 4 and 128 inclusive.
+  The third parameter may also be zero.
+* The minimum wave size is less than the maximum wave size.
+* The preferred wave size lies between the minimum and maximum.
```
tex3d, line 121: (if non-zero)

proposals/0014-wave-size-range.md
tex3d: This is a good point.  It's not necessary to expose this value in PSV data.

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+  maximum.
+* The shader range is the full 4-128 and the preferred value is one power of two
+  less than the current platform range value.
+* The shader range is the full 4-128 and the preferred value is one power of two
+  greater than the current platform range value.
+
+## Alternatives considered
+
+The preferred wave size is seen as useful, but unlike the simple range,
+ it requires adding new runtime data.
```
tex3d, line 258: As pointed out by Amar above, I don't think it requires any additional runtime data, just updated metadata in DXIL, but exposing a range already requires this as well.

proposals/0014-wave-size-range.md
llvm-beanz: From HLSL's perspective the distinction between runtime and driver isn't really significant. It may also vary based on which runtime/driver we're talking about.

Probably a more important distinction for us to be making is _when_ this can or should happen (i.e. after PSO creation).

proposals/0014-wave-size-range.md
llvm-beanz: I'd almost rather we make `0` an error. I can imagine in the future we might fix the bugs in DXC to allow `constexpr` values to be used for these arguments and it might be best to error on all values that are out of the [min, max] range so that logic errors in compile-time expressions get surfaced.

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
+# Wave Size Range
+
+* Proposal: [0014](0014-wave-size-range.md)
+* Author(s): [Greg Roth](https://github.com/pow2clk)
+* Sponsor: [Greg Roth](https://github.com/pow2clk)
+* Status: **Under Consideration**
+* Planned Version: Shader Model 6.8, validator version 1.8
```
llvm-beanz, line 7: Is the validator version relevant?

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+To store the additional values, an additional metadata tag indicating a tuple of
+ three values representing the parameters to `WaveSize` given by the user.
+In the case where no preferred wave size is specified, the value zero will
+ indicate that nothing was specified.
+
+### SPIR-V Additions
+
+To make use of this attribute, SPIRV would have to represent it in new SPIRV
+ operations that drivers can read to determine what the values are an how to
+ use to them to choose the best wave size for a given shader.
```
llvm-beanz, line 88: To be more explicit here. A new `ExecutionMode` value would need to be defined for the `OpExecutionMode` instruction to allow specifying `SubgroupSize` with 3 operands instead of 1.

I did a quick scan and I don't see a SPIR-V extension that supports this today, but it should be pretty straightforward. It is probably worth clearly stating that the SPIR-V extension is not planned as part of the work for implementing this feature.

proposals/0014-wave-size-range.md
llvm-beanz: > I'm wondering why 1 and 2 aren't valid wave sizes? (I suspect I'm missing something obvious)

IIRC, we require wave size of 4 so that derivative operations and quad ops work.

damyanp: There's a separate cap for those.  Not allowing WaveSize=1 prohibits a single threaded implementation, that I imagine could be interesting for some debugging scenarios.  Maybe not important, this is just ringing my arbitrary restriction bell.

llvm-beanz: We have a cap bit for the `Quad` ops, but not for the derivative functions (`ddx`, `ddy`, etc).

llvm-beanz: Wait... nvm. We do have a cap bit for derivatives in mesh and amplification shaders. Please ignore me.

proposals/0014-wave-size-range.md
alan-baker: Vulkan would handle the old WaveSize attribute on the API side and not in SPIR-V at all. `WaveSize(x)` would be best translated creating a pipeline with a required subgroup size and (likely) the full subgroups flag. So it is little incongruent to move part of it into SPIR-V. It could certainly be done, but there would be some awkwardness.

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,283 @@
...
+ and `prefWaveSize` is the optional preferred wave size representing the size
+ expected to be the most optimal for this shader.
+
+### DXIL Additions
+
+The existing `WaveSize` value is stored as a metadata scalar constant.
+To store the additional values, an additional metadata tag indicating a tuple of
+ three values representing the parameters to `WaveSize` given by the user.
+In the case where no preferred wave size is specified, the value zero will
+ indicate that nothing was specified.
```
llvm-beanz, line 82: We should include the DXIL constants addition (including the tag name and value), and the MD list formatting. Something like what is in the DXIL 1.8 spec proposal:

https://github.com/microsoft/hlsl-specs/pull/77/files#diff-33b9ed472d740f1163951807242915ff0e6e2b0f02d22753944e82a18bbaefbcR161

(we should also include that information in the DXIL 1.8 spec, but I can handle copying that over)

proposals/0014-wave-size-range.md
pow2clk: I did so to represent when the same sentence was extended to a second line. It keeps with the template recommendation of limiting line length while making clearer where such breaks were made just to keep the width down.

pow2clk (commented):
Forgot to submit these comments

proposals/0014-wave-size-range.md
llvm-beanz: @alan-baker, I think you're misunderstanding. In Vulkan the API allows you to specify the save size while creating the device, that's somewhat orthogonal to the HLSL `WaveSize` attribute. The existing `WaveSize` attribute should translate to the `OpExecutionMode` `SubgroupSize` property which is part of Vulkan 1.1 SubgroupDispatch.

This allows the shader to specify on an entry point the required subgroup size. That prevents the shader from being used if the Subgroup isn't the required size, which allows the author to make safe assumptions about subgroup size.

The extension here is to allow a shader to specify a range of subgroup sizes. That allows a shader author to make assumptions about the subgroup size within a specified range, which can allow them to ship less shader variants when needing to support specific subgroup sizes for optimal performance.

In both cases the requirement needs to be captured in the byte code to enforce that the shader cannot be used with a mismatched subgroup size.

alan-baker: Vulkan's functionality isn't about device creation. The subgroup size control is part of pipeline creation. There is a device property that specifies the min and max sizes supported by the device, but at pipeline creation time you can require a specific power-of-two size within that range (using [VkShaderStageRequiredSubgroupSizeCreateInfo](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkPipelineShaderStageRequiredSubgroupSizeCreateInfo)). 

Additionally, there are pipeline creation flags for compute-like shaders provided:
* `VK_SHADER_CREATE_ALLOW_VARYING_SUBGROUP_SIZE_BIT_EXT`: subgroup size can vary (incompatible with a required size)
* `VK_SHADER_CREATE_REQUIRE_FULL_SUBGROUPS_BIT_EXT`: require the shader to be launched with no inactive invocations.

This functionality was made core as part of Vulkan 1.3 (see [subgroup size control](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkPhysicalDeviceSubgroupSizeControlProperties)).

The SubgroupDispatch functionality came to SPIR-V from OpenCL and is not currently supported by Vulkan implementations. You can see this be the eventual dependence on the `Kernel` capability (via `DeviceEnqueue`).

Again, there could be a Vulkan/SPIR-V extension to allow some of that functionality for Vulkan, but it doesn't fit the current way Vulkan handles it. I understand wanting to capture the HLSL attribute as part of the compiled shader.

llvm-beanz: I don't think that changes what we need to do here for HLSL. If Vulkan doesn't support a shader requiring a specific subgroup size, then Vulkan extensions will be required both for supporting the existing SM 6.6 feature, and the SM 6.8 feature.

For SM 6.6's feature the SPIR-V SubgroupSize execution mode captures the required information (whether or not Vulkan supports the required capability, it is at least defined in SPIR-V). For SM 6.8 compatibility an additional SPIR-V execution mode (and capability, extension etc) will be required.

We're not actively driving changes in Vulkan or SPIR-V. We've included information in this proposal as a guide to anyone who might want to drive those efforts.

pow2clk (commented):
Lots of responses reflecting the most recent commit

proposals/0014-wave-size-range.md
```diff
@@ -254,9 +272,14 @@ For each wave size in the platform range,
 
 ## Alternatives considered
 
-The preferred wave size is seen as useful, but unlike the simple range,
- it requires adding new runtime data.
-That effort should be factored into the consideration of this alternative.
+Useful as it is, the preferred wave size parameter adds some level of testing, diagnostic, and other implementation complexity.
```
damyanp, line 275: nit: line 275 is longer than 80 columns

damyanp (approved):

proposals/0014-wave-size-range.md
llvm-beanz: I prefer disallowing things that are likely to mask mistakes, so I agree with @pow2clk here.

proposals/0014-wave-size-range.md
llvm-beanz: I think @tex3d's feedback here still stands though that this is a duplicate test criteria based on the first bullet.

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,320 @@
...
+
+Validation should confirm:
+
+* Each element in that is a power of two integer between 4 and 128 inclusive.
+  The third parameter may also be zero.
+* The minimum wave size is less than the maximum wave size.
+* The preferred wave size is zero or lies between the minimum and maximum.
+* Where validator version is less than 1.6, fail on shaders that use `WaveSize`.
+* Where validator version is less than 1.8, fail when the `WaveSize` metadata
+ tag has more than one parameter.
```
llvm-beanz, line 139:
```suggestion
* Where validator version is less than 1.7, fail on shaders that use `RangedWaveSize`.
* Validation should fail if an entry has both `RangedWaveSize` and `WaveSize` metadata.
```
Since we are using different tags for ranged and the old wave size, the validation rules should be simpler.

proposals/0014-wave-size-range.md
llvm-beanz: This looks like it is still in the current version

llvm-beanz (commented):
I think this is really close. We should renumber the document to 13 in preparation for merging.

proposals/0014-wave-size-range.md
pow2clk: Ah yes. I'd missed that in the interchange. Done.

proposals/0014-wave-size-range.md
pow2clk: I think this should be where the validator is less than 1.8.

I included the mention of `WaveSize` because even though it's a different spec, it seems we might, with minimal effort correct some testing holes in the process. Perhaps that would be better done with individual bugs though.

proposals/0014-wave-size-range.md
pow2clk: Woops. Missed this spot in yesterday's purge. Removed now.

proposals/0014-wave-size-range.md
pow2clk: Fixed.

pow2clk (commented):
Updating comment threads per most recent commit

proposals/0014-wave-size-range.md
llvm-beanz: This still needs correction.

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,354 @@
...
+
+Shader Model 6.6's `WaveSize` already introduces some improvement opportunities
+ on this approach.
+A shader could specify a wave size of 64 and each thread could load 4 depth
+ values and perform a the minimum and maximum calculations locally.
+These local results are then used in a wave minimum/maximum calculation and
+ writes the final result.
+This removes the need for atomics and the advance buffer clearing.
+
+With the addition of a
```
llvm-beanz, line 66: Hanging incomplete sentence.

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,354 @@
...
+This removes the need for atomics and the advance buffer clearing.
+
+With the addition of a 
+
+Lets say that we are running instead with a range, which is say [4,64]. But we can indicate that 64 is optimal. (Perhaps [4,64,64] to say which is our preferred wave size)
+
+So if we get wave64, SV_GroupWaveCount == 1, we run the console version on PC, great! This was the hint we provided and we get the best version of the shader.
+If SV_GroupWaveCount > 1, we can still do better than the current PC implementation, because we have SV_GroupWaveIndex. That is, we allocate a  per wave min/max value in group shared memory. Each wave writes its min/max into group shared memory. We do a group sync, all waves other than wave index 0 retire. Wave index 0 reads all the group shared values, does another min/max of those, and writes the results for the tile. No atomics, no pre-clearing of the buffer.
+
+So yeah we can definitely use both the range wave size and SV_GroupWaveIndex/Count in conjunction to bring the console implementation of the shader to PC, and inform the compiler to pick the wave size we know is going to do a better job and run faster, if it can.
```
llvm-beanz, line 73: nit: Please try to wrap text consistently.

pow2clk: I failed to push the commit that removed this block of text as I didn't mean to include it.

proposals/0014-wave-size-range.md
pow2clk: I think you might need to refresh. This was already removed.

pow2clk: Nevermind. Something is going on with my end. I don't see this locally.

pow2clk: Ah right. it's also part of the removed block. Pushing commit now.

llvm-beanz (approved):
I'm happy with this as-is. There are some unresolved threads from @tex3d.

proposals/0014-wave-size-range.md
tex3d: I'm not sure how this would represent a meaningful mistake.  It's more likely that it is intended to be parameterized in a uniform-syntax way - filled in from macros or something like that.  Rather than forcing a different syntax for the attribute for the degenerate case, we could support this case using the same syntax.

tex3d: I've filed #157 to follow up on this issue later.

proposals/0014-wave-size-range.md
tex3d: I've filed #157 to follow up on this issue later.

tex3d (approved):
I think we should merge the PR and add the proposal!

proposals/0014-wave-size-range.md
tex3d: Filed #158 to capture the last question.

proposals/0014-wave-size-range.md
```diff
@@ -0,0 +1,320 @@
...
+* If the preferred wave size value indicated by the third parameter of
+ `WaveSize` is not in the range specified by the first two parameters.
+* If multiple `WaveSize` attributes are applied to the same entry point
+ with different numbers or values of parameters,
+ the existing error indicating an attribute conflict is produced.
+* If more than three or fewer than one parameter is applied to `WaveSize`.
+* If negative values are provided for any of the `WaveSize` parameters.
+* If float values are provided for any of the `WaveSize` parameters.
+* If non-numerical values are provided for any of the `WaveSize` parameters.
+* If non-literal variables are provided for any of the `WaveSize` parameters.
```
llvm-beanz, line 126: Random thought. Do these values really need to be literals or could they be compile-time constant expressions? Is there any reason not to support something like:

```hlsl
static const int Val = 2;
[WaveSize(Val, Val * 4, Val * 2)]
```

I realize that we have some bugs in DXC that relate to evaluation of constant expressions, but we could certainly support this use case in Clang.

tex3d: I agree we should support this, and I believe we do support some variations of constant expressions in similar attribute cases.  This is probably an issue for follow up on what terminology we should use for these attribute values, and figure out what DXC currently supports.  Could you file an issue for later follow up so we can merge the proposal?

proposals/0014-wave-size-range.md
tex3d: It looks like most of the lines I highlighted are still there.  Did you miss these, or forget to push an update?

proposals/0014-wave-size-range.md
pow2clk: https://github.com/microsoft/hlsl-specs/issues/161

proposals/0014-wave-size-range.md
pow2clk: I missed some. Fixed.

PR #150, opened 2024-01-08
[0011] Clarify behavior of SpirvType parameters and add SpirvOpaqueType

### LINKED ISSUE #128: [0011] Enable vk::SpirvType to be used in interface arrays and structs

[0011: Inline SPIR-V](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0011-inline-spirv.md)

The compiler does not know the size or alignment of a `vk::SpirvType`. Because types that are used in an interface variables like globals or structured buffers must be fully laid out, the compiler needs to know the size and alignment of a `vk:SpirvType` if it is used in an aggregate type that is part of the interface.

**Additional context**

We might need to think how this will interact with the different layout rules that we support.

cassiebeckley:
`SpirvType` now takes size and alignment as parameters, and a new template `SpirvOpaqueType` is added that does not. Also lets users specify whether a parameter for the type-declaration instruction should take an id or an immediate literal value.

Fixes #128.

### DIFF
--- a/proposals/0011-inline-spirv.md
+++ b/proposals/0011-inline-spirv.md
@@ -157,11 +157,37 @@ test. Some of the difficulties are:
     they cannot use the same id for two different types. This can become hard to
     manage.
 
-This proposal deprecates the old mechanism, and replaces it with a new type
-`vk::SpirvType<int OpCode, ...>`. The template on the type contains the opcode
-and all of the parameters necessary for that opcode. The difficulty with this is
-that the operands are not just literal integer values. Sometimes they are
-another type.
+This proposal deprecates the old mechanism, and replaces it with two new types
+`vk::SpirvOpaqueType<uint OpCode, ...>` and
+`vk::SpirvType<uint OpCode, uint size, uint alignment, ...>`. For
+`SpirvOpaqueType`, the template on the type contains the opcode and all of the
+parameters necessary for that opcode. Each parameter may be one of three kinds
+of values:
+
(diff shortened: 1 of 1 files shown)


cassiebeckley:
@s-perron

s-perron (approved):

s-perron:
This is a pretty big change to the original proposal, but necessary to do what we need to do.

llvm-beanz (approved):
I think this LGTM. I wonder if there is some alignment we should aim for between this proposal and https://github.com/microsoft/hlsl-specs/pull/154.

cassiebeckley:
I'd considered using `alignas`, but it would require modifying the parser to have a different behavior than C++, which we were trying to avoid for the most part.

For example

```hlsl
typedef alignas(8) vk::SpirvType</* OpTypeInt */ 21, /* size */ 8, vk::ext_literal(8), vk::ext_literal(false)> uint8_t;
```

or

```hlsl
typedef vk::SpirvType</* OpTypeInt */ 21, /* size */ 8, vk::ext_literal(8), vk::ext_literal(false)> alignas(8) uint8_t;
```

would not be valid syntax.

llvm-beanz:
Right... because `alignas` isn't valid for typedefs :(

That said, we might be able to implement this feature in terms of `alignas` in the future because this horrible thing is completely valid (with C23):

```c++
template<int Align, int Size>
struct alignas(Align) OpaqueType
{
  _BitInt(Size) V;
};
```

PR #153, opened 2024-01-10
Introduce proposal for availability diagnostics

llvm-beanz:
This proposal introduces three modes for producing availability diagnostics during semantic analyisis. These diagnostics can be adjusted by the user to control the severity and allow for controlled tightening of rules around API exposure.

### DIFF
--- a/proposals/infra/INF-NNNN-availability-diagnostics.md
+++ b/proposals/infra/INF-NNNN-availability-diagnostics.md
@@ -49,8 +49,19 @@ As an example of the annotations in action take the following declaration from
 
 ```c++
 __attribute__((availability(shadermodel, introduced = 6.0)))
+__attribute__((availability(vulkan, introduced = 1.0)))
 __attribute__((clang_builtin_alias(__builtin_hlsl_wave_active_count_bits)))
 uint WaveActiveCountBits(bool Bit);
+
+// This function is available starting in SM 2.0 for pixel shaders, but SM 6.6
+// for compute, mesh and amplification shaders.
+__attribute__((availability(shadermodel_pixel, introduced = 2.0)))
+__attribute__((availability(shadermodel_compute, introduced = 6.6)))
+__attribute__((availability(shadermodel_mesh, introduced = 6.6)))
+__attribute__((availability(shadermodel_amplification, introduced = 6.6)))
+__attribute__((availability(vulkan, introduced = 1.0)))
+__attribute__((clang_builtin_alias(__builtin_hlsl_wave_active_count_bits)))
+uint ddx(bool Bit);
 ```
 
 > Note: the actual header uses macros to condense the attribute descriptions.
@@ -112,7 +123,7 @@ Unlike in the _default_ or _relaxed_ mode, the compiler will emit diagnostics
 for mismatched shader model version without the use of a call graph and
 regardless of reachability.
 
-A user enables strict mode by passing `-fhlsl-strict-diagnostics`. 
+A user enables strict mode by passing `-fhlsl-strict-diagnostics`.
 
 ### Comparison Against Existing Behavior
 
@@ -128,7 +139,7 @@ SamplerState samplerState : register(s0);
 
 FeedbackTexture2D<SAMPLER_FEEDBACK_MIN_MIP> map : register(u0);
 
-[numthreads(4, 4, 1)] 
+[numthreads(4, 4, 1)]
 void main(uint3 threadId : SV_DispatchThreadId) {
   float2 uv = threadId.xy;
   uv /= 256;
@@ -144,9 +155,11 @@ note: at 'call void @dx.op.writeSamplerFeedbackLevel(i32 176, %dx.types.Handle %
 Validation failed.
 ```
 
-This error displays the LLVM IR text for the error site. In a trivial example
-like the one above finding the error is easy. In a complex shader that could be
-thousands of lines long this is unmanageable.
+This error displays the LLVM IR text for the error site, and attempts to
+attribute a line location in the HLSL source, it can be error prone. In a
+trivial example like the one above finding the error may be straightforward, but
+it does involve recognizing textual LLVM IR and the HLSL source that generated
+it.
 
 With this proposal under the _relaxed_ mode, clang will emit the following
 warning:
@@ -181,7 +194,7 @@ void fn(uint3 threadId) {
   map.WriteSamplerFeedbackLevel(texture, samplerState, uv, threadId.x % 8);
 }
 
-[numthreads(4, 4, 1)] 
+[numthreads(4, 4, 1)]
 void main(uint3 threadId : SV_DispatchThreadId) { }
 ```
 
@@ -197,9 +210,6 @@ diagnostics. In the _strict_ mode, clang emits the following diagnostic:
 
 In this case the DXIL validator will emit no diagnostic since the call to the
 unavailable API is not present in the final output.
-
-In all three cases above the DXIL validator will also emit the same diagnostic
-as it does today.
 
 To illustrate the behavior for library shaders consider the following example:
 
@@ -228,7 +238,7 @@ When compiled with the `lib_6_3` profile under the _default_ mode, clang will
 emit the following error:
 
 ```
-<>:1:9: error: 'ddx' is available for pixel shaders beginning with Shader Model 2.0 
+<>:1:9: error: 'ddx' is available for pixel shaders beginning with Shader Model 2.0
    9 |   return ddx(f);
      |          ^~~
 ```
@@ -237,10 +247,10 @@ When compiled with the `lib_6_3` profile under the _strict_ mode, clang will
 emit the following errors:
 
 ```
-<>:1:9: error: 'ddx' is available for pixel shaders beginning with Shader Model 2.0 
+<>:1:9: error: 'ddx' is available for pixel shaders beginning with Shader Model 2.0
    9 |   return ddx(f);
      |          ^~~
-<>:6:9: error: 'WaveMultiPrefixSum' is available beginning with Shader Model 6.5 
+<>:6:9: error: 'WaveMultiPrefixSum' is available beginning with Shader Model 6.5
    9 |   return WaveMultiPrefixSum(f, 1.xxxx);
      |          ^~~~~~~~~~~~~~~~~~
 ```
--- a/proposals/infra/INF-NNNN-availability-diagnostics.md
+++ b/proposals/infra/INF-NNNN-availability-diagnostics.md
@@ -91,7 +91,7 @@ a _warning_.
 
 The relaxed mode does not issue diagnostics for `CallExpr` nodes that are inside
 functions which are not reachable from exported functions. A user enables
-relaxed mode by passing `-Wno-error-hlsl-availability`.
+relaxed mode by passing `-Wno-error=hlsl-availability`.
 
 ### Strict Diagnostic Mode
 
@@ -108,7 +108,11 @@ shader entry functions. If the callee of a `CallExpr` has availability
 annotations that signify that the API is unavailable for the target shader stage
 the compiler emits an _error_.
 
-A user enables strict mode by passing `-fhlsl-strict-diagnostics`.
+Unlike in the _default_ or _relaxed_ mode, the compiler will emit diagnostics
+for mismatched shader model version without the use of a call graph and
+regardless of reachability.
+
+A user enables strict mode by passing `-fhlsl-strict-diagnostics`. 
 
 ### Comparison Against Existing Behavior
 
@@ -162,9 +166,6 @@ following error:
       |       ^~~~~~~~~~~~~~~~~~~~~~~~~
 ```
 
-In all three cases above the DXIL validator will also emit the same diagnostic
-as it does today.
-
 To illustrate the difference between _default_ and _strict_ take the following
 example:
 
@@ -196,6 +197,57 @@ diagnostics. In the _strict_ mode, clang emits the following diagnostic:
 
 In this case the DXIL validator will emit no diagnostic since the call to the
 unavailable API is not present in the final output.
+
+In all three cases above the DXIL validator will also emit the same diagnostic
+as it does today.
+
+To illustrate the behavior for library shaders consider the following example:
+
+> [Godbolt Link](https://godbolt.org/z/rKMGz566M)
+```c++
+float d(float f) {
+  return ddx(f);
+}
+
+float dead(float f) {
+  return WaveMultiPrefixSum(f, 1.xxxx);
+}
+
+float also_dead(float f) {
+  return ddy(f)
+}
+
+[shader("vertex")]
+float main() : FOO {
+  float f = 3;
+  return d(f);
+}
+```
+
+When compiled with the `lib_6_3` profile under the _default_ mode, clang will
+emit the following error:
+
+```
+<>:1:9: error: 'ddx' is available for pixel shaders beginning with Shader Model 2.0 
+   9 |   return ddx(f);
+     |          ^~~
+```
+
+When compiled with the `lib_6_3` profile under the _strict_ mode, clang will
+emit the following errors:
+
+```
+<>:1:9: error: 'ddx' is available for pixel shaders beginning with Shader Model 2.0 
+   9 |   return ddx(f);
+     |          ^~~
+<>:6:9: error: 'WaveMultiPrefixSum' is available beginning with Shader Model 6.5 
+   9 |   return WaveMultiPrefixSum(f, 1.xxxx);
+     |          ^~~~~~~~~~~~~~~~~~
+```
+
+Neither case results in a diagnostic emission for the `ddy` call in `also_dead`
+because the function is not called and `ddy` is valid in the target shader model
+version.
 
 Because bytecode validation occurs late (after optimization), some more complex
 cases can occur. For example, users have reported situations that produce
--- a/proposals/infra/INF-NNNN-availability-diagnostics.md
+++ b/proposals/infra/INF-NNNN-availability-diagnostics.md
@@ -1,0 +1,209 @@
+<!-- {% raw %} -->
+
+# Strict Availability Diagnostics
+
+* Proposal: [INF-NNNN](INF-NNNN-availability-diagnostics.md)
+* Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
+* Sponsor: [Chris Bieneman](https://github.com/llvm-beanz)
+* Status: **Under Consideration**
+* Planned Version: 20YY (Clang-only)
+
+## Introduction
+
+Strict availability diagnostics introduces two new modes for diagnosing the use
+of unavailable shader APIs. Unavailable shader APIs are APIs that are exposed in
+HLSL code but are not available in the target shader stage or shader model
+version.
+
+## Motivation
+
+Today the enforcement of API availability is the responsibility of bytecode
+validators, runtimes and drivers operating on optimized shader code. Depending
+contextually on where this enforcement occurs this can result in errors being
+produced either late in compilation when mapping back to source locations is
+difficult or at runtime when the runtime or driver rejects the compiled shader.
+
+Strict availability diagnostics provides more actionable diagnostics for users
+by both having detailed error messages and accurate message locations. It also
+provides diagnostics early and allows implementations to generate meaningful
+diagnostics even without generation of debug information.
+
+> Note: The DXIL validator requires line table information to generate error
+> locations which can be inaccurate. Disabling line table information can have a
+> significant compile-time benefit.
+
+## Proposed solution
+
+### Availability Annotations
+
+In Clang, most shader APIs have header declarations that live in the default
+included `hlsl_intrinsics.h`. All shader API declarations have Clang
+availability attributes denoting version information for when the APIs are
+introduced, deprecated, and removed. This version annotation can be per-stage or
+for all stages. These annotations exist on declarations to power IDE tooling
+through clangd. Regardless of this proposal the availability annotations must be
+present.
+
+As an example of the annotations in action take the following declaration from
+`hlsl_intrinsics.h`:
+
+```c++
+__attribute__((availability(shadermodel, introduced = 6.0)))
+__attribute__((clang_builtin_alias(__builtin_hlsl_wave_active_count_bits)))
+uint WaveActiveCountBits(bool Bit);
+```
+
+> Note: the actual header uses macros to condense the attribute descriptions.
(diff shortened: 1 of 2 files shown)


proposals/infra/INF-NNNN-availability-diagnostics.md
```diff
@@ -0,0 +1,209 @@
...
+functions which are not reachable from exported functions.
+
+### Strict Diagnostic Mode
+
+The implementation of the strict diagnostic mode matches the relaxed mode,
+except that when a `CallExpr` references an unavailable API, the compiler emits
+an _error_.
+
+The strict mode does not issue diagnostics for `CallExpr` nodes that are inside
+functions which are not reachable from exported functions.
```
bogner, line 95: Technically the relaxed and strict modes can be implemented simply as a default-error warning, and selected explicitly with `-Werror=/-Wno-error=`. As such I'm not entirely convinced it makes sense to have a second way to do that via the `-fhlsl-diagnostic-mode` flag. I could be convinced it's worthwhile for consistency with the pedantic mode, but it's worth thinking about.

proposals/infra/INF-NNNN-availability-diagnostics.md
```diff
@@ -0,0 +1,209 @@
...
+availability annotation that marks it as unavailable for the target shader model
+and stage will produce an _error_.
+
+For library shaders, during parsing any callee of a `CallExpr` that has
+availability annotation that marks it as unavailable for the target shader model
+version will produce an _error_. After parsing the translation unit, an AST
+visitor will traverse to all `CallExpr` nodes that are reachable from annotated
+shader entry functions. If the callee of a `CallExpr` has availability
+annotations that signify that the API is unavailable for the target shader stage
+the compiler emits an _error_.
```
bogner, line 110: Are strict and pedantic at all different for library shaders? The explanations sound the same.

proposals/infra/INF-NNNN-availability-diagnostics.md
```diff
@@ -0,0 +1,209 @@
...
+
+Because bytecode validation occurs late (after optimization), some more complex
+cases can occur. For example, users have reported situations that produce
+validation errors when building un-optimized shaders for debugging proposes that
+do not appear with optimized shaders.
+
+One presentation of such a case is if a call to an unavailable API occurs under
+control flow. If the compiler optimizes away the control flow it may remove the
+API call. In these cases shaders that compile and verify successfully with DXC
+may produce warnings or errors with Clang.
```
bogner, line 207: The example of the difference between strict and pedantic is kind of intermixed with the explanation of how this is different from how things work in dxc. It might be best to put the strict/pedantic example immediately after the definition of pedantic, then have the discussion of the validator and the differences in behaviour between early and late diagnostics.

proposals/infra/INF-NNNN-availability-diagnostics.md
llvm-beanz: Yes, but I can elaborate.

proposals/infra/INF-NNNN-availability-diagnostics.md
llvm-beanz: I haven't addressed this comment (yet), although I further expanded examples. I'm thinking about how best to restructure the document.

damyanp (approved):
IMO it'd be good to get this out of PR mode and into resolving-issues mode so that we've got a sample proposal to look at in main.

proposals/infra/INF-NNNN-availability-diagnostics.md
```diff
@@ -0,0 +1,262 @@
...
+<>:9:6: error: 'WriteSamplerFeedbackLevel' is available beginning with Shader Model 6.6
+   9 |   map.WriteSamplerFeedbackLevel(texture, samplerState, uv, threadId.x % 8);
+     |       ^~~~~~~~~~~~~~~~~~~~~~~~~
+```
+
+In this case the DXIL validator will emit no diagnostic since the call to the
+unavailable API is not present in the final output.
+
+In all three cases above the DXIL validator will also emit the same diagnostic
+as it does today.
```
hekota, line 202: This is confusing as it follows the sentence "In this case the DXIL validator will emit no diagnostic".

proposals/infra/INF-NNNN-availability-diagnostics.md
```diff
@@ -0,0 +1,262 @@
+<!-- {% raw %} -->
```
hekota, line 1: Typo in title - diagnostics ?

proposals/infra/INF-NNNN-availability-diagnostics.md
```diff
@@ -0,0 +1,262 @@
...
+}
+```
+
+When compiled with the `lib_6_3` profile under the _default_ mode, clang will
+emit the following error:
+
+```
+<>:1:9: error: 'ddx' is available for pixel shaders beginning with Shader Model 2.0 
+   9 |   return ddx(f);
+     |          ^~~
```
hekota, line 233: It might be more clear to say ```error: 'ddx' is not available in library shaders``` 

Do we have a way to encode intrinsics availability depending on the shader stage? The ddx/ddy are tricky as they are available in pixel shaders since 2.0+, but for compute, mesh and amplification shaders since 6.6+.

proposals/infra/INF-NNNN-availability-diagnostics.md
```diff
@@ -0,0 +1,262 @@
...
+introduced, deprecated, and removed. This version annotation can be per-stage or
+for all stages. The availability annotations will need to exist on declarations
+regardless of this proposal, as they provide information to AST-based tooling
+such as IDE workflows.
+
+As an example of the annotations in action take the following declaration from
+`hlsl_intrinsics.h`:
+
+```c++
+__attribute__((availability(shadermodel, introduced = 6.0)))
```
Keenuts, line 51: Shall we have a similar decoration for Vulkan? (Or other APIs?)
```
__attribute__((availability(vulkan, introduced = 1.1)))
```

proposals/infra/INF-NNNN-availability-diagnostics.md
```diff
@@ -0,0 +1,262 @@
...
+Strict availability diagnostics provides more actionable diagnostics for users
+by both having detailed error messages and accurate message locations. It also
+provides diagnostics early and allows implementations to generate meaningful
+diagnostics even without generation of debug information.
+
+> Note: The DXIL validator requires line table information to generate error
+> locations which can be inaccurate. Disabling line table information can have a
+> significant compile-time benefit.
+
+## Proposed solution
```
Keenuts, line 35: Shall those proposals have a `Limits` sections?

For example, this wouldn't cover cases where the limitation comes from differences in the API we use:

For example:
```
groupshared int runtimeArray[0];
```

This is allowed in DXIL, unbounded, groupshared array which size is defined at runtime.
But this cannot be implemented in Vulkan/SPIR-V due to some restrictions that forbids runtime arrays for `groupshared` variables.
This not being a builtin/intrinsic, this attribute cannot be used, and we still need to have some ad-hoc availability diagnostic for such cases.

hekota: Unless there is a common pattern these special cases will probably need to be handled separately by targeted diagnostics for each API that differs.

proposals/infra/INF-NNNN-availability-diagnostics.md
hekota: We could add new clang attribute to annotate intrinsics availability per shader stage:

```c++
__attribute__((availability_shader_stage(pixel, introduced = 2.0)))
__attribute__((availability_shader_stage(compute, introduced = 6.6)))
float ddx(float);
```

proposals/infra/INF-NNNN-availability-diagnostics.md
tex3d: This shouldn't be legal DXIL, but apparently we didn't prevent it in the compiler or validator.  I believe this is a bug.

> This is allowed in DXIL, unbounded, groupshared array which size is defined at runtime.

Have you seen this in the wild?  What do you mean when you say "size is defined at runtime"?  How is it defined at runtime?  I am not aware of any officially supported mechanism to do this.

proposals/infra/INF-NNNN-availability-diagnostics.md
```diff
@@ -0,0 +1,262 @@
...
+Compiled with the `-T cs_6_4` flag this produces the validation error:
+```
+<>:11:3: error: Opcode WriteSamplerFeedbackLevel not valid in shader model cs_6_4.
+note: at 'call void @dx.op.writeSamplerFeedbackLevel(i32 176, %dx.types.Handle %1, %dx.types.Handle %2, %dx.types.Handle %3, float %8, float %9, float undef, float undef, float %11)' in block '#0' of function 'main'.
+Validation failed.
+```
+
+This error displays the LLVM IR text for the error site. In a trivial example
+like the one above finding the error is easy. In a complex shader that could be
+thousands of lines long this is unmanageable.
```
tex3d, line 149: Note that the error does have the source location, but since way the compiler is used here provides no source file name, you only get `<>:11:3:`.

proposals/infra/INF-NNNN-availability-diagnostics.md
tex3d: I agree that this is confusing.  If you use an intrinsic in a shader stage that is allowed in that shader stage, but in a newer shader model, then you should get a diagnostic that's relevant to the entry point type, not one that tells you about when it was first introduced elsewhere.

We also might need to add details about optional features required, especially if we want to move to a model where you must explicitly specify allowed features to prevent accidentally using optional features you didn't intend to use (which would be more consistent with SPIR-V targets).  That's directly relevant to this intrinsic because ddx is also usable in mesh and amplification shaders in SM 6.6+, but there it requires an optional feature bit to be set.

tex3d: I should mention that I don't agree with the error suggested above though.  I think it should say something more like: `error: 'ddx' is not available in vertex shaders`, since that's the entry function type that's violating the constraint.

If it were used in a compute shader, it should say something like: `error: 'ddx' is available for compute shaders beginning with Shader Model 6.6`.

proposals/infra/INF-NNNN-availability-diagnostics.md
Keenuts: It's an extension of what I could find on the docs 😊
Context: https://github.com/microsoft/DirectXShaderCompiler/issues/3667#issuecomment-2025620965

But this is one specific example, just wanted to check if those specifications should have a "limits" of "issue" section. This way we can list tradeoffs or known limits to we still have to deal with.

proposals/infra/INF-NNNN-availability-diagnostics.md
llvm-beanz: I think we can do this without adding a new attribute. The existing availability attribute has some sub-platform cases today for Apple's "App Extensions".

We can add `shadermodel_pixel` as a supported parameter to the existing attribute and get the support we need through one attribute.

proposals/infra/INF-NNNN-availability-diagnostics.md
llvm-beanz: Yea. I think this is out of scope for this document. Resource types are kinda special in that we allow unsized arrays, but all resource handles need to be fully scalarized, so they aren't really generated as unbounded arrays.

I think for the purposes of this document we should keep the focus narrow to API availability not other broader capability differences.

proposals/infra/INF-NNNN-availability-diagnostics.md
llvm-beanz: Yes, that's true for all the examples not just the DXC validator one. Godbolt gives the filename as `<source>`, I truncated that down to just `<>` for brevity since it isn't relevant.

While I think that generally this class of diagnostics (availability of DXIL operations) is one of the more-stable diagnostic locations for the DXIL validator, the diagnostic would be way more approachable to users if it was generated on the AST and referred to the HLSL source. That's the core goal of this proposal.

proposals/infra/INF-NNNN-availability-diagnostics.md
hekota: We are going to need to add at least `shadermodel_pixel`, `shadermodel_compute`, `shadermodel_mesh`, `shadermodel_amplification` just to cover just the derivate intrinsics. Let's see if we can get away with adding so many new values.

llvm-beanz: > We are going to need to add at least `shadermodel_pixel`, `shadermodel_compute`, `shadermodel_mesh`, `shadermodel_amplification` just to cover just the derivate intrinsics. Let's see if we can get away with adding so many new values.

I don't think that will be a problem. This isn't really a significant maintenance burden to the compiler.

llvm-beanz: The existing diagnostics in clang take this form:

```
<>:2:12: warning: 'WaveActiveCountBits' is only available on HLSL ShaderModel 6.0 or newer [-Wunguarded-availability-new]
    2 |     return hlsl::WaveActiveCountBits(b);
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~
<>/hlsl/hlsl_intrinsics.h:1374:34: note: 'WaveActiveCountBits' has been marked as being introduced in HLSL ShaderModel 6.0 here, but the deployment target is HLSL ShaderModel 5.0
 1374 | __attribute__((convergent)) uint WaveActiveCountBits(bool Val);
      |    
```
[Compiler Explorer](https://godbolt.org/z/sqKfbbK9j) (ignore the unrelated error)

It is probably nice to keep the error text brief and elaborate in a note.

tex3d: > The example above was compiled with the lib_6_3 profile, not a vertex shader.
While that's true about the profile, the reason it's an error is because it's used in a vertex shader entry function in that library.

The closest thing to a "library shader" is an exported function in a library.  That would have a different message still, such as: `error: 'ddx' is not available for use in exported functions`.  The term "library shader" isn't really useful to refer to any shader in a library, since it's not specific enough.

Ideally, there should be no difference between what's allowed in a vertex shader entry function compiled to vs_6_4 and that same function compiled into a lib_6_4 library, thus the diagnostic language should also be the same, or as close as possible (if there's some justified difference).

hekota: > We also might need to add details about optional features required, especially if we want to move to a model where you must explicitly specify allowed features to prevent accidentally using optional features you didn't intend to use (which would be more consistent with SPIR-V targets). That's directly relevant to this intrinsic because ddx is also usable in mesh and amplification shaders in SM 6.6+, but there it requires an optional feature bit to be set.

I would like to scope this proposal to HLSL intrincisc availability as related to shader model versions. We should have a separate discussion/proposal about availability diagnostics related to target runtimes (like Vulcan support @Keenuts mentioned below) and how to specify required optional features in HLSL.

llvm-beanz: Filed an issue to track clarifying and reworking the diagnostic text:
https://github.com/microsoft/hlsl-specs/issues/188

llvm-beanz: > I would like to scope this proposal to HLSL intrincisc availability as related to shader model versions. We should have a separate discussion/proposal about availability diagnostics related to target runtimes (like Vulcan support @Keenuts mentioned below) and how to specify required optional features in HLSL.

I don't agree with excluding Vulkan support.

I think we should have the same diagnostic infrastructure for all runtimes, and I think that is very reasonable without really expanding the design scope.

The only thing we should need to do is annotate the APIs with Vulkan availability in addition to Shader Model availability. I'm not suggesting that you we do that as part of this effort, but perhaps @keenuts can help identify and annotate the APIs we've already added in Clang.

hekota (approved):
LGTM!

tex3d (approved):
I think this is plenty for adding the proposal, where we can continue to discuss and iterate on more details after that.

bogner (approved):
This looks reasonable. It would probably be good to elaborate on the diagnostic text and add some examples of the "note" diagnostics we expect to be present, but I think it's reasonable to track that addition in https://github.com/microsoft/hlsl-specs/issues/188.

proposals/infra/INF-NNNN-availability-diagnostics.md
tex3d: Aren't there constraints on availability of these intrinsics in different execution environments for Vulkan as well, similar to different shader stages in DX?  Wouldn't we want to express that here as well?

proposals/infra/INF-NNNN-availability-diagnostics.md
```diff
@@ -0,0 +1,272 @@
...
+__attribute__((clang_builtin_alias(__builtin_hlsl_wave_active_count_bits)))
+uint WaveActiveCountBits(bool Bit);
+
+// This function is available starting in SM 2.0 for pixel shaders, but SM 6.6
+// for compute, mesh and amplification shaders.
+__attribute__((availability(shadermodel_pixel, introduced = 2.0)))
+__attribute__((availability(shadermodel_compute, introduced = 6.6)))
+__attribute__((availability(shadermodel_mesh, introduced = 6.6)))
+__attribute__((availability(shadermodel_amplification, introduced = 6.6)))
+__attribute__((availability(vulkan, introduced = 1.0)))
```
tex3d, line 62: Given the potential for Vulkan execution environment availability to differ from DX shader stages and shader models, might we want to have more specific terms for "shadermodel_*" that could apply separately to DX or Vulkan, invoking the applicable target in the name?  Do these shader models and versions have a 1:1 mapping to something for Vulkan targets?

proposals/infra/INF-NNNN-availability-diagnostics.md
llvm-beanz: I did update the doc with a `vulkan` annotation.

proposals/infra/INF-NNNN-availability-diagnostics.md
llvm-beanz: I’m not sure I understand what you’re saying. I’ve added a `vulkan` availability annotation, if we need per-stage vulkan annotations we can add those too.

proposals/infra/INF-NNNN-availability-diagnostics.md
llvm-beanz: I think the remaining issues here are tracked in #188, so I’ll resolve the comment.

proposals/infra/INF-NNNN-availability-diagnostics.md
tex3d: The main thing was that, to me, "shadermodel" sounds more generic than it should be, like it doesn't reference DX somehow to differentiate it from a different target.  In other words, if we have to add something like `vulkan_fragment`, maybe `shadermodel_pixel` should be named something more like `dx_shadermodel_pixel` to make it clear that this shadermodel is a DX-specific thing.

It could just be me not fully understanding the target topology here.  Perhaps shadermodel is something meaningful for Vulkan targets as well?

proposals/infra/INF-NNNN-availability-diagnostics.md
tex3d: Yes, I see that.  I was thinking that if we had shadermodel_pixel, and there are similar (but slightly different) constraints for Vulkan, like for "fragment", then wouldn't we expect some breakdown of `vulkan` into `vulkan_fragment`, etc...?  Or am I misunderstanding how this sort of thing would be described?

llvm-beanz: If there are APIs that are restricted by vulkan execution model (the SPIR-V terminology for DX shader stages), we should add appropriate vulkan values. I did a quick scan of the SPIR-V and GLSL extensions and nothing stood out, but it was by no means exhaustive.

proposals/infra/INF-NNNN-availability-diagnostics.md
llvm-beanz: “shadermodel” is already claimed for the DirectX shader model and appears in the target triple, as does “vulkan”. We could consider renaming “shadermodel” to “directx” although I think it is fine as it is. We’ve claimed the term so we don’t need a separate prefix.

proposals/infra/INF-NNNN-availability-diagnostics.md
tex3d: Yeah, this is fine for now.

I'm not sure what you meant by scanning extensions.  Aren't there plenty of operations only available under certain execution models (not needing to consider extensions)?  Like the equivalent of ddx is probably not available under all execution models, right?

proposals/infra/INF-NNNN-availability-diagnostics.md
tex3d: Actually, come to think of it, I don't think this is the best approach to mark this kind of feature support.  I think we should be identifying feature groupings and tagging operations with that feature group, which would be driven by the details of shader model and target.  For instance, ddx would have something more like `__attribute__((availability(hlsl_derivatives)))`, then `hlsl_derivatives` availability depends on shader model, stage, and potentially even available feature sets (like opting-in to mesh/amp shader derivatives).

What do you think?

llvm-beanz: The availability attribute is designed to operate on the target triple directly, and I think we should keep that pattern. If we want to add further groupings of features, we can do that directly in the hlsl_intrinsics header using preprocessor macros to construct the different attribute formulations. That allows us to keep unnecessary complexity out of the compiler and keeps our use of the attribute consistent with its design and implementation.

Keenuts: For the per-stage bit, commented on the other thread.
As for the `shadermodel` vs `directx/dx_shadermodel`, I'm fine with both. I agree with Chris that `shadermodel` is already a claimed DX specific name.
And no, we don't have a 1:1 mapping between shadermodel & vk version, which is why using the `cs_X_Y` for DXC when building SPIR-V saddens me, but that's another story.

So fine for me to have DX use `shadermodel_*` for this.

proposals/infra/INF-NNNN-availability-diagnostics.md
Keenuts: Thanks Tex! Yes, we'd also need to have a `_stage` suffix for the availability, as you have for `shadermodel_`.

Keenuts (approved):
Just the bit around `vulkan_stage` suffix, otherwise LGTM, thanks!

PR #155, opened 2024-01-17
Propose Mesh Nodes for Work Graphs

tex3d:
Mesh nodes is the mesh shader subset of the proposed graphics nodes functionality proposed in the current Work Graphs spec [here](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#mesh-nodes).

This proposal describes the HLSL and DXIL aspects for compiler implementation and refers to the Work Graphs spec for contextual details.  This only covers this mesh shader subset of the feature intended for an experimental preview.


proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,302 @@
...
+shaders provide the function of generating the mesh node dispatches on the GPU.
+
+Mesh nodes can also be used as one or more entrypoints in a work graph
+(standalone since they can't output to other nodes),
+in which case the input record (payload) is supplied from DispatchGraph()
+at the API just like any other work graph entrypoint input.
+
+A complete mesh node is defined at the API by first creating a
+[generic program](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#d3d12_generic_program_desc)
+including the mesh node function name, a pixel shader, and other optional state,
```
python3kgae, line 67: What are these optional states?

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,302 @@
...
+##### Input System Values
+
+For input values outside the input record, mesh nodes support the `broadcasting` launch [node shader system values](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#node-shader-system-values).
+These are the same [system values supported by mesh shaders](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/MeshShader.md#hlsl-attributes-and-intrinsics).
+
+| system value semantic | type    | description |
+|-----------------------|---------|-------------|
+| `SV_GroupThreadID`    | `uint3` | Thread ID within group |
+| `SV_GroupIndex`       | `uint`  | Flattened thread index within group |
+| `SV_GroupID`          | `uint3` | Group ID within dispatch |
```
amarpMSFT, line 181: SV_ViewID is another input available to mesh shaders.

amarpMSFT: Same comment applies to earlier above where the system values are listed.

proposals/xxxx-mesh-nodes.md
tex3d: You need to see the Work Graphs spec for complete details here.  This was just an attempt to briefly summarize the context into which this shader function fits.  There are other subobject types for defining parts of the graphics pipeline state which could be added to the program as well.  See subobject types listed [here](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#d3d12_state_subobject_type) for some relevant graphics-related state that might be used in addition to the shaders.

amarpMSFT (changes requested):

proposals/xxxx-mesh-nodes.md
amarpMSFT: One example is a pixel shader, which can be defined via ps_6_*.  Other examples are rasterizer state, blend state, depth stencil state, long list.  In the future we should add support for defining all those subobjects in DXIL so they don't have to be defined at the API.  For now, one has to define some things in the API since not everything can be expressed in HLSL.  As an example there is support for defining raytracing subobjects in HLSL already. The runtime extracts these and turns them in to DDI subobject descriptions that are equivalent to API ones, so we wouldn't even need drivers to change.

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,302 @@
...
+  // Cooperate across group to write output arrays...
+}
+```
+
+### Compiling Experimental Mesh Node Shaders
+
+As a kind of node shader in work graphs, mesh node shaders are functions
+defined in a library.
+To use the experimental feature, you must compile HLSL with options:
+`-T lib_6_9 -select-validator internal`.
```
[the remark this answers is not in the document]
tex3d: Ok, but that seems to be an API detail at this point, doesn't it?  That's why I heavily linked to the Work Graphs spec.

Also, what could you do without a pixel shader?  Is that valid?

proposals/xxxx-mesh-nodes.md
tex3d: Good catch!  However, considering this doesn't yet support ViewID attribute expansion validation data for the runtime, should we even have ViewID support in the prototype?  I could call out this system value as unsupported, or we could support it and call out the lack of space validation for ViewID-expanded attributes.

proposals/xxxx-mesh-nodes.md
python3kgae: It feels like more than just mesh here. Is it possible to call this something like graphics node instead of mesh node?

tex3d: This was basically a "dispatchmesh" launch node, which was shortened to "mesh" because it was cleaner and there wasn't a good reason not to.  DispatchMesh launches a graphics pipeline of a certain kind, which does require associated state.  The mesh node shader defined here is really just the mesh node shader portion, requiring the rest of the state to set up a graphics pipeline (or "program") to run.

The term "graphics node" would imply more generality which is proposed in the work graphs spec (see various links), including the ability to launch the full legacy graphics pipeline, starting with some vertex shader-like node instead of a mesh shader-like node.  Even those would need different NodeLaunch types to indicate the type of the pipeline, such as "draw" or "drawindexed", then "mesh" is still needed to distinguish the node type being defined here from the others.

This proposal focuses only on defining the mesh shader part of the graphics nodes proposal.  Hence, "mesh launch node" or "mesh node" for short.

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,329 @@
...
+[NodeLocalRootArgumentsTableIndex(4)]
+[NumThreads(4,4,1)]
+[NodeMaxDispatchGrid(16,16,1)]
+[OutputTopology("triangle")]
+void MyMeshNode(
+    DispatchNodeInputRecord<MyMeshRecord> nodeInput,
+    out vertices PerVertexValues verts[16],
+    out primitives PerPrimitiveValues prim[8],
+    out indices uint3 tris[16] )
+{
```
python3kgae, line 106: Is it OK to call a normal mesh shader entry inside a mesh node to share code?

tex3d: I don't think that would be a good idea, even if the compiler allows you to do so.  Technically, it might work out (by accident) with inlining/optimization/copy elimination, but there's nothing in the language to preserve special argument handling to avoid copying everything between shared memory record/arrays and parameters that will be considered thread-local for the called mesh shader entry.  We could adjust internals to better support that scenario, but I don't think that's a priority at this time.  References (and appropriately assigning a different address space to params) could make a shared function called by both entry function types work properly in the future.

damyanp:
Looks like the D3D spec now refers to a `NodeMaxInputRecordsPerGraphEntryRecord` attribute (https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#helping-mesh-nodes-work-better-on-some-hardware).

Suggest we get some text in the spec mentioning this, or an issue filed to add it later?

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,329 @@
...
+then referring to that program in a
+[work graph program node](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#d3d12_program_node),
+which may optionally override node properties, such as the node ID.
+
+See the [DispatchMesh Launch Nodes](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#dispatchmesh-launch-nodes)
+section of the work graphs spec, along with sections linked from there, to
+complete the picture.
+
+In the future, we may add subobjects to specify more of these details within
+the DXIL library for convenience.
```
pow2clk, line 80: This is another detail that I feel belongs in a dedicated section.

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,329 @@
...
+void MyMeshNode(
+    DispatchNodeInputRecord<MyMeshRecord> nodeInput,
+    out vertices PerVertexValues verts[16],
+    out primitives PerPrimitiveValues prim[8],
+    out indices uint3 tris[16] )
+{
+  float3 data = nodeInput.Get().perMeshConstant;
+  // compute numVerts, numPrims...
+  SetMeshOutputCounts(numVerts, numPrims);
+  // Cooperate across group to write output arrays...
```
pow2clk, line 110: I'm not sure it's clear that this implies that additional code should be here to do this. Maybe a dummy function called "writeOutputArrays()" or something.

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,329 @@
...
+for more detailed descriptions of the node function attributes.
+
+| Attribute | Required | Description |
+|--|--|--|
+| `[shader("node")]` | Y | Indicates a node shader entry point |
+| `[NodeLaunch("mesh")]` | Y | Signifies a mesh node |
+| `[NumThreads(x,y,z)]` | Y | Specifies the launch size of the threadgroup of the Mesh shader, just like with compute shader. The number of threads can not exceed `X * Y * Z = 128`. See [numthreads](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/MeshShader.md#numthreads) in the Mesh Shader spec. |
+| `[OutputTopology(`*topology*`)]` | Y | Specifies the topology for the output primitives of the mesh shader. *topology* must be `"line"` or "`triangle"`. See [outputtopology](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/MeshShader.md#outputtopology) in the Mesh Shader spec. |
+| `[NodeDispatchGrid(x,y,z)]` | Yes, unless `NodeMaxDispatchGrid` is defined | Declare a fixed dispatch grid size for use with this node. ([details](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#shader-function-attributes)) |
```
pow2clk, line 150: Nit: Since this is widening the column anyway, the rest might as well be "Yes" and "No" instead of just "Y" and "N"

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,329 @@
...
+| `[NodeID("nodeName")]` or `[NodeID("nodeName",arrayIndex)]` | N | Name for the node; uses function name if omitted. Optional `uint arrayIndex` overrides the default index of `0`. ([details](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#shader-function-attributes)) |
+| `[NodeLocalRootArgumentsTableIndex(index)]` | N | `uint index` indicates the record index into the local root arguments table bound when the work graph is used. May be omitted or set to `-1` (equivalent to omitting).  If omitted and a local root signature is used, the runtime will auto-assign the index. ([details](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#shader-function-attributes)) |
+| `[NodeShareInputOf("nodeName")]` or `[NodeShareInputOf("nodeName",arrayIndex)]` | N | Share the input of the specified NodeID with this node. ([details](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#shader-function-attributes)) |
+
+#### Function Parameters
+
+##### Input Record
+
+Zero or one input record declaration is supported for the mesh node entry.
```
pow2clk, line 161: Might bear mentioning here that the record must be read-only
```suggestion
Zero or one read-only input record declaration is supported for the mesh node entry.
```

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,329 @@
...
+Any other semantics on fields in the record structure are ignored.
+
+See the work graphs spec
+[Node input declaration](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#node-input-declaration)
+section under options for broadcasting launch for more detail.
+
+##### Input System Values
+
+For input values outside the input record, mesh nodes support the `broadcasting` launch [node shader system values](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#node-shader-system-values).
+These are the same [system values supported by mesh shaders](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/MeshShader.md#hlsl-attributes-and-intrinsics).
```
pow2clk, line 184: I think you wanted this anchor: 
```suggestion
These are the same [system values supported by mesh shaders](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/MeshShader.md#system-value-semantics).
```
I think the intent is to say that the meaning of those supported are the same as they are in mesh shaders since there is one omission in ViewID though it is, somewhat strangely, not listed in the system values section of that spec

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,329 @@
...
+These arrays address memory shared across the group of threads, just like
+`groupshared` memory.
+The intrinsic function [`SetMeshOutputCounts`](#setmeshoutputcounts) must be called before writing to
+any of the output arrays.
+
+| Mesh shader shared output arrays                      | Array dimension defines maximum number of | Required  |
+|-|-|-|
+| vertex [`indices`](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/MeshShader.md#vertex-indices)                   | primitives    | Required  |
+| attributes for [`vertices`](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/MeshShader.md#vertex-attributes)       | vertices      | Required  |
+| attributes for [`primitives`](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/MeshShader.md#primitive-attributes)  | primitives    | Optional  |
```
pow2clk, line 211: another nit: unlike above where the "Required" column is filled in with Y and N, here we have the "Required" and "Optional". I don't really care which, but I prefer consistency.

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,329 @@
...
+|-----------------------------|----------|-----------|----------|-------------|
+| `SV_Position`               | `float4`     | Y | `vertices`   |  |
+| `SV_RenderTargetArrayIndex` | `uint`       | N | `primitives` |  |
+| `SV_ViewPortArrayIndex`     | `uint`       | N | `primitives` |  |
+| `SV_ClipDistance`           | `float<1-4>` | N | `vertices`   |  |
+| `SV_CullDistance`           | `float<1-4>` | N | `vertices`   |  |
+| `SV_ShadingRate`            | `uint`       | N | `primitives` |  |
+| `SV_PrimitiveID`            | `uint`       | N | `primitives` | See [mesh shader spec](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/MeshShader.md#sv_primitiveid-in-the-pixel-shader) |
+| `SV_CullPrimitive`          | `uint`       | N | `primitives` | See [mesh shader spec](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/MeshShader.md#sv_cullprimitive) |
+
```
pow2clk, line 233: It may bear mentioning that all of the node-style outputs are invalid here.

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,329 @@
...
+3. If the compiler can prove that this function is not called, then the threadgroup doesn't have any output. If the shader writes to any of the shared output arrays, compilation and shader validation will fail. If the shader does not call any of these functions, the compiler will issue a warning, and no rasterization work will be issued.
+4. Only the input values from the first active thread are used.
+5. This call must dominate all writes to shared output arrays. In other words, there must not be any execution path that even appears to reach any writes to any output array without first having executed this call.
+
+See [SetMeshOutputCounts](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/MeshShader.md#setmeshoutputcounts)
+in the mesh shader spec for specific details, restrictions, and examples.
+
+### Interchange Format Additions
+
+To the `DXIL::NodeLaunchType` enum, add `Mesh` (4).
```
pow2clk, line 259: Mentioning how this is represented in the metatdata tag explicitly would make it clear how this applies to the interchange format

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,329 @@
...
+- `MaxOutputPrimitives` - based on `primitives` array size
+- `MeshOutputTopology` - `uint8_t` encoded `DXIL::MeshOutputTopology` based on `[OutputTopology(`*topology*`)]`
+
+#### Device Capability
+
+See the work graphs spec
+[here](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#graphics-nodes)
+for detail.
+
+Devices that support experimental `D3D_SHADER_MODEL_6_9` and experimental [`D3D12_WORK_GRAPHS_TIER_1_1`](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#d3d12_work_graphs_tier) are required to support these features as part of graphics nodes in work graphs.
```
pow2clk, line 325: Presumably mesh shaders would need to be supported too?

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,329 @@
...
+
+## Open Questions
+
+- Should vanilla mesh shaders be supported, or only mesh nodes?
```
pow2clk, line 329: I don't know what this implies. can you elaborate?

pow2clk (commented):
Lots of nitpicks! Sorry I forgot to push submit on this

proposals/xxxx-mesh-nodes.md
amarpMSFT: Apps don't have to check for mesh shader support, as it's implicitly a requirement for work graphs tier 1.1.  You could mention it that way (I'll also clarify in work graphs spec).

proposals/xxxx-mesh-nodes.md
amarpMSFT: This is asking if the path the initial prototype allowed should go forward -> allowing existing mesh shaders to be used, with payload acting as node input record.   I think the answer here has to be no, since existing mesh shaders don't include some of the metadata that the runtime needs in order to enable all ways of connecting into a graph.  For instance there isn't metadata indicating where SV_DispatchGrid is located in the input record, which the runtime needs to tell driver how to handle entrypoint input to a node, so existing mesh shaders can't be used at the entry of a graph.  Might as well keep it simple and only support mesh nodes.  Similarly we kept it simple and didn't expose vanilla compute shaders.

amarpMSFT (approved):

proposals/xxxx-mesh-nodes.md
tex3d: That's a nice idea, but writing this correctly is harder.  You do not want to pass the vertices, primitives, or indices arrays to a function to be written.  Rather, you need to do this in the main function, otherwise, you have per-thread copy-in/copy-out semantics on each array, when those arrays are essentially group shared.

I could write some code that assigns array elements values based on something like SV_GroupIndex that I add.  Like so:
```cpp
  // Cooperate across group to write output arrays...
  if (gi < numVerts) {
    verts[gi] = ComputeVertexValues(gi);
  }
  if (gi < numPrims) {
    prims[gi] = ComputePrimValues(gi);
    tris[gi] = ComputeTri(gi);
  }
```

proposals/xxxx-mesh-nodes.md
```diff
@@ -0,0 +1,361 @@
...
+[here](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#mesh-nodes)
+for detail.
+
+Devices that support mesh shaders, as well as experimental `D3D_SHADER_MODEL_6_9` and experimental [`D3D12_WORK_GRAPHS_TIER_1_1`](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#d3d12_work_graphs_tier) are required to support these features as part of graphics nodes in work graphs.
+
+## Open Questions
+
+- Should vanilla mesh shaders be supported, or only mesh nodes?
+  - No, only mesh nodes will be supported.  Ordinary mesh shaders lack key information, such as the `SV_DispatchGrid` location in the input record.
+- `[NodeMaxInputRecordsPerGraphEntryRecord(...)]` if proposed [here](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#helping-mesh-nodes-work-better-on-some-hardware).  Should this attribute be added to HLSL?
```
damyanp, line 357: typo? maybe that should read `is proposed [here]`?

damyanp (approved):
Looks good to get into the repo!

pow2clk (approved):
Looks good! I appreciate the responses to my nitpicks, but I think the most recent draft makes everything flow better in ways I can claim no credit for.

PR #162, opened 2024-01-23
[Wave Size Range] Renumber spec

pow2clk:
Miscommunication left this with an explicit number and circumstances put
it in before the one currently (also inappropriately) tentatively
numbered 13. This renumbers it to 13 to be sequential.

What I should have done was create it as NNNN- until just before
merging. That's what I'll do in the future.


tex3d:
> This renumbers it to 13 to be sequential.

Oops.  14.  😉

tex3d (approved):

tex3d:
> It would have helped to point out the other unmerged proposal inappropriately numbered 13. I found it and added a comment here: https://github.com/microsoft/hlsl-specs/pull/154/files#r1464102827

Or was it this one here?
https://github.com/microsoft/hlsl-specs/pull/147/files#diff-1795d7ee29efc4747efe2f6e162d167f020fbe1f0390c9c8800b009be56ce8a4R3

pow2clk:
Yes to both and the one in my control will be addressed as well.

llvm-beanz (approved):

bob80905 (approved):

PR #165, opened 2024-01-25
[Wave Size Range] Revise metadata representation

### LINKED ISSUE #158: Should we collapse degenerate Wave Size Range values in DXIL 1.8 metadata
Regarding the 0014-wave-size-range proposal : 
https://github.com/microsoft/hlsl-specs/blob/627ae31a2fc9835a7752645f42759ae30e4ee7f6/proposals/0014-wave-size-range.md?plain=1#L137

see [comment](https://github.com/microsoft/hlsl-specs/pull/149#discussion_r1458114151).

This concerns whether we should require all three values in DXIL metadata for wave size range for the degenerate case that has the same meaning as the DXIL 1.6 form with only one value.

We could generate the 1-value metadata form for the degenerate case for DXIL 1.8+ as well, instead of always requiring 3 values, even though they don't convey any useful information.

pow2clk:
The biggest change here is that the RangedWaveSize tag can have a first element non-zero with the rest zero to represent the old WaveSizeTag behavior. Added some clarification of metadata restrictions and error generation as well.

Fixes #158

### DIFF
--- a/proposals/0013-wave-size-range.md
+++ b/proposals/0013-wave-size-range.md
@@ -146,10 +146,10 @@ Validation should confirm:
 * The third element (preferred wave size) is greater or equal to the first
  element (minimum wave size) and less than or equal to the second element
  (maximum wave size) or zero.
-* Shaders with with versions less than 6.8 fail on shaders that use
- `kDxilRangedWaveSizeTag`.
-* Shaders with versions greater than or equal to 6.8 fail on shaders that use
- `kDxilWaveSizeTag`.
+* Shaders with target shader model less than 6.8 that use 
+  `kDxilRangedWaveSizeTag` should fail.
+* Shaders with target sahder model greater than or equal to 6.8 that use
+  `kDxilWaveSizeTag` should fail.
 
 ### Runtime Additions
 
--- a/proposals/0013-wave-size-range.md
+++ b/proposals/0013-wave-size-range.md
@@ -83,9 +83,12 @@ Where `minWaveSize` is the minimum wave size supported by the shader
 
 ### DXIL Additions
 
-The existing `WaveSize` value is stored as a single metadata 32-bit integer
- constant.
-The new metadata tag expands this to a tuple of three 32-bit integers.
+The existing metadata `kDxilWaveSizeTag`(11)
+ that takes only a single 32-bit integer constant is invalid in 6.8 shaders,
+ but will continue to be supported in earlier shader model versions.
+
+The new metadata `kDxilRangedWaveSizeTag`(23) takes a tuple
+ of three 32-bit integers.
 These values represent all potential parameters to `WaveSize`.
 If all three values are non-zero, they represent the minimum, maximum,
  and preferred wave sizes respectively.
@@ -96,9 +99,10 @@ If only the first value is non-zero, it represents the legacy wave size as
 In this case, the single non-zero value is effectively minimum, maximum and
  preferred size in that it represents the only wave size supported by the shader.
 
-|         Tag             | Constant |         Value           |
-|-------------------------|----------|-------------------------|
-|kDxilRangedWaveSizeTag   |    23    |MD list: (i32, i32, i32) |
+|         Tag             | Constant |         Value           | Shader Models |
+|-------------------------|----------|-------------------------|---------------|
+|kDxilWaveSizeTag         |    11    |          i32            |     <6.8      |
+|kDxilRangedWaveSizeTag   |    23    |MD list: (i32, i32, i32) |    >=6.8      |
 
 ### SPIR-V Additions
 
@@ -142,9 +146,9 @@ Validation should confirm:
 * The third element (preferred wave size) is greater or equal to the first
  element (minimum wave size) and less than or equal to the second element
  (maximum wave size) or zero.
-* Validators with versions less than 1.8 fail on shaders that use
+* Shaders with with versions less than 6.8 fail on shaders that use
  `kDxilRangedWaveSizeTag`.
-* Validators with versions greater than or equal to 1.8 fail on shaders that use
+* Shaders with versions greater than or equal to 6.8 fail on shaders that use
  `kDxilWaveSizeTag`.
 
 ### Runtime Additions
--- a/proposals/0013-wave-size-range.md
+++ b/proposals/0013-wave-size-range.md
@@ -141,14 +141,14 @@ Validation should confirm:
 * Each element of that tuple is a power-of-two integer between 4 and 128
  inclusive or zero.
 * However, the first element (minimum wave size) is not zero.
-* The first element (minimum wave size) is less than the second
- (maximum wave size).
+* The second element (maximum wave size) is greater than the first element
+ (minimum wave size) or zero.
 * The third element (preferred wave size) is greater or equal to the first
  element (minimum wave size) and less than or equal to the second element
  (maximum wave size) or zero.
-* Shaders with target shader model less than 6.8 that use 
+* Shaders with target shader model less than 6.8 that use
   `kDxilRangedWaveSizeTag` should fail.
-* Shaders with target sahder model greater than or equal to 6.8 that use
+* Shaders with target shader model greater than or equal to 6.8 that use
   `kDxilWaveSizeTag` should fail.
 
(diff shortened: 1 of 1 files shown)


proposals/0013-wave-size-range.md
```diff
@@ -83,11 +83,18 @@ Where `minWaveSize` is the minimum wave size supported by the shader
...
 
-The existing `WaveSize` value is stored as a metadata 32-bit integer constant.
-To store the additional values, an additional metadata tag indicating a tuple of
- three 32-bit integers representing the given parameters to `WaveSize`.
-In the case where no preferred wave size is specified, the value zero will
- indicate that nothing was specified.
+The existing `WaveSize` value is stored as a single metadata 32-bit integer
+ constant.
+The new metadata tag expands this to a tuple of three 32-bit integers.
+These values represent and and all parameters to `WaveSize`.
```
bob80905, line 89: and typo

pow2clk: woops! Not entirely sure what I was going to put here, but I'll put something useful in its place.

proposals/0013-wave-size-range.md
```diff
@@ -125,16 +132,20 @@ These are where new or slightly altered errors are produced:
...
+* Each element of that tuple is a power-of-two integer between 4 and 128
+ inclusive or zero.
+* However, the first element (minimum wave size) is not zero.
+* The first element (minimum wave size) is less than the second
+ (maximum wave size).
+* The third element (preferred wave size) is greater or equal to the first
+ element (minimum wave size) and less than or equal to the second element
+ (maximum wave size) or zero.
+* Validators with versions less than 1.8 fail on shaders that use
+ `kDxilRangedWaveSizeTag`.
```
bob80905, line 146: I have been told that RangedWaveSize tag should fail validation with dxil validator version < 1.8, and I have also been told that it should fail for shader models < 6.8. Is it one or the other, or both? @tex3d

proposals/0013-wave-size-range.md
```diff
@@ -83,11 +83,18 @@ Where `minWaveSize` is the minimum wave size supported by the shader
 
 ### DXIL Additions
 
-The existing `WaveSize` value is stored as a metadata 32-bit integer constant.
-To store the additional values, an additional metadata tag indicating a tuple of
- three 32-bit integers representing the given parameters to `WaveSize`.
-In the case where no preferred wave size is specified, the value zero will
- indicate that nothing was specified.
+The existing `WaveSize` value is stored as a single metadata 32-bit integer
+ constant.
```
tex3d, line 87: This is confusing.  It seems to imply that when only using the existing feature, there is a single metadata value.  Then, combined with wording below, you might be led to believe that we expand the same tag to 3 values.  I think it's a lot more clear if we can just explicitly say that we deprecate the `kDxilWaveSizeTag` (11) in SM 6.8 in favor of a new tag `kDxilRangedWaveSizeTag` (23), and here's the format of the value for that new tag...

Then, in the table below, if we could include both the old tag and the new tag, as well as a shader model column, it demonstrates that we still use the old tag for older shader models, and it has the same format there, but the new tag is always used for SM 6.8 and above.

proposals/0013-wave-size-range.md
tex3d: It's specifically for shader model < 6.8.  1.8 is the first validator version that accepts shader model 6.8, so prior versions will fail when they encounter the unrecognized tag, and the newer version will not allow you to set a validator version < 1.8 if the shader model is 6.8.

proposals/0013-wave-size-range.md
```diff
@@ -125,16 +132,20 @@ These are where new or slightly altered errors are produced:
...
+* However, the first element (minimum wave size) is not zero.
+* The first element (minimum wave size) is less than the second
+ (maximum wave size).
+* The third element (preferred wave size) is greater or equal to the first
+ element (minimum wave size) and less than or equal to the second element
+ (maximum wave size) or zero.
+* Validators with versions less than 1.8 fail on shaders that use
+ `kDxilRangedWaveSizeTag`.
+* Validators with versions greater than or equal to 1.8 fail on shaders that use
+ `kDxilWaveSizeTag`.
```
tex3d, line 148: Like the above comment implies, this should only be when the shader model is greater than or equal to 6.8.  That means even when validator version is 1.8, the `kDxilWaveSizeTag` is the correct tag to use for shader model 6.7 (or 6.6).

tex3d (approved):
LGTM.

proposals/0013-wave-size-range.md
```diff
@@ -125,16 +136,20 @@ These are where new or slightly altered errors are produced:
...
+ elements.
+* Each element of that tuple is a power-of-two integer between 4 and 128
+ inclusive or zero.
+* However, the first element (minimum wave size) is not zero.
+* The first element (minimum wave size) is less than the second
+ (maximum wave size).
+* The third element (preferred wave size) is greater or equal to the first
+ element (minimum wave size) and less than or equal to the second element
+ (maximum wave size) or zero.
+* Shaders with with versions less than 6.8 fail on shaders that use
```
bob80905, line 149: nit: double with, typo

proposals/0013-wave-size-range.md
```diff
@@ -144,70 +159,80 @@ No additions are needed here.
...
+3. The three parameter variant of `WaveSize` correctly produces a metadata tuple
+ pointed to by a `kDxilRangedWaveSizeTag` in the entry point attribute list with
+ all three values in their respective element locations.
+ No `kDxilWaveSizeTag` should be produced.
+4. That the PSV0 `MinimumExpectedWaveLaneCount` and `MaximumExpectedWaveLaneCount`
   values reflect those provided for the wave size range.
 
 Note that the above must all use literal values for the parameters on account of
- other compile-time contants being broken due to DXC bug 
+ other compile-time constants being broken due to DXC bug
```
bob80905, line 195: Do we want to specify the bug being referred to here?

proposals/0013-wave-size-range.md
bob80905: I might add, the wording is strange? Maybe a better way to say it is:
"Compilation with target shader model less than 6.8 will fail on shaders with the DxilWaveSize attribute".

pow2clk: I agree the wording is awkward, but I have a couple disagreements with your proposed text. 

Compilation isn't what's happening here necessarily since this is about validation failures.

Part of the awkwardness might be because the bullet points are meant to refer to the header part that says "Validation should confirm".  So I expect this to be read "Validation should confirm (that) Shaders with versions less than 6.8 fail ..." I'll take another run at it though.

bob80905 (approved):

proposals/0013-wave-size-range.md
```diff
@@ -125,16 +136,20 @@ These are where new or slightly altered errors are produced:
...
+ inclusive or zero.
+* However, the first element (minimum wave size) is not zero.
+* The first element (minimum wave size) is less than the second
+ (maximum wave size).
+* The third element (preferred wave size) is greater or equal to the first
+ element (minimum wave size) and less than or equal to the second element
+ (maximum wave size) or zero.
+* Shaders with target shader model less than 6.8 that use 
+  `kDxilRangedWaveSizeTag` should fail.
+* Shaders with target sahder model greater than or equal to 6.8 that use
```
bob80905, line 151: Nit: Typo! sahder

proposals/0013-wave-size-range.md
```diff
@@ -125,16 +136,20 @@ These are where new or slightly altered errors are produced:
...
- metadata.
-* Where validator version is greater than or equal to 1.8, fail when The tuple
- that the wave size range tag points to does not have exactly three elements.
+* The metadata `kDxilRangedWaveSizeTag` points to a tuple of exactly three
+ elements.
+* Each element of that tuple is a power-of-two integer between 4 and 128
+ inclusive or zero.
+* However, the first element (minimum wave size) is not zero.
+* The first element (minimum wave size) is less than the second
+ (maximum wave size).
```
tex3d, line 145:
```suggestion
* The second element (maximum wave size) is greater than the first value, or zero.
```

PR #167, opened 2024-02-13
[0011] Make SpirvType take type template parameter pack

cassiebeckley:
Instead of using an arbitrary unchecked list of parameters, require that `SpirvType` and `SpirvOpaqueType` take a type template parameter pack for the list of operands. For example, now `SpirvOpaqueType` is defined as

```
template<uint OpCode, typename... Operands>
struct SpirvOpaqueType {};
```

A type `vk::integral_constant<typename T, T v>` is defined in order to allow literal values as operands, and a type qualifier _spirv_immediate is defined in order to specify that a literal value should be embedded as an immediate value in the SPIR-V type-declaration instruction.

@s-perron and @llvm-beanz for review.

### DIFF
--- a/proposals/0011-inline-spirv.md
+++ b/proposals/0011-inline-spirv.md
@@ -171,12 +171,13 @@ kinds of values:
 
     For example, `123` can be passed in by using
     `vk::integral_constant<uint, 123>`.
-1.  An `integral_constant` with the `_spirv_immediate` type qualifier. This
-    value will be passed in to the type-declaration instruction as an immediate
-    literal value.
+1.  An instantiation of the `vk::Literal<typename T>` type template. `T` should
+    be an instantiation of `integral_constant`. The value of this constant will
+    be passed in to the type-declaration instruction as an immediate literal
+    value.
 
     For example, `123` can be passed in as an immediate literal by using
-    `_spirv_immediate vk::integral_constant<uint, 123>`.
+    `vk::Literal<vk::integral_constant<uint, 123> >`.
 1.  Any type. The id of the lowered type will be passed in to the
     type-declaration instruction.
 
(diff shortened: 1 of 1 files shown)


s-perron:
I'll delegate my review to @keenuts. In general, I am fine with this direction. It will make it easier to add it to clang and DXC.

proposals/0011-inline-spirv.md
```diff
@@ -178,15 +185,15 @@ SPIRV.html#OpTypeArray) takes an id for the element type and an id for the
 element length, so an array of 16 integers could be declared as
 
 ```
-vk::SpirvOpaqueType</* OpTypeArray */ 28, int, 16>
+vk::SpirvOpaqueType</* OpTypeArray */ 28, int, vk::integral_constant<uint, 16> >
```
Keenuts, line 188: If I understand correctly, the goal of adding this type is to know upfront the destination type of the constant without having to guess?
But why was `vk::ext_literal` bad? What issues did you find?

Keenuts (commented):
Sorry for the delay for the review!

proposals/0011-inline-spirv.md
cassiebeckley: iirc the specific critique from @llvm-beanz was that it was defined as a function that returned `void`, and we were "looking through" the function call in the AST to get the value. We could fix it by making a function return a value of, say, `vk::Literal<T>` type, but Chris suggested making it a type qualifier, which is what I've done here.

I do wonder if maybe this is a very specific thing to add a new type qualifier for – it's only relevant for the `SpirvType` template.

llvm-beanz (approved):
This looks good to me.

proposals/0011-inline-spirv.md
Keenuts: Ahh right, I see. Thanks you both for the explanation!
Got it now.

Keenuts (approved):

PR #172, opened 2024-03-18
Introduce proposal for c style library for HLSL compilation

coopp:
One of the HLSL compiler's primary usage modes is as a library. The upstream compiler will also need to expose a library for  toolchains to compile a shader.

This commit introduces a new proposal to add a c-style library for toolchains and other titles to use the HLSL compiler as a library.

The proposal is to address:
 Issue#63631

### DIFF
--- a/proposals/0016-c-interface-compiler-library.md
+++ b/proposals/0016-c-interface-compiler-library.md
@@ -13,11 +13,11 @@
 
 ## Introduction
 
-An effort is underway to modernize compilation of HLSL based shaders. This is
-commonly referred to as the upstream effort. This effort is a long term play
-that positions HLSL shader compilation in a place that provides the most
-stability and maintainability for the future by giving as many options to
-DirectX while the GPU/NPU landscape evolves.
+An effort is underway to modernize compilation of HLSL based shaders by adding
+HLSL compilation to clang. This effort is a long term play that positions
+HLSL shader compilation in a place that provides the most stability and
+maintainability for the future by giving as many options to DirectX while the
+GPU/NPU landscape evolves.
 
 There are multiple ways to compile a shader today.
 * dxc.exe - Executable that takes command line options and outputs bytecode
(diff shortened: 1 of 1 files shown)


llvm-beanz (commented):
Sorry for the delay in posting this. I had a few small comments and apparently forgot to hit the "submit review" button.

llvm-beanz (approved):
Approving since none of my comments really need to be blocking here. Feel free to mark them as resolved once you've read them whether you update the doc or not.

One thing I'd like to see as a next step is a catalog of the existing COM APIs and their capabilities. I don't think we'll want to bring all the capabilities over to the new API, but understanding what is there will be extremely informative for the design of what we need.

Can you look into that and add a new section before the `Detailed design` about the existing dxcompiler.dll and DXIL.dll APIs?

coopp:
--> "Can you look into that and add a new section before the Detailed design about the existing dxcompiler.dll and DXIL.dll APIs?" 

Yes, I'll work on this and put out  another PR with this information.

proposals/0016-c-interface-compiler-library.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Motivation
+
+One of the HLSL compiler's primary usage modes is as a library. The upstream
+compiler will also need to expose a library for toolchains to compile a shader.
+
+## Proposed solution
+
+The existing HLSL compiler's library support on Windows is COM based.
+This is a non-starter for the upstream effort.  COM doesn't fit any of the
```
damyanp, line 35: "upstream effort"

I wonder at what point we start referring to this just as "clang" or something like that, rather than "upstream".  Upstream tends to frame everything from the perspective of where we were, rather than of where we are / want to be.

damyanp (approved):

proposals/0016-c-interface-compiler-library.md
llvm-beanz: +1 this. I've been trying to socialize calling the new HLSL compiler "Clang" too. I think that really helps clarify our commitment to the open source project and community rather than drawing an arbitrary distinction.

PR #175, opened 2024-03-20
Expand definition of floating literals

llvm-beanz:
This adds a definition of floating literals that is consistent with existing HLSL documentation and other shader compilers. It is notably different from the implementations in DXC and FXC.

### DIFF
--- a/specs/language/lex.tex
+++ b/specs/language/lex.tex
@@ -203,10 +203,10 @@ source file names in an implementation defined way.
 followed by valid identifier characters and floating point literal suffixes
 (\texttt{e+}, \texttt{e-}, \texttt{E+}, \texttt{E-}, \texttt{p+}, \texttt{p-},
 \texttt{P+}, and \texttt{P-}). Preprocessing number tokens lexically include all
-\textit{integer-literal} and \textit{floating-point-literal} tokens.
+\textit{integer-literal} and \textit{floating-literal} tokens.
 
 \p Preprocessing numbers do not have types or values. Types and values are
-assigned to \textit{integer-literal}, \textit{floating-point-literal}, and
+assigned to \textit{integer-literal}, \textit{floating-literal}, and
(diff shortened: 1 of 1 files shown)


llvm-beanz:
A few additional notes for this PR. This adds a definition for floating literals that is:
* Consistent with the [official HLSL docs](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-appendix-grammar#floating-point-numbers).
* Consistent with [GLSL's specification](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.pdf): "When the suffix "lf" or "LF" is present, the literal has type double. Otherwise, the literal has type float."

**It is not consistent with FXC or DXC's behavior.**

DXC's implementation of literal types has been a significant source of bugs in the compiler(https://github.com/microsoft/DirectXShaderCompiler/issues/6147, https://github.com/microsoft/DirectXShaderCompiler/issues/3973, https://github.com/microsoft/DirectXShaderCompiler/issues/4683, https://github.com/microsoft/DirectXShaderCompiler/issues/5493, https://github.com/microsoft/DirectXShaderCompiler/issues/6410, and probably more). It has also caused [confusion for even highly technical users](https://github.com/shader-slang/slang/issues/1185), and is probably inconsistent across DXIL and SPIR-V.

llvm-beanz:
As an experiment to attempt to gauge the disruption this change would cause. DXC has 40 tests which are fragile to this change (see: https://github.com/llvm-beanz/DirectXShaderCompiler/commit/4c854da7f5c0d92b2b893bbdfaf27286705bbda5). Of those, several of them are fragile because they look for the `literal float` string, not because they impact correctness or code generation.

I'm trying to run some additional testing against other shaders that I have available to further gauge the potential impact on existing code.

damyanp (approved):

PR #176, opened 2024-03-21
Update 0015-extended-command-info.md to reflect what we shipped.

amarpMSFT:
Update the device capability section for ExtendedCommandInfo

### DIFF
--- a/proposals/0015-extended-command-info.md
+++ b/proposals/0015-extended-command-info.md
@@ -136,11 +136,43 @@ Validation should confirm:
 
 #### Runtime information
 
-No additions are needed here.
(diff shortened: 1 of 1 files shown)


damyanp (changes requested):
Is this documenting what we already shipped?  Or is this proposing new changes?

(It'd be great if the answer were in the PR description)

amarpMSFT:
yeah, describing what we shipped, was missing in this spec.

damyanp (approved):
I've not verified that the details in here are correct, hopefully someone more familiar with the code can do that, but approving based on the description of what this change is.

PR #177, opened 2024-03-21
remove stale copy/pasted text

amarpMSFT:
Removes irrelevant text from another spec mistakenly copied in.


damyanp (approved):

pow2clk (approved):
I added a description as required. We should enable that automated check for this repo.

Issue #73, opened 2023-08-03
Eliminate hlsl `literal` types


llvm-beanz:
**Is your feature request related to a problem? Please describe.**
HLSL literal types are magic. They convert through odd rules that are inscrutable to define, and often result in promotions to sizes larger than the user wants.

**Describe the solution you'd like**
Adopt C/C++ literal prefix rules and implicit conversion rules.

**Describe alternatives you've considered**
Fixing the issues with literal types seems unnecessarily complicated and would result in further confusion by users.

**Additional context**
Some recent bugs:

[vector of int64 instead of int32](https://godbolt.org/z/MMbYxP9cW)
```hlsl
RWByteAddressBuffer buffer : register(u0, space0); 
[numthreads(1, 1, 1)]
void main() {
    buffer.Store4(0, 0.xxxx);
}
```

[Some odd doubles](https://godbolt.org/z/hPWf5cfa5)
```hlsl
[numthreads(1, 1, 1)]
void main() {
  int b;
  bool c;
  float a = (b == 0 ? 38 : 32) * (c ? 0.05 : 1.0);
}
```

KStocky:
I shall add to this.

This will not compile https://shader-playground.timjones.io/48171e474758a6dc0f099fd9ca4c19c2 because a literal int and int are treated as different types.

However, Literal Int decays very easily to int when queried with some template metaprogramming. E.g. https://shader-playground.timjones.io/48064aafe35c1381d5afaee168525f7b

This is a valid workaround, however, I feel it should not be necessary.

I don't feel like devs should have to use enable_if everywhere to ensure that two types are in fact the same type.

llvm-beanz:
I've posted a spec PR that addresses this in the language spec by adopting the behavior that is documented on Microsoft's HLSL docs rather than the behavior that DXC implements:

https://github.com/microsoft/hlsl-specs/pull/175

We should consider implementing this behavior in DXC as part of HLSL 202x (it should be a pretty simple change to DXC).

llvm-beanz:
As per the discussion in #211, proposal [0017 - Conforming Literals](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0017-conforming-literals.md) has been accepted and merged into DXC as https://github.com/microsoft/DirectXShaderCompiler/pull/6572.

With this change `literal` types are no longer part of HLSL as of 202x.

PR #178, opened 2024-03-21
Add a proposal to describe 202x and 202y

llvm-beanz:
I've talked about 202x and 202y in passing but haven't really written down my thoughts. This proposal is the start of writing down how I think we should classify the two language versions.

### DIFF
--- a/proposals/0020-hlsl-202x-202y.md
+++ b/proposals/0020-hlsl-202x-202y.md
@@ -2,7 +2,7 @@
 
 # HLSL 202x and 202y
 
-* Proposal: [INF-NNNN](NNNN-hlsl-202x-202y.md)
+* Proposal: [0020](0020-hlsl-202x-202y.md)
 * Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
 * Sponsor: [Chris Bieneman](https://github.com/llvm-beanz)
 * Status: **Under Consideration**
--- a/proposals/infra/INF-NNNN-hlsl-202x-202y.md
+++ b/proposals/infra/INF-NNNN-hlsl-202x-202y.md
@@ -2,7 +2,7 @@
 
 # HLSL 202x and 202y
 
-* Proposal: [INF-NNNN HLSL 202x and 202y](NNNN-hlsl-202x-202y.md)
+* Proposal: [INF-NNNN](NNNN-hlsl-202x-202y.md)
 * Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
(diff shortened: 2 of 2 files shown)


damyanp:
I'm a little surprised to have a single proposal for 202x and 202y.  I'd have thought that 202x would likely be nailed down much sooner than 202y. Is your thinking that these could be separated in the future?  Or are they more inherently tied together than I understand?

dmpots (approved):
The proposal looks complete enough to merge for further discussion. My only question is whether this should be split into separate proposals for 202x and 202y.

coopp (approved):
I like this proposal.  Having both 202x and 202y called out as separate things really clears up confusion.

llvm-beanz:
> I'm a little surprised to have a single proposal for 202x and 202y. I'd have thought that 202x would likely be nailed down much sooner than 202y.

The proposal structure is maybe a bit odd for this kind of topic. This isn't really intending to propose a specific features or set of features as much as a guiding principle and approach for planning our next two feature sets.

> Is your thinking that these could be separated in the future?

I'm proposing them as effectively two release targets. We already have a whole bunch of features that we were kinda planning as candidates for 202x, and I'm suggesting we really push almost all of those out to 202y and provide a narrower scoped purpose for 202x.

> Or are they more inherently tied together than I understand?

I think it is helpful to have some idea of the features we want in 202y because that knowledge can help inform some of the work going on in Clang so that we know how aggressively to disable C++ features under the HLSL mode. That said, we might actually not know what features need to go in 202x to help bridge users to Clang until we get further along in development.

llvm-beanz:
> The proposal looks complete enough to merge for further discussion. My only question is whether this should be split into separate proposals for 202x and 202y.

I think ultimately we will want separate proposals for each feature that goes into 202x and 202y, and some unifying document that lists which features go into each version.

damyanp:
Ah, so the document contained in this PR will become a bit like an index to all the 202x and 202y proposals - and so when we write a proposal for #73 we'd update this doc to link to that?

damyanp (approved):

llvm-beanz:
> Ah, so the document contained in this PR will become a bit like an index to all the 202x and 202y proposals - and so when we write a proposal for #73 we'd update this doc to link to that?

Yea, and we can use this to frame out the motivation and intention behind each of the releases so that we know which bucket features might fit into. I suspect we'll have more features that we would want to get into 202y than we'll have time for, so eventually we'll need to acknowledge a "202z" although I hesitate to name that anything since it might be a "203z" instead...

PR #179, opened 2024-03-22
Add proposal for conforming literals

llvm-beanz:
This proposal seeks to codfiy the behavior documented in the HLSL documentation
https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-appendix-grammar#floating-point-numbers, and captured in #175 as the behavior for HLSL 202x.

### DIFF
--- a/proposals/NNNN-conforming-literals.md
+++ b/proposals/NNNN-conforming-literals.md
@@ -1,0 +1,90 @@
+<!-- {% raw %} -->
+
+# Conforming Literals
+
+* Proposal: [NNNN](NNNN-conforming-literals.md)
+* Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
+* Sponsor: [Chris Bieneman](https://github.com/llvm-beanz)
+* Status: **Under Consideration**
+* Planned Version: HLSL 202x
+* PRs: #175
+* Issues: #73, microsoft/DirectXShaderCompiler#6147,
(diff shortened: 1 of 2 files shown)


proposals/NNNN-conforming-literals.md
```diff
@@ -0,0 +1,90 @@
...
+### Problems with this solution
+
+This is a significant change in behavior which will cause subtle issues for
+existing shaders due to variations in precision of compile-time constant
+evaluation. This behavior difference will cause subtle bugs that will be
+challenging to diagnose in the midst of a larger compiler transition (i.e.
+adopting Clang).
+
+For that reason, this feature proposal targets HLSL 202x, with support for the
+new literal behavior in DXC.
```
coopp, line 83: Will this literal behavior support for HLSL 202x just be the default (not configurable)?  Also, ( not for this proposal doc of course ) will HLSL 202y also adopt this behavior?

coopp (approved):
Looks good to me.

proposals/NNNN-conforming-literals.md
llvm-beanz: Yea, this will be the only behavior supported when you pass `-HV 202x` to DXC or Clang (which will eventually become a real year), and then at some later point will become the default language version.

For 202y, the intent is that `y > x`, and that all behaviors of `x` will be inherited by `y`. I should make that clear in the other proposal document.

damyanp (approved):

PR #182, opened 2024-03-28
[0017] Adding notes about experiments

llvm-beanz:
This adds a bunch of detail about experiments that were run to assess the potential impact of this language change on existing shader code.

The TL;DR is that the thesis holdes that few shaders in production titles are expected to be significantly impacted.


proposals/0017-conforming-literals.md
```diff
@@ -82,6 +82,100 @@ adopting Clang).
...
+### Experimentation
+
+In writing this proposal a key thesis was proposed that the impact of this
+change on developers would be minimal because DXC's behavior for DXIL is
+undocumented, complex, and different from other shader compilers (including
+DXC's SPIR-V generator). The last point is important and relevant because a
+large percentage of shader code is compiled with multiple different shader
+compilers which implement different language semantics. Because of this, it is
+believed that shaders need to be resilient to subtle behavioral differences like
+the ones this change would cause.
```
damyanp, line 94:
```suggestion
believed that many existing shaders are engineered to be resilient to subtle behavioral differences like
the ones this change would cause.
```

proposals/0017-conforming-literals.md
```diff
@@ -82,6 +82,100 @@ adopting Clang).
...
+enough to produce minor visual artifacts. These differences are significant
+enough and represent inaccuracies in rendered output that we would expect a
+software developer to be concerned by and wish to address even though the
+overall image still rendered mostly correctly.
+
+##### Catastrophic Visible Differences
+
+We have 6 test cases, all from the same software title which had severe
+obviously incorrect rendering which would make the title unplayable. One
+important note about this title is that it is an Xbox & Windows exclusive title.
```
damyanp, line 164: May be worth re-iterating here the significance of it being xbox/windows exclusive.

damyanp (approved):

proposals/0017-conforming-literals.md
```diff
@@ -82,6 +82,100 @@ adopting Clang).
...
+
+#### Conclusions Drawn
+
+The testing here holds with the core thesis that developer impact will be
+minimal and that most shaders are resilient to the changes in precision. Further
```
dmpots, line 171: > developer impact will be minimal

I think I would say this slightly differently. That the number of titles impacted should be minimal, but the impact to a single developer could require them to rewrite some shaders. 

What would the fix be for the catastrophic failures? Would it be to use the previous version of hlsl with the `-HV` flag? Or use a suffix on some of the constants?

dmpots (approved):
Overall analysis and conclusion look good to me.

proposals/0017-conforming-literals.md
damyanp: Possibly, or to find out the cause of the issue since.  Easy to miss from the PR, but this is in the context of a proposal for a change in HLSL 202x, so DXC behavior for other language versions won't change.

damyanp: Recall that the current DXC behavior doesn't match the documentation: https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-appendix-grammar#floating-point-numbers:

> Use the “L” suffix to specify a full 64-bit precision floating-point literal. A 32-bit float literal is the default.
>
> For example, the compiler recognizes the following literal value as a 32-bit precision floating-point literal and ignores the lower bits:
>
>
> `double x = -0.6473313946860445;`
> The compiler recognizes the following literal value as a 64-bit precision floating-point literal:
>
> `double x = -0.6473313946860445L;`

llvm-beanz:
I've filed two issues to track additional work on this proposal:

https://github.com/microsoft/hlsl-specs/issues/183
https://github.com/microsoft/hlsl-specs/issues/184

proposals/0017-conforming-literals.md
llvm-beanz: @dmpots, are you okay with my updated wording here?

dmpots: Yes, that looks good. It clearly separates the expected impact across all titles vs. the work an individual title might have to do if impacted by the change. Thanks!

llvm-beanz: I filed #184 to track the remaining outstanding question here about any mitigations we may want to take for the catastrophic failures.

Since @dmpots is okay with the updated wording I'll resolve this thread.

PR #185, opened 2024-04-01
[0017] Update proposal with details on integers

llvm-beanz:
This adds the details for the C integer literal rules that this proposal intends to adopt.

### DIFF
--- a/proposals/0017-conforming-literals.md
+++ b/proposals/0017-conforming-literals.md
@@ -49,9 +49,9 @@ This proposal adopts this behavior for floating literals.
 
 Similarly this proposal adopts C rules for non-suffixed integer literals. The
 type of a base-10 non-suffixed integer literal is the first integer type from
-the list [`int`, `int64_t`] which can represent the specified literal value. The
-type of an octal, hexadecimal or binary non-suffixed integer literal is the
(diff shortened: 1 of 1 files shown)


damyanp (approved):

proposals/0017-conforming-literals.md
```diff
@@ -47,8 +47,12 @@ which states:
...
 This proposal adopts this behavior for floating literals.
 
-Similarly this proposal adopts 32-bit integer as the default representation for
-integer literals.
+Similarly this proposal adopts C rules for non-suffixed integer literals. The
+type of a base-10 non-suffixed integer literal is the first integer type from
+the list [`int`, `int64_t`] which can represent the specified literal value. The
+type of an octal, hexadecimal or binary non-suffixed integer literal is the
+first type from the list [`int`, `uint`, `int64_t`, `uint64_t`] that can
+represent the specified literal value.
```
bogner, line 55: It feels weird to mix `int` and `int64_t` here, though I understand the choice to use `int64_t` rather than deal with discussing data models. Might be best to explicitly use `int32_t` instead of `int` for consistency.

bogner (approved):

PR #193, opened 2024-04-08
Add proposal for root signature in clang.

python3kgae:
Move discussion from https://github.com/llvm/llvm-project/pull/83933


proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,205 @@
...
+in the translation unit.
+The second parsing will calculate the offset for serialized root signature and
+only run on the root signature for the entry function or local/global root
+signature which used in SubobjectToExportsAssociation.
+
+HLSLRootSignatureDecl will have method to parse the string for diagnostic and
+generate the SerializedRS mentioned above.
+
+A HLSLRootSignatureAttr will be created when meet RootSignature attribute in
+HLSL.
```
llvm-beanz, line 54: Why do we need both an attribute and a declaration AST node? The attribute AST node needs to be attached to a declaration so I'm not sure I understand the purpose that the extra declaration provides.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,205 @@
...
+
+For case compile to a standalone root signature blob, the
+HLSLRootSignatureAttr will be bind to a fake empty entry.
+
+In clang code generation, the HLSLRootSignatureAttr in AST will be translated
+into a global variable with struct type to express the layout and metadata to
+save things like static sampler, root flags, space and NumDescriptors in LLVM IR.
```
llvm-beanz, line 61: Why a struct type? Why not metadata? Given the need to express more than just a packing of types it seems like metadata is better suited to this problem.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,205 @@
...
+
+CGHLSLRuntime will generate metadata to link the global variable as root
+signature for given entry function.
```
llvm-beanz, line 64: What is this global variable? Are you expecting this is fully-serialized?

That would be unfortunate because it would mean any other programming language that wants to use the DirectX backend would also need to know how to serialize the binary representation of the root signature and embed it in the IR. It would be easier for other languages if they just needed to emit metadata to describe the root signature.

proposals/0021-root-signature-in-clang.md
python3kgae: Missed something when moving the doc :(
The declaration AST node is for normal root signature and local/global root signature.
The local/global root signature doesn't need the attribute.
I'll add local/global root signature part back.

python3kgae: Added back local/global root signature.

proposals/0021-root-signature-in-clang.md
python3kgae: The pro for use struct is that it saved the root signature in the serialized layout while you can still access the element when needed.
Then when emit serialized root signature, just pack the data into a vector of uint32_t.

proposals/0021-root-signature-in-clang.md
python3kgae: Unless what saved in the metadata is the original root signature string, generate metadata is just add another representation which other languages need to generate like serialized root signature layout.

If it is OK, I think save the string in LLVM IR is easier.
But not sure about share parsing code between clang and llvm.

llvm-beanz: No other language is going to use the root signature string that HLSL uses. That's a very HLSL-specific design decision.

Other languages might decide to do a structure-based interface, or (like Metal) a parameter-based interface. Forcing those frontends to generate the DirectX binary serialization or the HLSL string are both really awkward design decisions.

python3kgae: We can expose struct like
  struct ParsedRootSignature {
    dxbc::RootSignature::ContainerRootSignatureDesc RSDesc;
    SmallVector<dxbc::RootSignature::ContainerRootParameter, 8> RSParameters;
    SmallVector<dxbc::RootSignature::StaticSamplerDesc, 8> StaticSamplers;
    SmallVector<dxbc::RootSignature::v_1_1::ContainerDescriptorRange, 8>
        DescriptorRanges;
    SmallVector<
        std::variant<dxbc::RootSignature::ContainerRootConstants,
                     dxbc::RootSignature::v_1_1::ContainerRootDescriptor,
                     dxbc::RootSignature::ContainerRootDescriptorTable>,
        8>
        RSParamExtras;
  };
And offer function to generate the representation for root signature in llvm IR from a ParsedRootSignature .

It will be easier than create metadata directly.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,239 @@
...
+## Requirements
+
+1) Clang-dxc should parse existing root signature grammars and provide 
+diagnostics.
+2) Clang-dxc should translate textual root signatures into binary encoded 
+formats.
+3) Clang-dxc should support reading and writing binary encoded root 
+ignatures to and from DXContainer files.
+4) Clang-dxc should validate root signatures before emitting them.
+5) Clang-dxc should have a comprehensive test infrastructure to test all
```
llvm-beanz, line 33: nit: I don't really like calling these all `clang-dxc`. `clang-dxc` is a specific software component, it is the binary (or symlink) for the Clang DXC-compatible driver. We don't actually want to do this in the driver, and some of it shouldn't even necessarily be exposed through the driver.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,239 @@
...
+The string contains a collection of comma-separated clauses that describe root
+signature constituent components.
+
+## Motivation
+
+To support root signature in clang-dxc.
+
+## Requirements
+
+1) Clang-dxc should parse existing root signature grammars and provide
```
llvm-beanz, line 26: nit: Requirements aren't "should" sentences, they should be "must" or "shall". I realize this is a bit wishy-washy language, but "should" doesn't state a requirement it states a nice-to-have.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,239 @@
...
+formats.
+3) Clang-dxc should support reading and writing binary encoded root 
+ignatures to and from DXContainer files.
+4) Clang-dxc should validate root signatures before emitting them.
+5) Clang-dxc should have a comprehensive test infrastructure to test all 
+software components.
+
+Non-requirements
+1) Clang-dxc is not required to validate or manipulate root signatures read 
+from a DXContainer.
```
llvm-beanz, line 38: I in the workflow where we read a DXContainer in and re-emit it, we don't need to validate on the container, but we do need to be able to read the root signature and put it back into the IR module (which sounds like manipulating).

llvm-beanz (commented):
I think the latest update mostly captures the requirements. Are there changes you think need to be factored into the design based on the requirements?

proposals/0021-root-signature-in-clang.md
python3kgae: Updated

proposals/0021-root-signature-in-clang.md
python3kgae: Updated

proposals/0021-root-signature-in-clang.md
python3kgae: Updated

python3kgae:
> I think the latest update mostly captures the requirements. Are there changes you think need to be factored into the design based on the requirements?

I'm not sure what to do for the future versions.
And we have discussion about the IR representation. I'll update once we got conclusion on that.

llvm-beanz:
> I'm not sure what to do for the future versions. And we have discussion about the IR representation. I'll update once we got conclusion on that.

My read of the IR representation that you're proposing is that it is pretty tightly linked to the binary encoding that gets put into the container file. I think we should decouple those representations.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,270 @@
...
+```
+
+* LLVM IR for root signature saved in metadata.
+
+```
+
+  ; named metadata for entry root signature
+  !hlsl.entry.rootsignatures = !{!3}
+  …
+  !3 = !{ptr @main, !4}
```
damyanp, line 229: Thanks, seeing it like this helps me understand what's being talked about a bit more!

I think the schema you've followed here is to have a 1:1 mapping between the d3d struct and the metadata tuple?  I'm trying to work out what the contents of `!4` example corresponds to.  It has this data in it:

* 2
* 0
* 24
* 0
* 96
* 3
* !5, !12, !15

The RootFlags in your example is `ALLOW_INPUT_ASSEMBLER_INPUT_LAYOUT | DENY_VERTEX_SHADER_ROOT_ACCESS` which is `0x1 | 0x2` which is 3.  So I guess that the corresponds to the "3".

The example has 6 root parameters and 2 static samplers.  I'm not sure how this corresponds to the other numbers (2, 0, 24, 0, and 96).  The `!5` looks like it references metadata listing the parameters, the `!12` looks like it references a list of statics samplers.  I'm not sure what `!15` is.

It'd be great to understand what the schema is here.  Maybe the example here isn't meant to encode the root signature example above?



When coming up with schemas it would be good to consider what's involved in building up the root signature it describes, and what would be involved in lowering it to DXIL / SPIR-V target.  For example, could parameters be added by just appending some metadata and adding a reference to it to an existing list? Or would we need to chase around and update multiple tuples?  I think that the main consideration here is for code complexity rather than performance.

python3kgae: It is mapping struct like
```

  struct SerializedRS {
    ContainerRootSignatureDesc RSDesc;
    ContainerRootParameter  rootParameters[6];
    StaticSamplerDesc  samplers[2];

    // Extra part for each RootParameter in rootParameters.
    // RootConstants/RootDescriptorTable/RootDescriptor dependent on ParameterType.
    // For RootDescriptorTable, the extra part will be like
    // struct {
    //   RootDescriptorTable table;
    //   ContainerDescriptorRange ranges[NumDescriptorRangesForTheTable];
    // };

    struct {

      RootConstants b0;

      ContainerRootDescriptor t1;

      ContainerRootDescriptor u1;

      struct {

        ContainerRootDescriptor tab0;

        ContainerDescriptorRange tab0Ranges[2];

      } table0;

      struct {

        ContainerRootDescriptor tab1;

        ContainerDescriptorRange tab1Ranges[1];

      } table1;

      RootConstants b10;

    };

  };

```

python3kgae: For code complexity, we could take DxilMDHelper::EmitSignatureElement  in https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/DXIL/DxilMetadataHelper.cpp#L541 as example for write a struct to metadata.

To append more things, we used this tag thing in dxc https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/DXIL/DxilMetadataHelper.cpp#L1290
Not sure if it is the correct llvm approach though.

damyanp: Why would we include things like "RootParametersOffset" in the metadata?  That's the offset in bytes to the root parameters inside a binary serialization of the root signature?  I'm not sure that makes sense here.  So in the above example, for the metadata we have:

* 2: the version of the binary serialized root signature (this isn't the binary serialized root sig, so isn't relevant)
* 0: the number of parameters?? (but this should be 6?)
* 24: the offset in the binary serialized root signature (doesn't make sense here)
* 0: the number of static samplers (but this should be 2?)
* 96: another offset in the binary serialized root signature
* 3: the flags
* !5, !12, !15

I still don't understand what the !15 is for.

I think that trying to encode the binary serialized form of the root signature in metadata is missing the point of why we'd be using the metadata form in the first place.  

At the very least, if you strip out these unnecessary bits you end up with something more like:

```
!4 = !{ i32 3, !5, !12 }  # flags, parameters, static samplers
```

But I think that once you embrace that the metadata schema isn't just a transliteration of the serialized binary format into metadata nodes there's opportunity for further simplifications.

python3kgae: Updated the metadata.
All unused data is removed.

damyanp: Thanks, this is looking better now.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,264 @@
...
+  !4 = !{i32 2, i32 3, !5, !18}      ; version 1.1, flag 3, root params, statis samplers.
```
damyanp, line 230: How is this version number determined?  It isn't specified in the string passed to the attribute in the HLSL.

What is it used for?  Seems that this is more a detail when the RS is serialized.

llvm-beanz: I had a similar reaction to seeing the version, but I think you can specify the version from the frontend if you use the `rootsig_1_0` or `rootsig_1_1` target profile. So we would need it here.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,264 @@
...
+  !11 = !{i32 2, i32 1, i32 1, i32 0, i32 0}
+  !12 = !{i32 0, i32 8, i32 1, i32 0, i32 1}
+  !13 = !{i32 1, i32 -1, i32 1, i32 0, i32 1}
+  !14 = !{i32 0, i32 0, !15}  ; second table
+  !15 = !{!16}  ; table ranges.
+  !16 = !{i32 3, i32 4, i32 0, i32 1, i32 0}
+  !17 = !{i32 1, i32 0, i32 10, i32 0, i32 3}
+  !18 = !{!19, !20} ; static samplers.
+  !19 = !{i32 85, i32 1, i32 1, i32 1, float 0.000000e+00, i32 16, i32 4, i32 2, float 0.000000e+00, float 0x47EFFFFFE0000000, i32 1, i32 0, i32 0}
+  !20 = !{i32 21, i32 3, i32 1, i32 1, float 0.000000e+00, i32 16, i32 4, i32 2, float 0.000000e+00, float 0x47EFFFFFE0000000, i32 2, i32 0, i32 0}
```
llvm-beanz, line 246: IMO, this is a huge step in the right direction. I suspect there is some opportunity for further refinement, but this is _much_ simpler to read and specify in the IR.

Can you document in more detail what each value in the metadata corresponds to? Even trying to map this to the structures from the structure-based approach is extremely tedious. I’d really like to see something like how [DXIL.rst](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/DXIL.rst#shader-properties-and-capabilities) documents its metadata structures.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,264 @@
...
+  …
+  !3 = !{ptr @main, !4}
+  !4 = !{i32 2, i32 3, !5, !18}      ; version 1.1, flag 3, root params, statis samplers.
+  !5 = !{!6, !7, !8, !9, !14, !17}  ; root parameters.
+  !6 = !{i32 2, i32 0, i32 0, i32 1, i32 8}
+  !7 = !{i32 3, i32 0, i32 0, i32 0, i32 0}
+  !8 = !{i32 4, i32 0, i32 0, i32 0, i32 0}
+  !9 = !{i32 0, i32 0, !10}  ; first table
+  !10 = !{!11, !12, !13} ; table ranges.
+  !11 = !{i32 2, i32 1, i32 1, i32 0, i32 0}
```
damyanp, line 237: `!11` is the `CBV(b1)` entry in the descriptor table, and seems to correspond to `D3D12_DESCRIPTOR_RANGE1`:

* `2` = `RangeType` = `D3D12_DESCRIPTOR_RANGE_TYPE_CBV`
* `1` = `NumDescriptors`
* `1` = `BaseShaderRegister`
* `0` = `Flags`
* `0` = `OffsetInDescriptorsFromTableStart`

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,264 @@
...
+  !12 = !{i32 0, i32 8, i32 1, i32 0, i32 1}
+  !13 = !{i32 1, i32 -1, i32 1, i32 0, i32 1}
```
damyanp, line 239: If the last value in the tuple is OffsetInDescriptorsFromTableStart, then shouldn't the one in `!13` be `9` - so it's after the previous two ranges (1 + 8 = 9).

damyanp: Or maybe it should be "-1" to indicate "DESCRIPTOR_RANGE_OFFSET_APPEND".  I'm not sure if your design is to calculate the actual descriptor range offsets (which is implied by these having the value "1" in here) or if it is to just represent what was specified in the HLSL source (which would be the default value for OffsetInDescriptorsFromTableStart = DESCRIPTOR_RANGE_OFFSET_APPEND = -1 in this case).

proposals/0021-root-signature-in-clang.md
damyanp: This would be good information to capture in the spec.

proposals/0021-root-signature-in-clang.md
```diff
@@ -227,26 +227,85 @@ a list of FunctionDecls.
...
+  !16 = !{i32 0, i32 0, !17}
+  !17 = !{!18}
+  !18 = !{i32 3, i32 4, i32 0, i32 1, i32 0}
+
+  !19 = !{i32 1, i32 0, !20}
+  !20 = !{i32 10, i32 0, i32 3}
+
+  !21 = !{!22, !23}
+  !22 = !{i32 85, i32 1, i32 1, i32 1, float 0.000000e+00, i32 16, i32 4, i32 2, float 0.000000e+00, float 0x47EFFFFFE0000000, i32 1, i32 0, i32 0}
+  !23 = !{i32 21, i32 3, i32 1, i32 1, float 0.000000e+00, i32 16, i32 4, i32 2, float 0.000000e+00, float 0x47EFFFFFE0000000, i32 2, i32 0, i32 0}
```
damyanp, line 260: I've not had a chance to analyze this deeply, but how come there are now 23 metadata entries when there was previously 20?  I'd be hoping that changes would be made to decrease this number, not increase it.

python3kgae: It increased because now RootParameter has fixed format ( ParameterType | Visibility | Extra ), the difference part is moved into Extra metadata.
Before, it has different format for different Kind.

damyanp: Why did we have to do that?  This seems to be a move in the wrong direction.

python3kgae: OK. I'll move it back.

python3kgae: Updated.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,315 @@
...
+
+  !14 = !{i32 1, i32 0, i32 10, i32 0, i32 3}
+
+  !15 = !{!16, !17}
+  !16 = !{i32 85, i32 1, i32 1, i32 1, float 0.000000e+00, i32 16, i32 4, i32 2, float 0.000000e+00, float 0x47EFFFFFE0000000, i32 1, i32 0, i32 0}
+  !17 = !{i32 21, i32 3, i32 1, i32 1, float 0.000000e+00, i32 16, i32 4, i32 2, float 0.000000e+00, float 0x47EFFFFFE0000000, i32 2, i32 0, i32 0}
+
+```
+
+* Metadata format
```
damyanp, line 260: I think I understand what this is trying to show.  I suggest you add a column so we can see which line each of the rows in the table corresponds to.  eg:
Root Signature
| | Version | Flag | RootParameters | StaticSamplers |
|-| --- | --- | --- | --- |
| !3 | i32 2 | i32 3 | !4 | !15 |

python3kgae: Fixed.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,314 @@
...
+| !7 | i32 4 |  i32 0 | i32 0 | i32 0 | i32 0 |
+
+DescriptorTable
+| | ParameterType | Visibility | DescriptorRanges
+|-| --- | --- | --- | 
+| !8 | i32 0 |  i32 0 | !9, !10, !11 |
+| !12 | i32 0 |  i32 0 | !13 |
+
+DescriptorRange
+| | RangeType | NumDescriptors | BaseShaderRegister | Space | Flags |
```
damyanp, line 286: This is missing the `OffsetInDescriptorsFromTableStart` field.

python3kgae: We don't need it for metadata.
OffsetInDescriptorsFromTableStart could be calculated when serialize the root signature.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,314 @@
...
+       UAV(u1, numDescriptors = unbounded, flags = DESCRIPTORS_VOLATILE)),
+       DescriptorTable(Sampler(s0, space=1, numDescriptors = 4)),
+       RootConstants(num32BitConstants=3, b10),
+       StaticSampler(s1),
+       StaticSampler(s2, addressU = TEXTURE_ADDRESS_CLAMP, filter = FILTER_MIN_MAG_MIP_LINEAR )"
+       HLSLRootSignature 'main.RS'
+
+```
+
+* LLVM IR for the root signature saved in ConstantStruct.
```
damyanp, line 160: Are you ready now to make a decision over whether or not the ConstantStruct approach or the metadata approach is preferable?

To my eyes, the metadata approach is simpler, more concise and I can easily imagine further changes we can make to further simplify it or extend it.

proposals/0021-root-signature-in-clang.md
damyanp: This isn't the case.  Developers can (and do: I've done it!) explicitly specify this value to allow for overlapping ranges of descriptors.

python3kgae: Sorry. I mistook it for the offset in the serialized root signature. :(
Will add it.

proposals/0021-root-signature-in-clang.md
python3kgae: I was thinking ConsantStruct could save the serialization by just make the global variable with "RTS0" section. But it does not work because cannot have extra global variable when emit dxil. :(
I'll update the doc to use metadata.

python3kgae: Updated.

proposals/0021-root-signature-in-clang.md
python3kgae: Fixed.

proposals/0021-root-signature-in-clang.md
python3kgae: No.
Updated the global variable to metadata.

llvm-beanz:
@python3kgae, the documentation PR that you closed on LLVM had a lot of really valuable information. It's really a shame to drop that. Can you please do as I suggested [here](https://github.com/llvm/llvm-project/pull/83933#pullrequestreview-1986889387) and put up a PR with just the documentation of the root signatures without the implementation design?

It would also be extremely helpful to have the grammar for the local and global root signatures so that we can understand the required parsing impact.

python3kgae:
> @python3kgae, the documentation PR that you closed on LLVM had a lot of really valuable information. It's really a shame to drop that. Can you please do as I suggested [here](https://github.com/llvm/llvm-project/pull/83933#pullrequestreview-1986889387) and put up a PR with just the documentation of the root signatures without the implementation design?
> 
> It would also be extremely helpful to have the grammar for the local and global root signatures so that we can understand the required parsing impact.

Sure. I can do that.
Where should this separate doc live?

And any more issue with current PR?

proposals/0021-root-signature-in-clang.md
damyanp: Has this been addressed?  If so, can we resolve this conversation?

llvm-beanz:
> Sure. I can do that. Where should this separate doc live?

I can see two docs being useful. The "Usage" and "Root Signature Grammar" sections of your document seem like they should live under `clang/docs/HLSL`, and the "Serialized Format" section could live in `llvm/docs/DirectX`

Related: I filed an issue on myself to document the DXContainer data that I implemented (https://github.com/llvm/llvm-project/issues/88775).
 
> And any more issue with current PR?

We need the grammar for local and global root signatures documented here or in the PR to LLVM.

I'll give the document here a more detailed read today to see if I have any other questions.

proposals/0021-root-signature-in-clang.md
llvm-beanz, line 1: Can we instead put this under `proposals/infra/` as INF-0002? This isn't really a new language or tool feature it is just a new implementation approach to an existing feature so it would be nice to keep those separate.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,247 @@
...
+              "                 UAV(u1, numDescriptors = unbounded, " \
+              "                     flags = DESCRIPTORS_VOLATILE)), " \
+              "DescriptorTable(Sampler(s0, space=1, numDescriptors = 4)), " \
+              "RootConstants(num32BitConstants=3, b10), " \
+              "StaticSampler(s1)," \
+              "StaticSampler(s2, " \
+              "              addressU = TEXTURE_ADDRESS_CLAMP, " \
+              "              filter = FILTER_MIN_MAG_MIP_LINEAR )"
+
+```
```
bogner, line 119: This isn't actually an example of the root signature - it's just a define that matches the syntax of a root signature. It'd be helpful for readers here to see the various ways that a root signature is defined. That is, we can specify the root signature as an attribute, via `-rootsig-define`, and using `GlobalRootSignature` or `LocalRootSignature` types (are there more?). It would be helpful to point these out, especially as a way to explain what exactly the HLSLRootSignature decl would mean.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,247 @@
...
+       CBV(b0, space = 1, flags = DATA_STATIC), SRV(t0), UAV(u0),
+       DescriptorTable( CBV(b1), SRV(t1, numDescriptors = 8,flags = DESCRIPTORS_VOLATILE),
+       UAV(u1, numDescriptors = unbounded, flags = DESCRIPTORS_VOLATILE)),
+       DescriptorTable(Sampler(s0, space=1, numDescriptors = 4)),
+       RootConstants(num32BitConstants=3, b10),
+       StaticSampler(s1),
+       StaticSampler(s2, addressU = TEXTURE_ADDRESS_CLAMP, filter = FILTER_MIN_MAG_MIP_LINEAR )"
+       HLSLRootSignature 'main.RS'
+
+```
```
bogner, line 149: What does the actual AST node look like? Is it a string? It isn't clear what this example is showing.

proposals/0021-root-signature-in-clang.md
```diff
@@ -0,0 +1,247 @@
...
+
+Additionally, a new AST node, SubobjectToExportsAssociationDecl, should be 
+added to handle the SubobjectToExportsAssociation statement in HLSL. 
+The SubobjectToExportsAssociationDecl will contain a HLSLRootSignatureDecl and 
+a list of FunctionDecls.
+
+The default root signature version number is 2 which maps to rootsig_1_1.
+With option -force-rootsig-ver rootsig_1_0 could make it to 1.
+
+## Detailed design
```
bogner, line 95: This section is called "Detailed design", but it's currently a list of examples. If this is just examples we should label it as such, but it would also be helpful to have more detailed descriptions of the representations throughout.

proposals/0021-root-signature-in-clang.md
python3kgae: Updated.

proposals/0021-root-signature-in-clang.md
python3kgae: Added a Background part to show all these different root signatures.

proposals/0021-root-signature-in-clang.md
python3kgae: Changed to Examples.

proposals/0021-root-signature-in-clang.md
python3kgae: Updated.

proposals/0021-root-signature-in-clang.md
damyanp: @llvm-beanz - ok to resolve this conversation?

proposals/0021-root-signature-in-clang.md
damyanp: @llvm-beanz - ok to resolve this conversation?

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,683 @@
...
+  llvm::SmallVector<RootSignature::StaticSamplerDesc, 8> StaticSamplers;
+  llvm::SmallVector<RootSignature::ContainerDescriptorRange, 8>
+      DescriptorRanges;
+  llvm::SmallVector<
+      std::variant<RootSignature::ContainerRootConstants,
+                   RootSignature::ContainerRootDescriptor,
+                   RootSignature::ContainerRootDescriptorTable>,
+      8>
+      RSParamExtras;
+};
```
damyanp, line 378: I don't understand how this maps to root signatures.  For example, how would these structs represent this root parameter:

```
DescriptorTable(
    CBV(b1), 
    SRV(t1, numDescriptors = 8, flags = DESCRIPTORS_VOLATILE), 
    UAV(u1, numDescriptors = unbounded, flags = DESCRIPTORS_VOLATILE)),
```

ContainerRootDescriptorTable only has a NumDescriptorRanges field.

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,697 @@
...
+  for (const auto &Range : ParsedRS.DescriptorRanges) {
+    // push_back all the descriptor ranges.
+  }
+  
+  Code = serialization::DECL_HLSL_ROOT_SIGNATURE;
+}
+
+```
+Unless there's other correct way to serialize the data, share serialization 
+code between clang and llvm seems better.
```
llvm-beanz, line 428: The `clang::ASTRecordWriter` writes to `clang::ASTWriter::RecordDataImpl` which is a `llvm::SmallVectorImpl<uint64_t>`. The backing `llvm::SmallVector<uint64_t>` then gets encoded with the bitcode writer.

If you serialize using the backend serialization you're then going to need to convert it to an array of uint64_t's and let the bitcode serializer operate on those. That doesn't make sense from an architecture perspective and it doesn't seem to save any complexity to me.

proposals/infra/INF-0002-root-signature-in-clang.md
python3kgae: Added comment for DescriptorRanges.

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,697 @@
...
+  float MinLOD = 0;
+  float MaxLOD = hlsl::RootSignature::Float32Max;
+  uint32_t ShaderRegister;
+  uint32_t RegisterSpace = 0;
+  hlsl::RootSignature::ShaderVisibility ShaderVisibility =
+      hlsl::RootSignature::ShaderVisibility::All;
+};
+
+struct ParsedRootSignature {
+  dxbc::RootSignature::ContainerRootSignatureDesc RSDesc;
```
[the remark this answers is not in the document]
python3kgae: Updated. Now it only contains required information.

proposals/infra/INF-0002-root-signature-in-clang.md
python3kgae: Updated to use Record.AddString.

python3kgae: So the correct way is pack uint64_t for all the uint32_t data in a ParsedRootSignature?

Could you add pseudo code in your mind to help understanding?

llvm-beanz: The code you have above here does what I'm describing. When you call `push_back` to append each value into the ASTWriter you're adding them to the vector that gets serialized in the bitcode. When the bitcode writer writes the series of values it bitpacks the values into less storage as I described above.

That is the correct way to serialize binary data that is part of an AST node, rather than any method that could share the serialization code with LLVM.

Other than the comment in the spec here you've provided no explanation of how you would share serialization with LLVM.

python3kgae: I see.
If we share serialization code with LLVM, extra work will be done for build the offset data and store those offsets.  If not share, those could be saved for AST serialization.

llvm-beanz: I don’t understand. We shouldn’t encode the offsets into the AST serialization, they would be meaningless since the data gets bit-packed by the bitcode writer.

python3kgae: Add the issue for share code to the doc.

llvm-beanz: Are you still considering sharing the serialization with the DirectX backend as a viable option? If so can you please include information in the spec about how you would do that. What actually gets encoded in the serialized AST?

python3kgae: No. I'll remove that.

proposals/0021-root-signature-in-clang.md
llvm-beanz: I believe this comment has still not been addressed.

llvm-beanz: Oh, I see in your update you mixed documenting the schema with the example. I think we should separate the documentation of the schema from the example. The example is helpful for driving conversation, but is something we should not include in the technical documentation.

python3kgae: Added separate section for metadata format.

proposals/infra/INF-0002-root-signature-in-clang.md
damyanp: Ah, I see.  I don't understand why this design was chosen rather than a more direct one - eg where ContainerRootDescriptorTable might look like:

```c++
struct ContainerRootDescriptorTable {
  SmallVector<ContainerDescriptorRange> Ranges;
};
```

python3kgae: It was started from final serialization format.
Updated to save Ranges in ContainerRootDescriptorTable .

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+To support root signature in clang-dxc.
+
+## Requirements
+
+1) Compiler must parse existing root signature grammars and provide 
+diagnostics.
+2) Compiler must translate textual root signatures into binary encoded 
+formats.
+3) Compiler must support reading and writing binary encoded root 
+ignatures to and from DXContainer files.
```
coopp, line 31: missing 's' in signatures

python3kgae: Good catch. Fixed.

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+};
+
+LocalRootSignature MyLocalRootSignature = 
+{
+    "RootConstants(num32BitConstants = 4, b1)"  // Cube constants 
+};
+
+```
+
+There's -setrootsignature option in DXC which attach root signature part to an existing input DXIL container. It is a DXIL container level operation. Not AST or llvm IR.
```
coopp, line 121: nit: This line exceeds the 80 col limit.

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+    // push_back all the fields of static samplers.
+  }
+  for (const auto &Range : ParsedRS.DescriptorRanges) {
+    // push_back all the descriptor ranges.
+  }
+  
+  Code = serialization::DECL_HLSL_ROOT_SIGNATURE;
+}
+
+```
```
damyanp, line 404: I think that this code is now out of date with the latest versions of the structs in this document.

I don't know that you need the code here.  Are you saying that the root signature will the be serialized using standard LLVM bitcode?

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+[RootSignature(MyRS1)]
+float4 main(float4 coord : COORD) : SV_Target
+{
+…
+}
+
+```
+
+Because the RootSignature attribute in hlsl only have the string, it is easier
+to add a StringArgument to save the string first.
```
damyanp, line 421: > it is easier to add a StringArgument to save the string first.

I don't understand what this is trying to say.  Add a StringArgument to what?

proposals/infra/INF-0002-root-signature-in-clang.md
llvm-beanz: The pseudocode is also not correct. When writing arrays to bitcode (as with pretty much any binary serialization) you need to push the size of the array before the `N` elements so that when you decode you know how many elements to read.

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+A HLSLRootSignatureDecl will be created and added to the HLSLRootSignatureAttr 
+for diagnostic and saved to the HLSLEntryRootSignatureAttr for clang
```
damyanp, line 423: Why are there separate HLSLRootSignatureDecl and HLSLRootSignatureAttr nodes?  What's the difference between them?  Is this a normal pattern in Clang?

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+occur in Sema.
+
+To ensure that a used resource has a binding in the root signature, this
+information should be accessible in the backend to make sure legal dxil is
+generated.
+
+A new AST node HLSLRootSignatureDecl will be added to represent the root
+signature in the AST.
+To avoid parsing root signature string twice, HLSLRootSignatureDecl will 
+save the parsed root siganture instead of the string.
```
coopp, line 138: spelling: siganture

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+
+```
+
+Because the RootSignature attribute in hlsl only have the string, it is easier
+to add a StringArgument to save the string first.
+A HLSLRootSignatureDecl will be created and added to the HLSLRootSignatureAttr 
+for diagnostic and saved to the HLSLEntryRootSignatureAttr for clang 
+code generation.
+The HLSLRootSignatureDecl will save StringLiteral instead StringRef for 
+diagnostic.
```
damyanp, line 426: I don't understand the significance of using StringLiteral instead of StringRef here.  Is there something about StringLiteral that makes diagnostics easier?  What's the reason for the decision to use one over the other?

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+    "DescriptorTable(SRV(t1, numDescriptors = 2))"  // Static index and vertex buffers.
+};
+
+LocalRootSignature MyLocalRootSignature = 
+{
+    "RootConstants(num32BitConstants = 4, b1)"  // Cube constants 
+};
+
+```
+ only a HLSLRootSignatureDecl will be created.
```
damyanp, line 460: How are global/local root signatures represented in the AST?

python3kgae: As a HLSLRootSignatureDecl  node.

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+Additionally, a new AST node, SubobjectToExportsAssociationDecl, should be 
+added to handle the SubobjectToExportsAssociation statement in HLSL.
+```
+SubobjectToExportsAssociation MyLocalRootSignatureAssociation =
+{
+    "MyLocalRootSignature",    // Subobject name
+    "MyHitGroup;MyMissShader"  // Exports association 
+};
+```
+The SubobjectToExportsAssociationDecl will contain a HLSLRootSignatureDecl and
```
damyanp, line 471: Presumably something at some point validates that the subject name and the exports actually refer to things that exist.  When does this happen?

HLSLRootSignatureDecl contains the full rootsignature definition string, doesn't it, not just the name of the root signature?  Or does this AST node mean different things depending on the context it is referenced from?

proposals/infra/INF-0002-root-signature-in-clang.md
python3kgae: StringLiteral has source location which will help to emit diagnostic.

proposals/infra/INF-0002-root-signature-in-clang.md
python3kgae: HLSLRootSignatureDecl  is shared between RootSignature attribute and local/global root signature.
HLSLRootSignatureAttr  is only for the RootSignature attribute.

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+};
+enum class TextureAddressMode : uint32_t {
+  Wrap = 1,
+  Mirror = 2,
+  Clamp = 3,
+  Border = 4,
+  MirrorOnce = 5
+};
+
+struct ContainerRootDescriptor {
```
coopp, line 311: Many of the structs defined below have initialized default values, but there are some with members that are missing values.  Was this intentional?

Example:
ContainerRootDescriptor initializes RegisterSpace but not the others.
ContainerDescriptorRange is missing inits

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+{
+    "MyLocalRootSignature",    // Subobject name
+    "MyHitGroup;MyMissShader"  // Exports association 
+};
+```
+The SubobjectToExportsAssociationDecl will contain a HLSLRootSignatureDecl and 
+a list of FunctionDecls.
+
+
+The default root signature version number is 2 which maps to rootsig_1_1.
```
coopp, line 475: Seems like this should be higher up in the document.  Talking about the different versions of root signatures.  Are you calling this info out here because it is going to be written in the metadata at some point?

Also would the version also be stored in the deserialized version of the root signature?

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+       DescriptorTable( CBV(b1), SRV(t1, numDescriptors = 8,flags = DESCRIPTORS_VOLATILE),
+       UAV(u1, numDescriptors = unbounded, flags = DESCRIPTORS_VOLATILE)),
+       DescriptorTable(Sampler(s0, space=1, numDescriptors = 4)),
+       RootConstants(num32BitConstants=3, b10),
+       StaticSampler(s1),
+       StaticSampler(s2, addressU = TEXTURE_ADDRESS_CLAMP, filter = FILTER_MIN_MAG_MIP_LINEAR )"
+       HLSLRootSignature 'main.RS'
+
+```
+It is possible to avoid the root signature string in HLSLEntryRootSignatureAttr
```
coopp, line 583: I'm confused by this statement.  Are you pointing out an optimization here?  Leaving out / adding the root signature string in the printed output?  Is this for diagnostics?

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+
+```
+
+
+## Alternatives considered (Optional)
+
+* Save root signature string to AST.
+
+  Assuming the cost to parse root signature string is cheap,
+  HLSLRootSignatureDecl could follow common Clang approach which only save
```
coopp, line 677: Following the common clang approach is a good starting point.  More investigation would be needed to really bottom out on how expensive string parsing is and how often it needs to be parsed.  If it is parsed once into an object that can be passed to other places to be inspected that might not be too bad, but I don't fully understand how clang deals with strings in the AST yet.  Maybe it stays as a string forever and travels around to consumers of the AST.

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+void ASTDeclWriter::VisitHLSLRootSignatureDecl(HLSLRootSignatureDecl *D) {
+  VisitNamedDecl(D);
+  Record.AddSourceLocation(D->getKeywordLoc());
+  Record.AddString(D->getRootSignatureString());
+  Code = serialization::DECL_HLSL_ROOT_SIGNATURE;
+}
+```
+
+  Con:
+    Need to parse more than once.
```
coopp, line 711: Ahh.. So it will be parsed multiple times.  I see.

proposals/infra/INF-0002-root-signature-in-clang.md
python3kgae: Updated this part.

proposals/infra/INF-0002-root-signature-in-clang.md
python3kgae: Fixed.

proposals/infra/INF-0002-root-signature-in-clang.md
python3kgae: The name exist should be checked in Sema when creating the SubobjectToExportsAssociationDecl node. If cannot find the name, it should report error and SubobjectToExportsAssociationDecl will not be created. 

HLSLRootSignatureDecl  will have the name of the root signature (faked name for a root signature attribute) and a ParsedRootSignature object.

proposals/infra/INF-0002-root-signature-in-clang.md
python3kgae: Cannot find a good place to add it. Do you have suggestion about it?

The version stored in deserialized/serialized version of root signature.

proposals/infra/INF-0002-root-signature-in-clang.md
python3kgae: I'm pointing out that print the root signature string for HLSLEntryRootSignatureAttr when dump AST is optional for implementation.

proposals/infra/INF-0002-root-signature-in-clang.md
python3kgae: Not all members have default value.
Added the missing RegisterSpace init.

proposals/infra/INF-0002-root-signature-in-clang.md
python3kgae: Fixed.

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,717 @@
...
+    {
+    …
+    }
+
+```
+
+The other mechanism is to create a standalone root signature blob, perhaps to 
+reuse it with a large set of shaders, saving space. There're shader model 
+rootsig_1_0 androotsig_1_1 to generate the standalone root signature blob.
+The name of the define string is specified via the usual /E argument.
```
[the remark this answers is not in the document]
python3kgae: The target -Trootsig_1_0/1 is for compiling a root signature.

The macro name for the root signature string is treated as entry point name for this case.

damyanp:
Superseded by #209.

PR #197, opened 2024-04-12
Add details about existing nano-COM compiler library and include description of features and code snippet usage

coopp:
This commit includes details about the dxcapi nano-COM library.  

The point of this change is to ensure that there is understanding of what is supported and supplied in today's compiler library before designing any c-style exports.

The approach taken:
* What features are in the dxcapi COM library?
* What form is the data being passed around?  How is it created/manipulated etc?
* How do I compile a shader?
* How do I access reflection data?
* What are the other parts

There are still some missing snippets that will need to be added (maybe).
* How to use the linker... and what scenarios is it used for?
* How to optimize a shader?
* How to use the include handler?
* Perhaps others..

### DIFF
--- a/proposals/0016-c-interface-compiler-library.md
+++ b/proposals/0016-c-interface-compiler-library.md
@@ -63,12 +63,13 @@ supports the following features.
 * [PDB Symbol Access](#shader-pdbs)
 
 These features will also need to be supported by the new c-style library.
-Code snippets on how to compile a shader using the COM library are located
+Code snippets on how to use the COM library are located
 in [DXC api code snippets](#dxc-code-snippets).
 
+### DXC Library Api descriptions and code snippets
 
 The following interfaces are used to work with data being passed to/from the
-library.
+DXC library.
 
 #### Buffers
 
--- a/proposals/0016-c-interface-compiler-library.md
+++ b/proposals/0016-c-interface-compiler-library.md
@@ -49,8 +49,109 @@ will also be addressed in this proposal. The design will help migration of
 older DXC-based solutions to adopt clang as the preferred HLSL compiler.
 
 ## Detailed design
+The current DirectX shader compiler library is a nano-COM implementation that
+supports the following features.
 
-_The detailed design is not required until the feature is under review._
+* [Include Handlers](#include-handlers)
+* [Compiler](#compile-shader)
+* [Linker](#link-shader)
+* [Reflection Data Access](#shader-reflection)
+* [DXIL Container Access](#dxil-container-access)
+* [PDB Symbol Access](#shader-pdbs)
+
+The following interfaces are used to work with data being passed to/from the
+library.
+
+#### Buffers
+```c++
+// Structure for supplying bytes or text input to Dxc APIs. Represents both
+// text (DxcText) and non-text byte buffers (DxcBuffer).
+struct DxcBuffer {
+  LPCVOID Ptr;
+  SIZE_T Size;
+  UINT Encoding;
+};
+typedef DxcBuffer DxcText;
+
+// General purpose buffers
+struct IDxcBlob : public IUnknown {
+public:
+  LPVOID GetBufferPointer();
+  SIZE_T GetBufferSize();
+};
+
+// String buffers that guarantee null-terminated text and the stated encoding
+struct IDxcBlobEncoding : public IDxcBlob {
+public:
+  HRESULT GetEncoding(BOOL *pKnown, UINT32 *pCodePage);
+};
+
+struct IDxcBlobUtf16 : public IDxcBlobEncoding {
+public:
+  LPCWSTR GetStringPointer();
+  SIZE_T GetStringLength();
+};
+
+struct IDxcBlobUtf8 : public IDxcBlobEncoding {
+public:
+  LPCSTR GetStringPointer();
+  SIZE_T GetStringLength();
+};
+```
+
+### Features supported in DirectX shader compiler library (dxcapi.h)
+#### Include Handlers
+Provided to customize the handling of include directives. An implementation of
+the following interface is provided as a hook into compilation. A default
+implementation that reads include files from the filesystem can also be created
+using IDxcUtils::CreateDefaultIncludeHandler.
+
+```c++
+struct IDxcIncludeHandler : public IUnknown {
+  HRESULT STDMETHODCALLTYPE LoadSource(
+    LPCWSTR pFilename,            // candidate filename.
+    IDxcBlob **ppIncludeSource);  // resultant source object for included file,
+                                  // nullptr if not found.
+};
+```
+
+#### Compile Shader
+IDxcCompiler3 is the most current entrypoint for compiling a shader or 
+disassembling DXIL containers/bitcode.
+
+Provides support for:
+* Compiling a single entry point to the target shader model
+* Compiling a library to a library target (using -R lib_*)
+* Compiling a rootsignature (-T rootsig_*)
+* Preprocessing HLSL source (-P)
+* Disassembling DXIL container or bitcode
+
+```c++
+struct IDxcCompiler3 : public IUnknown {
+  HRESULT Compile(
+    const DxcBuffer *pSource,            // source text to compile
+    LPCWSTR *pArguments,                 // array of pointers to arguments
+    UINT32 argCount,                     // number of arguments
+    IDxcIncludeHandler *pIncludeHandler, // user-provided interface to handle
+                                         // #include directives (optional)
+    REFIID riid, LPVOID *ppResult);      // IDxcResult: status, buffer, and errors
+
+  HRESULT Disassemble(
+    const DxcBuffer *pObject,            // program to disassemble: dxil container or bitcode.
+    REFIID riid, LPVOID *ppResult);      // IDxcResult: status, disassembly text, and errors
+};
+```
+#### Link Shader
+TBD
+
+#### Shader Reflection
+TBD
+
+#### DXIL Container Access
+TBD
+
+#### Shader PDBs
+TBD
 
 ## Alternatives considered for supporting legacy toolchains
 
--- a/proposals/0016-c-interface-compiler-library.md
+++ b/proposals/0016-c-interface-compiler-library.md
@@ -829,9 +829,96 @@ if (errors && errors->GetStringLength() > 0) {
 ```
 
 ### Optimizing a shader
+A shader can be optimized by providing different passes lists to the optimizer.
+The code below shows how to get the optimizations performed and how to call
+into the IDxcOptimizer with a pass list.
+
+Passes can also be added as arguments for compilation.
 
 ```c++
-// How to optimize a shader
+    // Collect optimization passes and list them
+    arguments.push_back(L"-Odump");
+
+    com_ptr<IDxcResult> dumpResult;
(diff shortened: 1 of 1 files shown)


coopp:
I still have some more snippets to add to this document and probably expand on the descriptions of the supported features.  This is a first pass at getting some of this new content in front of the team.

proposals/0016-c-interface-compiler-library.md
```diff
@@ -50,7 +50,495 @@ older DXC-based solutions to adopt clang as the preferred HLSL compiler.
...
+library.
+
+#### Buffers
+```c++
+// Structure for supplying bytes or text input to Dxc APIs. Represents both
+// text (DxcText) and non-text byte buffers (DxcBuffer).
+struct DxcBuffer {
+  LPCVOID Ptr;
+  SIZE_T Size;
+  UINT Encoding;
```
damyanp, line 80: Is there a list of valid values for `Encoding` anywhere?  Presumably this comes from a enum or something?

proposals/0016-c-interface-compiler-library.md
```diff
@@ -50,7 +50,495 @@ older DXC-based solutions to adopt clang as the preferred HLSL compiler.
...
+  SIZE_T GetBufferSize();
+};
+
+// String buffers that guarantee null-terminated text and the stated encoding
+struct IDxcBlobEncoding : public IDxcBlob {
+public:
+  HRESULT GetEncoding(BOOL *pKnown, UINT32 *pCodePage);
+};
+
+struct IDxcBlobUtf16 : public IDxcBlobEncoding {
```
damyanp, line 97: Are these proper interfaces with uuid's that have been omitted from this spec, so that you can QI for IDxcBlobUtf16?

proposals/0016-c-interface-compiler-library.md
```diff
@@ -50,7 +50,495 @@ older DXC-based solutions to adopt clang as the preferred HLSL compiler.
...
+                              // or -Odump
+  DXC_OUT_REFLECTION = 8,     // IDxcBlob - RDAT part with reflection data
+  DXC_OUT_ROOT_SIGNATURE = 9, // IDxcBlob - Serialized root signature output
+  DXC_OUT_EXTRA_OUTPUTS  = 10,// IDxcExtraResults - Extra outputs
+
+  DXC_OUT_FORCE_DWORD = 0xFFFFFFFF
+} DXC_OUT_KIND;
+
+struct IDxcOperationResult : public IUnknown {
+  HRESULT GetStatus(HRESULT *pStatus);
```
damyanp, line 139: missing `}`

proposals/0016-c-interface-compiler-library.md
```diff
@@ -50,7 +50,495 @@ older DXC-based solutions to adopt clang as the preferred HLSL compiler.
...
+  HRESULT GetOutput(DXC_OUT_KIND dxcOutKind,
+                    REFIID iid, void **ppvObject,
+                    IDxcBlobUtf16 **ppOutputName);
+  UINT32 GetNumOutputs();
+  DXC_OUT_KIND GetOutputByIndex(UINT32 Index);
+  DXC_OUT_KIND PrimaryOutput();
+};
+```
+#### Helper library
+The shader compiler library also provides a helper library IDxcUtils that is
```
damyanp, line 152: There's a lib called IDxcUtils?  I'm not sure that's the case is it?  How do I get an IDxcUtils pointer?

damyanp: Looks like IDxcUtils fulfils a factory-like role in this API.

proposals/0016-c-interface-compiler-library.md
```diff
@@ -50,7 +50,495 @@ older DXC-based solutions to adopt clang as the preferred HLSL compiler.
...
+
+Provides support for:
+* Compiling a single entry point to the target shader model
+* Compiling a library to a library target (using -R lib_*)
+* Compiling a rootsignature (-T rootsig_*)
+* Preprocessing HLSL source (-P)
+* Disassembling DXIL container or bitcode
+
+```c++
+struct IDxcCompiler3 : public IUnknown {
```
damyanp, line 267: Checking: this really derives from IUnknown and not IDxcCompiler2?

proposals/0016-c-interface-compiler-library.md
```diff
@@ -50,7 +50,495 @@ older DXC-based solutions to adopt clang as the preferred HLSL compiler.
...
+    REFIID riid, LPVOID *ppResult);      // IDxcResult: status, buffer, and
+                                         // errors
+
+  HRESULT Disassemble(
+    const DxcBuffer *pObject,            // dxil container or bitcode
+    REFIID riid, LPVOID *ppResult);      // IDxcResult: status, disassembly
+                                         // text, and errors
+};
+```
+#### Shader Linker
```
damyanp, line 283: nit: although this renders correctly, it looks like github's syntax highlighting gets confused.  I think some vertical space between blocks might help it would a bit.

proposals/0016-c-interface-compiler-library.md
```diff
@@ -50,7 +50,495 @@ older DXC-based solutions to adopt clang as the preferred HLSL compiler.
...
+  // Assemble dxil in ll or llvm bitcode to DXIL container.
+  HRESULT AssembleToContainer(
+    IDxcBlob *pShader,               // shader to assemble
+    IDxcOperationResult **ppResult); // output status, buffer, and errors
+};
+```
+
+#### Shader Reflection
+A DXIL container can be inspected and different parts can be accessed using
+IDxcContinerReflection. ID3D12ShaderReflection can also be retrieved using
```
damyanp, line 331: Isn't ID3D12ShaderReflection part of the API we should be looking at as well?

proposals/0016-c-interface-compiler-library.md
coopp: I removed the uuids, but yes.  IDXcBlob is built on IUnknown. and Encoding is from Blob, and Utf16 is from Encoding. . I am not sure QI works all the way through as I have not looked at the interface impl.  If COM rules apply then I would think it would be possible.. But nano-COM sometimes changes things.

proposals/0016-c-interface-compiler-library.md
coopp: I showed how that can be obtained in the snippets.  CreateInstance is called to get that object

proposals/0016-c-interface-compiler-library.md
coopp: Yep!

proposals/0016-c-interface-compiler-library.md
coopp: Yes.  I wasn't sure how deep the rabbit hole goes, but you are correct.  This needs some attention here to cover ALL areas of functionality.  Obviously we cannot pop out that interface from a flat c-api. :)

proposals/0016-c-interface-compiler-library.md
damyanp: If QI doesn't work as you'd expect it to then it isn't nano-COM.

proposals/0016-c-interface-compiler-library.md
coopp: ID3D12ShaderReflection is implemented in DXC!  Interesting.  Guess that fits.  It is strange that the interface name is D3D12 runtime looking.

coopp: I'll add it to this spec so it covers all supported interfaces DXC exposes.

proposals/0016-c-interface-compiler-library.md
coopp: Added in latest push

proposals/0016-c-interface-compiler-library.md
coopp: Added vertical whitespace

proposals/0016-c-interface-compiler-library.md
coopp: Added interfaces and structs hopefully in a good order to understand usage.

proposals/0016-c-interface-compiler-library.md
```diff
@@ -148,9 +159,16 @@ struct IDxcResult : public IDxcOperationResult {
   DXC_OUT_KIND PrimaryOutput();
 };
 ```
+
 #### Helper library
 The shader compiler library also provides a helper library IDxcUtils that is
 used to create/consume the different data in the forms referred to above.
```
damyanp, line 165:
```suggestion
#### IDxcUtils
The shader compiler library also provides a set of utility functions via IDxcUtils. This can be
used to create/consume the different data in the forms referred to above.
```

I really don't like calling this a "helper library" since it isn't a library in the .lib or .dll sense of the word.

proposals/0016-c-interface-compiler-library.md
```diff
@@ -354,11 +374,326 @@ struct IDxcContainerReflection : public IUnknown {
...
 
+ID3D12ShaderReflection provides access to descriptions and other information
+about the many different parts of the shader.
+
+* [Shader Description](#shader-description)
+* [Buffers](#shader-buffers-description)
+* [Variables and type information](#shader-types=and-variables)
+* [Functions and parameters](#shader-functions-and-parameters)
+
+PIX uses this interface to obtain nice buffer formatting behavior for the UI.
```
damyanp, line 385: PIX isn't the only user of this interface, and the buffer formatting behavior is just one possible use of it.  Suggest making it clearer in the wording that this is just an example scenario.

proposals/0016-c-interface-compiler-library.md
```diff
@@ -492,11 +830,100 @@ if (errors && errors->GetStringLength() > 0) {
 ```
 
 ### Optimizing a shader
+A shader can be optimized by providing different passes lists to the optimizer.
```
damyanp, line 833: Would be great to have some insight into how widely this is used.  I think it is quite telling that Copilot is unable to write me any sensible example code that uses this interface.

llvm-beanz (approved):
I had a couple comments, but nothing that needs to block the PR. Feel free to resolve them once you've read them.

proposals/0016-c-interface-compiler-library.md
```diff
@@ -50,7 +50,922 @@ older DXC-based solutions to adopt clang as the preferred HLSL compiler.
...
+The current DirectX shader compiler library is a nano-COM implementation that
+supports the following features.
+
+* [Include Handlers](#include-handlers)
+* [Compiler](#shader-compiler)
+* [Linker](#shader-linker)
+* [Validation](#shader-validation)
+* [Reflection Data Access](#shader-reflection)
+* [DXIL Container Access](#dxil-container-access)
+* [PDB Symbol Access](#shader-pdbs)
```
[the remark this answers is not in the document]
coopp: Added a blurb calling out this missing info to the doc.

proposals/0016-c-interface-compiler-library.md
coopp: Agreed.  It was hard to find info on this to even write my snippet.  Nothing really out there about it.

proposals/0016-c-interface-compiler-library.md
coopp: Fixed

PR #198, opened 2024-04-15
[Language] Begin Basic.types

### LINKED ISSUE #195: Missing specification of foundational types
https://github.com/microsoft/hlsl-specs/tree/main/specs/language

The spec is missing a section describing the fundamental types defined in HLSL.

**Additional context**
https://github.com/microsoft/hlsl-specs/pull/168#discussion_r1556445951

llvm-beanz:
This begins drafting the Basic.types section which specifies the foundations of HLSL's data types. This includes some basic information about representations as well as the classifications of types and the basic behaviors of some types.

Fixes #195


specs/language/basic.tex
```diff
@@ -1,5 +1,97 @@
...
+\textit{cv-qualified} versions of these types are collectively called
+\textit{scalar types}.
+
+\p Vectors of scalar types declared with the built-in \texttt{vector<T,N>}
+template are \textit{vector types}.
+
+\p \textit{POD types}\footnote{POD is an initialism for Plain-Old-Data which is
+referenced in the C++ specification but never expanded.} are POD class types,
+scalar types, vector types, and arrays of such types. \textit{POD class types}
+are class types that only contain members POD types.
```
bogner, line 41: C++ moved away from the "POD" nomenclature in C++20 (they split it into "trivial" and "standard-layout" types - the old "POD" types are types that are both trivial and standard-layout). We should probably either follow their lead or not bother with this definition at all. I suspect we can get away with leaving this out, at least for now.

specs/language/basic.tex
```diff
@@ -1,5 +1,97 @@
...
 
+\Sec{Types}{Basic.types}
+
+\p The \textit{object representation} of an object of type \texttt{T} is the
+sequence of \textit{N} bytes taken up by the object of type \texttt{T}, where
+\textit{N} equals \texttt{sizeof(T)}\footnote{\texttt{sizeof(T)}, returns the
+size of an object of type \texttt{T} in bytes as stored in device memory.
+The size may be different in other memory spaces and no operator exists to query
+it.}. The \textit{object representation} of an object may be different based on
+the \textit{memory space} it is stored in (\ref{Intro.Memory.Spaces}).
```
bogner, line 11: The wording in this footnote is confusing. I think it's saying
> "sizeof" returns the size of the struct as-if it's stored in device memory, and determining the size if it's stored in another memory space is not possible

specs/language/basic.tex
```diff
@@ -1,5 +1,97 @@
...
+
+\p The \textit{value representation} of an object is the set of bits that hold
+the value of type \texttt{T}.
```
bogner, line 14: Should we mention padding? That is, the "object representation" presumably includes padding, and the "value representation" presumably does not, as in C++.

specs/language/basic.tex
```diff
@@ -1,5 +1,97 @@
...
+
+\p An \textit{object type} is a type that is not a function type, not a
+reference type, and not a void type.
+
+\p Arithmetic types (\ref{Basic.types.arithmetic}), enumeration types, and
+\textit{cv-qualified} versions of these types are collectively called
+\textit{scalar types}.
+
+\p Vectors of scalar types declared with the built-in \texttt{vector<T,N>}
+template are \textit{vector types}.
```
bogner, line 36: Should we specify that `1 <= N <= 4` in the vector built-in, or is that best left to be defined elsewhere?

specs/language/basic.tex
```diff
@@ -1,5 +1,97 @@
...
+standard signed integer types are collectively called \textit{signed integer
+types}.
+
+\p There are three \textit{standard unsigned integer types}: \texttt{uint16\_t},
+\texttt{uint32\_t}, and \texttt{uint64\_t}. Each of the unsigned integer types is
+explicitly named for the size in bits of the type's object representation. There
+is also the type alias \texttt{uint} which is an alias of \texttt{uint32\_t}. The
+standard unsigned integer types are collectively called \textit{unsigned integer
+types}\footnote{HLSL does not inherit C's default type rule, nor does it
+inherit the \texttt{unsigned} keyword.}.
```
bogner, line 58: I find "C's default type rule" fairly ambiguous - C has a lot of defaults when it comes to types. Is this simply saying that HLSL does not have C's `signed` and `unsigned` modifiers for integral types?

specs/language/basic.tex
```diff
@@ -1,5 +1,97 @@
...
+function that returns no value. Any expression can be explicitly converted to
+\texttt{void}.
+
+\Sub{Special Classes}{Basic.types.special}
+
+\p A class type may be an \textit{ordinary class} or a \textit{special class}.
+An ordinary class is a class type that contains no members of special class,
+array of special class, or cv-qualified versions of these types. Special class
+objects can only exist in thread memory: they cannot be stored to any other
+memory space (\ref{Intro.Memory.Spaces}).
```
bogner, line 93: This definition is recursive with no base case - it only defines that a class is a special class if it contains a special class, but doesn't explain why a class would be special otherwise.

specs/language/basic.tex
llvm-beanz: I struggled with this a bunch. Part of the reason I went with POD wording was because I felt it was clear to understand and the current C++ wording gets _way_ complicated because of resource types.

Technically resource types are trivially copyable, types containing resources are trivially copyable, but they also have magic special behavior and can't be used in some contexts.

It makes sense to me that C++ focuses on the ability to copy values of the types, but that doesn't quite give the distinction we need for HLSL.

specs/language/basic.tex
llvm-beanz: Yea, I think one of the problems with our definition of `sizeof` is that we don't allow you to specify which address space you want the size for. The value returned from `sizeof` is probably valid for most memory spaces, but it isn't valid for constant. For example:

```hlsl
// In a cbuffer this structure is 3 rows because of the
// packing rules, which makes it 48 bytes instead of 16.
struct FourDWords {
  float2 V;
  float A[2];
};
_Static_assert(sizeof(FourDWords) == sizeof(uint) * 4, "What you expect");
```

Also `sizeof` fails to compile if the type is a "special" type that only has a thread representation. For example:

```hlsl
struct SpecialType {
  uint U;
  RWBuffer<uint> B;
};
// error: invalid application of 'sizeof' to non-numeric type 'SpecialType'
_Static_assert(sizeof(SpecialType) == sizeof(uint), "Fails to compile");
```

llvm-beanz:
The PDF build is here: https://github.com/microsoft/hlsl-specs/actions/runs/8724175011/artifacts/1422465192

In case anyone doesn't like reading LaTeX.

specs/language/basic.tex
```diff
@@ -1,5 +1,112 @@
 \Ch{Basic Concepts}{Basic}
 
+\Sec{Types}{Basic.types}
+
+\p The \textit{object representation} of an object of type \texttt{T} is the
+sequence of \textit{N} bytes taken up by the object of type \texttt{T}, where
+\textit{N} equals \texttt{sizeof(T)}\footnote{\texttt{sizeof(T)}eturns the size
```
coopp, line 7: spelling: eturns.  Did you mean returns?

coopp (approved):
Looks good to me.  Glad this information is all captured now.  Very interesting read for sure!

specs/language/basic.tex
```diff
@@ -1,5 +1,113 @@
...
+and it is not an intangible type. It is a \textit{complete type} if the compiler
+has sufficient information to determine the size of an object of type
+\texttt{T}, or if the type is known to be an intangible type. An object may not
+be defined to have an \textit{incomplete} type.
+
+\p Arithmetic types (\ref{Basic.types.arithmetic}), enumeration types, and
+\textit{cv-qualified} versions of these types are collectively called
+\textit{scalar types}.
+
+\p Vectors of scalar types declared with the built-in \texttt{vector<T,N>}
```
damyanp, line 49: Presumably somewhere we have something like `use float4 = vector<float, 4>`?  Wondering if it'd be helpful to mention these names that most users of HLSL are likely more familiar with here?

damyanp (approved):

specs/language/basic.tex
```diff
@@ -1,5 +1,112 @@
 \Ch{Basic Concepts}{Basic}
 
+\Sec{Types}{Basic.types}
+
+\p The \textit{object representation} of an object of type \texttt{T} is the
+sequence of \textit{N} bytes taken up by the object of type \texttt{T}, where
+\textit{N} equals \texttt{sizeof(T)}\footnote{\texttt{sizeof(T)}eturns the size
+of the struct as-if it's stored in device memory, and determining the size if
```
tex3d: Is there a glossary entry for "device memory"?  I imagine by this term you mean something like host-visible device memory or some specific memory type that is a subset of this.  Non-host-visible device memory layouts (including group shared memory, local memory for registers/arrays, inter-stage attributes, inter-function parameters/payloads, etc...) could be optimized in ways that are not visible to the host application or directly observable by the shader.

llvm-beanz: "device memory" is defined in `Intro.Memory.Spaces` which is earlier in the document. The end of the next sentence has a reference to that section of the document.

specs/language/basic.tex
llvm-beanz: Yea, we have the typedefs in clang [here](https://github.com/llvm/llvm-project/blob/main/clang/lib/Headers/hlsl/hlsl_basic_types.h). We use `typedef` rather than `using` because `using` is an HLSL 2021 feature, so that makes the header compatible to older versions.

The C and C++ specs defines all the typedefs later in the document, which had been my plan. Basically anything that can be implemented in a header goes way late.

specs/language/basic.tex
llvm-beanz: I think the definition of `sizeof` is particularly problematic and we don't have a well enough defined memory model to cover all the relevant cases.

bogner (approved):
This looks much better - I find the "intangible types" terminology and definitional approach much clearer than the "special types" wording from before.

Note: While I see `classes.tex` in the sources here, it doesn't seem to be rendered in the artifact here: https://github.com/microsoft/hlsl-specs/actions/runs/8726130788/artifacts/1422979344. Similarly, the "Intangible Types" page there is blank - is that missing or do you intend to put it there as a placeholder to be filled in later?

llvm-beanz:
> Note: While I see `classes.tex` in the sources here, it doesn't seem to be rendered in the artifact here: https://github.com/microsoft/hlsl-specs/actions/runs/8726130788/artifacts/1422979344. Similarly, the "Intangible Types" page there is blank - is that missing or do you intend to put it there as a placeholder to be filled in later?

I meant to delete `classes.tex` and somehow re-added it. Clearly git and I don't get along. It is removed now.

The classes and intangible types chapters are placeholders to be filled out later. I thought it was better to have references to empty sections rather than a bunch of TODO comments to add references later.

specs/language/basic.tex
llvm-beanz: I'm going to mark this conversation as resolved since I don't know that we have an actionable change here. I think we have some bigger issues around clearly defining the memory model for HLSL (and DXIL) that we'll also need to figure out over the coming months. We'll also need to revisit this wording as we try to close in on a better definition for `sizeof`.

specs/language/basic.tex
llvm-beanz: I've filed an issue to track adding this documentation:
https://github.com/microsoft/hlsl-specs/issues/203

bharadwajy:
Several (may be all?) of the definitions of the HLSL language concepts specified here appear (by design and choice, I presume) very similar to those of C++, for example [types](https://en.cppreference.com/w/cpp/language/type). If so, would it be prudent to defer to the public C++ specification for concepts that are the same in HLSL and C++ while providing details of concepts and support that is specific to HLSL and differ from C++? This could prevent ambiguous interpretation of concepts supported in both the languages owing to potentially different verbiage.

llvm-beanz:
> Several (may be all?) of the definitions of the HLSL language concepts specified here appear (by design and choice, I presume) very similar to those of C++, for example [types](https://en.cppreference.com/w/cpp/language/type). If so, would it be prudent to defer to the public C++ specification for concepts that are the same in HLSL and C++ while providing details of concepts and support that is specific to HLSL and differ from C++? This could prevent ambiguous interpretation of concepts supported in both the languages owing to potentially different verbiage.

In general, I think this is really great and important feedback. I've tried to cite the C and C++ specifications where possible. You'll note glossary references to `isoC23`, and `isoCPP` in this PR. I've also cited `IEEE754` where I could instead of defining the floating point type representations. Unfortunately a lot of our behaviors are subtly different from C and C++, so citing the existing specifications doesn't really make sense.

For example in this PR, HLSL defines size-specified integer types, and the floating point type have explicitly defined sizes too. While we do satisfy C's guarantee that `float` is <= the size of `double`, in HLSL we can make a stronger guarantee. We also have the min-precision types which have completely different challenges. Because of those differences neither C nor C++'s definitions of arithmetic types fit HLSL.

Probably about the only direct overlap in this PR between HLSL and C++ is the definitions of _object representation_, _value representation_ and _object type_. I deliberately chose to define those terms (even if slightly differently) so that the definition is in our document. My main reasoning is that it would suck when reading the spec to need to jump to another document to find a single-sentence term definition, and those definitions can move around between versions of the specification.

There is some shared language around Classes, but HLSL classes are _very_ different from C++ so I really don't want to cite C++ too much in that part of the document because I don't want to mislead people into thinking the languages align more than they do.

For the purpose of this PR I think we have very little that we can directly inherit from C or C++ beyond what I've already cited without a whole lot of extra qualifying language (i.e. "It's like C++ but with these differences..."). If you think I missed something where we could more concisely just cite the C or C++ specification please let me know.

specs/language/basic.tex
llvm-beanz: I'm going to mark this as resolved for this thread. The additional documentation will be handled in a separate PR.

PR #202, opened 2024-04-17
Fix whitespace

llvm-beanz:
One part of this is just a whitespace no-op, but putting the \define on its own line fixes some formatting in the PDF to properly indent the item.


dmpots (approved):

damyanp (approved):
significant whitespace!!!

PR #204, opened 2024-04-17
[INF-0001] Add environment parameter for per-stage availability annotations

### LINKED ISSUE #189: [INF-0001] Consider environment as a parameter to availability attribute
[INF-0001 Availability Diagnostics](https://github.com/microsoft/hlsl-specs/blob/main/proposals/infra/INF-0001-availability-diagnostics.md)

There was some back and forth in two threads [here](https://github.com/microsoft/hlsl-specs/pull/153#discussion_r1550567652) and [here](https://github.com/microsoft/hlsl-specs/pull/153#discussion_r1547497568) about the representation of the environment component of the target triple. In LLVM we use the environment to model DirectX shader stages, which are analogues to SPIR-V execution models.

The number of permutations that we need to support here could grow over time and get large. In fact, the permutations that Apple has added already have exploded in recent years. It might be worth considering adding an `environment` parameter to the attribute to reduce future explosion of permutations.

hekota:
Proposal to add environment parameter to the Clang availability attribute and use it for per-stage availability annotations.

Also rename intrinsic -> function to keep the naming consistent.

Fixes #189

### DIFF
--- a/proposals/infra/INF-0001-availability-diagnostics.md
+++ b/proposals/infra/INF-0001-availability-diagnostics.md
@@ -62,8 +62,8 @@ model 6.0+. This AST annotation allows clangd to communicate to users the
 availability information to language server protocol clients. This can drive
 more informed auto-complete, refactoring, and in-editor diagnostics.
 
-The Clang availability attribute works well when a function availability depends
-only on the shader model version. However, the availability of some HLSL
+The Clang availability attribute works well when a function's availability
+depends only on the shader model version. However, the availability of some HLSL
 functions depends not only on the shader model version but also on the target
 shader stage. For example the derivative functions `ddx` and `ddy` were
 introduced in Shader Model 2.0 for use only in pixel shaders. In Shader Model
@@ -76,7 +76,7 @@ identical to the environment component of the `llvm::Triple`. If the
 `environment` parameters is present, the declared availability attribute would
 apply only for targets with the same environment.
 
-Using the new `environment` parameter the per-stage availability annotation for
+Using the new `environment` parameter, the per-stage availability annotation for
 the derivative function `ddx` would look like this:
 
 ```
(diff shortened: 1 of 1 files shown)


llvm-beanz (approved):
LGTM. I like the new syntax!

proposals/infra/INF-0001-availability-diagnostics.md
```diff
@@ -72,6 +62,40 @@ model 6.0+. This AST annotation allows clangd to communicate to users the
 availability information to language server protocol clients. This can drive
 more informed auto-complete, refactoring, and in-editor diagnostics.
 
+The Clang availability attribute works well when a function availability depends
```
damyanp, line 65:
```suggestion
The Clang availability attribute works well when a function's availability depends

```

proposals/infra/INF-0001-availability-diagnostics.md
```diff
@@ -72,6 +62,40 @@ model 6.0+. This AST annotation allows clangd to communicate to users the
...
+6.6 their support was extended to compute, mesh and amplification shaders, and
+they are not available in any other shader stage.
+
+In order to encode this information we propose adding a new `environment`
+parameter to the Clang availability attribute. The allowed values would be
+identical to the environment component of the `llvm::Triple`. If the
+`environment` parameters is present, the declared availability attribute would
+apply only for targets with the same environment.
+
+Using the new `environment` parameter the per-stage availability annotation for
```
damyanp, line 79:
```suggestion
Using the new `environment` parameter, the per-stage availability annotation for

```

proposals/infra/INF-0001-availability-diagnostics.md
```diff
@@ -270,26 +294,26 @@ may produce warnings or errors with Clang.
...
+'a' is not available in S shader environment on HLSL Shader Model x.y
 ```
 
-If intrinsic `a` is available in shader stage `S` in shader model higher than
-the target shader model the diagnostic message should be:
+If function `a` is available in shader stage `S` in shader model higher than the
+target shader model the diagnostic message should be:
 ```
-'a' is available in S shaders beginning with Shader Model x.y
+'a' is only available in S shader environment on HLSL Shader Model x.y or newer
```
damyanp, line 316:
```suggestion
'a' is only available in 'S' shader environment on HLSL Shader Model x.y or newer

```

Wondering about quoting the name of the environment here to make it more clear what the variable thing is?

damyanp (approved):
LGTM, I have a few nit picky comments that I don't feel strongly about.

proposals/infra/INF-0001-availability-diagnostics.md
damyanp: While I'm picking at this, is "HLSL Shader Model" the correct term to use?  HLSL can be used in contexts where Shader Model doesn't apply, and also non-HLSL languages can be used with D3D where Shader Model does apply.

hekota: > Wondering about quoting the name of the environment here to make it more clear what the variable thing is?

I'd prefer to leave it unquoted. It will resolve to "...in pixel shader environment..." and such and it is based on the compilation target. The 'a' comes from the input file so that's why it is quoted.

PR #206, opened 2024-04-19
Update conversion ranks for minimum precision types

llvm-beanz:
This just updates the conversion rank for minimum precision types which are now represented in the Basic.types section.

I've iterated on this a few times and reduced it down to just the min16{int|uint|float} types because those are the only types DXC actually supports. Using any of the other types produces a warning and they get treated as the corresponding min16 type.

The promotion orders are as follows:

min16int < int16_t < int32_t < int64_t
min16uint < uint16_t < uint32_t < uint64_t
min16float < half < float < double


coopp (approved):
The rank of this review shall have a minimum precision value of 'Approved'.

damyanp (approved):

PR #207, opened 2024-04-19
Add doc for packoffset.

python3kgae:
Added design doc for packoffset to start the discussion.

Right now, cbuffer llvm IR and metadata is not ready yet.
Need to update the related part once cbuffer got full support.

Also need to decide how to support 16bit types.

### DIFF
--- a/proposals/infra/INF-0003-packoffset.md
+++ b/proposals/infra/INF-0003-packoffset.md
@@ -1,0 +1,141 @@
+<!-- {% raw %} -->
+
+# Strict Availability Diagnostics
+
+* Proposal: [INF-0003](INF-0003-packoffset.md)
+* Author(s): [Xiang Li](https://github.com/python3kgae)
+* Sponsor: 
+* Status: **Under Consideration**
+* Planned Version: 
+
+## Introduction
+
+packoffset attribute is used to change the layout of a cbuffer.
+
+```
+packoffset( c[Subcomponent][.component] )
+```
+
+It will overwrite the layout of a cbuffer with the packoffset attributes on
+the constants.
+
+Here are several examples of manually packing shader constants.
+
+```
+cbuffer MyBuffer
+{
+    float4 Element1 : packoffset(c0);
+    float1 Element2 : packoffset(c12);
+    float1 Element3 : packoffset(c1.y);
+}
+```
+without the packoffset, Element2 will be in c1.x.
+
+
+### Limitations
+
+There're some limitations when apply packoffset on a constant.
+
+1. packoffset is only allowed in a constant buffer.
+```
+float c : packoffset(c0); // invalid for apply packoffset on a global constant.
+```
+
+2. cannot mix packoffset elements with nonpackoffset elements in a cbuffer.
+```
+cbuffer A {
+  float a : packoffset(c0);
+  float b;  // invalid for mix packoffset elements and nonpackoffset elements.
+}
+```
+3. Pack subcomponents of vectors and scalars whose size is large enough to prevent crossing register boundaries.
+```
+cbuffer A {
+  float2 a : packoffset(c0.w); // invalid for a.x in c0.w and a.y in c1.x 
+                               // which crossed register boundary.
+}
+```
+4. struct, matrix, array cannot have component.
+```
+struct S {
+  float a;
+};
+
+cbuffer A{
+  S s : packoffset(c2.y); // invalid for offset to component y
+  int a[2] : packoffset(c3.z); // invalid for offset to component z
+  float2x2 m : packoffset(c6.z); // invalid for offset to component z
+  S s2 : packoffset(c12); // valid
+  int a2[2] : packoffset(c13); // valid
+  float2x2 m2 : packoffset(c16); // valid
+}
+```
+
+## Motivation
+
+To support all HLSL features available in compute profile.
+
+## Proposed solution
+
+### AST
+
+A new attribute HLSLPackOffsetAttr will be created with:
+```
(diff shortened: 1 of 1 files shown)


proposals/infra/INF-0003-packoffset.md
```diff
@@ -0,0 +1,141 @@
...
+## Proposed solution
+
+### AST
+
+A new attribute HLSLPackOffsetAttr will be created with:
+```
+def HLSLPackOffset: HLSLAnnotationAttr {
+  let Spellings = [HLSLAnnotation<"packoffset">];
+  let LangOpts = [HLSL]
+  let Args = [IntArgument<"RegNum">, IntArgument<"Component", /*optional*/ 1>]
```
damyanp, line 87: Does this automatically understand how to convert something like `c2.w`  into `2` and `3`? Or do you have to write specialized code for it?

proposals/infra/INF-0003-packoffset.md
```diff
@@ -0,0 +1,141 @@
...
+```
+HLSLBufferDecl cbuffer MyBuffer
+  VarDecl Element1 float4
+    HLSLPackOffsetAttr   0 0
+  VarDecl Element2 float
+    HLSLPackOffsetAttr  12 0
+  VarDecl Element3 float
+    HLSLPackOffsetAttr   1 1
+  
+```
```
damyanp, line 101: If I write `packedoffset(t0)` where will that error be detected and reported?

proposals/infra/INF-0003-packoffset.md
```diff
@@ -0,0 +1,141 @@
...
+cbuffer CB {
+  float a : packoffset(c2);
+  float b : packoffset(c4);
+  float2 c : packoffset(c2.z);
+}
+```
+will get layout struct like
+```
+struct CBLayout {
+  float4 padding0[2]; // c0/c1
```
damyanp, line 118: With padding like this I wonder how something like this would work out:

```
cbuffer A {
  float 4 B : packoffset(c1);
}

float4 C : register(c0);
```

damyanp: Or similarly:
````
cbuffer A {
    float4 B : packoffset(c1);
}

cbuffer C {
  float4 D : register(c0);
}
````

File: proposals/infra/INF-0003-packoffset.md [hunk 4]
```diff
@@ -0,0 +1,141 @@
...
+    HLSLPackOffsetAttr  12 0
+  VarDecl Element3 float
+    HLSLPackOffsetAttr   1 1
+  
+```
+
+### llvm IR
+
+To apply the packoffset to a cbuffer, the layout of the struct constructed for 
+the cbuffer requires reordering of its fields, and padding should be introduced
```
damyanp, line 106: Might be worth explaining a bit about this layout struct concept?  Seems like this sort of background is needed to understand this proposal.

proposals/infra/INF-0003-packoffset.md
python3kgae: No. You have to write code to parse the attribute argument string.
Make it integer could save the result of the parse and checking.

There're 2 steps to create a HLSLPackOffsetAttr.
1. Parse the '(c0)' part in packoffset(c0) to build a ParsedAttribute with 'c0' as a String when meet packoffset.
2. Check the 'c0' and create a HLSLPackOffsetAttr when process ParsedAttribute.

For a normal attribute like [numthread(1,1,1)], the step 1 is generated automatically.
But packoffset is a HLSL annotation attribute which have to write code for step1. In the code, we could decide the argument as String even the td file say the argument is IntArgument.

The benefit of make the argument as IntArgument is no need to parse the string once the HLSLPackOffsetAttr is created.

proposals/infra/INF-0003-packoffset.md
python3kgae: When build HLSLPackOffsetAttr from a ParsedAttribute.

proposals/infra/INF-0003-packoffset.md
python3kgae: They're in different cbuffer.
The C is in global CB where B is in cbuffer A.

proposals/infra/INF-0003-packoffset.md
python3kgae: Added section about layout struct.

damyanp (approved):
Seems like a good start to me.  Will be interested in what the plan is for answering the open questions and for getting the finalized cbuffer IR. Do you think there's meaningful implementation that can be started without these resolved?

python3kgae:
> Seems like a good start to me.  Will be interested in what the plan is for answering the open questions and for getting the finalized cbuffer IR. Do you think there's meaningful implementation that can be started without these resolved?

Yes. The AST part should be OK to be done without the IR and metadata support.

proposals/infra/INF-0003-packoffset.md
```diff
@@ -0,0 +1,171 @@
+<!-- {% raw %} -->
+
+# Strict Availability Diagnostics
```
bogner, line 3: Looks like you forgot to update the title

bogner (approved):
The attribute approach seems sensible to me.

proposals/infra/INF-0003-packoffset.md
python3kgae: Fixed .

llvm-beanz:
This PR broke the GitHub pages build. The document contains a header to start a raw block that isn't terminated. You need to add this to the end of the document:

`<!-- {% endraw %} -->`

PR #208, opened 2024-04-19
Add Integer Literals chapter

llvm-beanz:
Adds a new chapter to describe the integer literal behavior as conforming to C. This documents the feature as proposed in:

https://github.com/microsoft/hlsl-specs/blob/main/proposals/0017-conforming-literals.md

### DIFF
--- a/specs/language/lex.tex
+++ b/specs/language/lex.tex
@@ -236,7 +236,83 @@ LL(2) parser.}.
   vector-literal
 \end{grammar}
 
-%\Sub{Integer Literals}{Lex.Literal.Int}
+\Sub{Integer Literals}{Lex.Literal.Int}
+
+\begin{grammar}
+  \define{integer-literal}\br
+  decimal-literal \opt{integer-suffix}\br
+  octal-literal \opt{integer-suffix}\br
+  hexadecimal-literal \opt{integer-suffix}\br
+
+  \define{decimal-literal}\br
+  nonzero-digit\br
+  decimal-literal digit\br
+
+  \define{octal-literal}
+  \terminal{0}\br
+  octal-literal octal-digit\br
+
+  \define{hexadecimal-literal}\br
+  \terminal{0x} hexadecimal-digit\br
+  \terminal{0X} hexadecimal-digit\br
+  hexadecimal-literal hexadecimal-digit\br
+
+  \define{nonzero-digit} \textnormal{one of}\br
+  \terminal{1 2 3 4 5 6 7 8 9}\br
+
+  \define{octal-digit} \textnormal{one of}\br
+  \terminal{0 1 2 3 4 5 6 7}\br
+
+  \define{hexadecimal-digit} \textnormal{one of}\br
+  \terminal{0 1 2 3 4 5 6 7 8 9}\br
+  \terminal{a b c d e f}\br
+  \terminal{A B C D E F}\br
+
+  \define{integer-suffix}\br
+  unsigned-suffix \opt{long-suffix}\br
+  long-suffix \opt{unsigned-suffix}\br
+
+  \define{unsigned-suffix} \textnormal{one of}\br
+  \terminal{u U}\br
+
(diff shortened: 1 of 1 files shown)


llvm-beanz:
Here's an image of the complicated part of the rendered text. This covers everything except the grammar bits that aren't too hard to read in the source:
Screenshot 2024-04-30 at 11 43 09 AM

specs/language/lex.tex
```diff
@@ -236,7 +236,83 @@
...
+    none & \texttt{int32\_t} & \texttt{int32\_t} \\
+         & \texttt{int64\_t} & \texttt{uint32\_t} \\
+         &         & \texttt{int64\_t} \\
+         &         & \texttt{uint64\_t} \\
+    \hline
+    \texttt{u} or \texttt{U} & \texttt{uint32\_t} & \texttt{uint32\_t} \\
+                             & \texttt{uint64\_t} & \texttt{uint64\_t} \\
+    \hline
+    \texttt{l} or \texttt{L} & \texttt{uint32\_t} & \texttt{uint32\_t} \\
+                             & \texttt{uint64\_t} & \texttt{uint64\_t} \\
```
bogner, line 305: Looks like you listed the `u` cases twice by mistake

bogner (approved):

llvm-beanz:
Updated rendering from the latest doc push:

Screenshot 2024-04-30 at 3 58 50 PM

farzonl:
im coming a bit late to this, is there a reason why we don't document short\int16_t  or ushort\uint16_t here?

farzonl (approved):

tex3d:
> There is no literal suffix for integers smaller than `int`, and this is just about the literal handling. `short` and `ushort` aren't valid keywords in HLSL, we do have `int16_t` and `uint16_t` which are documented in **[Basic.types]**.

Is there a precedent for a literal suffix for these types?  Can we adopt something?  Otherwise, when adapting code to comply with the conforming literals language changes in HLSL 202x, code like `i + 1` where `i` is `int16_t` would have to be adapted in an awkward way to `i + (int16_t)1`.

tex3d (approved):

llvm-beanz:
> Is there a precedent for a literal suffix for these types?
There is no precedent in C. This is probably because C basically implicitly converts all integer types smaller than `int` to `int` before all operations (yay Usual Arithmetic Conversions).

>Can we adopt something? Otherwise, when adapting code to comply with the conforming literals language changes in HLSL 202x, code like `i + 1` where `i` is `int16_t` would have to be adapted in an awkward way to `i + (int16_t)1`.

We _could_ come up with something ourselves, but I actually think it is better to not create something non-conformant here. Maybe it makes sense to allow literals to start as 16-bit types, and scale up because implicit promotions are cheap. I’m unconvinced we have a good enough reason to deviate from C/C++ here.

devshgraphicsprogramming:
> 

so why doesn't a C++ compiler give me a lot of warnings about uint32_t being casted to uint16_t when DXC does?

> I don't know that we're going to make you happy since the feedback you're giving us is self-inconsistent. It seems like you want us to align the language with how you want it to work rather than aligning based on C/C++ or existing usage.

You're right. Sorry my bad.

llvm-beanz:
> so why doesn't a C++ compiler give me a lot of warnings about uint32_t being casted to uint16_t when DXC does?

C/C++ has the most hated conversion feature ever [usual arithmetic conversions](https://timsong-cpp.github.io/cppwp/n4868/expr#arith.conv). Basically all integer values smaller than `int` are promoted to `int` or `unsigned int` implicitly following the [integer promotion rules](https://timsong-cpp.github.io/cppwp/n4868/conv.prom), and more rules identify the common type from there.

Because this means a _whole lot_ of extra conversions for shorts, most C/C++ compilers don't warn on conversions that are caused by the usual arithmetic conversions.

HLSL does not have usual arithmetic conversions because that would be _awful_ for performance on SIMD processors.

PR #209, opened 2024-04-19
[INF-0002] Add proposal for root signature handling in Clang

damyanp:
This change adds an initial proposal for root signature handling in Clang.

This change proposes adding:

* New AST nodes to represent the root signature
* A metadata representation of the root signature so it can be stored in LLVM IR
* Validation and diagnostic generation for root signatures during semantic
  analysis
* Conversion of the metadata representation to the binary serialized format.

### DIFF
--- a/proposals/infra/INF-0002-root-signature-in-clang.md
+++ b/proposals/infra/INF-0002-root-signature-in-clang.md
@@ -1,0 +1,371 @@
+<!-- {% raw %} -->
+
+# Implementation of Root Signatures in Clang
+
+* Proposal: [0002](INF-0002-root-signture-in-clang.md)
+* Author(s): [Xiang Li](https//github.com/python3kgae), [Damyan
+  Pepper](https://github.com/damyanp)
+* Sponsor: Damyan Pepper
+* Status: **Under Consideration**
+* Impacted Project(s): Clang
+
+<!--
+*During the review process, add the following fields as needed:*
+
+* PRs: [#NNNN](https://github.com/microsoft/DirectXShaderCompiler/pull/NNNN)
+* Issues:
+  [#NNNN](https://github.com/microsoft/DirectXShaderCompiler/issues/NNNN)
+  -->
+
+## Introduction
+
+[Root
+Signatures](https://learn.microsoft.com/en-us/windows/win32/direct3d12/root-signatures-overview)
+can be [specified in
+HLSL](https://learn.microsoft.com/en-us/windows/win32/direct3d12/specifying-root-signatures-in-hlsl)
+and included in the generated DXContainer in a binary serialized format.
+Support for this functionality needs to be added to Clang.
+
+This change proposes adding:
+
+* New AST nodes to represent the root signature
+* A metadata representation of the root signature so it can be stored in LLVM IR
+* Validation and diagnostic generation for root signatures during semantic
+  analysis
+* Conversion of the metadata representation to the binary serialized format.
+
+## Motivation
+
+### What are Root Signatures?
+In DirectX HLSL, resources can be associated with registers.  For example:
+
+``` c++
+StructuredBuffer<float4> b1 : register(t0);
+StructuredBuffer<float4> b2 : register(t1);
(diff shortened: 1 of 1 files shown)


proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,371 @@
...
+```
+
+At this point, final validation is performed to ensure that the root signature
+itself is valid. One key validation here is to check that each register is only
+bound once in the root signature. Validation is performed this late to avoid
+requiring every frontend implementation to perform the same validation.
+
+> **Open Question** - although it seems important to perform the above
+> validation here for IR that didn't come directly from clang, maybe there'd
+> still be value in performing the same validation in Sema? Maybe it's ok to
```
python3kgae, line 291: Checking that each register is bound only once in the root signature should be performed initially in Sema. This allows for early error reporting and ensures we have a correct AST. 

What cannot be done in Sema is checking that all used resources are bound into the root signature. This is because optimization may remove code related to resource usage

damyanp: I think it is important that it is checked here as well, so I think you're proposing we should perform the check in multiple places?  It'd be great if we could assume that the IR representation was correct, but I don't know if we can really do that.

python3kgae: Yes. Unless we can make sure all frontends generate legal root signature metadata, we'll have to validate it in DirectX backend.

damyanp:
Raw notes from a design review:

* DXR State Objects - should be called HLSL Subobjects
* The Decl is needed to support multiple types of root signature (local/global/attribute)
* We do need to do local/global root signatures here to really understand how it the decl works in.
* Idea: could we detect/optimize case if multiple entry points use the same root signature string? [optimization]
* Need details on error reporting and diagnostics - especially as this feeds into requirements for the data structures.
* Need details of how metadata binding to root signatures and entry points / subobjects work.
* Metadata that's a list of root signatures is useful? Maybe? Need to undertstand how it is used.
* Example rootsig should have multiple descriptor ranges to be a good example.
* Show how descriptor table's flags would be included.
* All diagnostics that can be done in sema should be done there.
* We can't re-use d3d code for this so we need to implement ourselves.

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,393 @@
...
+
+This change proposes adding:
+
+* New AST nodes to represent the root signature
+* A metadata representation of the root signature so it can be stored in LLVM IR
+* Validation and diagnostic generation for root signatures during semantic
+  analysis
+* Conversion of the metadata representation to the binary serialized format.
+
+## Motivation
```
coopp, line 37: This section is blank.  Should it be removed?

proposals/infra/INF-0002-root-signature-in-clang.md
```diff
@@ -0,0 +1,393 @@
...
+When parsed will produce a the equivalent of:
+
+```c++
+parsedRootSignature = RootSignature{
+  RootFlags(ALLOW_INPUT_ASSEMBLER_INPUT_LAYOUT),
+  RootCBV(0, 1), // register 0, space 1
+  StaticSampler(1, 0), // register 1, space 0
+  DescriptorTable({
+    SRV(0, 0, unbounded), // register 0, space 0, unbounded
+    UAV(5, 1, 10) // register 5, space 10, 10 descriptors
```
coopp, line 223: UAV(5, 1, 10) <--- The second parameter here is indicating space 1.  You mentioned space 10 in your comment.  Am I missing something?

coopp (approved):
Just a small comment in a code comment that might be incorrect. -- Looks good to me.

damyanp:
Remaining issues:

- [ ] [INF-0002] Fill out sections on Validations in Sema and Codegen
- [ ] [INF-0002] Fill out detailed metadata schema design

PR #213, opened 2024-04-22
Minor clarifications to Barrier for ALL_MEMORY and equivalent forms

tex3d:
- Added text clarifying ALL_MEMORY behavior.
- Added note clarifying that using the new barrier intrinsic with equivalent arguments is not translated to the legacy DXIL op for lower shader model targets.
- Minor formatting adjustments.

Part of https://github.com/microsoft/DirectXShaderCompiler/issues/6538.

### DIFF
--- a/proposals/0018-work-graphs.md
+++ b/proposals/0018-work-graphs.md
@@ -551,9 +551,11 @@ void GroupMemoryBarrierWithGroupSync() {
 }
 ```
 
-> Note: These equivalent forms are not currently mapped to the previous DXIL
-> Barrier operation on prior shader models.  The new barrier is only available
-> on SM 6.8 and above.
+> Note: The new barrier is only available on shader model 6.8 and above.
+> Although there are some cases where there is an equivalent DXIL operation
+> for prior shader models, there is not a complete mapping from one to the
+> other, which would only make it partially available. This design allows for a
+> consistent set of rules that can easily be validated early on in compilation.
 
 ### New Structure Attributes
 
--- a/proposals/0018-work-graphs.md
+++ b/proposals/0018-work-graphs.md
@@ -488,7 +488,8 @@ enum SEMANTIC_FLAG {
 
 /// @brief Request a barrier for a set of memory types and/or thread group
 /// execution sync.
-/// @param MemoryTypeFlags Flag bits as defined by MEMORY_TYPE_FLAG.
+/// @param MemoryTypeFlags Flag bits as defined by MEMORY_TYPE_FLAG. Specifying
+/// ALL_MEMORY means all valid memory types given the context.
 /// @param SemanticFlags Flag bits as defined by SEMANTIC_FLAG.
 ///
 /// `Barrier` must be called from thread group uniform control flow when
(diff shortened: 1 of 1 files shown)


proposals/0018-work-graphs.md
```diff
@@ -543,9 +549,12 @@ void GroupMemoryBarrier() { Barrier(GROUP_SHARED_MEMORY, GROUP_SCOPE); }
 void GroupMemoryBarrierWithGroupSync() {
   Barrier(GROUP_SHARED_MEMORY, GROUP_SCOPE | GROUP_SYNC);
 }
-
 ```
 
+> Note: These equivalent forms are not currently mapped to the previous DXIL
+> Barrier operation on prior shader models.  The new barrier is only available
+> on SM 6.8 and above, for now.
```
llvm-beanz, line 556: I don't entirely know that this note's clarification is really required. We didn't provide qualifying statements like this to any of the other language additions in this proposal.

At a minimum we should keep the spec to what we've committed to and separate out future plans in a separate spec. Putting wording like "for now" or "yet" into a spec implies to a reader that this will change, and this change isn't on our roadmap.

```suggestion
> on SM 6.8 and above.
```

proposals/0018-work-graphs.md
```diff
@@ -543,9 +549,12 @@ void GroupMemoryBarrier() { Barrier(GROUP_SHARED_MEMORY, GROUP_SCOPE); }
 void GroupMemoryBarrierWithGroupSync() {
   Barrier(GROUP_SHARED_MEMORY, GROUP_SCOPE | GROUP_SYNC);
 }
-
 ```
 
+> Note: These equivalent forms are not currently mapped to the previous DXIL
+> Barrier operation on prior shader models.  The new barrier is only available
+> on SM 6.8 and above.
```
llvm-beanz, line 556: I'm still struggling with this note. I don't think it adds clarity to the specification.

Is there a concern that the above text implied that these intrinsics would be available in older shader models that we can address by clarifying the wording elsewhere in the specification?

tex3d: There is a long-standing convention for mapping HLSL features to shader models where the compiler will map the feature if it's possible without significant complications or performance penalties.

We describe the new Barrier operations as a superset of the prior operations and list the equivalent forms to the previous HLSL intrinsics - "The pseudo-code below shows implementing the existing HLSL memory barrier functions using the new `Barrier` function.".

This describes the subset of parameter values that map directly to functionality in existing shader models.  This note is included to make it clear that even though the compiler *could* map these uses directly to existing operations on prior shader models, it does not do so.

As for alternative means of clarification, we could explicitly define something like the availability attributes on the Barrier declarations in this spec.  Is that preferable?  Even with that, one might ask *why* the equivalent forms can't be made to compile down to prior shader models, and I suppose our answer would be: because we decided not to implement that.  The note helps avoid that question, since it addresses it directly.

damyanp: Since it seems that there is some confusion over what the obvious thing to do here would be, I think some kind of a note is worthwhile.  How about:

> Note: The new barrier is only available on shader model 6.8 and above. Although there are some cases where there is an equivalent DXIL operation for prior shader models, there is not a complete mapping from one to the other, which would only make it partially available. This design allows for a consistent set of rules that can easily be validated early on in compilation.

damyanp (approved):
I agree that the "note" could be adjusted a bit so that it reinforces and clarifies the design decisions that have been made, but otherwise LGTM.

proposals/0018-work-graphs.md
tex3d: I've updated the note as suggested.  @llvm-beanz, please resolve if this is satisfactory.

llvm-beanz (approved):

PR #215, opened 2024-04-24
Cleanup C++ references in the language spec

llvm-beanz:
This just cleans up the C++ references in the document. It includes the following changes:

* Made the name in the glossaries human readable rather than the ISO/IEC number, and included the year.
* Changed two places where \glsdesc was called to \gls
* Extended the description text to have the ISO/IEC number and full standard title.
* Formalize general C/C++ references to C11 and C++11.

With this change references from `\gls{...}` will expand to a readable name (e.g. "ISO C Standard (2011)") rather than the ISO/IEC number.

This change also makes general references to C and C++ (the isoC and isoCPP glossary entries) point to C11 and C++11 respectively. We still have an explicit reference to C23 for integer value behaviors (C23 adopted two's compliement).

The intent of basing on C11 and C++11 is that HLSL 202x is unlikely to support most C++ features beyond C++98, but C11 and C++11 contained a lot of improvements and clarifications in wording. Also C++11 features are highly requested, so HLSL 202y will almost certainly gain C++11 (and maybe later version) features.

I did not want to base on newer C++ because the differences between HLSL and C++17+ are substantial, so the benefit of citing a base specification is limited in those cases.

We should continue to cite newer specifictions by version explicitly where it is helpful.

### DIFF
--- a/specs/language/basic.tex
+++ b/specs/language/basic.tex
@@ -77,7 +77,7 @@ minimum precision unsigned integer types are collectively called
 
 \p The signed integer types and unsigned integer types are collectively called
 \textit{integer types}. Integer types inherit the object representation of
-integers defined in \gls{isoC23}\footnote{C23 adopts two's compliment as the
+integers defined in \gls{isoC23}\footnote{C23 adopts two's complement as the
 object representation for integer types.}. Integer types shall satisfy the
 constraints defined in \gls{isoCPP}, section \textbf{basic.fundamental}.
 
--- a/specs/language/basic.tex
+++ b/specs/language/basic.tex
@@ -77,9 +77,9 @@ minimum precision unsigned integer types are collectively called
 
 \p The signed integer types and unsigned integer types are collectively called
 \textit{integer types}. Integer types inherit the object representation of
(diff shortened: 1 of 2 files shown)


specs/language/basic.tex
```diff
@@ -77,9 +77,9 @@
 
 \p The signed integer types and unsigned integer types are collectively called
 \textit{integer types}. Integer types inherit the object representation of
-integers defined in \glsdesc{isoC23}\footnote{C23 adopts two's compliment as the
+integers defined in \gls{isoC23}\footnote{C23 adopts two's compliment as the
```
bharadwajy, line 80:
```suggestion
integers defined in \gls{isoC23}\footnote{C23 adopts two's complement as the
```

bharadwajy (approved):

damyanp:
> This change also makes general references to C and C++ (the isoC and isoCPP glossar entries)

Should that be "glossary"?

damyanp (approved):

dmpots (approved):
Small typo in description *whwere*

> Changed two places whwere \glsdesc was called to \gls

PR #216, opened 2024-04-24
Semantic index assignment spec

bob80905:
A spec on how semantic and index pairs are assigned to entry point function parameters, how signature points are identified, and how diagnostics are emitted.


llvm-beanz:
@bob80905, I believe this proposal is better suited to wg-hlsl since this deals with implementation details not the language or shader model features.

bob80905:
> @bob80905, I believe this proposal is better suited to wg-hlsl since this deals with implementation details not the language or shader model features.

Moved over to:
https://github.com/llvm/wg-hlsl/pull/112

PR #217, opened 2024-04-25
[0006] Resource element type validation

bob80905:
This spec describes how the compiler will validate resource element types.


damyanp:
Can you put the proposal number in the PR title please?

proposals/0022-validating-resource-container-elements.md
```diff
@@ -0,0 +1,123 @@
+* Proposal: [0022](0022-validating-resource-container-elements.md)
+* Author(s): [Joshua Batista](https://github.com/bob80905)
+* Sponsor: TBD
+* Status: **Under Consideration**
+* Planned Version: Shader Model 6.8
```
damyanp, line 5: Is SM 6.8 the right planned version to put here?

proposals/0022-validating-resource-container-elements.md
```diff
@@ -0,0 +1,123 @@
...
+type, and determine whether all leaf element types are the set of acceptable primitives
+(i.e., int, uint, half, float, or double.) If, for example, the resource element type is
+an array of other resources, an error should be emitted.
+
+## Detailed design
+
+Initially, the function will start with the input resource element type. Due to the nature
+of the problem, the function can be recursively reapplied to any vector or matrix-like structures
+within the input type. Formally, the resource element type is valid if the function returns true
+on the given input type, or if the function returns true on all fields within the given input type.
```
damyanp, line 35: For me, I'd find this easier to follow if there were some examples.

proposals/0022-validating-resource-container-elements.md
```diff
@@ -0,0 +1,123 @@
...
+* Status: **Under Consideration**
+* Planned Version: Shader Model 6.8
+* PRs:
+* Issues:
+  [#75676](https://github.com/llvm/llvm-project/issues/75676)
+
+## Introduction
+
+There's a specific closed set of types that are valid as element types for an HLSL resource, 
+which includes int/uint of sizes 16/32/64, half, float, and double. If someone writes
```
damyanp, line 13: It sounds like arrays of these types are also valid?

proposals/0022-validating-resource-container-elements.md
```diff
@@ -0,0 +1,123 @@
...
+from DXC.
+
+### Interchange Format Additions
+
+* What DXIL changes does this change require?
+None
+* What Metadata changes does this require?
+None
+* How will SPIRV be supported?
+No need
```
damyanp, line 71: As these questions aren't really relevant, I wonder if this should really be using the infra template?

damyanp (changes requested):
I think this might work better if it were treated as an infra proposal rather than a language / shader model proposal.

The problem statement in the introduction doesn't seem to match the solution design. It sounds like only the int/uint/floats of 16/32/64 sizes are supported, but then it seems that there's something about arrays and structs happening as well.  Some examples, along we proposed diagnostic outputs, might help me understand what the goal is here.

proposals/0022-validating-resource-container-elements.md
tex3d, line 1: I think this was based on the [shader model feature template](https://github.com/microsoft/hlsl-specs/blob/main/proposals/templates/shader-model-feature-template.md) instead of the [infrastructure template](https://github.com/microsoft/hlsl-specs/blob/main/proposals/templates/infrastructure-template.md), which is what I think it should have been based on.

proposals/0022-validating-resource-container-elements.md
```diff
@@ -2,30 +2,41 @@
...
 
 ## Introduction
 
-There's a specific closed set of types that are valid as element types for an HLSL resource, 
-which includes int/uint of sizes 16/32/64, half, float, and double. If someone writes 
-`RWBuffer<MyCustomType>` we should reject it and give the user a nice diagnostic. 
+There's a specific closed set of types that are valid as element types for an HLSL
+resource, which includes int/uint of sizes 16/32/64, half, float, and double. Arrays of
+these types are not valid. If someone writes `RWBuffer<MyCustomType>` we should reject it
+and give the user a nice diagnostic.
```
damyanp, line 17: There's a key bit of information omitted from here which is that it sounds like you want to support user defined types that are equivalent to one of the supported types.

In the examples you have:

```hlsl
struct x {
	int i;
};

struct bad {
	x a;
	bool b;	
};

RWBuffer<x> valid;
RWBuffer<bad> invalid;
```

But...is it because of the `bool` that is is invalid?  How about these examples:

```hlsl
struct a {
   int aa;
   int ab;
};

struct b {
   x bx;  // (same x as in example above)
};

struct c {
  a ca;
  float4 cb;
};
```

bob80905: Sure, I can add some extra clarification, but if I understand correctly, I believe x should be allowed, bad should not be allowed because of the bool, a should be allowed, as well as b and c.

proposals/0022-validating-resource-container-elements.md
```diff
@@ -58,22 +58,42 @@ field is saved, and the source location is used in the diagnostic output.
...
+   int ab;
+};
+
+struct b {
+   x bx;
+};
+
+struct c {
+  a ca;
+  float4 cb;
```
damyanp, line 76: Interesting.  I updated by comment because I put `float4` in there, and realized that `vector<float, 4>` probably isn't a valid type here. Is it?

proposals/0022-validating-resource-container-elements.md
damyanp: The main thing I'm looking for here is an explanation in the introduction or motivation section on what the rule is actually meant to be.  From the description of that rule I should be able to determine which of those examples are correct.  The examples can be included to help people understand what the rule is, but I don't see anything in the current text saying what the rule is, beyond:

> There's a specific closed set of types that are valid as element types for an HLSL resource, which includes int/uint of sizes 16/32/64, half, float, and double. Arrays of these types are not valid.

From this description none of the struct examples are valid.


Then you can discuss the proposed solution for validating this rule.

llvm-beanz (commented):
In DXC we generate somewhat odd errors for some of these cases (see: https://godbolt.org/z/jnTY1PM9Y), but they're all very custom special case.

Can we implement these errors in terms of type traits? If so, can we find a way to make these errors not special cases in the compiler?

A C++ example of the kind of thing I'm thinking of would be something like this:

```c++
#include <type_traits>

template<typename T>
class RWBuffer {
    static_assert(std::is_standard_layout<T>::value, "RWBuffer requires...");
};

struct GoodStruct {
    int X[5];
};

struct BadStruct {
    int &X;
};

RWBuffer<int> I; // Fine
RWBuffer<GoodStruct> G; // Fine
RWBuffer<BadStruct> B; // Error!
```
[Compiler Explorer](https://godbolt.org/z/foh5sxGv4)

With that in mind, what are the different traits we need to be able to specify as requirements for the template parameters of resource types?

llvm-beanz:
Because this is stuck in my head...

I don't know the full set of requirements that we would need for RWBuffer, but some examples would be:
```c++
template<typename T>
class RWBuffer {
    // T must be a complete type.
    // this builtin exists in Clang for C++ and should work for us.
    // DXC produces this error: https://godbolt.org/z/bWonbjxTz.
    static_assert(__is_complete_type(T),
                  "type parameter must be complete type."); 
    
    // I coined this term in the language spec to refer to handle types like
    // resources where we don't know their size
    // (see: https://github.com/microsoft/hlsl-specs/pull/198).
    // Not implemented in Clang.
    // DXC produces this error: https://godbolt.org/z/x37K1dE1e
    static_assert(!__is_intangible_type(T),
                  "type parameter must not be an intangible type.");
    
    // This is from Damyan's example, where we require that all fields of the
    // flattened type be the same type.
    // Not implemented in Clang.
    // DXC produces this error: https://godbolt.org/z/GGjsbjWT3
    static_assert(__is_homogenious_aggregate(T),
                  "type parameter must be an object where all fields are the same type.");
};
```

bob80905:
@beanz This is extremely useful feedback! I do think I could definitely take advantage of those builtins you mentioned, it would greatly simplify the implementation of the diagnostic function that determines whether or not the resource element type is valid (we can call it DiagnoseResourceElementType for now).
However, just asserting doesn't seem to capture all the information I would like. I would definitely want to point to the specific element in a struct that is non-homogenous in the diagnostic, for example. 
There are probably other rules that are missing like you mentioned, so I will need to read up on how DXC defines valid resource element types. But those builtins you mentioned look like a great start.

llvm-beanz:
> I do think I could definitely take advantage of those builtins you mentioned, it would greatly simplify the implementation of the diagnostic function that determines whether or not the resource element type is valid (we can call it DiagnoseResourceElementType for now).

I think we should try not to have a custom function in Clang to do this diagnostic. Someday it would be nice to move RWBuffer out of the ExternalSemaSource into a header, and we can't do that if we rely on custom baked rules to validate the type.

> However, just asserting doesn't seem to capture all the information I would like. I would definitely want to point to the specific element in a struct that is non-homogenous in the diagnostic, for example.

While that may be nice, we don't do that in DXC today, and I'm not sure that is worth the extra complexity. We should aim for a simple solution that is more general purpose if we can.

> There are probably other rules that are missing like you mentioned, so I will need to read up on how DXC defines valid resource element types. But those builtins you mentioned look like a great start.
👍

bharadwajy:
I agree with @llvm-beanz that type validation done using C++ 20 traits (and concepts, if need be) is the way to go with a goal to move the functionality to a header eventually.

bob80905:
> > However, just asserting doesn't seem to capture all the information I would like. I would definitely want to point to the specific element in a struct that is non-homogenous in the diagnostic, for example.
> 
> While that may be nice, we don't do that in DXC today, and I'm not sure that is worth the extra complexity. We should aim for a simple solution that is more general purpose if we can.

I should note, in one of your examples you had:
```
struct differentbad {
	RWBuffer<int> B;
};

RWBuffer<differentbad> invalid2;
```
Which emits a very informative diagnostic:

error: 'RWBuffer<int>' is an object and cannot be used as a type parameter
RWBuffer<differentbad> invalid2;
         ^
note: usage of 'RWBuffer<int>' found in field 'B' of type 'differentbad'

I mentioned before that this doesn't seem possible with the type_trait approach, so this design would be sacrificing some diagnostic specificity.

llvm-beanz:
> Which emits a very informative diagnostic:
> 
> error: 'RWBuffer' is an object and cannot be used as a type parameter RWBuffer invalid2; ^ note: usage of 'RWBuffer' found in field 'B' of type 'differentbad'
> 
> I mentioned before that this doesn't seem possible with the type_trait approach, so this design would be sacrificing some diagnostic specificity.

I would argue that diagnostic makes no sense. “RWBuffer” is an object? so is every struct. Why is that object different from other objects?

The real reason `differentbad` can’t be used as a parameter for RWBuffer is because it contains a resource handle, and resource handles can’t be read from resources because they don’t have an object representation that can be stored.

There are some limitations to the quality of concept error messages at the moment, but fixing that problem seems like a better approach than building a custom error reporting system that needs to be hard-coded in the compiler.

proposals/0022-validating-resource-container-elements.md
```diff
@@ -0,0 +1,103 @@
+* Proposal: [0022](0022-validating-resource-container-elements.md)
```
llvm-beanz, line 1: This should be an infrastructure proposal rather than a feature proposal.

proposals/0022-validating-resource-container-elements.md
bob80905: It shouldn't be valid because c is not a homogenous type, having both float4 and ints inside struct a.

damyanp (changes requested):
From just reading the spec it isn't clear what the plan is for how these new type traits will be used.  Chris's comments suggest using `static_assert`.  Is that the plan?  So the errors will be static_assert style errors?  Or something else?

farzonl:
> C++ 20

Most of these type traits were added as part of c++11 [is_standard_layout](https://en.cppreference.com/w/cpp/types/is_standard_layout) or exist in clang proper [__is_complete_type](https://github.com/llvm/llvm-project/blob/b9702bb12f19b7bdb7f4420f94ee603d29d1bf1b/clang/include/clang/Basic/TokenKinds.def#L555) 

Do we need to use `concept` to achieve the goals of this proposal? My concern is that LLVM is still on c++17 and the most recent changes to `concepts` will land in clang 19 which is not what most folks are building with https://clang.llvm.org/cxx_status.html

bob80905:
> > C++ 20
> 
> Most of these type traits were added as part of c++11 [is_standard_layout](https://en.cppreference.com/w/cpp/types/is_standard_layout) or exist in clang proper [__is_complete_type](https://github.com/llvm/llvm-project/blob/b9702bb12f19b7bdb7f4420f94ee603d29d1bf1b/clang/include/clang/Basic/TokenKinds.def#L555)
> 
> Do we need to use `concepts` to achieve the goals of this proposal? My concern is that LLVM is still on c++17 and the most recent changes to `concepts` will land in clang 19 which is not what most folks are building with https://clang.llvm.org/cxx_status.html

I don't think using `concepts` are absolutely necessary, and using the example Chris gave, it seems this can be implemented without using `concepts` at all.

bob80905:
> From just reading the spec it isn't clear what the plan is for how these new type traits will be used. Chris's comments suggest using `static_assert`. Is that the plan? So the errors will be static_assert style errors? Or something else?

Yes, I'll make a note in the spec that static asserts will be used on these type traits to validate the types. I think the plan will indeed be to use static asserts, and static-assert-style errors will be emitted.

damyanp:
This isn't being actively worked on at the moment, and needs to be retargeted to wg-hlsl - so closing it.

PR #218, opened 2024-04-27
Add Basic.Decl

llvm-beanz:
This specifies more of the basic language behaviors including declarations and definitions.

HLSL has some small implicit differences from C & C++ for declarations and definitions. Notably the handling of global variable declarations, which are implcitly const and external unless `static` is specified.

### DIFF
--- a/specs/language/basic.tex
+++ b/specs/language/basic.tex
@@ -77,7 +77,6 @@ external in HLSL.},
 \item it is an \textit{alias-declaration} (\ref{Decl}),
 \item it is a \textit{using-declaration} (\ref{Decl}),
 \item it is a \textit{static\_assert-declaration} (\ref{Decl}),
-\item it is an \textit{attribute-declaration} (\ref{Decl}),
 \item it is an \textit{empty-declaration} (\ref{Decl}),
 \item or a \textit{using-directive} (\ref{Decl}).
 \end{itemize}
--- a/specs/language/basic.tex
+++ b/specs/language/basic.tex
@@ -1,5 +1,116 @@
 \Ch{Basic Concepts}{Basic}
 
+\begin{note}
+  \p HLSL inherits a great deal if its behavior from C and C++. Some of that was
+  intentional by aligning syntax early in the development of the language, some
+  of it was a side effect of DXC being implemented on top of Clang.
+
+  \p This chapter includes a lot of definitions that are inherited from C and C++.
+  Some are identical to C or C++, others are slightly different. HLSL is neither
+  a subset nor a superset of C or C++, and cannot be simply described in terms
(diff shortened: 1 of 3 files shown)


specs/language/basic.tex
```diff
@@ -1,5 +1,122 @@
...
+\item it is a \textit{static\_assert-declaration} (\ref{Decl}),
+\item it is an \textit{attribute-declaration} (\ref{Decl}),
+\item it is an \textit{empty-declaration} (\ref{Decl}),
+\item or a \textit{using-directive} (\ref{Decl}).
+\end{itemize}
+
+\p The two examples below is adapted from \gls{isoCPP} \textbf{[basic.def]}. All
+but one of the following are definitions:
+\begin{HLSL}
+int f(int x) { return x+a; } // defines f and x
```
coopp, line 88: nit: The implementation for f( ) is x + a here... But a isn't called out as being defined.  Did you mean to put x + 1 here instead?

specs/language/basic.tex
```diff
@@ -1,5 +1,122 @@
...
+static int i;                // defines N::i
+}
+\end{HLSL}
+
+\p All of the following are declarations:
+\begin{HLSL}
+int a;                       // declares a
+const int c;                 // declares c
+X anX;                       // declares anX
+int f(int);                  // declares g
```
coopp, line 107: Where is 'g' declared?  I am missing something?

coopp (approved):
Just a few small things, but it looks good to me.

specs/language/basic.tex
llvm-beanz: doh!

bogner (approved):

specs/language/basic.tex
```diff
@@ -1,5 +1,121 @@
...
+\begin{HLSL}
+int a;                       // declares a
+const int c;                 // declares c
+X anX;                       // declares anX
+int f(int);                  // declares f
+struct S;                    // declares S
+typedef int Int;             // declares Int
+using N::d;                  // declares d
+using Float = float;         // declares Float
+cbuffer CB {                 // does not declare CB
```
damyanp, line 111: Does `CB` have any significance after it appears in `cbuffer CB`?  It's not declared....so what is it?

damyanp (approved):

specs/language/basic.tex
llvm-beanz: Yea... we need to write up an explanation of that elsewhere in the spec. It isn't a declared name in the HLSL translation unit, but all the declarations under it get grouped in reflection. It behaves a lot more like a grouping attribute than a declaration.

llvm-beanz: Resolving conversation based on my comment above.

PR #240, opened 2024-05-16
A bit of wordsmithing

bharadwajy:
Started to just fix the typo in "HLSL inherits a great deal if its behavior from C and C++" - but would like to suggest additional edits.

### DIFF
--- a/specs/language/basic.tex
+++ b/specs/language/basic.tex
@@ -1,15 +1,14 @@
 \Ch{Basic Concepts}{Basic}
 
 \begin{note}
-  \p HLSL inherits a great deal if its behavior from C and C++. Some of that was
-  intentional by aligning syntax early in the development of the language, some
-  of it was a side effect of DXC being implemented on top of Clang.
+  \p A significant portion of HLSL semantics is inherited from C and C++. Some
+  of this is a result of intentional adoption of syntax early in the development
+  of the language and some a side-effect of Clang-based implementation of DXC.
 
   \p This chapter includes a lot of definitions that are inherited from C and C++.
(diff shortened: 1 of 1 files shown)


specs/language/basic.tex
```diff
@@ -1,15 +1,14 @@
 \Ch{Basic Concepts}{Basic}
 
 \begin{note}
-  \p HLSL inherits a great deal if its behavior from C and C++. Some of that was
-  intentional by aligning syntax early in the development of the language, some
-  of it was a side effect of DXC being implemented on top of Clang.
+  \p A significant portion of HLSL semantics is inherited from C and C++. Some
```
llvm-beanz, line 4: “is inherited” or “are inherited” is passive voice, which is generally good to avoid.

Maybe:

> HLSL inherits a significant portion of its language semantics from C and C++.

specs/language/basic.tex
```diff
@@ -1,15 +1,14 @@
...
+  of this is a result of intentional adoption of syntax early in the development
+  of the language and some a side-effect of Clang-based implementation of DXC.
```
llvm-beanz, line 6:
```suggestion
  of the language and some a side-effect of the Clang-based implementation of DXC.
```

specs/language/basic.tex
bharadwajy: > “is inherited” or “are inherited” is passive voice, which is generally good to avoid.

Agreed.

> Maybe:
> 
> > HLSL inherits a significant portion of its language semantics from C and C++.

Modified as suggested.

llvm-beanz (approved):

bob80905 (approved):

PR #244, opened 2024-05-21
[Overload.Res] Add overload resolution sections

llvm-beanz:
This starts to flesh out more complete overload resolution rules. A lot of this is extremely similar to C++ however we have some key differences. Those differences include:

* HLSL supports overload resolution in 5 contexts, whereas C++ supports it in 7 (DXC actually only supports 2 of those contexts).
* HLSL has two additional conversion sequence ranks.
* HLSL has more rules to disambiguate some conversions.


bogner (approved):

hekota (approved):
LGTM!

PR #249, opened 2024-05-29
Add Linkage section

hekota:
Create chapter file for Declarations and add section about linkage. Enumerate the language defaults and where it is implementation dependent. Add empty chapter file for Statements to preserve the chapter ordering.

### DIFF
--- a/specs/language/declarations.tex
+++ b/specs/language/declarations.tex
@@ -14,31 +14,33 @@
 
 \p The following entities in HLSL have \textit{external linkage}:
 \begin{itemize}
-    \item entry point functions
-    \item functions marked with \texttt{export} keyword
-    \item global variables that are not marked \texttt{static} or \texttt{groupshared} \footnote{These are not really linked with other translation units but rather their values are loaded indirectly based on cbuffer mapping.}
-    \item enumerations
-    \item classes or template classes, their member functions, static data members, nested classes and enumerations
+  \item entry point functions
+  \item functions marked with \texttt{export} keyword
+  \item global variables that are not marked \texttt{static} or \texttt{groupshared} \footnote{These are not really linked with other translation units but rather their values are loaded indirectly based on cbuffer mapping.}
+  \item static data members of classes or template classes
 \end{itemize}
 
 \p Linkage of functions (including template functions) that are not entry points or marked with \texttt{export} keyword is implementation dependent. \footnote{In DXC today functions that are not entry points or exported have \textit{internal linkage} by default. This can be overriden by \texttt{-default-linkage} compiler option.}
-    
+
 \Sub{Internal Linkage}{Decl.Linkage.Internal}
 
 \p Entities with \textit{internal linkage} can be referred to from all scopes in the current translation unit.
 
 \p The following entities in HLSL have \textit{internal linkage}:
 \begin{itemize}
-    \item global variables marked as \texttt{static} or \texttt{groupshared}
+  \item global variables marked as \texttt{static} or \texttt{groupshared}
+  \item all entities declared in an unnamed namespace or a namespace within an unnamed namespace
+  \item enumerations
+  \item classes or template classes, their member functions, and nested classes and enumerations
 \end{itemize}
 
 \Sub{No Linkage}{Decl.Linkage.NoLinkage}
 
 \p An entity with \textit{no linkage} can be referred to only from the scope it is in.
 
-\p Any of the following entites declared at block scope have no linkage:
+\p Any of the following entites declared at function scope or block scopes derived from function scope have no linkage:
 \begin{itemize}
   \item local variables
   \item local classes and their member functions
-  \item other names declared at block scope such as typedefs, enumerations, and enumerators.
+  \item other entities declared at function scope or block scopes derived from function scope that such as typedefs, enumerations, and enumerators
 \end{itemize}
--- a/specs/language/declarations.tex
+++ b/specs/language/declarations.tex
@@ -1,0 +1,8 @@
+\Ch{Declarations}{Decl}
+\Sec{Function Definitions}{Decl.Function}
+\Sub{Visibility and Linkage}{Decl.Function.Visibility}
+
+\p Functions have \textit{internal linkage} by default.
+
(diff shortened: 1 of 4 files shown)


specs/language/placeholders.tex
```diff
@@ -4,12 +4,6 @@
 % forward references can be inserted into the specification as it is being
 % written to force updating the references as they change.
 
-\Ch{Statements}{Stmt}
-\Sec{Label Statements}{Stmt.Label}
```
damyanp, line 8: Wondering why Statements was pulled out into its own file, but Classes wasn't?  I can see why Declarations was because this change adds some real content to that.

damyanp (approved):

specs/language/placeholders.tex
llvm-beanz: I think @hekota did that just to preserve the chapter ordering, which seems fine to me.

damyanp: Gotcha, that makes sense.

specs/language/declarations.tex
```diff
@@ -0,0 +1,44 @@
...
+
+\p Entities with \textit{external linkage} can be referred to from the scopes in the other translation units and enable linking between them.
+
+\p The following entities in HLSL have \textit{external linkage}:
+\begin{itemize}
+    \item entry point functions
+    \item functions marked with \texttt{export} keyword
+    \item global variables that are not marked \texttt{static} or \texttt{groupshared} \footnote{These are not really linked with other translation units but rather their values are loaded indirectly based on cbuffer mapping.}
+    \item enumerations
+    \item classes or template classes, their member functions, static data members, nested classes and enumerations
```
llvm-beanz, line 21: Do enumerations and classes have external linkage in HLSL?

I think they likely have _internal linkage_ and we duplicate their definitions and use the One Definition Rule to resolve them to a single instance.

I think a static data member of a class is the only thing here that would truly have _external linkage_.

specs/language/declarations.tex
```diff
@@ -0,0 +1,44 @@
...
+
+\p Linkage of functions (including template functions) that are not entry points or marked with \texttt{export} keyword is implementation dependent. \footnote{In DXC today functions that are not entry points or exported have \textit{internal linkage} by default. This can be overriden by \texttt{-default-linkage} compiler option.}
+    
+\Sub{Internal Linkage}{Decl.Linkage.Internal}
+
+\p Entities with \textit{internal linkage} can be referred to from all scopes in the current translation unit.
+
+\p The following entities in HLSL have \textit{internal linkage}:
+\begin{itemize}
+    \item global variables marked as \texttt{static} or \texttt{groupshared}
```
llvm-beanz, line 32: Unnamed namespaces and members of unnamed namespaces also are given internal linkage.

llvm-beanz (commented):
In the cases of _external_ and _internal_ linkage there's also a property where a declaration inherits the linkage of its encompassing declaration. For example:

```
namespace PublicStuff {
  void Fn(int); // external linkage
  
  class Obj {
    void MemberFn(); // external linkage
  }; // external linkage
}

namespace {
  void Fn(int); // internal linkage
  
  class Obj {
    void MemberFn(); // internal linkage
  }; // internal linkage
}
```

This isn't true for all declarations. For example declarations in function-scope and scopes that derive from function-scope are always no-linkage unless they are `static`.

specs/language/declarations.tex
hekota: DXC does not support unnamed namespaces. Should we support them in Clang?

hekota:
> In the cases of _external_ and _internal_ linkage there's also a property where a declaration inherits the linkage of its encompassing declaration. For example:
> 
> ```
> namespace PublicStuff {
>   void Fn(int); // external linkage
>   
>   class Obj {
>     void MemberFn(); // external linkage
>   }; // external linkage
> }
> 
> namespace {
>   void Fn(int); // internal linkage
>   
>   class Obj {
>     void MemberFn(); // internal linkage
>   }; // internal linkage
> }
> ```

DXC currently does not support unnamed namespaces, that's why I did not include it here.

> This isn't true for all declarations. For example declarations in function-scope and scopes that derive from function-scope are always no-linkage unless they are `static`.

I believe this is described in the No Linkage section under the term block-scope. I can add more details to that section to make it clearer.

specs/language/declarations.tex
llvm-beanz: That's an interesting question... It hadn't occurred to me that DXC didn't support them. They work in Clang today because we haven't done anything to actively disable them. My inclination is that unless there is a reason not to support them, we should just plan them as a 202x or 202y feature. I would lean toward 202x, just because I think it is trivial to enable this in DXC:

```diff
diff --git a/tools/clang/lib/Parse/ParseDeclCXX.cpp b/tools/clang/lib/Parse/ParseDeclCXX.cpp
index de46dfdb43c..be072a0715d 100644
--- a/tools/clang/lib/Parse/ParseDeclCXX.cpp
+++ b/tools/clang/lib/Parse/ParseDeclCXX.cpp
@@ -94,7 +94,8 @@ Decl *Parser::ParseNamespace(unsigned Context,
     }
   }
   // HLSL Change Starts
-  else if (getLangOpts().HLSL) {
+  else if (getLangOpts().HLSL &&
+           getLangOpts().HLSLVersion < hlsl::LangStd::v202x) {
     Diag(Tok, diag::err_expected) << tok::identifier;
   }
   // HLSL Change Ends
```

Curious if @Tex3d has any historical context for why anonymous namespaces were disabled other than matching FXC.

hekota: I've updated the spec to mention that entities in unnamed namespaces have internal linkage.

llvm-beanz (approved):
LGTM. I think we may need to tweak some of this in the future but I think this is pretty good.

specs/language/declarations.tex
tex3d: I believe the original context was just to match FXC support for HLSL.  FXC can't parse an anonymous namespace, so we disabled it originally in DXC.  We never prioritized re-enabling it, so it's still that way.

PR #252, opened 2024-06-10
Add default inlining spec

pow2clk:
A specification to define the current behavior of HLSL default function inlining. This includes how it is depended on and how it can and cannot be affected by compiler attributes and flags.

### DIFF
--- a/proposals/infra/INF-NNNN-default-function-inlining.md
+++ b/proposals/infra/INF-NNNN-default-function-inlining.md
@@ -6,7 +6,7 @@
 * Author(s): [Greg Roth](https://github.com/pow2clk)
 * Sponsor: [Greg Roth](https://github.com/pow2clk)
 * Status: **Under Consideration**
-* Impacted Project(s): DXC, DXC-Clang
+* Impacted Project(s): DXC, Clang
 
 ## Introduction
 
@@ -41,6 +41,8 @@ As parameters, that mapping can't be resolved.
 All functions that are not entry functions will be inlined at the call site.
 During code generation, all non-entry functions' linkage is set to external
 and the `AlwaysInline` attribute is added for called functions.
+During codege, `stacksave` and `stackrestore` instructions should be ommited
+as they won't be needed once the calls are inlined.
 
 The `always-inline` pass then inlines these functions.
 It is called early in the process to accommodate the dependent passes.
@@ -52,8 +54,7 @@ If lifetime markers are enabled, the inlined function will add them for the
 static allocas,
 scoping them to the inlined function.
 
-After inlining, the HLPreprocess step erases stacksave and stackrestore
-intrinsics along with any users thereof and moves all allocas in non-entry
+After inlining, the HLPreprocess step moves all allocas in non-entry
 functions into the entry function entry block.
 
 This leaves functions that have no runtime interface without any callers.
--- a/proposals/infra/INF-NNNN-default-function-inlining.md
+++ b/proposals/infra/INF-NNNN-default-function-inlining.md
@@ -1,0 +1,97 @@
+<!-- {% raw %} -->
+
+# Default Function Inlining
+
+* Proposal: [INF-NNNN](INF-NNNN-default-function-inlining.md)
+* Author(s): [Greg Roth](https://github.com/pow2clk)
+* Sponsor: [Greg Roth](https://github.com/pow2clk)
+* Status: **Under Consideration**
+* Impacted Project(s): DXC, DXC-Clang
+
+## Introduction
+
+By default all functions called within an HLSL shader are inlined
+at each call site.
+This allows for additional optimizations, legalization, and flexibility.
+
+## Motivation
+
+Inlining called functions avoids having to fully define a calling convention
+outside of entry point functions whose interface concerns how they interact
+with the runtime rather than a calling convention internal to HLSL.
+
+This allows optimizations specific to the call site.
+For example, where a function parameter takes a constant value,
+constant folding could precalculate values available at compile time that were
+calculated in the function using those parameters.
+Additionally, dead code that results from impossible control flow branches could
+be eliminated.
+
+It allows the elimination of redundant copying of parameters.
+Inlined functions lose their parameters, exposing the copies into and out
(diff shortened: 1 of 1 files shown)


proposals/infra/INF-NNNN-default-function-inlining.md
```diff
@@ -0,0 +1,97 @@
...
+that there is no parameter passing.
+It does need to come after parameters are legalized.
+
+If lifetime markers are enabled, the inlined function will add them for the
+static allocas,
+scoping them to the inlined function.
+
+After inlining, the HLPreprocess step erases stacksave and stackrestore
+intrinsics along with any users thereof and moves all allocas in non-entry
+functions into the entry function entry block.
```
llvm-beanz, line 57: Why do we need this? We should just not generate these intrinsics for HLSL since they won't be supported. I think the only place in clang CodeGen where we could emit these is `CodeGenFunction::EmitAutoVarAlloca`, and OpenMP has special handling to avoid emitting them, we should evaluate something similar for HLSL.

proposals/infra/INF-NNNN-default-function-inlining.md
```diff
@@ -0,0 +1,97 @@
+<!-- {% raw %} -->
+
+# Default Function Inlining
+
+* Proposal: [INF-NNNN](INF-NNNN-default-function-inlining.md)
+* Author(s): [Greg Roth](https://github.com/pow2clk)
+* Sponsor: [Greg Roth](https://github.com/pow2clk)
+* Status: **Under Consideration**
+* Impacted Project(s): DXC, DXC-Clang
```
damyanp, line 9: I think we just say "Clang" for the Clang project, unless "DXC-Clang" is referring to something more specific?

damyanp: Is this proposing changes to DXC as well?

pow2clk: I'm not proposing changes to DXC, though this largely documents how things are in DXC without much consideration of changes in clang just yet. I would prefer a more specific term than "clang" to refer to the modernized compiler, but I won't fight that battle now.

proposals/infra/INF-NNNN-default-function-inlining.md
pow2clk: That seems reasonable. This is how it was done in DXC and that's what I was going off of.

pow2clk: I can see how the stacksave and stackrestore instructions can be simply omitted this way. The allocas in inlined functions would still have to be moved though unless @llvm-beanz has another proposal of where to do that?

llvm-beanz: @pow2clk, why do we need to do that? The DXIL spec doesn’t specify that as a requirement. Even if it did, that would be a DXIL target requirement that we should do in the DXIL backend.

pow2clk: In DXC, the inlined functions get removed pretty early, so any information they have would be lost. I think that's the default behavior of the always inline pass. Even if we don't run that pass as early as we have historically, I would think it would still be before the DXIL backend

llvm-beanz: But why do we need to do anything to the allocas?

pow2clk: I'm mainly documenting what DXC did. I'd have to dig deeper to come up with the exact reasons why. In the meantime, I'd like to get this in. If I remove this for now and add a PR to address the whole HLPreprocess pass of it all, would you approve the rest of the spec as it is?

llvm-beanz: The solution we propose for Clang shouldn't be documentation of what DXC does. We should be writing a spec for what we need to do in Clang. For the "Proposed solution" you probably don't need this level of detail.

What I'm most interested is less the nuts-and-bolts of how it is implemented and rather what it should be doing.

Do we mark functions as always_inline or call instructions? Are there cases where we do one or the other? When will the inlining pass run? Are there other future considerations we need to think about (i.e. noinline).

I'm completely in favor of merging in-progress specs, but the best time to review something is as it is being proposed. So if you don't want me to review implementation details, remove them from the document.

proposals/infra/INF-NNNN-default-function-inlining.md
```diff
@@ -0,0 +1,98 @@
...
+
+It allows code that is only legalizable because of inlining.
+This includes called functions that take resourced parameters that need
+to map to unique, non-static global resources.
+As parameters, that mapping can't be resolved.
+
+## Proposed solution
+
+All functions that are not entry functions will be inlined at the call site.
+During code generation, all non-entry functions' linkage is set to external
```
tex3d, line 42: Why is all non-entry function's linkage set to external?

proposals/infra/INF-NNNN-default-function-inlining.md
pow2clk: I started with documenting DXC because that's what the facts on the ground are. I'm not familiar enough with modern LLVM to be a good source for how that behavior ought to change. I thought we could modify it as appropriate from there as we already have. 

I took infrastructure specs to consist of nuts and bolts. Explaining what it should do at a high level sounds like a language spec and a rather short one at that. 

I welcome the review you gave. I just don't have a good answer for it now and I was interested in other input you have on the spec, if there's nothing else, then I think it makes sense to move on. If this is just the first of a series of comments, then we might as well resolve it now so we can move onto the next one.

llvm-beanz: It seems to me like this specification has more questions than answers and I don’t really have a clear idea what you’re proposing. I see two concrete things in this document:

1) Apply the always_inline attribute to function declarations.
2) Documenting DXC’s pass structures and behaviors without explanation of why they matter.

I don’t think the later is necessary as I don’t see how it helps implement what we need to build, and if the only change we’re proposing is the former, do we need a spec to propose putting an attribute on function declarations?

PR #268, opened 2024-06-26
Add `export` keyword to HLSL language spec

hekota:
Add `export` keyword to HLSL language specs.

The spec describes two ways the `export` keyword can be used.

1. On individual function declarations
```
export void f() {}
```
2. On a group of function declaration:
```
export {
   void f1();
   void f2() {}
}
```

DXC currently supports only the first case, but since Clang has support for both, we might as well support it in HLSL 202y too. 

This spec update does not yet include detailed rules for when a function can or cannot be exported, such as when it has resource argument or semantic annotations. That will be covered by llvm/llvm-project#93330.

This change also adds more sections under Declarations (based on C++ spec layout).

Contributes to: llvm/llvm-project#92812

### DIFF
--- a/specs/language/declarations.tex
+++ b/specs/language/declarations.tex
@@ -1,13 +1,30 @@
 \Ch{Declarations}{Decl}
 \Sec{Preamble}{Decl.Pre}
+\p Declarations generally specify how names are to be interpreted. Declarations have the form
+\begin{grammar}
+  \define{declaration-seq}\br
+  \textit{declaration}\br
+  \textit{declaration-seq declaration}
+
+  \define{declaration}\br
+  \textit{name-declaration}\br
+  \textit{special-declaration}
+  
+  \define{name-declaration}\br
+  ...
+
+  \define{special-declaration}\br
+  \textit{export-declaration-group}\br
+  ...
+\end{grammar}
+
 \Sec{Specifiers}{Decl.Spec}
 \Sub{General}{Decl.Spec.General}
 \p The specifiers that can be used in a declaration are
 \begin{grammar}
   \define{decl-specifier}\br
-  \texttt{...}\br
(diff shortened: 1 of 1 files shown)


bogner (approved):

hekota:
image
image

specs/language/declarations.tex
```diff
@@ -1,7 +1,75 @@
...
+  
+  \define{function-declaration-seq}\br
+  \textit{function-declaration} \opt{function-declaration-seq}
+\end{grammar}
+
+\p The \texttt{export} specifier denotes that every \textit{function-declaration} included in \textit{function-declaration-seq} has \textit{external linkage} (\ref{Decl.Linkage.External}).
+
+\p The \textit{export-declaration-group} declaration cannot appear directly or indirectly within an unnamed namespace.
+
+\p Functions with \textit{external linkage} can also be decladed with an \texttt{export} specifier (\ref{Decl.Spec.Fct}).
```
bob80905, line 69: typo: decladed

specs/language/declarations.tex
```diff
@@ -1,7 +1,75 @@
...
+
+\p If a function is part of \textit{export-declaration-group} then all redeclarations of the same function must also be part on a \textit{export-declaration-group} or be declared with an \texttt{export} specifier (\ref{Decl.Spec.Fct}).
```
bob80905, line 71: typo: part *of* a

bob80905:
For completion, do you also want to define `function-declaration` ? It doesn't have a grammar definition.

hekota:
> For completion, do you also want to define `function-declaration` ? It doesn't have a grammar definition.

I believe that is beyond the scope of the task llvm/llvm-project#92812.

bob80905 (approved):

PR #271, opened 2024-07-04
Add proposal for Vulkan cooperative matrix

s-perron:
This propsal will expose the
[VK_KHR_cooperative_matrix](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_cooperative_matrix.html)
extension in HLSL.

### DIFF
--- a/proposals/0021-vk-coop-matrix.md
+++ b/proposals/0021-vk-coop-matrix.md
@@ -26,8 +26,8 @@ The solution to this problem is to add a new class,
 `vk::khr::CooperativeMatrix`, which will be defined in a header file
 "vk/khr/cooperative_matrix.h". This class will create an object with SPIR-V type
 `OpTypeCooperativeMatrixKHR`. Functions are added that will expose the SPIR-V
-the operations that take cooperative matrices as operands. All functions are
-defined to match the corresponding operations in the
+operations that take cooperative matrices as operands. All functions are defined
+to match the corresponding operations in the
 (SPV_KHR_cooperative_matrix)[https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_cooperative_matrix.html]
 extension.
 
@@ -44,9 +44,9 @@ the SPIR-V specification. The required enums will be defined in "vk/spirv.h".
 
 ### `vk::khr::CooperativeMatrix`
 
-The `CooperativeMatrix` class will defined in a header file as a wrapper around
-a `vk::SpirvType` that will expand to the appropriate SPIR-V type. This class
-will have the following interface.
+The `CooperativeMatrix` class will be defined in a header file as a wrapper
+around a `vk::SpirvType` that will expand to the appropriate SPIR-V type. This
+class will have the following interface.
 
 ```c++
 
@@ -146,7 +146,7 @@ class CooperativeMatrix {
        MemoryAccessMask memoryAccessMask = MemoryAccessMaskNone);
 
   // Constructs a cooperative matrix with all values initialized to v. Note that
-  // all active threads must have the same value for v.
+  // all threads in scope must have the same value for v.
   static CooperativeMatrix Splat(ComponentType v);
 
   // Returns the result of OpCooperativeMatrixLengthKHR on the current type.￼
@@ -224,7 +224,7 @@ The header file will check that the targeted Vulkan version is at least Vulkan
 
 Interactions with other HLSL features are implicitly compiler errors. The
 interface enforces all SPIR-V validation rules, and the compiler will issue
-errors if these rules are voilated.
+errors if these rules are violated.
 
 This will be tested by adding SPIR-V codegen tests that will verify that the
 correct code is generated when the header file is used.
--- a/proposals/0021-vk-coop-matrix.md
+++ b/proposals/0021-vk-coop-matrix.md
@@ -83,67 +83,83 @@ class CooperativeMatrix {
 
   // Store the cooperative matrix using OpCooperativeMatrixStoreKHR to
   // data using the given memory layout, stride, and memory access mask.
+  // `NonPrivatePointer` and `MakePointerVisible` with the workgroup scope
+  // will be added to the memory access mask to make the memory coherent.
   //
   // This function uses a SPIR-V pointer because HLSL does not allow grouphsared
   // memory object to be passed by reference. The pointer is a hack to get
   // around that.
-  template <MemoryAccessMask memoryAccessMask, class Type>
-  void Store(WorkgroupSpirvPointer<Type> data, CooperativeMatrixLayout layout,
-             uint32_t stride);
+  template <MemoryAccessMask memoryAccessMask, CooperativeMatrixLayout layout,
+            class Type>
+  void Store(WorkgroupSpirvPointer<Type> data, uint32_t stride);
 
   // Same as above, but uses MemoryAccessMaskNone for the memory access mask.
-  template <class Type>
-  void Store(WorkgroupSpirvPointer<Type> data, CooperativeMatrixLayout layout,
-             uint32_t stride);
+  template <CooperativeMatrixLayout layout, class Type>
+  void Store(WorkgroupSpirvPointer<Type> data, uint32_t stride);
 
   // Store the cooperative matrix using OpCooperativeMatrixStoreKHR to
   // data[index] using the given memory layout, stride, and memory access mask.
-  template <MemoryAccessMask memoryAccessMask, class Type>
-  void Store(RWStructuredBuffer<Type> data, uint32_t index,
-             CooperativeMatrixLayout layout, uint32_t stride);
+  // `NonPrivatePointer` and `MakePointerVisible` with the QueueFamily scope
+  // will be added to the memory access mask to make the memory coherent.
+  //
+  template <MemoryAccessMask memoryAccessMask, CooperativeMatrixLayout layout,
+            class Type>
+  void Store(globallycoherent RWStructuredBuffer<Type> data, uint32_t index,
+             uint32_t stride);
 
   // Same as above, but uses MemoryAccessMaskNone for the memory access mask.
-  template <class Type>
-  void Store(RWStructuredBuffer<Type> data, uint32_t index,
-             CooperativeMatrixLayout layout, uint32_t stride)
+  template <CooperativeMatrixLayout layout, class Type>
+  void Store(globallycoherent RWStructuredBuffer<Type> data, uint32_t index,
+             uint32_t stride);
 
   // Loads a cooperative matrix using OpCooperativeMatrixStoreKHR from
   // data using the given memory layout, stride, and memory access mask.
+  // `NonPrivatePointer` and `MakePointerAvailable` with the workgroup scope
+  // will be added to the memory access mask to make the memory coherent.
   //
   // This function uses a SPIR-V pointer because HLSL does not allow grouphsared
   // memory object to be passed by reference. The pointer is a hack to get
   // around that.
-  template <MemoryAccessMask memoryAccessMask, class Type>
+  template <MemoryAccessMask memoryAccessMask, CooperativeMatrixLayout layout,
+            class Type>
   static CooperativeMatrix Load(WorkgroupSpirvPointer<Type> data,
-                                CooperativeMatrixLayout layout,
                                 uint32_t stride);
 
   // Same as above, but uses MemoryAccessMaskNone for the memory access mask.
-  template <class Type>
+  template <CooperativeMatrixLayout layout, class Type>
   static CooperativeMatrix Load(WorkgroupSpirvPointer<Type> data,
-                                CooperativeMatrixLayout layout,
                                 uint32_t stride);
 
   // Loads a cooperative matrix using OpCooperativeMatrixLoadKHR from
   // data[index] using the given memory layout, stride, and memory access mask.
-  template <MemoryAccessMask memoryAccessMask, class Type>
-  static CooperativeMatrix Load(RWStructuredBuffer<Type> data, uint32_t index,
-                                CooperativeMatrixLayout layout,
+  // `NonPrivatePointer` and `MakePointerAvailable` with the QueueFamily scope
+  // will be added to the memory access mask to make the memory coherent.
+  //
+  template <MemoryAccessMask memoryAccessMask, CooperativeMatrixLayout layout,
+            class Type>
+  static CooperativeMatrix Load(globallycoherent RWStructuredBuffer<Type> data,
+                                uint32_t index,
+
                                 uint32_t stride);
 
   // Same as above, but uses MemoryAccessMaskNone for the memory access mask.
-  template <class Type>
-  static CooperativeMatrix Load(RWStructuredBuffer<Type> data, uint32_t index,
-                                CooperativeMatrixLayout layout,
-                                uint32_t stride);
+  template <CooperativeMatrixLayout layout, class Type>
+  static CooperativeMatrix Load(globallycoherent RWStructuredBuffer<Type> data,
+                                uint32_t index, uint32_t stride);
 
   // Loads a cooperative matrix using OpCooperativeMatrixLoadKHR from
   // data[index] using the given memory layout, stride, and memory access mask.
-  template <class Type>
-  static CooperativeMatrix
-  Load(StructuredBuffer<Type> data, uint32_t index,
-       CooperativeMatrixLayout layout, uint32_t stride,
-       MemoryAccessMask memoryAccessMask = MemoryAccessMaskNone);
+  // No memory access bits are added to the mask. Since the memory is readonly,
+  // there should be no need.
+  template <MemoryAccessMask memoryAccessMask, CooperativeMatrixLayout layout,
+            class Type>
+  static CooperativeMatrix Load(StructuredBuffer<Type> data, uint32_t index,
+                                uint32_t stride);
+
+  // Same as above, but uses MemoryAccessMaskNone for the memory access mask.
+  template <CooperativeMatrixLayout layout, class Type>
+  static CooperativeMatrix Load(StructuredBuffer<Type> data, uint32_t index,
+                                uint32_t stride);
 
   // Constructs a cooperative matrix with all values initialized to v. Note that
   // all threads in scope must have the same value for v.
@@ -167,7 +183,6 @@ class CooperativeMatrix {
       vk::integral_constant<uint, columns>, vk::integral_constant<uint, use> >;
 
   [[vk::ext_extension("SPV_KHR_cooperative_matrix")]]
-  [[vk::ext_extension("SPV_KHR_vulkan_memory_model")]]
   [[vk::ext_capability(/* CooperativeMatrixKHRCapability */ 6022)]]
   [[vk::ext_capability(/* VulkanMemoryModel */ 5345)]]
   SpirvMatrixType _matrix;
@@ -259,7 +274,9 @@ correct code is generated when the header file is used.
 7.  The `Get` and `Set` functions are used instead of `operator[]` because
     `operator[]` returns a reference, which is not available in HLSL.
 8.  We chose to default the memory operand on the load and store function to
-    None. This choice was arbitrary.
+    make workgroup and RWStructuredBuffers coherent. This is in line with the
+    test in the GLSLang test suite. We expect this to be the most common usage.
+    More Load and Store function can be added later if this is not true.
 9.  For the multiply-add function, we preferred to avoid flag parameters, and we
     added multiple versions of the function. We feel this provides better
     readability. If we were to pass the operand as a parameter, it would have to
--- a/proposals/0021-vk-coop-matrix.md
+++ b/proposals/0021-vk-coop-matrix.md
@@ -417,7 +417,7 @@ To be able to pass a GroupShared array by reference, we introduce a new type and
 function to `vk/spirv.h`.
 
 ```
-template <typename T>
+template <typename PointeeType>
 using WorkgroupSpirvPointer = const vk::SpirvOpaqueType<
     /* OpTypePointer */ 32,
     vk::Literal<vk::integral_constant<uint, StorageClassWorkgroup> >,
--- a/proposals/0021-vk-coop-matrix.md
+++ b/proposals/0021-vk-coop-matrix.md
@@ -167,7 +167,9 @@ class CooperativeMatrix {
       vk::integral_constant<uint, columns>, vk::integral_constant<uint, use> >;
 
   [[vk::ext_extension("SPV_KHR_cooperative_matrix")]]
+  [[vk::ext_extension("SPV_KHR_vulkan_memory_model")]]
   [[vk::ext_capability(/* CooperativeMatrixKHRCapability */ 6022)]]
+  [[vk::ext_capability(/* VulkanMemoryModel */ 5345)]]
   SpirvMatrixType _matrix;
   // clang-format on
 };
@@ -222,6 +224,12 @@ instruction.
 The header file will check that the targeted Vulkan version is at least Vulkan
 1.1, and issue an error if it is not.
 
+The SPIR-V specification requires that the Vulkan memory model is used when
+cooperative matrices are used. The Vulkan memory model extension and capability
+are added to the header file to indicate this. The compiler will have to target
+the Vulkan memory model when the cooperative matrix header file is included. It
+is the responsibility of the compiler to determine how to enforce this.
+
 Interactions with other HLSL features are implicitly compiler errors. The
 interface enforces all SPIR-V validation rules, and the compiler will issue
 errors if these rules are violated.
@@ -251,7 +259,7 @@ correct code is generated when the header file is used.
 7.  The `Get` and `Set` functions are used instead of `operator[]` because
     `operator[]` returns a reference, which is not available in HLSL.
 8.  We chose to default the memory operand on the load and store function to
-    None. This choice was arbitrary. 
+    None. This choice was arbitrary.
 9.  For the multiply-add function, we preferred to avoid flag parameters, and we
     added multiple versions of the function. We feel this provides better
     readability. If we were to pass the operand as a parameter, it would have to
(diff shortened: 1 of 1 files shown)


proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,264 @@
...
+
+// Store the cooperative matrix using OpCooperativeMatrixStoreKHR to data[i] using
+// memory layout RowMajorKHR and the given stride and memory access mask.
+void StoreRowMajor(RWStructuredBuffer<ComponentType> data, uint32_t index,
+uint32_t stride, MemoryAccessMask memoryAccessMask);
+
+// Store the cooperative matrix using OpCooperativeMatrixStoreKHR to data[i] using
+// memory layout ColumnMajorKHR and the given stride and memory access mask.
+void StoreColumnMajor(RWStructuredBuffer<ComponentType> data, uint32_t index,
+uint32_t stride, MemoryAccessMask memoryAccessMask);
```
s-perron, line 92: These need to be extended to be more like the loads. In particular, we want to be able to store to an array in groupshared memory.

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,264 @@
...
+a `vk::SpirvType` that will expand to the appropriate SPIR-V type. This class
+will have the following interface.
+
+```c++
+
+// The base cooperative matrix class. The template arguments correspond to the
+// operands in the OpTypeCooperativeMatrixKHR instruction.
+template <typename ComponentType, Scope scope, uint rows, uint columns,
+CooperativeMatrixUse use>
+class CooperativeMatrix {
```
s-perron, line 56: There are some operations that the spir-v spec allows that are missing:

- [ ] Constructing a coop  matrix from a single scalar
- [ ] OpConvertFToU
- [ ] OpConvertFToS
- [ ] OpConvertSToF
- [ ] OpConvertUToF
- [ ] OpUConvert
- [ ] OpSConvert
- [ ] OpFConvert
- [ ] OpBitCast
- [ ] accessing individual elements (This will add access chains that spirv-opt will have to optimize).

s-perron: All of the operations have been added to the header file implementation in DXC. Still need to update this PR.

proposals/0021-vk-coop-matrix.md
s-perron: Still working on how to allow groupshared memory. I don't see how to do with without allowing more that is legal. In this case, spirv-val will be the backstop, but that is not ideal.

There are also some issues with spirv-opt not optimizing correctly when the groupshared memory is used. I'll need to get another fix into spirv-opt.

s-perron: We cannot have a wrapper function that will load and store from group shared memory because HLSL does not allow us to pass that type of memory by reference. If someone want to do a loads and stores they will have to access the ext_instruction and SPIR-V types directly.

llvm-beanz: Adding pointers/references to the type system is kinda a pain, but I wonder if we could do something slightly hacky that doesn't actually change the language to address cases like this.

We do have the ability to represent `ref` passed parameters in DXC's intrinsics, so maybe we could have a builtin that converts a `ref groupshared T` to a `GroupsharedAddr<T>` builtin object, that has no methods and can't be used for anything other than passing into some limited set of builtin functions.

Thoughts?

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,346 @@
+<!-- {% raw %} -->
+
+# Vulkan cooperative matrix extension
+
+*   Proposal: [0021](0021-vk-coop-matrix.md)
+*   Author(s): [Steven Perron](https://github.com/s-perron)
+*   Sponsor: TBD
+*   Status: **Accepted**
+*   Required Version: Vulkan 1.1
```
Keenuts, line 9: SPIR-V spec requires SPIR-V 1.6, so seems like Vulkan 1.3 should be the required version as older version don't require SPIR-V 1.6 support. (even if the VK extension has no min VK version requirement as long as VK_KHR_get_physical_device_properties2 is present)

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,346 @@
...
+Users have no user-friendly way to write Vulkan shaders that use cooperative
+matrices. This is a useful feature that users would like to use.
+
+## High-level description
+
+The solution to this problem is to add a new class,
+`vk::khr::CooperativeMatrix`, which will be defined in a header file
+"vk/khr/cooperative_matrix.h". This class will create an object with Spir-V type
+`OpTypeCooperativeMatrixKHR`. Functions are added that will expose the Spir-V
+the operations that take cooperative matrices as operands. All functions are
```
Keenuts, line 29:
```suggestion
`OpTypeCooperativeMatrixKHR`. Functions are added that will expose the SPIR-V operations that take cooperative matrices as operands. All functions are
```

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,346 @@
...
+
+  // Store the cooperative matrix using OpCooperativeMatrixStoreKHR to data[i]
+  // using memory layout ColumnMajorKHR and the given stride and memory access
+  // mask.
+  template <class Type>
+  void StoreColumnMajor(RWStructuredBuffer<Type> data, uint32_t index,
+                        uint32_t stride, MemoryAccessMask memoryAccessMask);
+
+  // Constructs a cooperative matrix with all values initialized to v. Note that
+  // all active threads must have the same value for v.
```
Keenuts, line 114: My understanding of the constraint is: If you use a cooperative matrix type operation, execution should be uniform across the whole scope, not just active threads:

(For OpTypeCooperativeMatrixKHR)
```
All dynamic instances of an instruction with an operand or result that is an object of this type must be executed such that all the invocations in the Scope instance are active or none of them are.
```

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,346 @@
...
+attribute `vk::ext_instruction`. The `GetLength()` function is a wrapper around
+a builtin function, which the compiler expands the appropriate SPIR-V
+instruction.
+
+The header file will check that the targeted Vulkan version is at least Vulkan
+1.1, and issue an error if it is not.
+
+Interactions with other HLSL features are implicitly compiler errors. The
+interface enforces all SPIR-V validation rules, and the compiler will issue
+errors if these rules are voilated.
```
Keenuts, line 218:
```suggestion
errors if these rules are violated.
```

proposals/0021-vk-coop-matrix.md
s-perron: Interesting idea. I started thinking about this, and I realized we can do something similar using only inline spir-v. See https://godbolt.org/z/WecY1a9K5 for how we can define an opaque pointer type that can be passed around, but only really used in inline spir-v defined function and builtin functions.

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,346 @@
...
+a `vk::SpirvType` that will expand to the appropriate SPIR-V type. This class
+will have the following interface.
+
+```c++
+
+// The base cooperative matrix class. The template arguments correspond to the
+// operands in the OpTypeCooperativeMatrixKHR instruction.
+template <typename ComponentType, Scope scope, uint rows, uint columns,
+          CooperativeMatrixUse use>
+class CooperativeMatrix {
```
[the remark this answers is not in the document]
s-perron: Only a default constructor that leaves it undefined. We cannot add constructors in HLSL (https://godbolt.org/z/WqrEz85xr). Some of the awkwardness is caused by limitations in HLSL.

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,346 @@
...
+  // Note: The SPIR-V specification allows the load and store function to access
+  // workgroup data (groupshared in HLSL). This call does not accept group
+  // shared data because HLSL cannot pass the groupshared array by reference. It
+  // is impossible to implement a function that will load and store from group
+  // sharded data. This can be revisited when references are added to HLSL.
+
+  // Store the cooperative matrix using OpCooperativeMatrixStoreKHR to data[i]
+  // using memory layout RowMajorKHR.￼
+  template <class Type>
+  void StoreRowMajor(RWStructuredBuffer<Type> data, uint32_t index);
```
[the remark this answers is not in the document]
s-perron: It should have to have a stride. 

The OpCooperativeMatrixLoadKHR and OpCooperativeMatrixStoreKHR say the stride is optional. I did not read the requirements in the Cooperative Matrix Layout section. Both layouts require a stride. I'll remove these when I redo the load and store instruction.

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,346 @@
...
+  // Store the cooperative matrix using OpCooperativeMatrixStoreKHR to data[i]
+  // using memory layout ColumnMajorKHR.￼
+  template <class Type>
+  void StoreColumnMajor(RWStructuredBuffer<Type> data, uint32_t index);
+
+  // Store the cooperative matrix using OpCooperativeMatrixStoreKHR to data[i]
+  // using memory layout RowMajorKHR and the given stride and memory access
+  // mask.
+  template <class Type>
+  void StoreRowMajor(RWStructuredBuffer<Type> data, uint32_t index,
```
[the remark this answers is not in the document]
s-perron: It is not necessary. This was just my personal preference for functions with fewer parameters. If there will be more layouts, it might be better to make it a parameter.

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,346 @@
...
+  // Store the cooperative matrix using OpCooperativeMatrixStoreKHR to data[i]
+  // using memory layout ColumnMajorKHR and the given stride and memory access
+  // mask.
+  template <class Type>
+  void StoreColumnMajor(RWStructuredBuffer<Type> data, uint32_t index,
+                        uint32_t stride, MemoryAccessMask memoryAccessMask);
+
+  // Constructs a cooperative matrix with all values initialized to v. Note that
+  // all active threads must have the same value for v.
+  static CooperativeMatrix splat(ComponentType v);
```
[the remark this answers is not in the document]
s-perron: Can't add constructors. :shrug:

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,346 @@
...
+  // Apply OpIMul or OFMul, depending on ComponentType, in a element by element
+  // manner.
+  CooperativeMatrix operator*(CooperativeMatrix other);
+
+  // Apply OpSDiv, OpUDiv or OFDiv, depending on ComponentType, in a element by
+  // element manner.
+  CooperativeMatrix operator/(CooperativeMatrix other);
+
+  // Apply OpMatrixTimesScalar in a element by element manner.
+  CooperativeMatrix operator*(ComponentType scalar);
```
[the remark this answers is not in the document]
s-perron: About the left multiple, we cannot overload non-member operators (https://godbolt.org/z/dvq59jEaG).

I don't think we can overload +=, -=, etc.  (https://godbolt.org/z/nv44W86Pq).

proposals/0021-vk-coop-matrix.md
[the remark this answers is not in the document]
llvm-beanz: I _think_ we'll probably add constructors in HLSL 202y. They'd be pretty hard to add in DXC for a lot of not-great reasons. My initial PR to propose them is [here](https://github.com/microsoft/hlsl-specs/pull/5). GitHub closed it because the original fork doesn't exist anymore.

proposals/0021-vk-coop-matrix.md
s-perron: I added an opaque pointer type, and a function to get the address of the groupshared memory.

proposals/0021-vk-coop-matrix.md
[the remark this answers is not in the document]
s-perron: I added some extra design notes that explain why some decision were made.

proposals/0021-vk-coop-matrix.md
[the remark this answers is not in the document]
s-perron: I reworked the load and store function. The layout is now a parameter, and there is always a stride.

proposals/0021-vk-coop-matrix.md
s-perron: Fixed.

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,426 @@
...
+
+  // Store the cooperative matrix using OpCooperativeMatrixStoreKHR to
+  // data using the given memory layout, stride, and memory access mask.
+  //
+  // This function uses a SPIR-V pointer because HLSL does not allow grouphsared
+  // memory object to be passed by reference. The pointer is a hack to get
+  // around that.
+  template <MemoryAccessMask memoryAccessMask, class Type>
+  void Store(WorkgroupSpirvPointer<Type> data, CooperativeMatrixLayout layout,
```
jeffbolznv, line 91: Since the layout must be a constant instruction in SPIR-V, should it be a template parameter?

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,426 @@
...
+
+where `v` must be a object in GroupShared memory.
+
+For example,
+
+```
+groupshared float shared_data[64];
+...
+WorkgroupSpirvPointer<float> scalar_ptr = vk::GetGroupSharedAddress(shared_data[0]);
+WorkgroupSpirvPointer<float[64]> array_ptr = vk::GetGroupSharedAddress(shared_data);
```
jeffbolznv, line 359: Are there any additional limitations on this WorkgroupSpirvPointer type? I'm concerned that use of this may lead to needing VariablePointer support (e.g. if these are loaded from a local variable).

proposals/0021-vk-coop-matrix.md
s-perron, line 1: This capability requires the vulkan memory model. I'll need to look into adding an option for that. :(.

proposals/0021-vk-coop-matrix.md
s-perron: There might be some uses. I'll look into what is possible.

proposals/0021-vk-coop-matrix.md
s-perron: It needs to be constant, instead of a literal, so it works. However, making it a template parameter would give a better error message. I'll check that out.

proposals/0021-vk-coop-matrix.md
s-perron: I made the type `const`, which will stop the need for variable pointers in all reasonable situations. I have a test for on the DXC pr where noinline is used on a function with a pointer as a parameter. This would require variable pointers, and I show how to add it. I don't expect people to do that.

Otherwise, I expect the legalization pass to optimize away all function and private storage class variables.

If it is used on another storage class, the validator will complain with a poor error.

proposals/0021-vk-coop-matrix.md
s-perron: I added the vulkan memory model extension and capability to header file. This will indicate that the vulkan memory model should be used.

I also updated DXC to generate code for the vulkan memory model whenever it sees the vulkan memory model capability.

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,457 @@
...
+
+  // Same as above, but uses MemoryAccessMaskNone for the memory access mask.
+  template <CooperativeMatrixLayout layout, class Type>
+  static CooperativeMatrix Load(WorkgroupSpirvPointer<Type> data,
+                                uint32_t stride);
+
+  // Loads a cooperative matrix using OpCooperativeMatrixLoadKHR from
+  // data[index] using the given memory layout, stride, and memory access mask.
+  // `NonPrivatePointer` and `MakePointerAvailable` with the QueueFamily scope
+  // will be added to the memory access mask to make the memory coherent.
```
jeffbolznv, line 136: Is this saying they're unconditionally coherent? Most of the time you want noncoherent, for performance. In the GLSL extension, you get coherent if and only if the buffer/member is qualified with the coherent keyword.

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,346 @@
...
+  // using memory layout RowMajorKHR.￼
+  template <class Type>
+  void StoreRowMajor(RWStructuredBuffer<Type> data, uint32_t index);
+
+  // Store the cooperative matrix using OpCooperativeMatrixStoreKHR to data[i]
+  // using memory layout ColumnMajorKHR.￼
+  template <class Type>
+  void StoreColumnMajor(RWStructuredBuffer<Type> data, uint32_t index);
+
+  // Store the cooperative matrix using OpCooperativeMatrixStoreKHR to data[i]
```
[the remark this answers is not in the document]
jeffbolznv: I think this is fine, thanks.

proposals/0021-vk-coop-matrix.md
s-perron: I cannot have an overload distinguishes between something marked globallycoherent or not. If we wanted both, I'd have to have two separate functions with different names.

I went with this because all of the `spv.coopmat.*` tests in glslang us coherent for everything, I assumed that was the common case.

Let me know the best implementation. You know better how it will be used. I'm just implementing it. 

FYI: I interpreted that the workgroup memory should always be coherent from the GLSL tests:

https://github.com/KhronosGroup/glslang/blob/7c4d91e7819a1d27213aa3499953d54ae1a00e8f/Test/spv.coopmat.comp#L112-L113

The expected result is

https://github.com/KhronosGroup/glslang/blob/7c4d91e7819a1d27213aa3499953d54ae1a00e8f/Test/baseResults/spv.coopmat.comp.out#L352C1-L353C112

jeffbolznv: > I went with this because all of the spv.coopmat.* tests in glslang us coherent for everything, I assumed that was the common case.

That was probably just me testing that the right qualifier was emitted and then copy/pasting it everywhere.

Can you just let the shader author choose by setting the right bits in memoryAccessMask? Maybe automatically add it for shared memory because it's free-ish.

> FYI: I interpreted that the workgroup memory should always be coherent from the GLSL tests:

Right, the GLSL spec says "Shared variables are implicitly coherent".

llvm-beanz (approved):
I'm going to approve this just to be clear that this is fully approved and the direction is _awesome_.

I know there are some details still being worked out, but the direction of this proposal is fantastic and I'm eager for its inclusion.

Thank you everyone for all the amazing work on this!

proposals/0021-vk-coop-matrix.md
s-perron: For RWStructuredBuffers, I added a CoherentLoad and CoherentStore. The problem with just letting people set the bits is that it gets ugly because of the scope argument or other extra arguments depending on which memory operands are used.

jeffbolznv: Right, that makes sense. Adding an overload is fine with me.

s-perron:
@jeffbolznv @Keenuts I think I have this done. Let me know if there are any other issues. I'll be on vacation for a week. I'll do any  fixups when I return. I hope to merge at that time. I have the implementation done, and ready to be reviewed as soon as this merges.

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,466 @@
...
+  // Apply OpSDiv, OpUDiv or OFDiv, depending on ComponentType, in a element by
+  // element manner.
+  CooperativeMatrix operator/(CooperativeMatrix other);
+
+  // Apply OpMatrixTimesScalar in a element by element manner.
+  CooperativeMatrix operator*(ComponentType scalar);
+
+  // Store the cooperative matrix using OpCooperativeMatrixStoreKHR to
+  // data using the given memory layout, stride, and memory access mask.
+  // `NonPrivatePointer` and `MakePointerVisible` with the workgroup scope
```
jeffbolznv, line 86: The Av/Vis bits are backward - stores should do MakePointerAvailable, loads should do MakePointerVisible.

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,466 @@
...
+  // will be added to the memory access mask to make the memory coherent.
+  //
+  // This function uses a SPIR-V pointer because HLSL does not allow groupshared
+  // memory object to be passed by reference. The pointer is a hack to get
+  // around that.
+  //
+  // The layout and stride will be passed to the SPIR-V instruction as is. The precise meaning can be found in the specification for SPV_KHR_cooperative_matrix.
+  template <MemoryAccessMask memoryAccessMask, CooperativeMatrixLayout layout,
+            class Type>
+  static CooperativeMatrix Load(WorkgroupSpirvPointer<Type> data,
```
jeffbolznv, line 130: If the shader needs to use a start index into the array, can it do `vk::GetGroupSharedAddress(shared_data[index]);`?

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,466 @@
...
+  // data using the given memory layout, stride, and memory access mask.
+  // `NonPrivatePointer` and `MakePointerVisible` with the workgroup scope
+  // will be added to the memory access mask to make the memory coherent.
+  //
+  // This function uses a SPIR-V pointer because HLSL does not allow groupshared
+  // memory object to be passed by reference. The pointer is a hack to get
+  // around that.
+  //
+  // The layout and stride will be passed to the SPIR-V instruction as is. The precise meaning can be found in the specification for SPV_KHR_cooperative_matrix.
+  template <MemoryAccessMask memoryAccessMask, CooperativeMatrixLayout layout,
```
jeffbolznv, line 94: sort of a nitpick, but the strongly typed MemoryAccessMask enum will be clunky (and technically undefined behavior) if the shader wants to use more than one bit. I wonder if we still even need memoryAccessMask after how coherent has been resolved?

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,466 @@
...
+            class Type>
+  void Store(globallycoherent RWStructuredBuffer<Type> data, uint32_t index,
+             uint32_t stride);
+
+  // Same as above, but uses MemoryAccessMaskNone for the memory access mask.
+  template <CooperativeMatrixLayout layout, class Type>
+  void Store(globallycoherent RWStructuredBuffer<Type> data, uint32_t index,
+             uint32_t stride);
+
+  // Loads a cooperative matrix using OpCooperativeMatrixStoreKHR from
```
jeffbolznv, line 118: s/Store/Load

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,466 @@
...
+      vk::integral_constant<uint, columns>, vk::integral_constant<uint, use> >;
+
+  [[vk::ext_extension("SPV_KHR_cooperative_matrix")]]
+  [[vk::ext_capability(/* CooperativeMatrixKHRCapability */ 6022)]]
+  [[vk::ext_capability(/* VulkanMemoryModel */ 5345)]]
+  SpirvMatrixType _matrix;
+  // clang-format on
+};
+
+// Cooperative matrix that can be used in the "a" position of a multiple add
```
jeffbolznv, line 201: s/multiple/multiply (and a couple more below)

proposals/0021-vk-coop-matrix.md
```diff
@@ -0,0 +1,466 @@
...
+    reference. Also, the compiler will not be able to apply the correct layout
+    to the opaque pointer, so the workaround used for groupshared variables will
+    not work in general.
+6.  The `GetLength()` cannot be implemented using a `vk::ext_instruction`
+    function because the opcode expects an id of a type, and that cannot be
+    defined in inline SPIR-V.
+7.  The `Get` and `Set` functions are used instead of `operator[]` because
+    `operator[]` returns a reference, which is not available in HLSL.
+8.  We chose to default the memory operand on the load and store function to
+    make workgroup and RWStructuredBuffers coherent. This is in line with the
```
jeffbolznv, line 286: out of date

proposals/0021-vk-coop-matrix.md
s-perron: done.

proposals/0021-vk-coop-matrix.md
s-perron: Yes it is possible, and necessary even if index is 0, to get the correct pointer type. See the corresponding DXC PR for examples: https://github.com/microsoft/DirectXShaderCompiler/pull/6720/files#diff-586bab68d22cbac8d952abb2f83103a319ec3184d6419b61a50fbbf49500deee

proposals/0021-vk-coop-matrix.md
s-perron: I kept the version with the memory access mask just. There are a couple things people could do with it, even if it is not common. The extra testing is not that onerous. It will also future proof a little.

Yes the use of the bit mask is clunky, but I expect the overloads without the memory access mask to be used most of the time. It is only ugly for those who need to so something special.

I change the type to uint32_t to avoid the undefined behaviour.

proposals/0021-vk-coop-matrix.md
s-perron: done.

proposals/0021-vk-coop-matrix.md
s-perron: I removed point 8.

jeffbolznv (approved):
LGTM

proposals/0021-vk-coop-matrix.md
```diff
@@ -285,7 +411,45 @@ There will be template specializations for all pairs of the following types:
...
 *   `uint64_t`
 
-### Spir-V enums
+### SPIR-V pointers
+
+To be able to pass a GroupShared array by reference, we introduce a new type and
+function to `vk/spirv.h`.
+
+```
+template <typename T>
```
Keenuts, line 420:
```suggestion
template <typename PointeeType>
```
?

Keenuts (approved):

PR #276, opened 2024-07-17
Add a proposal on how to handle the loop unroll factor in HLSL 202x

farzonl:
This PR gives an intro to the issue at hand, the motifications for unification and a proposed solution.

### DIFF
--- a/proposals/NNNN-hlsl-loop-unroll-factor.md
+++ b/proposals/NNNN-hlsl-loop-unroll-factor.md
@@ -1,0 +1,38 @@
+# HLSL Loop Unroll Factor
+
+* Proposal: [NNNN](NNNN-hlsl-loop-unroll-factor.md)
+* Author(s): [Farzon Lotfi](https://github.com/farzonl)
+* Sponsor: [Farzon Lotfi](https://github.com/farzonl)
+* Status: **Under Consideration**
+* Impacted Projects: DXC & Clang
+
+## Introduction
+
+This proposal seeks to unify the diverging behaviors between the agreed upon 
+[spec for HLSL loop unroll](https://github.com/microsoft/hlsl-specs/pull/263)
+and the current behavior of DXC.
+
+The new spec wants to treat the unroll factor as a hint to the compiler for 
+partial loop unroll. The current behavior in dxc is that the unroll factor 
+specifies the maximum number of times the loop is to execute. 
+
+The DXC behavior was determined to violate user expectations by overriding 
+the existing loop termination conditions. Further the DXC behavior diverges 
+from how both clang and openCL treat the loop unroll factor which would have 
+made our port of HLSL loop unroll a bunch special cases instead of just 
+syntactic sugar that could sit on top of the existing loop unroll 
+implementations that exist in LLVM.
+
+## Motivation
+The HLSL compilers transition to clang has resulted in a compat break between 
+the [HLSLS loop unroll implementation in clang](https://github.com/llvm/llvm-project/pull/93879)
+and the one in DXC. While there is an expectation that the new compiler will 
+not be fully compatible with the previous compilers, These compat breaks should
+ be minimized when possible.
+
+ ## Proposed solution
+In [HLSL 202X](0020-hlsl-202x-202y.md) the DXC implementation of the loop 
+unroll factor should no longer alter the  number of iteration executions. 
+Ideally the loop unroll factor will also become a compiler hint for partial 
+loop unroll. If this can not be done because of concerns of invalid DXIL 
+generation then the loop unroll factor as a feature should be removed from DXC.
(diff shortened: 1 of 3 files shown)


proposals/NNNN-hlsl-loop-unroll-factor.md
```diff
@@ -0,0 +1,38 @@
...
+the [HLSLS loop unroll implementation in clang](https://github.com/llvm/llvm-project/pull/93879)
+and the one in DXC. While there is an expectation that the new compiler will 
+not be fully compatible with the previous compilers, These compat breaks should
+ be minimized when possible.
+
+ ## Proposed solution
+In [HLSL 202X](0020-hlsl-202x-202y.md) the DXC implementation of the loop 
+unroll factor should no longer alter the  number of iteration executions. 
+Ideally the loop unroll factor will also become a compiler hint for partial 
+loop unroll. If this can not be done because of concerns of invalid DXIL
```
bob80905, line 37: What invalid DXIL could be generated as a result of treating the loop unroll factor as a compiler hint? Are there specific examples that come to mind?

farzonl: Maybe the lanaguage needs to be more clear.  Today if unroll doesn't happen that could be invalid DXIL. @llvm-beanz and @tex3d alerted me to this.

 The change with the  unroll factor would be to allow for partial unrolls. We don't do partial unrolling today. So an unroll of n today in HLSL would be a full unroll just up to the number n. If there is code now that still remains in a loop body that could be invalid DXIL.

dmpots: A simple example is sample offsets, which upto sm6.6 had to be immedate values
https://godbolt.org/z/d6PsancE4

```
SamplerState s0;
Texture2D<float> t0;

float main(float2 uv : A) : SV_Target
{
    float result = 0;

    [loop]
    for (int i = 0; i < 3; ++i)
    {
        result += t0.Sample(s0, uv, i); // Fails to compile without unroll
    }

    return result;
}
```

proposals/NNNN-hlsl-loop-unroll-factor.md
```diff
@@ -0,0 +1,38 @@
...
+and the one in DXC. While there is an expectation that the new compiler will 
+not be fully compatible with the previous compilers, These compat breaks should
+ be minimized when possible.
+
+ ## Proposed solution
+In [HLSL 202X](0020-hlsl-202x-202y.md) the DXC implementation of the loop 
+unroll factor should no longer alter the  number of iteration executions. 
+Ideally the loop unroll factor will also become a compiler hint for partial 
+loop unroll. If this can not be done because of concerns of invalid DXIL 
+generation then the loop unroll factor as a feature should be removed from DXC.
```
dmpots, line 38: It would be good to clarify what "removed" means here. Are you thinking this would this be a hard error? Or just ignore the value and issue a warning.

dmpots (approved):

proposals/NNNN-hlsl-loop-unroll-factor.md
```diff
@@ -0,0 +1,38 @@
+# HLSL Loop Unroll Factor
+
```
farzonl, line 2: I need a proposal number.

proposals/NNNN-hlsl-loop-unroll-factor.md
llvm-beanz: I think we should separate out cases where DXIL is invalid without `[unroll]` (or `[flatten]`) being respected.

For those cases the compiler _should_ automatically unroll during legalization. It's possible that is too hard to get right in DXC so we would only implement that correctly in Clang, but we should get it right in Clang.

proposals/NNNN-hlsl-loop-unroll-factor.md
damyanp: Process is that you pick this _just_ before completing the, PR based on what's in main at the time.

proposals/NNNN-hlsl-loop-unroll-factor.md
```diff
@@ -0,0 +1,38 @@
...
+The DXC behavior was determined to violate user expectations by overriding 
+the existing loop termination conditions. Further the DXC behavior diverges 
+from how both clang and openCL treat the loop unroll factor which would have 
+made our port of HLSL loop unroll a bunch special cases instead of just 
+syntactic sugar that could sit on top of the existing loop unroll 
+implementations that exist in LLVM.
+
+## Motivation
+The HLSL compilers transition to clang has resulted in a compat break between 
+the [HLSLS loop unroll implementation in clang](https://github.com/llvm/llvm-project/pull/93879)
```
damyanp, line 28:
```suggestion
the [HLSL loop unroll implementation in clang](https://github.com/llvm/llvm-project/pull/93879)
```

damyanp (approved):

proposals/NNNN-hlsl-loop-unroll-factor.md
farzonl: To lower the code impact it makes sense to follow standard deprecation procedures. Ignore the value and issue a warning makes the most sense.

proposals/NNNN-hlsl-loop-unroll-factor.md
farzonl: 0022

PR #277, opened 2024-07-18
Shader Execution Reordering (SER) proposal

rasmusnv:
Add Shader Execution Reordering (SER) proposal for consideration.

### DIFF
--- a/proposals/0025-shader-execution-reordering.md
+++ b/proposals/0025-shader-execution-reordering.md
@@ -724,7 +724,10 @@ Parameter                           | Definition
 
 This variant of `ReorderThread` reorders threads based on a generic
 user-provided hint. Similarity of hint values should indicate expected
-similarity of subsequent work being performed by threads. The resolution of
+similarity of subsequent work being performed by threads. In general, more
+significant bits of the hint value are more important than less significant
+bits for determining coherency, though scheduling behavior is implementation
+specific. The resolution of
 the hint is implementation-specific. If an implementation cannot resolve all
 values of `CoherenceHint`, it is free to ignore an arbitrary number of least
 significant bits. The thread ordering resulting from this call may be
--- a/proposals/0025-shader-execution-reordering.md
+++ b/proposals/0025-shader-execution-reordering.md
@@ -80,7 +80,7 @@ the programming model is limited to HLSL and DXIL.
 This section describes the HLSL additions for `HitObject` and `ReorderThread`
 in detail.
 The canonical use of these features involve changing a `TraceRay` call to the
-following sequence:
+following sequence that is functionally equivalent:
 
 ```C++
 HitObject Hit = HitObject::TraceRay( ..., Payload );
--- a/proposals/0025-shader-execution-reordering.md
+++ b/proposals/0025-shader-execution-reordering.md
@@ -724,10 +724,10 @@ Parameter                           | Definition
 
 This variant of `ReorderThread` reorders threads based on a generic
 user-provided hint. Similarity of hint values should indicate expected
-similarity of subsequent work being performed by threads. In general, more
-significant bits of the hint value are more important than less significant
-bits for determining coherency, though scheduling behavior is implementation
-specific. The resolution of
+similarity of subsequent work being performed by threads. More significant
+bits of the hint value are more important than less significant bits for
+determining coherency. Specific scheduling behavior may vary by
+implementation. The resolution of
 the hint is implementation-specific. If an implementation cannot resolve all
 values of `CoherenceHint`, it is free to ignore an arbitrary number of least
 significant bits. The thread ordering resulting from this call may be
--- a/proposals/0025-shader-execution-reordering.md
+++ b/proposals/0025-shader-execution-reordering.md
@@ -1,7 +1,7 @@
 
 # Shader Execution Reordering (SER)
 
-* Proposal: [NNNN](NNNN-shader-execution-reordering.md)
+* Proposal: [0025](0025-shader-execution-reordering.md)
 * Author(s): [Rasmus Barringer](https://github.com/rasmusnv), Robert Toth,
 Michael Haidl, Simon Moll, Martin Stich
 * Sponsor: [Tex Riddell](https://github.com/tex3d)
--- a/proposals/0025-shader-execution-reordering.md
+++ b/proposals/0025-shader-execution-reordering.md
@@ -115,10 +115,11 @@ impact any other `HitObject` in the shader. A shader can have any number of
 active `HitObject`s at a time. Each `HitObject` will take up some hardware
 resources to hold information related to the hit or miss, including ray,
 intersection attributes and information about the shader to invoke. The size
-of a `HitObject` is unspecified. As a consequence, `HitObject` cannot be
-stored in structured buffers. Ray payload structs must not have members of
-`HitObject` type. `HitObject` supports assignment (by-value copy) and can be
-passed as arguments to and returned from local functions.
+of a `HitObject` is unspecified, and thus considered intangible. As a
+consequence, `HitObject` cannot be stored in memory buffers, in groupshared
+memory, in ray payloads, or in intersection attributes. `HitObject` supports
+assignment (by-value copy) and can be passed as arguments to and returned from
+local inlined functions.
 
 A `HitObject` is default-initialized to encode a NOP-HitObject (see `HitObject::MakeNop`).
 A NOP-HitObject can be used with `HitObject::Invoke` and `ReorderThread` but
--- a/proposals/0025-shader-execution-reordering.md
+++ b/proposals/0025-shader-execution-reordering.md
@@ -829,12 +829,12 @@ only of shaders, but also of the code after the loop.
 ```C++
 for( int bounceCount=0; ; bounceCount++ )
 {
-    HitObject hit;    // initialize to NOP-hitobject
+    HitObject hit;    // initialize to NOP-HitObject
 
     // Have threads conditionally participate in the next bounce depending on
     // loop iteration count and path throughput. Instead of having
     // non-participating threads break out of the loop here, we let them
-    // participate in the reordering with a NOP-hitobject first.
+    // participate in the reordering with a NOP-HitObject first.
     // This will cause them to be grouped together and execute the work after
     // the loop with higher coherence. Note that the values of 'bounceCount'
     // might differ between threads in a wave, because reordering may group
@@ -873,7 +873,10 @@ for( int bounceCount=0; ; bounceCount++ )
 
 ## Reorder points
 
-A _reorder point_ is a point in the control flow of a shader or sequence of
+The existing DXR API define _repacking points_ at `TraceRay` and `CallShader`.
+This proposal aims to formalize the concept and suggests renaming them to _reorder points_.
+
+A reorder point is a point in the control flow of a shader or sequence of
 shaders where the system may arbitrarily change the physical arrangement of
 threads executing on the GPU, usually with the goal of increasing coherence.
 That is, the system may choose to migrate thread contexts that arrive at a
@@ -918,18 +921,17 @@ scenarios.
 
 While it is understood that reordering at `TraceRay` and `CallShader` is done
 at the discretion of the driver, `HitObject::TraceRay` and `HitObject::Invoke`
-are intended to be used in combination with `ReorderThread`.
-The desired behavior, as communicated from a shader to the driver, is inferred
-from the presence of `ReorderThread`.
-If no `ReorderThread` call is made, it implies that the driver should minimize
-its efforts to reorder for hit coherence.
-Conversely, the existence of a `ReorderThread` call implies that more sophisticated
-reordering will be beneficial, and that reordering conceptually occurs at the
-`ReorderThread` call.
-This describes the intended behavior communicated from an application, though
-the actual behavior remains at the driver's discretion.
-For example, an implementation may choose to defer the reordering operation
-implied by `ReorderThread` to the next reorder point.
+are intended to be used in conjunction with `ReorderThread`.
+Reordering at `HitObject::TraceRay` and `HitObject::Invoke` is permitted but the
+driver should minimize its efforts to reorder for hit coherence and instead
+prioritize reordering through `ReorderThread`.
+
+Some implementations may achieve best performance when `HitObject::TraceRay`,
+`ReorderThread`, and `HitObject::Invoke` are called back-to-back.
+This case is semantically equivalent to DXR 1.0 `TraceRay` but with defined
+reordering characteristics.
+The back-to-back combination of `ReorderThread` and `HitObject::Invoke` may
+similarly see a performance benefit on some implementations.
 
 For performance reasons, it is crucial that the DXIL-generating compiler does
 not move non-uniform resource access across reorder points in general, and across
@@ -983,9 +985,17 @@ Due to the existence of non-coherent caches on most modern GPUs, an
 application must take special care when communicating information across
 reorder points through memory (UAVs).
 
-Specifically, if UAV stores or atomics are performed on one side of a reorder
-point, and on the other side the data is read via non-atomic UAV reads, the
-following is required:
+This proposal includes a new coherence scope for reorder points. The scope
+is limited to communication within the same dispatch index. Specifically,
+if UAV stores or atomics are performed on one side of a reorder point, and
+on the other side the data is read via non-atomic UAV reads, the following
+is required:
+- The UAV must be declared `[reordercoherent]`.
+- The UAV writer must issue a `Barrier(UAV_MEMORY, REORDER_SCOPE)` between
+the write and the reorder point.
+
+When communicating both between threads and across reorder points, global coherency
+can be utilized:
 - The UAV must be declared `[globallycoherent]`.
 - The UAV writer must issue a `DeviceMemoryBarrier` between the write and the
 reorder point.
@@ -1030,6 +1040,9 @@ Some examples follow.
 
 ### Example: Common computations that rely on large raygeneration state
 
+In this example, a large number of surface events are tracked in the ray generation shader.
+Only the result is communicated to the invoked shader via the payload.
+
 ```C++
 struct IorData
 {
@@ -1056,6 +1069,9 @@ for( ... )
 
 ### Example: Do common computations with hit-coherence
 
+In this example, common code runs in the raygeneration shader, after
+the thread has been reordered for hit coherence.
+
 ```C++
 hit = HitObject::TraceRay( ... );
 ReorderThread( hit );
@@ -1067,11 +1083,17 @@ HitObject::Invoke( hit, payload );
 
 ### Example: Same surface shader but different behavior in the raygeneration shader
 
+This example demonstrates varying reordering behavior and shadow logic, despite
+using the same shader code.
+
 ```C++
-// Primary ray wants perfect shadows and does not need to reorder as it is
-// coherent enough.
+// Primary ray wants perfect shadows and does not need to explicitly
+// reorder for hit coherence as it is coherent enough.
 ray = GeneratePrimaryRay();
 hit = HitObject::TraceRay( ... );
+// NOTE: Although ReorderThread is not explicitly invoked here,
+// reordering can still occur at any reorder point based on
+// driver-specific decisions.
 RayDesc shadowRay = SampleShadow( hit );
 payload.shadowTerm = HitObject::TraceRay( shadowRay ).IsHit() ? 0.0f : 1.0f;
 HitObject::Invoke( hit, payload );
@@ -1086,6 +1108,9 @@ HitObject::Invoke( hit, payload );
 
 ### Example: Unified shading
 
+No hit shader is invoked in this example; however, reordering can still
+improve data coherence.
+
 ```C++
 hit = HitObject::TraceRay( ... );
 
@@ -1096,8 +1121,12 @@ ReorderThread( hit );
 
 ### Example: Coherently break render loop on miss
 
-Rationale: Executing the miss shader when not needed is unnecessarily
-inefficient on some architectures.
+Executing the miss shader when not needed is unnecessarily inefficient
+on some architectures. In this example, miss shader execution is skipped.
+
+Note that behavior can vary. Other architectures may have better efficiency
+when `HitObject::TraceRay`, `ReorderThread` and `HitObject::Invoke` are
+called back-to-back (see [Reorder Points](#reorder-points)).
 
 ```C++
 for( ;; )
@@ -1114,6 +1143,10 @@ for( ;; )
 
 ### Example: Two-step shading, single reorder
 
+In this example, shading is performed in two steps.
+The first step gathers material information, while the second step dispatches a common surface shader.
+This approach can help reduce shader permutations.
+
 ```C++
 hit = HitObject::TraceRay( ... );
 ReorderThread( hit );
@@ -1125,12 +1158,22 @@ HitObject::Invoke( hit, payload );
 // Alter the hit object to point to a unified surface shader.
 hit.SetShaderTableIndex( payload.surfaceShaderIdx );
 
-// Invoke unified surface shading. We are already hit-coherent so not worth
-// reordering again.
+// Invoke unified surface shading. We are already hit-coherent so it is not
+// worth explicitly reordering again.
+// Reordering may still occur at any reorder point based on driver-specific
+// decisions.
 HitObject::Invoke( hit, payload );
 ```
 
 ### Example: Live state optimization
+
+In this example, logic is added to compress and uncompress part of the
+payload across `ReorderThread`.
+This can make sense if live state is more expensive across `ReorderThread`.
+
+Some implementations may favor cases where `HitObject::TraceRay`, `ReorderThread`
+and `HitObject::Invoke` are called back-to-back (see [Reorder Points](#reorder-points)),
+so performance profiling is necessary.
 
 ```C++
 hit = HitObject::TraceRay( ... );
@@ -1139,6 +1182,21 @@ uint compressedNormal = CompressNormal( payload.normal );
 ReorderThread( hit );
 payload.normal = UncompressNormal( compressedNormal );
 
+HitObject::Invoke( hit, payload );
+```
+
+### Example: Back-to-back calls
+
+This example demonstrates the back-to-back arrangement of `HitObject::TraceRay`,
+`ReorderThread`, and `HitObject::Invoke`.
+For some architectures, this arrangement is the most efficient, as it can be
+recognized as a single reorder point, reducing call overhead
+(see [Reorder Points](#reorder-points)).
+Additional logic between these calls should only be added when necessary.
+
+```C++
+hit = HitObject::TraceRay( ... );
+ReorderThread( hit );
 HitObject::Invoke( hit, payload );
 ```
 
@@ -1178,7 +1236,7 @@ XXX + 5  | HitObject_Invoke | Represents the invocation of the CH/MS shader repr
 XXX + 6  | ReorderThread | Reorders the current thread. Optionally accepts a `HitObject` arg, or `undef`.
 XXX + 7  | HitObject_IsMiss | Returns `true` if the `HitObject` represents a miss.
 XXX + 8  | HitObject_IsHit | Returns `true` if the `HitObject` represents a hit.
-XXX + 9  | HitObject_IsNop | Returns `true` if the `HitObject` is a Nop-HitObject.
+XXX + 9  | HitObject_IsNop | Returns `true` if the `HitObject` is a NOP-HitObject.
 XXX + 10 | HitObject_RayFlags | Returns the ray flags set in the HitObject.
 XXX + 11 | HitObject_RayTMin | Returns the TMin value set in the HitObject.
 XXX + 12 | HitObject_RayTCurrent | Returns the current T value set in the HitObject.
@@ -1410,7 +1468,7 @@ Matrix getters use the `hitobject_StateMatrix` dxil intrinsic and the return typ
 :------------         |:----------- |:--------- |:-----------
  HitObject_IsMiss            | `i1`    | scalar | Returns `true` if the `HitObject` represents a miss.
  HitObject_IsHit             | `i1`    | scalar | Returns `true` if the `HitObject` represents a hit.
- HitObject_IsNop             | `i1`    | scalar | Returns `true` if the `HitObject` is a Nop-HitObject.
+ HitObject_IsNop             | `i1`    | scalar | Returns `true` if the `HitObject` is a NOP-HitObject.
  HitObject_RayFlags          | `i32`   | scalar | Returns the ray flags set in the HitObject.
  HitObject_RayTMin           | `float` | scalar | Returns the TMin value set in the HitObject.
  HitObject_RayTCurrent       | `float` | scalar | Returns the current T value set in the HitObject.
--- a/proposals/0025-shader-execution-reordering.md
+++ b/proposals/0025-shader-execution-reordering.md
@@ -987,21 +987,28 @@ removed from the wave, again observing `A <= B`.
 
 Due to the existence of non-coherent caches on most modern GPUs, an
 application must take special care when communicating information across
-reorder points through memory (UAVs).
+reorder points through memory (UAVs). This proposal introduces a new
+coherence scope for communication between reorder points within the same
+dispatch index. The following additions are made:
+- A new `[reordercoherent]` storage class for UAVs.
+- A new `REORDER_SCOPE = 0x8` member in the `SEMANTIC_FLAG` enum for use in barriers.
 
-This proposal includes a new coherence scope for reorder points. The scope
-is limited to communication within the same dispatch index. Specifically,
-if UAV stores or atomics are performed on one side of a reorder point, and
-on the other side the data is read via non-atomic UAV reads, the following
-is required:
-- The UAV must be declared `[reordercoherent]`.
-- The UAV writer must issue a `Barrier(UAV_MEMORY, REORDER_SCOPE)` between
-the write and the reorder point.
+Specifically, if UAV stores or atomics are performed on one side of a
+reorder point, and on the other side the data is read via non-atomic
+UAV reads, the following steps are required:
+1. The UAV must be declared `[reordercoherent]`.
+2. The UAV writer must issue a `Barrier(UAV_MEMORY, REORDER_SCOPE)` between the write and the reorder point.
 
-When communicating both between threads and across reorder points, global coherency
-can be utilized:
-- The UAV must be declared `[globallycoherent]`.
-- The UAV writer must issue a `DeviceMemoryBarrier` between the write and the
+Note that these steps are required to ensure coherence across any reorder point.
+For example, between a write performed before `ReorderThread` or `TraceRay` and a
+subsequent read in the same shader, or between shader stages (such as data written
+in the closesthit shader and read in the raygeneration shader).
+
+When communicating both between threads with different dispatch index and
+across reorder points the reorder coherence scope is insufficient.
+Instead, global coherency can be utilized as follows:
+1. The UAV must be declared `[globallycoherent]`.
+2. The UAV writer must issue a `DeviceMemoryBarrier` between the write and the
 reorder point.
 
 ## Separation of ReorderThread and HitObject::Invoke
--- a/proposals/NNNN-shader-execution-reordering.md
+++ b/proposals/NNNN-shader-execution-reordering.md
@@ -207,7 +207,7 @@ Parameter                           | Definition
 ---------                           | ----------
 `Return: HitObject` | The `HitObject` that contains the result of the initialization operation.
 `RayQuery Query` | RayQuery from which the hit is created.
-`attr_t CommittedCustomAttribs` | See `ReportHit` for definition. If a closesthit shader is invoked from this `HitObject`, `attr_t` must match the attribute type of the closesthit shader. The size of `attr_t` must not exceed `MaxAttributeSizeInBytes` specified in the `D3D12_RAYTRACING_SHADER_CONFIG`.
+`attr_t CommittedCustomAttribs` | See the `Attributes` parameter of `ReportHit` for definition. If a closesthit shader is invoked from this `HitObject`, `attr_t` must match the attribute type of the closesthit shader.
 
 ---
 
@@ -257,8 +257,8 @@ Parameter                           | Definition
 Execute closesthit or miss shading for the hit or miss encapsulated in a
 `HitObject`. The `HitObject` fully determines the system values accessible
 from the closesthit or miss shader.
-A NOP-HitObject, or a `HitObject` constructed from a `RayQuery` without
-setting a shader table index, will not invoke a shader and is effectively
+A NOP-HitObject or a `HitObject` constructed from a `RayQuery` without
+setting a shader table index will not invoke a shader and is effectively
 ignored.
 
 The `HitObject::Invoke` call counts towards the trace recursion depth. It must
@@ -278,7 +278,7 @@ static void HitObject::Invoke(
 
 Parameter                           | Definition
 ---------                           | ----------
-`HitObject Hit` | `HitObject` that encapsulates information about the closesthit or miss shader to be executed. If the `HitObject` is a NOP-HitObject, or a `HitObject` constructed from a `RayQuery` without setting a shader table index, then neither a closeshit nor a miss shader will be executed.
+`HitObject Hit` | `HitObject` that encapsulates information about the closesthit or miss shader to be executed. If the `HitObject` is a NOP-HitObject or a `HitObject` constructed from a `RayQuery` without setting a shader table index, then neither a closeshit nor a miss shader will be executed.
 `payload_t Payload` | The ray payload. `payload_t` must match the type expected by the shader being invoked. See [Interaction With Payload Access Qualifiers](#interaction-with-payload-access-qualifiers) for details.
 
 > It is legal to call `HitObject::TraceRay` with a different payload type than
@@ -624,7 +624,7 @@ that information to augment reordering decisions before invoking a hit shader.
 
 `RootConstantOffsetInBytes` must be a multiple of 4.
 
-If the `HitObject` is a NOP-HitObject, or a `HitObject` constructed from a
+If the `HitObject` is a NOP-HitObject or a `HitObject` constructed from a
 `RayQuery` without setting a shader table index, the return value is zero.
 
 ---
--- a/proposals/NNNN-shader-execution-reordering.md
+++ b/proposals/NNNN-shader-execution-reordering.md
@@ -541,7 +541,7 @@ Returns 0 if the `HitObject` does not encode a hit.
 uint HitObject::GetHitKind();
 ```
 
-Returns the hit kind of a hit.
+Returns the hit kind of a hit. See `HitKind` for definition of possible values.
 
 Returns 0 if the `HitObject` does not encode a hit.
 
--- a/proposals/NNNN-shader-execution-reordering.md
+++ b/proposals/NNNN-shader-execution-reordering.md
@@ -644,13 +644,17 @@ caller -> anyhit -> (closesthit|miss) -> caller
 
 With `HitObject`, control is returned to the caller after
 `HitObject::TraceRay` completed and hence caller is inserted between anyhit
-and (closesthit|miss):
+and (closesthit|miss).
+
+The flow for `HitObject::TraceRay` is:
 
 ```C++
 caller -> anyhit -> caller
            ^  |
            |__|
 ```
+
+The flow for `HitObject::Invoke` is:
 
 ```C++
 caller -> (closesthit|miss) -> caller
--- a/proposals/NNNN-shader-execution-reordering.md
+++ b/proposals/NNNN-shader-execution-reordering.md
@@ -10,7 +10,7 @@ Michael Haidl, Simon Moll, Martin Stich
 
 ## Introduction
 
-This proposal introduces `ReorderThread` for raygeneration shaders to
+This proposal introduces `ReorderThread`, a builtin function for raygeneration shaders to
 explicitly specify where and how shader execution coherence can be improved.
 Separately, `HitObject` is introduced to decouple traversal, intersection
 testing and anyhit shading from closesthit and miss shading. This separation
@@ -23,13 +23,13 @@ Many raytracing workloads suffer from divergent shader execution and divergent
 data access because of their stochastic nature. Improving coherence with
 high-level application-side logic has many drawbacks, both in terms of
 achievable performance (compared to a hardware-assisted implementation) and
-developer effort. The existing DXR API allows implementations to dynamically
+developer effort. The DXR API already allows implementations to dynamically
 schedule shading work triggered by `TraceRay` and `CallShader`, but does not
-offer a way for the application to control scheduling in any way.
+offer a way for the application to control that scheduling in any way.
 
 Furthermore, the current fused nature of `TraceRay` with its combined
 execution of traversal, intersection testing, anyhit shading and closesthit or
-miss shading imposes various restrictions on the programming model that again
+miss shading imposes various restrictions on the programming model that, again,
 can increase the amount of developer effort and decrease performance. One
 aspect is that common code, e.g., vertex fetch and interpolation, must be
 duplicated in all closesthit shaders. This can cause more code to be generated
@@ -40,7 +40,7 @@ which must be transferred back to the caller through the payload.
 
 ## Proposed Solution
 
-Shader Execution Reordering (SER) introduces a new HLSL primitive,
+Shader Execution Reordering (SER) introduces a new HLSL built-in intrinsic,
 `ReorderThread`,
 that enables application-controlled reordering of work across the GPU for
 improved execution and data coherence.
@@ -67,7 +67,7 @@ information about the hit, such as the distance to the closest hit. Finally,
 
 The proposed extension to HLSL should be relatively straightforward to adopt by
 current DXR implementations: `HitObject` merely decouples existing `TraceRay`
-functionality into two distinct stages; the traversal stage and the shading
+functionality into two distinct stages: the traversal stage and the shading
 stage.
 For SER's `ReorderThread`, the minimal allowed implementation is simply a
 no-op, while implementations that already employ more sophisticated scheduling
@@ -189,7 +189,7 @@ Parameter                           | Definition
 Behaves like the overload of `HitObject::FromRayQuery` that takes only a
 `RayQuery` as argument, with custom attributes associated with
 COMMITTED_PROCEDURAL_PRIMITIVE_HIT. It is ok to always use this overload, even
-for COMMITTED_TRIANGLE_HIT. For anything other than a procedural hit the
+for COMMITTED_TRIANGLE_HIT. For anything other than a procedural hit, the
 specified attributes are ignored.
 
 ```C++
@@ -209,7 +209,7 @@ Parameter                           | Definition
 
 #### HitObject::MakeMiss
 
-Construct a `HitObject` representing a miss, without tracing a ray. It is
+Construct a `HitObject` representing a miss without tracing a ray. It is
 legal to construct a `HitObject` that differs from the result that would be
 obtained by passing the same parameters to `HitObject::TraceRay`, e.g.,
 constructing a miss for a ray that would have hit some geometry if it were
@@ -226,7 +226,7 @@ Parameter                           | Definition
 ---------                           | ----------
 `Return: HitObject` | The `HitObject` that contains the result of the initialization operation.
 `uint RayFlags` | Valid combination of Ray flags as specified by `TraceRay`. Only defined ray flags are propagated by the system.
-`uint MissShaderIndex` | The miss shader index, used to calculate the address of the shader table record. The miss shader index must reference a valid shader table record. Only the bottom 16 bits of this value are used.
+`uint MissShaderIndex` | The miss shader index, used to calculate the address of the shader table record. The miss shader index must reference a valid shader table record. Only the least significant 16 bits of this value are used.
 
 ---
 
@@ -595,7 +595,7 @@ HitGroupRecordAddress =
     HitGroupRecordIndex; // from shader: HitObject::SetShaderTableIndex(RecordIndex)
 ```
 
-> If the `HitObject` encodes a miss, the index relates to the miss table and only the bottom 16 bits are used:
+> If the `HitObject` encodes a miss, the index relates to the miss table and only the least significant 16 bits are used:
 
 ```C++
 MissRecordAddress =
--- a/proposals/NNNN-shader-execution-reordering.md
+++ b/proposals/NNNN-shader-execution-reordering.md
@@ -209,6 +209,8 @@ Parameter                           | Definition
 `RayQuery Query` | RayQuery from which the hit is created.
 `attr_t CommittedCustomAttribs` | See the `Attributes` parameter of `ReportHit` for definition. If a closesthit shader is invoked from this `HitObject`, `attr_t` must match the attribute type of the closesthit shader.
 
+The size of `attr_t` must not exceed `MaxAttributeSizeInBytes` specified in the `D3D12_RAYTRACING_SHADER_CONFIG`.
+
 ---
 
 #### HitObject::MakeMiss
--- a/proposals/NNNN-shader-execution-reordering.md
+++ b/proposals/NNNN-shader-execution-reordering.md
@@ -95,7 +95,7 @@ shading is then invoked in a more coherent context.
 
 Note that this is a very basic example. Among other things, it is possible to
 query information about the hit to influence `ReorderThread` with additional
-hints. See [Separation of ReorderThread and HitObject::Invoke](#separation-of-reorderthread-and-hitobjectinvoke) 
+hints. See [Separation of ReorderThread and HitObject::Invoke](#separation-of-reorderthread-and-hitobjectinvoke)
 for more elaborate examples.
 
 ### HitObject HLSL Additions
@@ -134,20 +134,20 @@ may call `TraceRay`).
 Executes traversal (including anyhit and intersection shaders) and returns the
 resulting hit or miss information as a `HitObject`. Unlike `TraceRay`, this
 does not execute closesthit or miss shaders. The resulting payload is not part
-of the `HitObject`. See 
+of the `HitObject`. See
 [Interaction with Payload Access Qualifiers](#interaction-with-payload-access-qualifiers)
 for details.
 
 ```C++
-template<payload_t> 
-static HitObject HitObject::TraceRay( 
-    RaytracingAccelerationStructure AccelerationStructure, 
-    uint RayFlags, 
-    uint InstanceInclusionMask, 
-    uint RayContributionToHitGroupIndex, 
-    uint MultiplierForGeometryContributionToHitGroupIndex, 
-    uint MissShaderIndex, 
-    RayDesc Ray, 
+template<payload_t>
+static HitObject HitObject::TraceRay(
+    RaytracingAccelerationStructure AccelerationStructure,
+    uint RayFlags,
+    uint InstanceInclusionMask,
+    uint RayContributionToHitGroupIndex,
+    uint MultiplierForGeometryContributionToHitGroupIndex,
+    uint MissShaderIndex,
+    RayDesc Ray,
     inout payload_t Payload);
 ```
 
@@ -173,7 +173,7 @@ separately, which in turn allows invoking a shader. In case of a procedural
 primitive, attributes are to be specified with an overload.
 
 ```C++
-static HitObject HitObject::FromRayQuery( 
+static HitObject HitObject::FromRayQuery(
     RayQuery Query);
 ```
 
@@ -193,8 +193,8 @@ for COMMITTED_TRIANGLE_HIT. For anything other than a procedural hit the
 specified attributes are ignored.
 
 ```C++
-template<attr_t> 
-static HitObject HitObject::FromRayQuery( 
+template<attr_t>
+static HitObject HitObject::FromRayQuery(
     RayQuery Query,
     attr_t CommittedCustomAttribs);
 ```
@@ -266,9 +266,9 @@ This function introduces [Reorder Points](#reorder-points), also in cases
 where no shader is invoked.
 
 ```C++
-template<payload_t> 
-static void HitObject::Invoke( 
-    HitObject Hit, 
+template<payload_t>
+static void HitObject::Invoke(
+    HitObject Hit,
     inout payload_t Payload);
 ```
 
@@ -584,13 +584,12 @@ Sets the index used for shader table lookups.
 
 Parameter                           | Definition
 ---------                           | ----------
-`uint RecordIndex` | The index into the hit or miss shader table. If the
-`HitObject` is a NOP-HitObject, the value is ignored.
+`uint RecordIndex` | The index into the hit or miss shader table. If the `HitObject` is a NOP-HitObject, the value is ignored.
 
 > If the `HitObject` encodes a hit, the index relates to the hit group table and only the bottom 28 bits are used:
 
 ```C++
-HitGroupRecordAddress = 
+HitGroupRecordAddress =
     D3D12_DISPATCH_RAYS_DESC.HitGroupTable.StartAddress + // from: DispatchRays()
     D3D12_DISPATCH_RAYS_DESC.HitGroupTable.StrideInBytes * // from: DispatchRays()
     HitGroupRecordIndex; // from shader: HitObject::SetShaderTableIndex(RecordIndex)
@@ -683,7 +682,7 @@ The following example shows a common pattern of combining `HitObject` and
 ```C++
 // Trace a ray without invoking closesthit/miss shading.
 HitObject hit = HitObject::TraceRay( ... );
- 
+
 // Reorder by hit point to increase coherence of subsequent shading.
 ReorderThread( hit );
 
@@ -718,8 +717,7 @@ void ReorderThread( HitObject Hit );
 
 Parameter                           | Definition
 ---------                           | ----------
-`HitObject Hit` | `HitObject` that encapsulates the hit or miss according to
-which reordering should be performed.
+`HitObject Hit` | `HitObject` that encapsulates the hit or miss according to which reordering should be performed.
 
 ---
 
@@ -739,14 +737,8 @@ void ReorderThread( uint CoherenceHint, uint NumCoherenceHintBitsFromLSB );
 
 Parameter                           | Definition
 ---------                           | ----------
-`uint CoherenceHint` | User-defined value that determines the desired
-ordering of a thread relative to others.
-`uint NumCoherenceHintBitsFromLSB` | Indicates how many of the least
-significant bits in `CoherenceHint` the implementation should try to take
-into account. Applications should set this to the lowest value required to
-represent all possible values of `CoherenceHint` (at the given
-`ReorderThread` call site). All threads should provide the same value at a
-given call site to achieve best performance.
+`uint CoherenceHint` | User-defined value that determines the desired ordering of a thread relative to others.
+`uint NumCoherenceHintBitsFromLSB` | Indicates how many of the least significant bits in `CoherenceHint` the implementation should try to take into account. Applications should set this to the lowest value required to represent all possible values of `CoherenceHint` (at the given `ReorderThread` call site). All threads should provide the same value at a given call site to achieve best performance.
 
 ---
 
@@ -776,22 +768,15 @@ described in
 
 ```C++
 void ReorderThread( HitObject Hit,
-                    uint CoherenceHint, 
+                    uint CoherenceHint,
                     uint NumCoherenceHintBitsFromLSB );
 ```
 
 Parameter                           | Definition
 ---------                           | ----------
-`HitObject Hit` | `HitObject` that encapsulates the hit or miss according to
-which reordering should be performed.
-`uint CoherenceHint` | User-defined value that determines the desired
-ordering of a thread relative to others.
-`uint NumCoherenceHintBitsFromLSB` | Indicates how many of the least
-significant bits in `CoherenceHint` the implementation should try to take
-into account. Applications should set this to the lowest value required to
-represent all possible values of `CoherenceHint` (at the given
-`ReorderThread` call site). All threads should provide the same value at a
-given call site to achieve best performance.
+`HitObject Hit` | `HitObject` that encapsulates the hit or miss according to which reordering should be performed.
+`uint CoherenceHint` | User-defined value that determines the desired ordering of a thread relative to others.
+`uint NumCoherenceHintBitsFromLSB` | Indicates how many of the least significant bits in `CoherenceHint` the implementation should try to take into account. Applications should set this to the lowest value required to represent all possible values of `CoherenceHint` (at the given `ReorderThread` call site). All threads should provide the same value at a given call site to achieve best performance.
 
 ---
 
@@ -814,12 +799,12 @@ for( int bounceCount=0; ; bounceCount++ )
     // that estimate into a coherence hint for reordering. Note that whether
     // this estimate is correct or not only affects performance, not
     // correctness.
-    bool probablyFinished = russianRoulette(albedo) 
+    bool probablyFinished = russianRoulette(albedo)
                             || bounceCount >= maxBounces;
     uint coherenceHints = probablyFinished ? 1 : 0;
 
     // Reorder based on the hit, while taking into account how likely we are to
-    // exit the loop this round. 
+    // exit the loop this round.
     ReorderThread( hit, coherenceHints, 1 );
 
     // Invoke shading for the current hit. Due to the reordering performed
@@ -840,7 +825,7 @@ for( int bounceCount=0; ; bounceCount++ )
 The following pseudocode shows another example of reordering used in the
 context of path tracing with a slightly different loop structure. In this
 case, NOP-HitObject are used in reordering to ensure coherent execution not
-only of shaders, but also of the code after the loop. 
+only of shaders, but also of the code after the loop.
 
 
 ```C++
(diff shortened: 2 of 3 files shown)


damyanp (approved):
Looks ready to take into the repo for further review. Before completing the PR, please pick the next free proposal number (at time of writing this comment, that would be 0021) and rename the file and the reference to it on line 4 appropriately.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+code snippet:
+
+```C++
+int MyFunc(int coherenceCoord)
+{ 
+    int A = WaveActiveBallot(true); 
+    if (WaveIsFirstLane()) 
+        ReorderThread(coherenceCoord, 32); 
+    int B = WaveActiveBallot(true); 
+    return A - B;
```
Flakebi, line 958: There are trailing spaces on these lines :)

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+### Divergence behavior
+
+An implementation may only reorder threads whose control flow arrives at a
+reorder point. Threads that do not arrive at a reorder point are guaranteed to
+not migrate, even if neighboring threads in the same wave do migrate.
+
+From the physical wave's point of view, an implementation may remove, replace,
+or do nothing with threads that arrive at a reorder point. It is also legal
+for an implementation to replace terminated threads in a wave if any of its
+threads arrive at a reorder point. This may indirectly affect threads that
```
Flakebi, line 947: It’s possible to add threads even if no threads terminated before. Maybe we can reformulate that to something like
> It is also legal for an implementation to add threads to a wave if any of its threads arrive at a reorder point.

?

One case where threads could be added added without threads terminating first is for odd dispatch sizes. For example if a wave had 32 threads and a dispatch with 48 threads is started:
```hlsl
int MyFunc(int coherenceCoord)
{
    int A = WaveActiveBallot(true);
    // Wave 0: A = 32
    // Wave 1: A = 16
    if (WaveIsFirstLane())
    {
        ReorderThread(coherenceCoord, 32);
        // Possible reordering:
        // Thread 0 from Wave 0 is sorted into Wave 1 as Thread 16
    }
    int B = WaveActiveBallot(true);
    // Wave 0: B = 31
    // Wave 1: B = 17
    // For Wave 1 it appears like a thread was added
    return A - B;
}
```

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+`HitObject::TraceRay` must not be invoked if the maximum trace recursion depth
+has been reached.
+
+This function introduces [Reorder Points](#reorder-points).
+
+---
+
+#### HitObject::FromRayQuery
+
+Construct a `HitObject` representing the committed hit in a `RayQuery`. It is
```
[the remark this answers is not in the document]
rasmusnv: a little bit later in the same section: "If no hit is committed in the RayQuery, the HitObject returned is the NOP object."

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+Parameter                           | Definition
+---------                           | ----------
+`uint CoherenceHint` | User-defined value that determines the desired
+ordering of a thread relative to others.
+`uint NumCoherenceHintBitsFromLSB` | Indicates how many of the least
+significant bits in `CoherenceHint` the implementation should try to take
+into account. Applications should set this to the lowest value required to
+represent all possible values of `CoherenceHint` (at the given
+`ReorderThread` call site). All threads should provide the same value at a
+given call site to achieve best performance.
```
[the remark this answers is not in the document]
rasmusnv: Thanks. Someone was a little too aggressive with applying a fixed width.

proposals/NNNN-shader-execution-reordering.md
rasmusnv: Then again, it doesn't apply to threads that are inactive from control flow. I think a thread that never started the dispatch could be considered to be terminated.

proposals/NNNN-shader-execution-reordering.md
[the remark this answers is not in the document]
Flakebi: Oh, I somehow missed that on the second read, thanks!

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+Executes traversal (including anyhit and intersection shaders) and returns the
+resulting hit or miss information as a `HitObject`. Unlike `TraceRay`, this
+does not execute closesthit or miss shaders. The resulting payload is not part
+of the `HitObject`. See 
+[Interaction with Payload Access Qualifiers](#interaction-with-payload-access-qualifiers)
+for details.
+
+```C++
+template<payload_t> 
+static HitObject HitObject::TraceRay(
```
rdrabins, line 143: We are interested in adding explicit communication from application in terms of what it wants (or needs) from the `TraceRay` call regarding reordering. I would assume we have to pass extra values to the call through extra args or through call attributes/decorations (similar to `[unroll]`) instead of relying on implicit behavior (which differs between vendors - some do implicitly sort/reorder, some don't).

This could be implemented using three possible values (using attribute form of expressing it in examples):

1. `PreferNoReorder` - the default value with the behavior according to the current spirit. Application doesn't need the reordering happen but the call still is a reorder point and implementation can decide to reorder. Note: This could be also simply spelled out in the spec to be the default if `Reorder` calls are present.
2. `Reorder` - with this value, it is equivalent to:
   ```c++
   HitObject hit = [PreferNoReorder]HitObject::TraceRay( ... );
   ReorderThread( hit ); 
   ```
   the value should be limited to shader stages supporting `ReorderThread(...)`.
3. `NoReorder` - the value means that implementation is required to _not_ reorder. This may be necessary for algorithm correctness in some cases (like, a wave remaining the same before and after the `TraceRay` call). Note: There should be a note mentioning that this _can_ degrade performance and thus should be only used for correctness, otherwise see `PreferNoReorder`.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+static HitObject HitObject::MakeNop();
+```
+
+Parameter                           | Definition
+---------                           | ----------
+`Return: HitObject` | The `HitObject` that contains the result of the initialization operation.
+
+---
+
+#### HitObject::Invoke
```
rdrabins, line 251: Like in comment for `TraceRay`, we are interested in adding more communication about reordering.
We'd consider set of values:

1. PreferNoReorder - (same as for `TraceRay`) the default value with the behavior according to the current spirit. Application doesn't need the reordering happen but the call still is a reorder point and implementation can decide to reorder. Note: This could be also simply spelled out in the spec to be the default if Reorder calls are present.
2. Reorder - with this value, the `[Reorder]HitObject::Invoke( hit, payload )` call would be equivalent to:
   ```c
   ReorderThread( hit ); 
   HitObject::Invoke( hit, payload );
   ```
the value should be limited to shader stages supporting ReorderThread(...).

2. `Reorder(hint,bits)` - calling `[Reorder(hint,bits)]HitObject::Invoke( hit, payload )` would be equivalent to:
   ```c
   ReorderThread( hit, hint, bits); 
   [PreferNoReorder]HitObject::Invoke( hit, payload );
   ```
   Like above the value should be limited to shader stages supporting ReorderThread(...).

We don't propose `NoReorder` here. In case of traceRay, implementation already has to have the inlined raytracing in place so they can probably use it for implementing that. 

The shader invocation depending on implementation may necessary involve reordering as some kind of function call implementation byproduct. 
Such implementation has to deal with difficult choices in compilation. e.g.:
```c++
   ReorderThread(hit, hint, bits);
   color = tex.Sample(sampler, UV);
   HitObject::Invoke( hit, payload );
```
- Is sampler benefiting the reordering before it or not?
- Should I reorder twice (at reorder and at invoke) or nop at `ReorderThread` and reorder according to hint while Invoking the shader?

when separation of `ReorderThread` isn't important part of algorithm, It's easier if application writes it like that:
```c++
   color = tex.Sample(sampler, UV);
   [Reorder(hint, bits)]HitObject::Invoke( hit, payload );
```

proposals/NNNN-shader-execution-reordering.md
Flakebi: I see your point. I think a clearer formulation would be nice. Then it gives less room for misunderstandings.

Without your explanation, I understood it in a way that no new threads can become active (as they can only be removed or replaced), so for a thread that *does not reorder*, I could always `assert(WaveActiveCountBits(true) <= activeThreadsAtShaderStart)`, but that would have been incorrect.

proposals/NNNN-shader-execution-reordering.md
rasmusnv: The proposed [Reorder] is equivalent to TraceRay followed by ReorderThread. We believe this could introduce confusion to an already complex API, especially regarding when and where reordering occurs. Instead, the implementation can reasonably perform straightforward analysis to recognize sequences like TraceRay/ReorderThread/Invoke, reducing the need for additional API complexity.

Regarding [NoReorder] we agree with @Flakebi about the feasibility of such an option. While RayQuery can be implemented like this (as you mention in a separate comment) it is not certain that any implementation can support calling anyhit and intersection shaders while not modifying active threads.

We can explicitly make HitObject::TraceRay and Invoke not reorder. Then the app can control reordering by adding or omitting ReorderThreads. If the app wants the implementation to decide, legacy TraceRay can be used. Note that this does not actually guarantee anything about active threads before/after the call. It is more about the trade-off between overhead and coherence.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+reorder points through memory (UAVs).
+
+Specifically, if UAV stores or atomics are performed on one side of a reorder
+point, and on the other side the data is read via non-atomic UAV reads, the
+following is required:
+- The UAV must be declared `[globallycoherent]`.
+- The UAV writer must issue a `DeviceMemoryBarrier` between the write and the
+reorder point.
+
+## Separation of ReorderThread and HitObject::Invoke
```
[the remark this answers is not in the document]
rasmusnv: TraceRayReorderInvoke is similar in spirit to the attributes.

The first point is not clear to us. It is already the case that the whole wave doesn't have to wait for all rays to finish at any reorder point.

We are not quite sure why the compiler is not able to move the Sample call after ReorderThread. There are no side effects in ReorderThread and TraceRay was already a reorder point, meaning that restrictions on memory visibility apply already. A sampled texture is also read-only and is not affected by side-effects.

Our stance is that it should be easy for an implementation to merge 2-3 calls into one if it is deemed beneficial, without adding further API surface. Any shader code in between the calls is typically not affected by any side effects that may occur in TraceRay or Invoke. The only exception would be RW UAVs that are globally coherent and are accessed in between the calls. However, it is always possible to move code across ReorderThread since it does not cause or observe side effects.

proposals/NNNN-shader-execution-reordering.md
rasmusnv: Again, we do not see a real need for this. If attributes are present at these calls, there is a risk that the user would simply specify [Reorder] on both TraceRay and Invoke, which does not actually mean what we want.

For questions like if Sample benefits from a particular location w.r.t. reordering, the assumption should be that it has an impact and that the application performs the access in the coherence scope where it makes the most sense. However, it does not necessarily hinder fusion. If it is without dependencies, it can still be moved before or after the fused call.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+    if (WaveIsFirstLane()) 
+        ReorderThread(coherenceCoord, 32); 
+    int B = WaveActiveBallot(true); 
+    return A - B; 
+}
+```
+
+In this example, a number of different things could happen:
+- If the implementation does not honor `ReorderThread` at all, the function
+will most likely return zero, as the set of threads before and after the
```
pow2clk, line 964: Again I'm seeing these uncertain phrases as obstacles to testing. Can we declare that it _will_ return zero if it is ignored?

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+conditional reorder would be the same. 
+- If the implementation reorders threads invoking `ReorderThread` but does not
+replace them, B will likely be less than A for threads not invoking
+`ReorderThread`, while the reordered threads will likely resume execution with
+a newly formed full wave, thereby obtaining `A <= B`.
+- If the implementation replaces threads in a wave, the threads not
+participating in the reorder may possibly be joined by more threads than were
+removed from the wave, again observing `A <= B`.
```
pow2clk, line 972: What determines these different possible outcomes? Is it implementation or situation dependent or both?

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+
+```DXIL
+%dx.types.HitObject = type { i8 * }
+```
+
+Central to `dx.types.HitObject` is its value semantics;
+an assignment is a copy.
+As the size of `dx.types.HitObject` is unknown at the DXIL level, the driver
+compiler will employ type replacement early in the lowering process. This is
+analogous to how `dx.types.Handle` is lowered. DXC will guarantee that the
```
pow2clk, line 1156: Under the circumstances, this document will have more longevity if we refer to it as the "offline compiler" or "Dxil generating compiler" rather than DXC.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+type is
+trivially replaceable, e.g., prevent SROA from splitting up the type.
+
+All intrinsics take `dx.types.HitObject` by value. Any intrinsic that produces
+a new `HitObject` will return a new `dx.types.HitObject` value.
+Intrinsics that modify a `HitObject` take the `dx.types.HitObject` by value and return a modified `dx.types.HitObject` value.
```
pow2clk, line 1162: That's just `dx.op.hitObject_SetShaderTableIndex` at present I think?

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+    %dx.types.RayQueryHandle)      ; ray query
+    nounwind argmemonly
+
+```
+This is used for the HLSL overload of `HitObject::FromRayQuery` that only takes `RayQuery`.
+
+#### HitObject_FromRayQueryWithAttrs
+
+```DXIL
+declare %dx.types.HitObject @dx.op.hitObject_FromRayQuery.AttrT(
```
pow2clk, line 1236: This seems meant to match `HitObject_FromRayQueryWithAttrs` in the table above? The names don't match.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+XXX + 20 | HitObject_WorldToObject4x3 | Returns the world to object space transformation matrix in 4x3 form.
+XXX + 21 | HitObject_GeometryIndex | Returns the geometry index committed on hit.
+XXX + 22 | HitObject_InstanceIndex | Returns the instance index committed on hit.
+XXX + 23 | HitObject_InstanceID | Returns the instance id committed on hit.
+XXX + 24 | HitObject_PrimitiveIndex | Returns the primitive index committed on hit.
+XXX + 25 | HitObject_HitKind | Returns the HitKind of the hit.
+XXX + 26 | HitObject_ShaderTableIndex | Returns the shader table index set for this HitObject.
+XXX + 27 | HitObject_SetShaderTableIndex | Returns a HitObject with updated shader table index.
+XXX + 28 | HitObject_LoadLocalRootTableConstant | Returns the root table constant for this HitObject and offset.
+XXX + 29 | HitObject_Attributes | Returns the attributes set for this HitObject.
```
pow2clk, line 1197: Names are all capitalized here, but lowercase below. I think they should be lowercase.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+ HitObject_ObjectToWorld3x4  | `float` | matrix | Returns the object to world space transformation matrix in 3x4 form.
+ HitObject_ObjectToWorld4x3  | `float` | matrix | Returns the object to world space transformation matrix in 4x3 form.
+ HitObject_WorldToObject3x4  | `float` | matrix | Returns the world to object space transformation matrix in 3x4 form.
+ HitObject_WorldToObject4x3  | `float` | matrix | Returns the world to object space transformation matrix in 4x3 form.
+ HitObject_GeometryIndex     | `i32`   | scalar | Returns the geometry index committed on hit.
+ HitObject_InstanceIndex     | `i32`   | scalar | Returns the instance index committed on hit.
+ HitObject_InstanceID        | `i32`   | scalar | Returns the instance id committed on hit.
+ HitObject_PrimitiveIndex    | `i32`   | scalar | Returns the primitive index committed on hit.
+ HitObject_HitKind           | `i32`   | scalar | Returns the HitKind of the hit.
+ HitObject_ShaderTableIndex  | `i32`   | scalar | Returns the shader table index set for this HitObject.
```
pow2clk, line 1325: as above, I think HitObject should be hitObject.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+    i32)                      ; num coherence hint bits from LSB
+    nounwind
+```
+
+#### Generic State Value getters
+
+State value getters return scalar, vector or matrix values dependent on the provided opcode.
+Scalar getters use the `hitobject_StateScalar` dxil intrinsic and the return type is the state value type.
+Vector getters use the `hitobject_StateVector` dxil intrinsic and the return type is the vector element type.
+Matrix getters use the `hitobject_StateMatrix` dxil intrinsic and the return type is the matrix element type.
```
pow2clk, line 1302: I think it might be clearer to leave these out of the table above and give them their own special section. It already looks like a new topic because of the table. I think dropping them to the bottom instead of in the middle of the list of intrinsics would be less confusing.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+declare void @dx.op.hitObject_Attributes.AttrT(
+    i32,                           ; opcode
+    %dx.types.HitObject,           ; hit object
+    AttrT*)                        ; attributes
+    nounwind argmemonly
+```
+
+`AttrT` is the user-defined intersection attribute struct type. See `ReportHit` for definition.
+
+#### HitObject_SetShaderTableIndex
```
pow2clk, line 1378: This is out of order with the table above

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+
+Returns a HitObject with updated shader table index.
+
+```DXIL
+declare %dx.types.HitObject @dx.op.hitObject_SetShaderTableIndex(
+    i32,                           ; opcode
+    %dx.types.HitObject,           ; hit object
+    i32)                           ; record index
+    nounwind readnone
+```
```
pow2clk, line 1388: For all of these, it would be useful to document any parameter restrictions that you'd like the validator to check. For example, perhaps you want ray vector elements to be restricted to between 0.0 and 1.0?

pow2clk (commented):
I'll disclose that I'm not a DXR expert. A lot of my feedback is copy editing, trying to make the document easier to read. I feel a bit stronger about my thoughts on testability and DXIL op feedback.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+Due to the existence of non-coherent caches on most modern GPUs, an
+application must take special care when communicating information across
+reorder points through memory (UAVs).
+
+Specifically, if UAV stores or atomics are performed on one side of a reorder
+point, and on the other side the data is read via non-atomic UAV reads, the
+following is required:
+- The UAV must be declared `[globallycoherent]`.
+- The UAV writer must issue a `DeviceMemoryBarrier` between the write and the
+reorder point.
```
Flakebi, line 987: Requiring global memory coherence can be a large hit to performance. Especially since a driver can decide not to sort if the workload is coherent (or sort in a more local manner).
Using `globallycoherent` and `DeviceMemoryBarrier` makes it impossible for a driver to omit the synchronization work when it’s not needed.

I propose that we add a new UAV attribute and barrier scope that relates the synchronization to reordering. So,
- The UAV must be declared `[reordercoherent]`
- The UAV writer must issue a `Barrier(UAV_MEMORY, THREAD_REORDER_SCOPE)` between the write and the
reorder point.

That way, when a driver decides not to reorder, the barrier can be a no-op.

rasmusnv:
@microsoft-github-policy-service agree company="NVIDIA"

damyanp (changes requested):
It looks like there's a large number of unresolved conversations on this PR - these either need to be moved into issues or addressed in the spec before we can merge it.

proposals/NNNN-shader-execution-reordering.md
rasmusnv: This seems useful at first glance, but expanding the memory model may overly broaden the scope of this proposal. It might be better suited as a separate proposal.

It is also unclear what use case [reordercoherent] is addressing. A UAV marked as [reordercoherent] can only communicate within the same DispatchRaysIndex, across any reordering. This essentially makes it a local scratch pad, which could be more efficiently handled using live state in the shader program (or as payload values if communicated with other shaders).

proposals/NNNN-shader-execution-reordering.md
Anteru: So you're saying that some implementations can support ray query, but can't execute the shaders per lane without a resort? That's an interesting limitation which is currently not expressed in the specification, as there seem to be implied uses where the thread assignment gets _not_ reordered by `TraceRay` even in the presence of any hit shaders being called by `TraceRay`. For example, if I want to keep quads together, you're saying there's no way to do that right now with `TraceRay` as explained in the SER specification?

My assumption is that any implementation support ray-query can maintain the thread assignment.

The complication in the "straightforward analysis" is that tooling may move instructions around between `TraceRay`, `ReorderThread`, `Invoke`, and some implementations may prefer to use the `ReorderThread` into either `TraceRay` _or_ `Invoke`, and then it's undecidable which one to do if there's a `ReorderThread` in the middle of it, even though the implementation may know that re-ordering in `TraceRay` for example will provide no benefit - and it can't communicate this. `ReorderThread` is a fine API if all re-ordering would be limited to just that API call, but the reality is that there are multiple re-order points, and the specification language for SER specifically uses comments to explain where no reordering should happen - which is unfortunately not something the compiler knows about. I.e. in samples like:

```
// Invoke unified surface shading. We are already hit-coherent so not worth
// reordering again.
HitObject::Invoke( hit, payload );
```

proposals/NNNN-shader-execution-reordering.md
Flakebi: > This seems useful at first glance, but expanding the memory model may overly broaden the scope of this proposal. It might be better suited as a separate proposal.

It needs one more enum value in [`SEMANTIC_FLAG`](https://microsoft.github.io/hlsl-specs/proposals/0018-work-graphs.html#barrier) and (optionally? see below) a `[reordercoherent]` marker on UAVs.

Vulkan also has a separate scope `ShaderCallKHR` for this.

Requiring global coherence is surely fine for some implementations but may incur a performance hit in other implementations.

> It is also unclear what use case [reordercoherent] is addressing. A UAV marked as [reordercoherent] can only communicate within the same DispatchRaysIndex, across any reordering.

The marker was meant to connect the barrier to certain UAVs. This allows UAVs that are not explicitly marked to skip synchronization.
I suppose an alternative is to not require marking UAVs but expect all bound UAVs to be synchronized with a reorder-scope barrier.
Although this may be expensive if reordering happens globally. Then this would be equivalent to mark all UAVs as `[globallycoherent]`.

rasmusnv: > The marker was meant to connect the barrier to certain UAVs. This allows UAVs that are not explicitly marked to skip synchronization.

This part I understand. What I would like to understand is the use case for a reorder-scope barrier. Since it can only be used to synchronize within the same DispatchRaysIndex, the data that is written/read to/from the reorder-scope coherent UAV can instead be stored as live state or in payload during intermediate steps and then be written once at the end (if needed). This seems more efficient overall.

proposals/NNNN-shader-execution-reordering.md
rasmusnv: > So you're saying that some implementations can support ray query, but can't execute the shaders per lane without a resort? That's an interesting limitation which is currently not expressed in the specification, as there seem to be implied uses where the thread assignment gets not reordered by TraceRay even in the presence of any hit shaders being called by TraceRay. For example, if I want to keep quads together, you're saying there's no way to do that right now with TraceRay as explained in the SER specification?

TraceRay is a reorder point so there are no guarantees about active threads afterwards. This doesn't necessarily imply that a full reorder will happen, but some amount of changes in active threads may still happen. As mentioned, RayQuery does not invoke shaders, so it is not exactly comparable to TraceRay that itself calls anyhit and intersection shaders.

> The complication in the "straightforward analysis" is that tooling may move instructions around between TraceRay, ReorderThread, Invoke

Could you clarify what the tooling is here? If you mean dxc I believe we can restrict movement of things like resource accesses.

> some implementations may prefer to [f]use the ReorderThread into either TraceRay or Invoke, and then it's undecidable which one to do if there's a ReorderThread in the middle of it, even though the implementation may know that re-ordering in TraceRay for example will provide no benefit - and it can't communicate this

Could you expand on this? If the driver knows that fusing with TraceRay will provide no benefit it seems logical that it should fuse with Invoke.

> ReorderThread is a fine API if all re-ordering would be limited to just that API call, but the reality is that there are multiple re-order points, and the specification language for SER specifically uses comments to explain where no reordering should happen - which is unfortunately not something the compiler knows about. I.e. in samples like:

Since there is no ReorderThread call before the Invoke, it makes sense to assume that no extra effort should be made to reorder the Invoke, beyond what is required for the underlying architecture. If the underlying architecture requires an extra reorder for the Invoke anyway, it may want to utilize information provided in the most recent ReorderThread call.

Anteru: To your starting point: That's exactly the problem. `TraceRay` _can_ be a reorder point, but if you happen to run on an implementation which doesn't, you can write code which relies on that and you can't express it any more. I don't like "implicit" defaults because implicit is platform dependent and thus hurts portability. I'm sure there's already code out there which assumes no reordering happens because there's only one any hit shader present for example and empirically it didn't. As to stories about implicit assumptions, `if (AMD) waveSize=64` was another implicit thing for quite some time :)

The problem is the tooling can move code across and there's no guarantee that running it through SPIR-V opt or not will not move around, and the problem I'm mentioning is not that the _driver_ knows that fusing with `TraceRay` has no benefit; it's only the app who knows this. I.e. in a case like:

```
TraceRay();
SomeCalculationWhichWillBeIncoherentNoMatterWhatReorderingHappensInTraceRay();
ReorderThread();
SomethingExpensiveWhichIsNowCoherent();
Invoke();
SomethingElseWhichRemainsCoherent();
```

The application may know reordering in `TraceRay` provides no possible benefit; but it can't tell the implementation. There are implementations which would _always_ reorder in `TraceRay` and `Invoke`, and _additionally_ in `ReorderThread` if requested by the application. Now you have an unsolvable problem: You have 3 re-order points, one of which is _explicit_ and two are _implicit_, and there's no way the application can tell you which of the implicit ones is "less valuable".

To the last point, "assuming no extra effort" is not something I like to assume, because assumptions _will_ vary between implementations and in the end we'll be just figuring out what the most common assumption was.

(Sorry, a bit out of order here :) )

As to moving around stuff: I'm not aware of DXC restricting moving of instructions around, nor, in the future, SPIR-V tools or custom processing paths. Knowing if `TraceRay` is assumed to be "useless" provides information to the compiler in terms of code motion opportunities - it can move things around freely as it can assume a sample call for example is the same cost before/after `TraceRay`; when it's a reordering `TraceRay`, moving a sample before it will be likely prohibitive if the state needs to get saved/restored.

My main concern is here that we'll end up with multiple implementations with multiple defaults as to when to sort and in which call, and what turns out to be best for an application will be determined by the platform they happen to develop on, and when they move from one implementation which happens to sort a lot in `TraceRay` by default they'll realize "oh `ReorderThread` doesn't seem to help here" and leave it out, then on the next implementation they'll be "Oh I really need `ReorderThread` to make this work", but that was because the second implementation didn't do any measurable reorder, etc. In the end we'll be all app-opting where/when to sort even though the application was written with certain assumptions in mind they simply could not communicate to us other than comments.

rasmusnv: Our proposed solution is to specify that HitObject::TraceRay and HitObject::Invoke will not perform heavyweight reordering. However, they still serve as reorder points, giving the driver flexibility in how execution is scheduled. Shaders that rely on having the same sets of active threads before and after these calls are not valid.

In your example, this means that the first incoherent calculation is assumed to not be worth reordering. The only reordering that may occur is related to the efficiency of the HitObject::TraceRay operation itself, which is left to the driver’s discretion. Shader code that benefits from and receive coherence is, by definition, the code that follows ReorderThread. It is then up to the driver to decide whether it is beneficial to fuse ReorderThread with Invoke, depending on the intervening code. My assumption is that it’s better not to fuse them if there is non-uniform resource access in between, as this makes it easier for the application to predict behavior based on code changes.

Regarding tooling, it is important to ensure that non-uniform resource access remains aligned with ReorderThread. Moving accesses across it will defeat its purpose.

Anteru: That requires us to specify it though: "TraceRay and Invoke will not perform 'heavy-weight' reordering", I assume only in the presence of `ReorderThreads`, and thus behave as-if we always specified the `[NoReorder]` flag on them (well, nearly, because you're saying not heavy-weight, which would have to be defined as well - what does heavyweight mean?) I think it's clear by this point that we're back to the original reason we wanted this expressed clearly and explicitly, to not rely on interpretation of what heavyweight means or not.

Which brings me back to the question I raised - are there implementations which simply _cannot_ guarantee that `TraceRay` maintains a 1:1 mapping? If so, that makes `[NoReorder]` impossible to implement, but I don't see why that should be the case given you can always fall back to inlining all shaders and using `RayQuery`, and most likely applications would only use that in cases where there is no reason to reorder to begin with (i.e. only one any hit shader.)

rasmusnv: > That requires us to specify it though

Yes, we can specify this clearly and it can be phrased in a better way. The important part is what the app conveys to the driver and what the driver is allowed to do in return. The app does not explicitly expect any reordering for HitObject::TraceRay and HitObject::Invoke but the driver is still allowed flexibility for scheduling the work efficiently. The driver can account for things like the number of shaders and how to run them, and whether to fuse ReorderThread with another call (which effectively moves the reorder point).

> Which brings me back to the question I raised - are there implementations which simply cannot guarantee that TraceRay maintains a 1:1 mapping?

Forcing 1:1 mapping is very restrictive and not something we'd be comfortable with. It is certainly not always the case today. It is also unclear why this would be useful given that the app has little control over thread arrangement to begin with.

Inlining is not generally a valid solution as shaders are compiled separately and pipeline creation should not consume excessive time for recompilation. It can also become problematic if there are many call sites that need inlining of the same shader(s).

jasilvanus: Regarding a strict `[NoReorder]`, agreed that we shouldn't allow the app to force a 1:1 thread mapping.

The driver can't always inline everything (consider a pipeline with many shaders and many `TraceRay` calls in nested CHS/Miss, and a large recursion limit, or early compilation of collections), and without inlining such a guarantee would put an unreasonable burden on the implementation.
Note that this issue is not just about the set of active threads immediately before/after a TraceRay, but also about reconvergence with other threads that did not execute the `TraceRay`, see the earlier example by @Flakebi. 

However, a `[PreferNoReorder]` hint on reorder points that tells the driver it shouldn't spend any additonal effort on reordering for coherence would be fine in that regard.

proposals/NNNN-shader-execution-reordering.md
rasmusnv: This has now been clarified.

proposals/NNNN-shader-execution-reordering.md
rasmusnv: The `Reorder points` section has been updated with a more detailed account for expected reordering behavior. Let me know if this aligns with our discussion or if there is still a need for attributes.

proposals/NNNN-shader-execution-reordering.md
Flakebi: Thanks!
(I can’t mark this as resolved myself)

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+* Sponsor: [Tex Riddell](https://github.com/tex3d)
+* Status: **Under Consideration**
+* Impacted Projects: DXC
+
+## Introduction
+
+This proposal introduces `ReorderThread` for raygeneration shaders to
+explicitly specify where and how shader execution coherence can be improved.
+Separately, `HitObject` is introduced to decouple traversal, intersection
+testing and anyhit shading from closesthit and miss shading. This separation
```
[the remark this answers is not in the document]
rasmusnv: Hopefully more clear now.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+schedule shading work triggered by `TraceRay` and `CallShader`, but does not
+offer a way for the application to control scheduling in any way.
+
+Furthermore, the current fused nature of `TraceRay` with its combined
+execution of traversal, intersection testing, anyhit shading and closesthit or
+miss shading imposes various restrictions on the programming model that again
+can increase the amount of developer effort and decrease performance. One
+aspect is that common code, e.g., vertex fetch and interpolation, must be
+duplicated in all closesthit shaders. This can cause more code to be generated
+in a divergent execution environment. Furthermore, `TraceRay`'s nature
```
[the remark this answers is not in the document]
rasmusnv: Yes. Hopefully it's better now.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+## Proposed Solution
+
+Shader Execution Reordering (SER) introduces a new HLSL primitive,
+`ReorderThread`,
+that enables application-controlled reordering of work across the GPU for
+improved execution and data coherence.
+
+Additionally, the introduction of `HitObject` allows separation of traversal,
+anyhit shading and intersection testing from closesthit and miss shading. This
+separation enables the otherwise orthogonal `ReorderThread` to improve
```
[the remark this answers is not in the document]
rasmusnv: Fixed.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+`ReorderThread` and shader table-based closesthit and miss shading to be combined with
+`RayQuery`.
+
+The proposed extension to HLSL should be relatively straightforward to adopt by
+current DXR implementations: `HitObject` merely decouples existing `TraceRay`
+functionality into two distinct stages; the traversal stage and the shading
+stage.
+For SER's `ReorderThread`, the minimal allowed implementation is simply a
+no-op, while implementations that already employ more sophisticated scheduling
+strategies are likely able to reuse existing mechanisms to implement support
```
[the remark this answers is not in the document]
rasmusnv: The "hint" is that it should be possible for existing implementations to take advantage of the new API. Or do you mean something else?

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+```C++
+HitObject Hit = HitObject::TraceRay( ..., Payload );
+ReorderThread( Hit );
+HitObject::Invoke( Hit, Payload );
+```
+
+This snippet traces a ray and stores the result of traversal, intersection
+testing and anyhit shading in `Hit`. The call to `ReorderThread` improves
+coherence based on the information inside the `Hit`. Closesthit or miss
+shading is then invoked in a more coherent context.
```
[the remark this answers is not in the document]
rasmusnv: I don't think it's possible to write a test for coherence after ReorderThread. An implementation could switch to no-op behavior for ReorderThread for any reason, such as the dispatch being small, or if we are close to the end of a dispatch.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+
+Note that this is a very basic example. Among other things, it is possible to
+query information about the hit to influence `ReorderThread` with additional
+hints. See [Separation of ReorderThread and HitObject::Invoke](#separation-of-reorderthread-and-hitobjectinvoke) 
+for more elaborate examples.
+
+### HitObject HLSL Additions
+
+```C++
+HitObject
```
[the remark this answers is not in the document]
rasmusnv: Are you saying you want this to appear earlier in the document?

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+```
+
+The `HitObject` type encapsulates information about a hit or a miss. A
+`HitObject` is constructed using `HitObject::TraceRay`,
+`HitObject::FromRayQuery`, `HitObject::MakeMiss`, or `HitObject::MakeNop`. It
+can be used to invoke closesthit or miss shading using `HitObject::Invoke`,
+and to reorder threads for shading coherence with `ReorderThread`.
+
+The `HitObject` has value semantics, so modifying one `HitObject` will not
```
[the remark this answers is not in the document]
rasmusnv: RayQuery I think would be the main point of confusion, which is a reference.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+    uint MultiplierForGeometryContributionToHitGroupIndex, 
+    uint MissShaderIndex, 
+    RayDesc Ray, 
+    inout payload_t Payload);
+```
+
+See `TraceRay` for parameter definitions.
+
+The returned `HitObject` will always encode a hit or a miss, i.e., it will
+never be a NOP-HitObject.
```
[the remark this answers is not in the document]
rasmusnv: Fixed.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+---------                           | ----------
+`Return: HitObject` | The `HitObject` that contains the result of the initialization operation.
+`uint RayFlags` | Valid combination of Ray flags as specified by `TraceRay`. Only defined ray flags are propagated by the system.
+`uint MissShaderIndex` | The miss shader index, used to calculate the address of the shader table record. The miss shader index must reference a valid shader table record. Only the bottom 16 bits of this value are used.
+
+---
+
+#### HitObject::MakeNop
+
+Construct a NOP-HitObject that represents neither a hit nor a miss. This is
```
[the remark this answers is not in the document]
rasmusnv: Added a summary to the beginning of "HitObject HLSL Additions".

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+
+---
+
+#### HitObject::GetHitKind
+
+```C++
+uint HitObject::GetHitKind();
+```
+
+Returns the hit kind of a hit.
```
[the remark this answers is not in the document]
rasmusnv: Fixed.

proposals/NNNN-shader-execution-reordering.md
rasmusnv: The idea is that the intrinsic name will have the actual attribute type appended instead.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+wants to participate in reordering without executing a closesthit or miss
+shader.
+
+```C++
+static HitObject HitObject::MakeNop();
+```
+
+Parameter                           | Definition
+---------                           | ----------
+`Return: HitObject` | The `HitObject` that contains the result of the initialization operation.
```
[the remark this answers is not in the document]
rasmusnv: Fixed.

proposals/NNNN-shader-execution-reordering.md
rasmusnv: I think it's difficult to implement strict tests for wave behavior regarding reordering. An implementation is not required to remove the calls to ReorderThread, even if the reordering itself is skipped. If it's simply used as a continuation point we could see a difference in the number of lanes active. What is the goal of the testing you have in mind?

proposals/NNNN-shader-execution-reordering.md
rasmusnv: It could both be up to the implementation and situation.

proposals/NNNN-shader-execution-reordering.md
rasmusnv: Fixed.

proposals/NNNN-shader-execution-reordering.md
```diff
@@ -0,0 +1,1388 @@
...
+`RayQuery Query` | RayQuery from which the hit is created.
+
+---
+
+#### HitObject::FromRayQuery with custom attributes
```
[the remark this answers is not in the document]
rasmusnv: Merged the overloads into the same section.

proposals/NNNN-shader-execution-reordering.md
rasmusnv: Moved.

proposals/NNNN-shader-execution-reordering.md
rasmusnv: I moved them last, but I think I disagree about leaving them out of the initial opcode table. That table is supposed to list all opcodes and their values. The later table doesn't have the values.

proposals/NNNN-shader-execution-reordering.md
rasmusnv: Done.

rasmusnv:
I have picked proposal number 0025. All pending changes have been made. If anything is left to discuss or change I suggest we open separate issues after the merge.

damyanp (approved):
> I have picked proposal number 0025. All pending changes have been made. If anything is left to discuss or change I suggest we open separate issues after the merge.

This seems like a good approach to me, it's getting quite hard to keep track of everything in this PR! Thank you to everyone for all the detailed comments and responses!

I'm going to mark this as approved, but I think we'll still need to get all the conversations marked as resolved before the PR can be merged, and this repo is set to require two approvals as well.

For resolving the conversations we should follow up with the original authors and/or file issues (and link them in the conversation before resolving it) if necessary.

proposals/0025-shader-execution-reordering.md
```diff
@@ -0,0 +1,1460 @@
...
+```
+
+The `HitObject` type encapsulates information about a hit or a miss. A
+`HitObject` is constructed using `HitObject::TraceRay`,
+`HitObject::FromRayQuery`, `HitObject::MakeMiss`, or `HitObject::MakeNop`. It
+can be used to invoke closesthit or miss shading using `HitObject::Invoke`,
+and to reorder threads for shading coherence with `ReorderThread`.
+
+The `HitObject` has value semantics, so modifying one `HitObject` will not
+impact any other `HitObject` in the shader. A shader can have any number of
```
jaladreips, line 114: Is there a reason for HitObject objects having value semantics while RayQuery objects have reference semantics?

proposals/0025-shader-execution-reordering.md
```diff
@@ -0,0 +1,1460 @@
...
+active `HitObject`s at a time. Each `HitObject` will take up some hardware
+resources to hold information related to the hit or miss, including ray,
+intersection attributes and information about the shader to invoke. The size
+of a `HitObject` is unspecified. As a consequence, `HitObject` cannot be
+stored in structured buffers. Ray payload structs must not have members of
+`HitObject` type. `HitObject` supports assignment (by-value copy) and can be
+passed as arguments to and returned from local functions.
```
tex3d, line 121: Description of restriction should mention that the type is `intangible`, and specify more locations that are disallowed.

```suggestion
intersection attributes and information about the shader to invoke. The size
of a `HitObject` is unspecified, and thus considered intangible. As a
consequence, `HitObject` cannot be stored in memory buffers, in groupshared
memory, in ray payloads, or in intersection attributes. `HitObject` supports
assignment (by-value copy) and can be passed as arguments to and returned from
local inlined functions.
```

proposals/0025-shader-execution-reordering.md
```diff
@@ -0,0 +1,1460 @@
...
+
+A `HitObject` is default-initialized to encode a NOP-HitObject (see `HitObject::MakeNop`).
+A NOP-HitObject can be used with `HitObject::Invoke` and `ReorderThread` but
+does not call shaders or provide additional information for reordering.
```
tex3d, line 125: > does not call shaders or provide additional information for reordering.

Should it be specified that both operations do nothing (`ReorderThread` causes no reordering)?  Or is it valid for an implementation to do thread reordering here when `HitObject` is a NOP-HitObject?

Flakebi: I read it as that it is valid to do reordering, though no hint or any other info are available. So, likely `ReorderThread` does no reordering. But we don’t want to *guarantee* that it does nothing for the same reason that we don’t want to guarantee reconvergence with a `[NoReorder]` attribute.

proposals/0025-shader-execution-reordering.md
```diff
@@ -0,0 +1,1460 @@
...
+---
+
+#### ReorderThread with coherence hint
+
+This variant of `ReorderThread` reorders threads based on a generic
+user-provided hint. Similarity of hint values should indicate expected
+similarity of subsequent work being performed by threads. The resolution of
+the hint is implementation-specific. If an implementation cannot resolve all
+values of `CoherenceHint`, it is free to ignore an arbitrary number of least
+significant bits. The thread ordering resulting from this call may be
```
rdrabins, line 729: We'd like to have a clearer description of how hit is mapping to coherence.
Could we add the sentence that higher bit of hint has more impact on coherence then lower ones.

e.g. consider 3 hints, and chose the most coherent pair of hint:
`A = 0b100000`
`B = 0b111111`
`C = 0b011111`

For example 3 methods for extracting coherence by defining opposite divergence metric:
1. If higher bit has more coherence then all lower bits after it, `A`,`B` pair is most coherent. Divergence more or less xor(...)
2. If its arithmetic difference then `A` and `C`. Divergence is more or less abs(substract(...)).
3. It could be also divergence = bitcount(xor) that tries to find most similar bits. then B and C is most coherent.

3 doesn't cope well with "it is free to ignore an arbitrary number of least
significant bits", but 1 and 2 fit well into definition.

I'm mentioning that because we discuss this feature only as reordering and forget that based on the ordering, the system needs to extract waves to execute.
Without understanding what is the actual "distance" between the hint codes its hard. It would be good to have that understanding on both sides of GPU.

proposals/0025-shader-execution-reordering.md
```diff
@@ -0,0 +1,1460 @@
...
+payload.giData = GlobalIlluminationCacheLookup( hit );
+
+HitObject::Invoke( hit, payload );
+```
+
+### Example: Same surface shader but different behavior in the raygeneration shader
+
+```C++
+// Primary ray wants perfect shadows and does not need to reorder as it is
+// coherent enough.
```
rdrabins, line 1072: We'd appreciate using language that doesn't suggest that there is no reorder at trace ray.
It should be the expectation that impl should only minimize this. 
// Primary ray wants perfect shadows and does not need to reorder more than required and best performant for given implementation
// The domain of shadow rays is expected to be coherent enough.

proposals/0025-shader-execution-reordering.md
rasmusnv: It is valid to reorder even with a NOP-HitObject. At the very least this makes sense if not all threads in a wave have a NOP-HitObject. Hints can still be used as well.

proposals/0025-shader-execution-reordering.md
rasmusnv: The latest commit includes clarifications like this.

proposals/0025-shader-execution-reordering.md
```diff
@@ -0,0 +1,1519 @@
...
+is required:
+- The UAV must be declared `[reordercoherent]`.
+- The UAV writer must issue a `Barrier(UAV_MEMORY, REORDER_SCOPE)` between
+the write and the reorder point.
+
+When communicating both between threads and across reorder points, global coherency
+can be utilized:
+- The UAV must be declared `[globallycoherent]`.
+- The UAV writer must issue a `DeviceMemoryBarrier` between the write and the
+reorder point.
```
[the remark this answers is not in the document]
rasmusnv: From my point of view, the `DeviceMemoryBarrier` call makes sure that all writes have completed before the reorder point gets to move the work to a different thread. When reading after the reorder point, there's no ordering problem. The only risk is that stale data is read from a local cache. That is mitigated by the `globallycoherent` attribute, causing the local cache to be bypassed.

proposals/0025-shader-execution-reordering.md
rasmusnv: I believe arithmetic difference is the most flexible and intuitive definition for users. This approach allows a hint to incorporate both hierarchical aspects (with more important elements in MSBs) and ranges of bits that represent closeness along some axis. I agree it should be spelled out.

jasilvanus: Agreed that it makes sense to add some details what we mean with a hierarchical hint, and giving an example.

However, I think the numerical value of the XOR makes more sense than arithmetic difference. We could also mention lexicographical sorting (even if the implementation doesn't explicitly sort) which many users should be familiar with. 

Consider:

```
hint1 = 0b1000
hint2 = 0b0111
hint3 = 0b1111
```

If the bits in the hint are intended to be used hierarchically (as per my understanding), then `hint2` is closer to `hint1` than to `hint3`. That is reflected in the numerical value of the XOR: `hint2 ^ hint1 = 0b111`, while `hint2 ^ hint3 = 0b1000`. However, using arithmetic difference, `hint2` would be closer to `hint3`.

proposals/0025-shader-execution-reordering.md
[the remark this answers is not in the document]
jasilvanus: Let's separate the `reordercoherent` and `globallycoherent` discussions. I think `globallycoherent` in DXR semantics are a bit problematic because of the synchronization limitations, but that's not the focus of this change.

For `reordercoherent`, are you saying that you agree that not requiring a barrier between the reorder point and the load *does penalize* some implementations, but we accept that in favour of a simpler API that is closer to what e.g. WG require?

proposals/0025-shader-execution-reordering.md
rasmusnv: XOR prioritizes the strictly hierarchical use case. The problem with XOR is when a bit range represents similarity along some linear axis and we want to group based on closeness. 

Take for example the following X,Y hint pairs along some linear axis of similarity:
```
X  Y  Arith XOR
5  0  5     5
5  1  4     4
5  2  3     7
5  3  2     6
5  4  1     1
5  5  0     0
5  6  1     3
5  7  2     2
5  8  3     13
5  9  4     12
```

In this use case, arithmetic difference produces a desirable metric but not XOR.

proposals/0025-shader-execution-reordering.md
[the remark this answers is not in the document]
rasmusnv: What I'm trying to say is that we are not trying to add acquire and release semantics to HLSL with this spec proposal.

What you are suggesting is calling `Barrier` both before and after the reorder point. First of all, note that you would have to have a barrier after each and every reorder point that wants to read the data after a single write since each reorder point will be a new thread. So a single write can result in a many required barriers, which may be difficult for the user. Second, there is no distinction between acquire and release with `Barrier` which means that all possible implementations are harmed (e.g., ordering + cache flush at both sides). Finally, it is not clear that there is any case where a high-frequency cache flush would be better than adopting the `globallycoherent` semantics of bypassing local caches, which all implementations already supports.

proposals/0025-shader-execution-reordering.md
```diff
@@ -0,0 +1,1519 @@
...
+### Memory coherence and visibility
+
+Due to the existence of non-coherent caches on most modern GPUs, an
+application must take special care when communicating information across
+reorder points through memory (UAVs).
+
+This proposal includes a new coherence scope for reorder points. The scope
+is limited to communication within the same dispatch index. Specifically,
+if UAV stores or atomics are performed on one side of a reorder point, and
+on the other side the data is read via non-atomic UAV reads, the following
```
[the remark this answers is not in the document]
rasmusnv: It is mentioned that a shader transition is another reorder point so the same rules apply. Does that cover it? Do you have a relevant reference for Vulkan?

proposals/0025-shader-execution-reordering.md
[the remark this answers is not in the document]
jasilvanus: > What I'm trying to say is that we are not trying to add acquire and release semantics to HLSL with this spec proposal.

Agreed.

> 
> What you are suggesting is calling `Barrier` both before and after the reorder point. First of all, note that you would have to have a barrier after each and every reorder point that wants to read the data after a single write since each reorder point will be a new thread. 

Nit: Let's not mix the thread terminology (as in the `globallycoherent` discussion) with the reorder terminology. We can say that at reorder points, `reordercoherent` synchronization is necessary. If we want to give it a name, maybe "execution locus" is a better term instead of "thread"?

> So a single write can result in a many required barriers, which may be difficult for the user. 

Agreed.

> Second, there is no distinction between acquire and release with `Barrier` which means that all possible implementations are harmed (e.g., ordering + cache flush at both sides). 

In the worst case, putting a `reordercoherent` barrier immediately after every reorder point (CallShader/TraceRay/ReportHit/HitObject ops / shader entry) should be equivalent to the current proposed behavior without the barrier on the reading side. Because it immediately follows a reorder point, it is clear that only reads can be affected, and if the implementation uses the bypass strategy nothing needs to be done. 
An implementation that doesn't use cache bypass would have to add exactly all these barriers, and the explicit one just makes that explicit.

> Finally, it is not clear that there is any case where a high-frequency cache flush would be better than adopting the `globallycoherent` semantics of bypassing local caches, which all implementations already supports.

Agreed that this is totally unclear in general, but I think when designing a specification that is supposed to be used for an extended lifetime, including on not yet existing hardware, one should be cautious to unnecessarily restrict the freedom of possible implementations.

To be clear, I am not demanding to change it to require the barrier on the reading side, I'm rather pointing out a potential issue. If there is consensus that the single-barrier solution is preferable, I'm totally fine with that.

proposals/0025-shader-execution-reordering.md
jasilvanus: I was interpreting the hints to have hierarchical semantics, but checking the spec again I couldn't find any explicit mentioning of that, so fine with me.

Again, good point that we should explain this.

rdrabins: I view it like in hint one encodes closeness in two metrics:
1. discreet space, any difference on given bits mean you are different, on this and that property. e.g.
   - you will use resource with this and that indices
   - shader will enter some if for some particular effect
2. some numeric that is arithmetic based on which you can calculate distance. the only example comes to my mind,  as @rasmusnv, mentions some linear things. I think the example that will be used most likely is parts of morton coded directions and origins.

now, the problem with arithmetic is that it potentially see the two hints with all bits different as the closest, so it breaks the 1. ... totally

on the other hand xor is succesfully used for checking morton builder.
even more crude, the closer code is considered the one that has lower first different bit (maybe I should put that on a list two, actually). so xor IMO should work great for 1. and good for 2. IMHO.

proposals/0025-shader-execution-reordering.md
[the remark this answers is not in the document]
jasilvanus: We re-discussed internally and are fine with the proposed solution of only requiring a barrier on the write side. It may penalize some implementations, but it's really a very special corner case where it makes a difference, and it's questionable whether it's worth the effort optimizing for that.

proposals/NNNN-shader-execution-reordering.md
derlemsft: In last week's meeting we talked about how the default behavior is no reorder, but implementations can allow reordering if explicitly specified by user. Hitobjecttraceray followed by hitobjectinvoke is a prefer no reorder, because no reorder thread is called directly. No attributes or parameters needed. 

Are we ok to close this comment?

proposals/NNNN-shader-execution-reordering.md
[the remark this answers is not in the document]
derlemsft: In last week's meeting we agreed to no monolithic API, but would make adjustments to the spec for clarity. 
Options going forward:

- In spec we could say calling the three functions is the same as calling trace ray semantically
- In spec explicitly call out that compiler and driver might merge these functions, and putting anything in between the function calls might prevent that.
- Have example code as a helper function. In the example code that is traceray, reorder, invoke, and the implementation does those three in a row, and it's just a sample code for the case where the app wants
- Add a helper library to compiler, and ship some helper functions as a hlsl header

@rdrabins given the above options, which would intel prefer?

proposals/NNNN-shader-execution-reordering.md
derlemsft: Same as above, generally agreed no attributes.

proposals/NNNN-shader-execution-reordering.md
derlemsft: From last week's meeting, AMD proposal for a Reorder Coherence domain: just add new enums for UAV attribute, and barrier ( ) enum param.

@rasmusnv @Flakebi can we confirm the decision here?

proposals/0025-shader-execution-reordering.md
rasmusnv: I agree that there are some interesting aspects to specifying XOR between the two, but I think we need to consider something like this in the context of the entire system to see if it's reasonable.

Let's say a sort is performed that will put the hint values in descending order. Chunks of this sorted work is distributed to various execution units for dispatch. This should work for both arithmetic distance (it's an arithmetic sort) and XOR distance (the MSB has the highest weight on its order). The possible approximation with respect to the distance metric is at chunk boundaries.

As mentioned, this still leaves the question of how to prioritize dispatch groups within each execution unit when values are not a perfect match. This assumes that we are doing something sophisticated, rather than simply grabbing the next N sorted items.

In the earlier example, the three values would be ordered as:

```
B = 0b111111
A = 0b100000
C = 0b011111
```

Let's assume that we only have to group 2 items. We either pick A&B or A&C depending on if it's a XOR or arithmetic distance metric. Another choice is a strict **order** metric, which is actually the simplest in this case. That would indicate that we start with A&B.

Then consider the case where we have to group more work, say 8. Let's say that we have some duplicates (x-factor indicates the number of duplicates):

B = 0b111111 x 4
A = 0b100000 x 1
C = 0b011111 x 4

Would we be better served grouping B&C, B&A, A&C? I think this is not clear. What I'm trying to say is that the exact distance metric for dispatch grouping might be best left unspecified.

Maybe a suitable formulation would be that the system should in general try to dispatch hint values in arithmetic order (NOTE: not distance), but the exact mapping to waves is left unspecified.

Thoughts?

Anteru: I would prefer if the expectation was we'd group A&B in your second example, based on the `XOR` distance. Given the grouping is allowed to throw away lower-order bits, it already implies higher order bits are more important, and it will be likely used to set "feature bits" anyways by ISVs because "arithmetic distance" between shaders is not really something you can calculate. But you can say "those two shaders are going to do X" and we want to group by similarity, so the more similar bits are set the more likely they are similar, with higher bits used to indicate "more expensive" things.

Undefined means it'll be defined by implementations, after all.

proposals/NNNN-shader-execution-reordering.md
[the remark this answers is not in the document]
Anteru: From the options above, the last one seems most preferable - a function we can reliably pattern match and "officially sanctioned".

proposals/NNNN-shader-execution-reordering.md
rasmusnv: Yes. "Memory coherence and visibility" has been updated with the proposal.

Flakebi: Yes, thanks for that.
The current text reads fine, I think it would be good to add that the new `REORDER_SCOPE` value is added to the `SEMANTIC_FLAG` enum and that `[reordercoherent]` is a new storage class that can be annotated on buffers.
Just to be clear to readers that these two are added in this proposal.

proposals/0025-shader-execution-reordering.md
```diff
@@ -0,0 +1,1519 @@
...
+strategies are likely able to reuse existing mechanisms to implement support
+for SER. No DXR runtime changes are necessary, since the proposed extension to
+the programming model is limited to HLSL and DXIL.
+
+## Detailed Design
+
+This section describes the HLSL additions for `HitObject` and `ReorderThread`
+in detail.
+The canonical use of these features involve changing a `TraceRay` call to the
+following sequence:
```
rdrabins, line 83: we've spoken last time about this being equivalent functionally.
The current sentence "almost" says that. Could we have this more explicitly?
```suggestion
The canonical use of these features involve changing a `TraceRay` call to the
following sequence that is functionally equivalent:
```

proposals/0025-shader-execution-reordering.md
tex3d: @rasmusnv, I think this is a good explanation, thanks.

proposals/0025-shader-execution-reordering.md
skeelyamd: One thing to note here is that the XOR codes can be converted to arithmetic codes while preserving distance orderings at the cost of using a few more bits.  One such mapping is shown below.

image

In the a second example above we would have:

Label | XOR code | Distance preserving arithmetic code
---|---|---
B | 63 | 364
A | 32 | 243
C | 31 | 121

With this transform we can use an arithmetic distance, yet B&A remain closer than A&C.

Anteru: I'm not convinced that it's easier for _applications_ to produce an arithmetic distance given applications will know features, not distances. I.e. a bit corresponds to a particular property, it's hard to imagine how an application would calculate an arithmetic distance from a set of feature bits. To me we got this backwards - now you'd have to take a "feature based metric" from the app, convert it to arithmetic distance using the reverse encoding, then unpack it to XOR again for sorting - what did we win? Would you mind posting the formula for both directions in this case?

skeelyamd: For clarity, in the above the application isn't computing an arithmetic _distance_.  It's converting an xor _code_ into an arithmetic _code_ directly, such that the distance orderings are preserved.

The code conversion above reinterprets the xor code bits in trinary.  So where a bit would normally be interpreted to have value 2^i the conversion interprets it as 3^i.  So xor code 0b0101 (aka 5) becomes 3^2 + 3^0 = 10.  This inserts a gap between each new option bit sufficient to preserve distance orderings.

In application code this could be a change from `xor_code |= option_bit;` to `arith_code += option_value;`.  Of course we could imagine more complex scenarios than simply accumulating option constants.

I don't think it is possible to preserve arithmetic distance orderings using an xor code.  Maybe it can be done but it's not obvious to me how to do it.  So unfortunately I can't post the formula for the opposite direction.

>take a "feature based metric" ... what did we win?

Since there appear to be valid arguments for both arithmetic and xor distances what we won is having a single consistent code space (arithmetic) which the application can use to encode either arithmetic or xor distances.  The implementations can always sort and group by arithmetic distance and the application has the freedom to construct its codes to reflect either distance ordering.

rdrabins: The most important thing about arithmetic is that for me its not even good for arithmetic based HW if it drops bits.
So impl has to do rounding when doing dropping.

e.g.
original codes given by app:
A 101000000
B 100111111
C 101111111
if implementation drops 6 bits it sees:
A 101__________
B 100__________
C 101__________

For arithmetic the closest codes changes from A and B to A and C when bits are dropped.
for XOR it stays A and C.
Simply arithmetic is not stable for dropping of bits.


I wonder if we could pass literal with mask that says at least how the structure of hint looks like. for instance if hint by app is of structure:

```c
struct HintStructA {
  uchar featureA : 1;
  uchar featureB : 1;
  uchar LittleEnum : 2;
  uchar featureD : 1;
  uchar LittleArithCode : 3;
};
```
than ReorderThreads instead of having important bit count argument could have mask (compile time constant!) that marks on which bit each field starts. 
For above `hintStructA`: it would be `0b11101100`;

Dropping bits impl may be dropping whole fields. App can express its wish of understanding the hint including the ones we are discussing above:
e.g. 
arithmetic 8bit 0b10000000
xor            8bit 0b11111111

rasmusnv: The biggest disconnect for me is how these concepts map to the implementation. As mentioned earlier, arithmetic order and xor order is the same. The problem appears when we start talking about distances.

Distances are only useful if wave dispatch isn't strictly in order. We would have to extract the most similar hint pairs iteratively with a greedy approach. But that would leave the remaining work even less coherent, so it seems questionable as a global optimization. It's also unclear to me how such an algorithm would be parallelized and how hint distance would be weighted against thread occupancy.

If we specify the behavior as arithmetic or xor distance I'm afraid that apps will try strange mappings to convert between the two, wasting bits and being less efficient, for no gain on many implementations that don't care about distance.

Having a mask that indicates the start of each field is interesting and would allow the user to express intent exactly without introducing waste. If the mask is added by hand it may be confusing for developers, but having it generated from a struct may be a sweet spot. I think we would have to introduce a struct attribute for hints (similar to `[raypayload]`) which only allows bit fields.

I propose that we either go with @rdrabins' struct idea or that we don't specify anything about distances in the spec and instead talk about order. I have a slight preference to the latter since I don't see exactly how the information would be used in practice and since it adds complexity. The advantage of the struct is that it strictly gives more information about intent.

proposals/NNNN-shader-execution-reordering.md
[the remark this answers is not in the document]
derlemsft: Msft team has decided on the following.

- No built-in TraceRayAndReorder helper will be shipped
- A note will be made in the documentation/spec about the equivalence of the code
- A header is something we will consider, but at a later date

proposals/0025-shader-execution-reordering.md
derlemsft: In design meeting we agreed to review DXIL spec and revisit.

proposals/0025-shader-execution-reordering.md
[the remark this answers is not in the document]
derlemsft: @rasmusnv AMD requested that the spec wording is more explicitly clarified for reordercoherent. No other outstandings, resolving thread.

proposals/NNNN-shader-execution-reordering.md
derlemsft: AMD requested spec wording be updated so it is clear that reorder scope will not be used to synchronize between different threads

proposals/0025-shader-execution-reordering.md
tex3d: **Regarding "distance" metrics:**

We have some interesting discussions about different ways to compare hint values and compute distances, but as @rasmusnv [points out](https://github.com/microsoft/hlsl-specs/pull/277/files/bb617dfc02fd7d9ab67c532097a8f487b8fa0e98#r1900819781):
> The biggest disconnect for me is how these concepts map to the implementation.

Language specifying that higher bits of the hint value are generally considered more important than lower bits might help, but this likely needs qualifiers to temper expectations.

For instance, an implementation may combine buckets linearly to get to some size after sorting ([as suggested by rasmusnv](https://github.com/microsoft/hlsl-specs/pull/277/files/bb617dfc02fd7d9ab67c532097a8f487b8fa0e98#discussion_r1853742351)), which doesn't guarantee that groups of threads are reliably split along the highest-bit lines first, or even that the closest buckets by any "distance" metric are combined.

Example - selecting 16 threads from buckets sorted in ascending order:
```
A 0b0000 x 8
B 0b0100 x 8
C 0b0110 x 4
D 0b0111 x 4
```
A simple, greedy schedular might schedule A&B together before scheduling C&D, when globally it appears that B&C&D should be combined instead, since they have the same highest bit, and are even closest in arithmetic space.  If you use the concept of "distance", then in both XOR and in arithmetic space, B&C&D should be combined instead of A&B, yet that's not what this simple scheduler does.

I'm not saying this is how it *should* work, but I think this should be considered a valid implementation.

Given this, I suspect the language in the spec will need to remain somewhat vague to avoid reliance on any particular concept of "distance".

I suggest something like this:
> In general, more significant bits of the hint value are more important than less significant bits for determining coherency, though scheduling behavior is implementation specific.

---
**Regarding structured hints:**

This thread started on the basis of an insufficiently clear definition of how the hint values impact scheduling coherence.  If we cannot clearly define how more complex structured hint definitions map to behavior, the problem gets worse, not better.  The way I see it, a more complex hint definition sets up additional expectations for users that aren't likely to be met by implementations.

proposals/0025-shader-execution-reordering.md
```diff
@@ -0,0 +1,1460 @@
...
+---------                           | ----------
+`HitObject Hit` | `HitObject` that encapsulates the hit or miss according to which reordering should be performed.
+
+---
+
+#### ReorderThread with coherence hint
+
+This variant of `ReorderThread` reorders threads based on a generic
+user-provided hint. Similarity of hint values should indicate expected
+similarity of subsequent work being performed by threads. The resolution of
```
tex3d, line 727:
```suggestion
similarity of subsequent work being performed by threads. In general, more
significant bits of the hint value are more important than less significant
bits for determining coherency, though scheduling behavior is implementation
specific. The resolution of
```

llvm-beanz:
Design meeting: assigning to @derlemsft as the PM ushering this through.

proposals/0025-shader-execution-reordering.md
rasmusnv: @tex3d I committed your suggested change.

proposals/0025-shader-execution-reordering.md
```diff
@@ -724,7 +724,10 @@ Parameter                           | Definition
 
 This variant of `ReorderThread` reorders threads based on a generic
 user-provided hint. Similarity of hint values should indicate expected
-similarity of subsequent work being performed by threads. The resolution of
+similarity of subsequent work being performed by threads. In general, more
```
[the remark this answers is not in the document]
tex3d: I'm not sure how we resolve these issues without something vague, unless we prescribe only specific allowed interpretations.
- If they are always more important, that may imply XOR distance comparisons for grouping, but an arithmetic ordering may violate this expectation when grouping threads.
- They are more important in the sense that if bits are dropped, it will be some arbitrary number of low bits, making the lower bits less important.
- If they are not always more important, then what language do we use?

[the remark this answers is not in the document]
tex3d: What about something like this:

> Thread grouping may be based on an arithmetic ordering of threads by hint value, or some other method.  Since an arbitrary number of lowest bits may be ignored, lower bits are less important than higher bits in the hint value.  However, thread grouping is not required to minimize any notion of distance between hint values within groups of threads, since this would greatly complicate thread grouping implementations.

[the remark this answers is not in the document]
tex3d: Then again, if all parties are happy with @llvm-beanz's addition of `Specific scheduling behavior may vary by implementation` covering the divergence from any notion of distance, we could just accept that.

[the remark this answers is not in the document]
llvm-beanz: Is there any situation where `1 << 12` would be a "less important" hint than `1 << 2`?

These are hints, so we don't specifically dictate what the implementation _must_ do with the hint, but being clear about the ordering importance of bits seems like something we should all be able to agree on.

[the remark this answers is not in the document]
tex3d: See my [last comment on the other thread](https://github.com/microsoft/hlsl-specs/pull/277#discussion_r1917520467) for an example of how threads with a different high bit could be grouped together while threads where the high bit is the same and only the lower bits are different are separated, when it could have grouped threads by the high bit value first instead.  If the high bit is always more important for grouping threads, the behavior observed from a simple greedy sorting scheduler may be unexpected.

In any case, I already said we could go with your suggested edit if everyone is satisfied with that.

proposals/0025-shader-execution-reordering.md
derlemsft: @rdrabins @rasmusnv resolving this thread if there are no other comments

Jasper-Bekkers:
If reordering is optional, I would suggest renaming the call to `MaybeReorderThread` so that it's clear to the caller that implementations are free to do this, or not.

proposals/0025-shader-execution-reordering.md
```diff
@@ -724,7 +724,10 @@ Parameter                           | Definition
...
+significant bits of the hint value are more important than less significant
+bits for determining coherency, though scheduling behavior is implementation
+specific. The resolution of
```
tex3d, line 730: @llvm-beanz suggestion:

```suggestion
similarity of subsequent work being performed by threads. More significant
bits of the hint value are more important than less significant bits for
determining coherency. Specific scheduling behavior may vary by
implementation. The resolution of
```

tex3d (approved):

Flakebi:
There are two comment threads above ([one](https://github.com/microsoft/hlsl-specs/pull/277#discussion_r1852057774) and [two](https://github.com/microsoft/hlsl-specs/pull/277#discussion_r1913955640)) about clarifications in the wording.
The threads were marked as resolved, but it would still be nice if the formulations can be clarified before this is merged.

rasmusnv:
> There are two comment threads above ([one](https://github.com/microsoft/hlsl-specs/pull/277#discussion_r1852057774) and [two](https://github.com/microsoft/hlsl-specs/pull/277#discussion_r1913955640)) about clarifications in the wording. The threads were marked as resolved, but it would still be nice if the formulations can be clarified before this is merged.

Please see the latest commit for clarifications.

Flakebi:
Thanks for update, looks good to me now :)

rasmusnv:
I think the PR is ready for merge. I resolved the remaining conversations as it looked like no changes were required at this point. For further changes we can open focused issues. I also picked up the next available number, 27. (Please ignore the close and re-open, that was a mistake).

damyanp:
Thank you everyone!  I'm going to go ahead an merge this PR.  Options for further discussion:

* open an issue in this repo against this proposal
* submit a PR with a proposed change to the proposal
* reach out directly to us if the above options are not possible

damyanp:
> If reordering is optional, I would suggest renaming the call to `MaybeReorderThread` so that it's clear to the caller that implementations are free to do this, or not.

https://github.com/microsoft/hlsl-specs/pull/378

PR #230, opened 2024-05-03
INF-0004 Specify diagnostics for resource binding, specifically on resource types and binding prefixes.

bob80905:
This PR adds a specification for what diagnostics the compiler should emit for each possible case of the binding of a resource. The goal of the PR is to prevent any ambiguity about what errors will be emitted in any case where a resource is bound. This excludes templated types in resources, and is solely focused on the outer type and the binding prefix. That is, given
`RWBuffer<T> : register(x3)`
The spec specifies what diagnostic gets emitted here, but doesn't address the potential for the invalid type of T. The spec focuses on the binding prefix that's used, and the resource type "RWBuffer".

PR #278, opened 2024-07-19
Add infrastructure proposal to discuss in-proc and out-of-proc support HLSL compiler library

coopp:
This commit adds an infrastructure proposal talking about a possible out of proc solution for the HLSL shader compiler library. It discusses at a very high level what an out of process architecture may look like and how it could be considered an optional feature as well.


proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Introduction
+
+An effort is underway to bring compilation support into clang for HLSL based
+shaders.  A C style library will be built to enable applications to compile
+a shader from their own processes in addition to being able to launching the
+clang process.
+
+There are two proposed ways one could consume the shader compiler library and
+compile a shader today.
```
damyanp, line 21: > There are two proposed ways one could consume the shader compiler library and compile a shader today.

Who is the "we" in the sentence?

Is "shader compiler library" the "C style library" mentioned previously?

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,138 @@
...
+
+* **In Process** - An application links against the shader compiler library to
+bring in support for shader compilation as a static library
+(full implementation) or a dynamic library. Compilation is performed within the
```
damyanp, line 25: In my mind there's a huge difference between static and dynamic here. I think that for the API we're thinking of here static linking is not on the table.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,138 @@
+<!-- {% raw %} -->
+
+# In-process only or In-process and Out-of-process support for the compiler library
```
damyanp, line 3: "library" is quite overloaded in this space.  Wonder if "API" would more directly indicate what this is trying to solve.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,138 @@
...
+
+* Proposal: [0004](INF-0004-inproc-outofproc-compiler-library-support.md)
```
damyanp, line 5: This proposal should start off numbered NNNN.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,138 @@
...
+
+There are two proposed ways one could consume the shader compiler library and
+compile a shader today.
+
+* **In Process** - An application links against the shader compiler library to
+bring in support for shader compilation as a static library
+(full implementation) or a dynamic library. Compilation is performed within the
+calling application's process.
+
+The DirectX Shader library does not currently support being linked as a
```
damyanp, line 28: > The DirectX Shader library does not currently support being linked as a

What does "the DirectX Shader library" refer to?  Is it the same as the "shader compiler library" or the "C style library" mentioned earlier?  Or is this referring to DXC's API?

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,138 @@
...
+static library.  It is a COM interface library that brings in compilation
+support dynamically. Both static and dynamic in-proc solutions will be 
+supported in the HLSL clang based library.
```
damyanp, line 31: > Both static and dynamic in-proc solutions will be supported in the HLSL clang based library.

Proposing that the new API will support static and dynamic is a pretty big decision being made here and so shouldn't be tucked away in the introduction.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,138 @@
...
+
+* **Out of Process** - An application links against the shader compiler library and
+uses the apis to compile which create one or more processes on the
+application's behalf to compile one or more shaders. Compilation is performed
+outside the calling application's process and isolated away to a worker
+process.
+
+In-Process must be supported because it is already a method that applications
+use today to consume the DirectX Shader Compiler library.
```
damyanp, line 40: > In-Process must be supported because it is already a method that applications use today to consume the DirectX Shader Compiler library.

I disagree with this logic.  Just because it is supported by DXC does not mean that this is the right solution for clang.

Maybe we'll still decide that we need to support in-proc, but if we do I think we need something stronger than "this is what DXC did" as a reason to do it.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,138 @@
...
+
+Out of process is considered an optional feature as the library could ship
+without it in the first release.
+
+This document is a discussion about an out of process design and to evaluate if
+we need to build the system immediately.
```
damyanp, line 46: I think that this document needs to pick one of these possibilities:

* in-proc only
* out-of-proc only
* both in- and out-of-proc options provided

Maybe some more detail on what is meant by in- or out-of-proc would help us understand the decision.

FWIW, going in to this, "out-of-proc only" is my preferred option, so this document shouldn't assume that in-proc is required.

damyanp (changes requested):
I've not reviewed past the Introduction since I think we need to get aligned on the assumptions being made there before we should look at the rest of this.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
damyanp: (suggest leaving any renaming of this until _just before_ this PR is completed so that it doesn't confuse the diffs)

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: Updated the document to be more consistent in the terms used.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: Updated the name in the document contents.  Will update the file name when the review is close to being complete to avoid diff issues.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: Made the proposal actually 'propose' a solution rather than be wishy washy about all possible options.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: Picked!

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: I removed the leaning on DXC's past as evidence to support a direction.  It should be top of mind for features, but for architecture we can be more creative as to what we want the new clang based architecture to be.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: Removed confusion.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: Fixed up terminology usage.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,113 @@
...
+shaders.  A C style api will be built to enable applications to compile
+a shader from their own processes in addition to being able to launching the
+clang process.
+
+This document is to propose building only the out-of-process support which
+gives the caller a more performant and secure solution.
+
+Visual Studio uses an out of process architecture for compilation which helps
+better utilize the power of the machine performing the compilation as well as
+isolating the code editor from any crashes that may occur during compilation.
```
damyanp, line 25: I don't quite get what Visual Studio has to do with anything here?

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,113 @@
...
+into the application's process. All work is performed within the application's
+process.
+
+* **Out of Process** - An application calls an api that launches one or more
+processes to perform the compilation. All work is performed outside the
+calling application's process.
+
+## Motivation
+
+### Why prefer an out of process solution?
```
damyanp, line 40: Also:
* cleanup between compiler invocations
* handling of catastrophic errors (if compiler aborts on error then what does that do to the calling process?)

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,113 @@
...
+and some of that state could leak to other compilation sessions if it is not
+protected. This issue was also called out when an experiment called
+[llvm-buildozer](https://reviews.llvm.org/D86351) was created.  The author was
+interested in multiple threads re-entering the main compiler entry point in an
+effort to reduce compilation times. The effort showed that the unsafe bits
+could be fixed up and made thread safe allowing the experiment to be built.
+I have not checked to see the thread safe fixed bits are in the current llvm
+source.
+
+## Proposed solution
```
damyanp, line 61: I think many of the details in here would belong in a detailed design section. I'm going to try and avoid reviewing the details here based on the experience with https://github.com/microsoft/hlsl-specs/pull/230.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,113 @@
...
+* Author(s): [Cooper Partin](https://github.com/coopp)
+* Sponsor: [Cooper Partin](https://github.com/coopp)
+* Status: **Under Consideration**
+* Impacted Project(s): (Clang)
+
+* Issues:
+
+## Introduction
+
+An effort is underway to bring compilation support into clang for HLSL based
```
damyanp, line 15: We should explain here _why_ we're building this API. Why would an app prefer to use the API rather than spawning a process?

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Detailed design
+
+A detailed design of the out-of-process system has not been created.
+There is an interesting example of an approach for a general purpose system
+called 'Compiler Services'. This was presented as an option to support editors
+and work with compiler databases. The proposal was called
+[Clang C++ Services](https://github.com/chandlerc/llvm-designs/blob/master/ClangService.rst)
+
+## Resources
```
damyanp, line 107: The template has an "Alternatives Considered" option, which might be a good place to discuss why inproc-only or inproc+outproc were not chosen as the solution.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: I was using it as an example.  It compiling things using out of process compilation.  I could remove this if you think it is too confusion.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: ok.. I can do that.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: ok.. Will do.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
damyanp: It adds confusion.  Pretty much _anything_ that compiles by launching the compiler as a separate .exe is doing out of proc compilation.

coopp: ha!  I didn't think about that!  Good point!

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: Fixed

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: fixed

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: Fixed

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,130 @@
...
+processes to perform the compilation. All work is performed outside the
+calling application's process.
+
+## Motivation
+
+### Why prefer an out of process solution?
+
+* Process creation on Windows based systems is expensive.  Launching clang.exe
+over and over again to compile shaders becomes a lengthy operation on 
+Windows platforms.
```
damyanp, line 54: > Launching clang.exe over and over again to compile shaders....

This is actually describing one possible out-of-proc solution.  In-proc would not involve launching new processes.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
```diff
@@ -0,0 +1,130 @@
...
+
+* Security is a concern when compiling a shader in a process space that is
+considered protected.  An out of process design would move that concern to an
+isolated process.
+
+* In-process support using multiple threads have known thread-safety concerns
+for the clang compiler.  The clang compiler holds some state at compile time
```
damyanp, line 61:
```suggestion
* Running multiple instances of clang In-process across multiple threads has known thread-safety concerns
for the clang compiler.  The clang compiler holds some state at compile time
```

damyanp (approved):
I think this is probably ready to have a design discussion about.

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: fixed

proposals/infra/INF-0004-inproc-outofproc-compiler-library-support.md
coopp: I removed this.  You are correct that it is referring to a solution and that isn't want the list is supposed to communicate.

bogner (approved):
The title mentions in-proc, out-of-proc, or both, but the content of the proposal is all about out-of-proc - should we update the title to match?

Otherwise, the content looks like it's in good shape to start with. LGTM.

coopp:
This PR is being abandoned in favor of this one: https://github.com/llvm/wg-hlsl/pull/3 which is going into a different repo.

PR #287, opened 2024-07-26
Complete detailed design in resource binding diagnostic infrastructure spec

bob80905:
This PR adds a detailed specification for the register annotation diagnostic infrastructure, going into detail on the analysis and diagnostic emisison steps. It's a follow up PR of https://github.com/microsoft/hlsl-specs/pull/230

### DIFF
--- a/proposals/infra/INF-0005-register-types-and-diagnostics.md
+++ b/proposals/infra/INF-0005-register-types-and-diagnostics.md
@@ -178,32 +178,32 @@ struct Bar{
   RWBuffer<int> a;
   RWBuffer<int> b;
 };
-Bar x : register(t9) : register(t10);
+Bar x : register(u9) : register(u10);
 // error: binding type 'u' cannot be applied more than once
 //
 
 // Binding 'c' when it should be packoffset in non-global scope:
 //
 // - warn_hlsl_register_type_c_packoffset
-// warning: binding type 'c' ignored in buffer declaration. Did you mean 'pack_offset'?
+// warning: binding type 'c' ignored in buffer declaration. Did you mean 'packoffset'?
 cbuffer g_cbuffer { float f : register(c2); }
-// warning: binding type 'c' ignored in buffer declaration. Did you mean 'pack_offset'?
+// warning: binding type 'c' ignored in buffer declaration. Did you mean 'packoffset'?
 //
 
 // Binding 'b' ignored / being used in the legacy way:
 //
 // - warn_hlsl_deprecated_register_type_b
-// warning: binding type 'b' only applies to constant buffers. The "bool constant" binding type has been removed.
+// warning: binding type 'b' only applies to constant buffers. The "bool constant" binding type is no longer supported
 float f : register(b0);
-// warning: binding type 'b' only applies to constant buffers. The "bool constant" binding type has been removed.
+// warning: binding type 'b' only applies to constant buffers. The "bool constant" binding type is no longer supported
 //
 
 // Binding 'i' ignored (also wording):
 //
 // warn_hlsl_deprecated_register_type_i
-// warning: binding type 'i' ignored. The "integer constant" binding type has been removed
+// warning: binding type 'i' ignored. The "integer constant" binding type is no longer supported
 float f : register(i0);
-// warning: binding type 'i' ignored. The "integer constant" binding type has been removed
+// warning: binding type 'i' ignored. The "integer constant" binding type is no longer supported
 //
 ```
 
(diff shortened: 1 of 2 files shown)


proposals/infra/INF-0005-register-types-and-diagnostics.md
```diff
@@ -130,21 +130,275 @@ is no corresponding member, a warning will be emitted that is treated as an erro
...
+Bar x : register(t9) : register(t10);
+// error: binding type 'u' cannot be applied more than once
+//
+
+// Binding 'c' when it should be packoffset in non-global scope:
+//
+// - warn_hlsl_register_type_c_packoffset
+// warning: binding type 'c' ignored in buffer declaration. Did you mean 'pack_offset'?
+cbuffer g_cbuffer { float f : register(c2); }
+// warning: binding type 'c' ignored in buffer declaration. Did you mean 'pack_offset'?
```
damyanp, line 191: I think it is spelled `packoffset`.

```suggestion
// - warn_hlsl_register_type_c_packoffset
// warning: binding type 'c' ignored in buffer declaration. Did you mean 'packoffset'?
cbuffer g_cbuffer { float f : register(c2); }
// warning: binding type 'c' ignored in buffer declaration. Did you mean 'packoffset'?

```

damyanp, line 191: I think it is spelled `packoffset` - without the underscore.

proposals/infra/INF-0005-register-types-and-diagnostics.md
```diff
@@ -130,21 +130,275 @@ is no corresponding member, a warning will be emitted that is treated as an erro
...
+//
+
+// Mismatched register bindings:
+//
+// - err_hlsl_binding_type_mismatch
+// error: binding type '%select{t|u|b|s|c}0' only applies to '%select{srv resources|uav resources|constant buffer resources|sampler state|constant buffer data}0'
+float f : register(t0);
+// error: binding type 't' only applies to 'srv' resources.
+RWBuffer<float> f : register(c3);
+// error: binding type 'c' only applies to constant buffer data.
```
damyanp, line 163: I'm wondering if "constant buffer data" is the correct term to use here? It only applies to offsets in $Globals, doesn't it?

If you use it inside a `cbuffer` then you get a warning to use `packoffset`.  I think earlier in the document it talks about "offset into $Globals", I wonder if some of that text could make it here?

proposals/infra/INF-0005-register-types-and-diagnostics.md
```diff
@@ -130,21 +130,275 @@ is no corresponding member, a warning will be emitted that is treated as an erro
...
+// Multiple bindings with the same register type:
+//
+// - err_hlsl_duplicate_register_annotation
+// error: binding type '%select{t|u|b|s|c|i}' cannot be applied more than once
+struct Bar{
+  RWBuffer<int> a;
+  RWBuffer<int> b;
+};
+Bar x : register(t9) : register(t10);
+// error: binding type 'u' cannot be applied more than once
```
damyanp, line 183: I think in this example it should be `'t'`?  Although there should be another error because there aren't any `t` resources in that UDT.

proposals/infra/INF-0005-register-types-and-diagnostics.md
```diff
@@ -130,21 +130,275 @@ is no corresponding member, a warning will be emitted that is treated as an erro
...
+// - warn_hlsl_register_type_c_packoffset
+// warning: binding type 'c' ignored in buffer declaration. Did you mean 'pack_offset'?
+cbuffer g_cbuffer { float f : register(c2); }
+// warning: binding type 'c' ignored in buffer declaration. Did you mean 'pack_offset'?
+//
+
+// Binding 'b' ignored / being used in the legacy way:
+//
+// - warn_hlsl_deprecated_register_type_b
+// warning: binding type 'b' only applies to constant buffers. The "bool constant" binding type has been removed.
```
damyanp, line 197: > ...has been removed.

Has been removed from what?  Wonder if "is no longer supported" might be better?

proposals/infra/INF-0005-register-types-and-diagnostics.md
```diff
@@ -130,21 +130,275 @@ is no corresponding member, a warning will be emitted that is treated as an erro
 because this behavior was permissible in legacy versions of the compiler. These warnings
 are also part of the same warning group, `LegacyConstantRegisterBinding`.
 
-`DiagnoseHLSLRegisterAttribute` will also be responsible for emitting a diagnostic
-if any other invalid register type is detected. If `DiagnoseHLSLRegisterAttribute`
-finds any critical errors, the attribute, `HLSLResourceBindingAttr`, won't be added
+`DiagnoseHLSLRegisterAttribute` will also be responsible for emitting a diagnostic
```
damyanp, line 133: There's something strange about the line endings (maybe) in this file?  These diffs are odd.  Plus if I try and use the "Add a suggestion" feature then the diff thinks that the entire line has changed.

Any ideas what may have happened?

proposals/infra/INF-0005-register-types-and-diagnostics.md
```diff
@@ -130,21 +130,275 @@ is no corresponding member, a warning will be emitted that is treated as an erro
...
+    if ContainsNumeric is not set:
+      emit warn_hlsl_user_defined_type_missing_member
+  else
+    emit err_hlsl_binding_type_invalid  
+```
+
+## Behavioral Differences
+
+This infrastructure will introduce some behavioral differences between `clang` and `dxc`.
+The whole approach of setting flags based on decl characteristics and using them
```
damyanp, line 389: > The whole approach of setting flags based on decl characteristics and using them to drive diagnostics is an approach that wasn't taken in DXC.

This implementation detail isn't relevant to any behavioral differences.

proposals/infra/INF-0005-register-types-and-diagnostics.md
```diff
@@ -130,21 +130,275 @@ is no corresponding member, a warning will be emitted that is treated as an erro
...
+to drive diagnostics is an approach that wasn't taken in DXC. Secondly, as mentioned above, 
+the `disallow-legacy-binding-rules` warning group did not exist in `dxc`, and neither
+did any of the warnings that are contained in that group. Those warnings are being 
+introduced to `clang-dxc`. Many of these warnings will be treated as errors, causing some
```
damyanp, line 393: nit: "clang-dxc" is the driver.  We're introducing these new warnings and errors to clang. It's possible to light up all this proposed functionality without going through clang-dxc.

proposals/infra/INF-0005-register-types-and-diagnostics.md
```diff
@@ -130,21 +130,275 @@ is no corresponding member, a warning will be emitted that is treated as an erro
...
+// Binding 'i' ignored (also wording):
+//
+// warn_hlsl_deprecated_register_type_i
+// warning: binding type 'i' ignored. The "integer constant" binding type has been removed
+float f : register(i0);
+// warning: binding type 'i' ignored. The "integer constant" binding type has been removed
+//
+```
+
+In DXC, the analysis and diagnostic emission steps would happen in DiagnoseRegisterType(),
```
damyanp, line 211: I think there's too much detail in this section. The list of diagnostics is useful, but I'd rather just review the implementation after we've agreed on the specific diagnostics rather than review both this textual description of it and the implementation itself.

I've not reviewed anything from here until the Behavioral Differences section.

damyanp (changes requested):
Couple of notes on the proposed diagnostics. I've not reviewed the description of the implementation, I'll wait to see the implementation PR for that.

bogner (approved):
Looks good. Like Damyan says, I don't think we need the Analysis and Diagnostics Emission sections - they're basically just pseudo code at this point which will be easier to review and engage with in the implementation PR itself.

proposals/infra/INF-0005-register-types-and-diagnostics.md
bob80905: I think the `offset` part is an implementation detail that doesn't need to be mentioned in the diagnostic, but I could totally see changing the message to:
`...only applies to $Globals constant buffer data`
Thoughts?

proposals/infra/INF-0005-register-types-and-diagnostics.md
bob80905: I actually intended the bindings to use `u`, not `t`. Fixed.

proposals/infra/INF-0005-register-types-and-diagnostics.md
bob80905: Agreed

proposals/infra/INF-0005-register-types-and-diagnostics.md
damyanp: I'm not sure I agree that the offset part is an implementation detail - the developer absolutely needs to know what that does!

However, for a diagnostic I think it's ok to lean on just trying to describe what it has to apply to.  

Did we find a snappy way to classify things that go into $Globals?  Is it "global data" or "numeric globals" perhaps?

proposals/infra/INF-0005-register-types-and-diagnostics.md
bob80905: I think I just left some whitespace originally, merged that in, and then Chris opened a PR to fix it while this new PR was up.

proposals/infra/INF-0005-register-types-and-diagnostics.md
damyanp: Any thoughts on resolution to this one?

bob80905: How about `default cbuffer globals`?

damyanp: TBH, I think any mention of cbuffer and $Globals is probably confusing here.  How about: 

`error: binding type 'c' only applies to variables in the global scope`

PR #288, opened 2024-07-30
[Editorial] Move Decl.Linkage to Basic.Linkage

llvm-beanz:
This is purely an editorial re-ordering. No changes in language just moving the section from Decl earlier to Basic so that other concepts in basic for specifying the program model will flow more naturally.


hekota (approved):
LGTM!

PR #290, opened 2024-07-31
[INF-0004] Flesh out debug layer bytecode validation behavior.

amarpMSFT:


proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -127,38 +127,107 @@ apply the hash. Instead it should apply the `PREVIEW` sentinel value from the
...
+Flags are set via `ID3D12BytecodeOptions`:
+
+```C++
+[uuid(9291ad45-9a10-43a5-9c06-f5aa558e89a7), object, local, pointer_default(unique)]
+interface ID3D12BytecodeOptions
+    : IUnknown
+{
+    HRESULT SetBytecodeFlags(
+        [annotation("_In_")] D3D12_BYTECODE_FLAGS flags
+    );
```
amarpMSFT, line 212: Should there be any other valdiation controls?  Like allowing behavior tweaks on the validator in some way? (e.g validator version setting?)

proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -127,38 +127,107 @@ apply the hash. Instead it should apply the `PREVIEW` sentinel value from the
...
 
-Similarly the runtime could also support performing DXIL validation that fails
-shader creation if an error is found. It would support similar app controls as
-the debug layer, except defaulting to validation off.
+Apps can configure how the runtime validates bytecode passed to it. 
+
+The validation referred to here is implementation exposed by `dxcompiler.dll` 
+or `dxil.dll`.  This is the same validation used when the compiler
+endorses bytecode at compile time by applying a hash.  So it would typically be 
+redundant to validate again  But it can be useful to validate shaders that have
```
llvm-beanz, line 164:
```suggestion
redundant to validate again, but it can be useful to validate shaders that have 
```

proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -127,38 +127,107 @@ apply the hash. Instead it should apply the `PREVIEW` sentinel value from the
...
-Similarly the runtime could also support performing DXIL validation that fails
-shader creation if an error is found. It would support similar app controls as
-the debug layer, except defaulting to validation off.
+Apps can configure how the runtime validates bytecode passed to it. 
+
+The validation referred to here is implementation exposed by `dxcompiler.dll` 
+or `dxil.dll`.  This is the same validation used when the compiler
+endorses bytecode at compile time by applying a hash.  So it would typically be 
+redundant to validate again  But it can be useful to validate shaders that have 
+the `BYPASS` hash, or have the option to force validation for whatever reason.
```
llvm-beanz, line 165: nit: unnecessarily wordy
```suggestion
the `BYPASS` hash, or have the option to force validation. 
```

llvm-beanz (approved):
LGTM with a few really small nits.

@amarpMSFT, can you add yourself to the authors list at the top of the file?

jenatali (approved):

proposals/infra/INF-0004-validator-hashing.md
amarpMSFT: Ok, so we need HLSL work item to add this no-hash option.

llvm-beanz: I'm not sure I follow why we would add a no-hash option. Shouldn't we either run the validator and hash or not run the validator and thus not hash?

proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -127,38 +128,108 @@ apply the hash. Instead it should apply the `PREVIEW` sentinel value from the
...
+redundant to validate again, but it can be useful to validate shaders that have 
+the `BYPASS` hash, or have the option to force validation. 
+
+If there is bytecode to validate, the runtime and/or debug layer attempt to 
+load one of these dlls, `dxcompiler.dll` first. If either is available, 
+bytecode validation is possible. 
+
+The following flags in `d3d12.h` show the validation control options.  
+`ID3D12Device` exposes `ID3D12BytecodeOptions` to set the flags, shown 
+further below.
```
tex3d, line 174: I feel like the API details should be placed under a new `## Detailed Design` section. immediately preceding the `## Appendix 1: DXIL Hashing Algorithm` section.  In this section, the need for, and expected functionality, of the proposed interface could be discussed.

proposals/infra/INF-0004-validator-hashing.md
amarpMSFT: This is for when the runtime calls the validator.  It's just interested in whether the bytecode is valid or not, no need to hash.

proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -127,38 +128,130 @@ apply the hash. Instead it should apply the `PREVIEW` sentinel value from the
...
+```cpp
+ static const UINT32 DxcValidatorFlags_Default = 0;
+ static const UINT32 DxcValidatorFlags_InPlaceEdit =
+    1; // Validator is allowed to update shader blob in-place.
+ static const UINT32 DxcValidatorFlags_RootSignatureOnly = 2;
+ static const UINT32 DxcValidatorFlags_ModuleOnly = 4;
++static const UINT32 DxcValidatorFlags_SkipHash =
++    8; // Skip hash update; no modification or copy of input blob.
+ static const UINT32 DxcValidatorFlags_ValidMask = 0xF;
+```
```
tex3d, line 147: Since this was confusing as a diff (which wasn't even entirely correct for the mask change), maybe we should only list the new flag.

proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -127,38 +128,130 @@ apply the hash. Instead it should apply the `PREVIEW` sentinel value from the
...
+
+The runtime attempts to use the `DxcValidatorFlags_ValidMask` flag,
```
tex3d, line 149: This should have been the `DxcValidatorFlags_SkipHash` flag, which is the newly added one.

```suggestion
The runtime attempts to use the `DxcValidatorFlags_SkipHash` flag,
```

proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -123,42 +124,140 @@ should be that all shaders are validated and hashed using what is today the
...
+by a "final" validator.
+
+### IDxcValidator API Update
+
+A new `IDxcValidator` API flag is needed to avoid container hashing and
+unnecessary overhead when used by runtime validation.
+
+Proposed change to flags in `dxcapi.h`:
+
+```cpp
```
llvm-beanz, line 138: If you specify this as `diff` instead of `cpp` the rendering will actually render as an inline unified diff.

llvm-beanz: Or just strip it down as @tex3d suggested below (I guess I should read the other comments before commenting).

proposals/infra/INF-0004-validator-hashing.md
llvm-beanz: I still question the need for this. Don't we have a flag to write in place? That should avoid increased memory impact, and if we're not writing the output we just wrote some memory that gets thrown away.

I'm unsure that the cost of calculating the hash for something that happens only in the debug layer is actually worth the extra support complexity of a new user exposed interface option.

amarpMSFT: There is an  DxcValidatorFlags_InPlaceEdit flag.  The runtime could try to make sure it's only ever calling validate with it's own copy of bytecode, never passing in the APIs caller's memory.  But even if it did that, the runtime would lose track of what BYPASS the shader had originally since it would now be stomped by an unneeded hash.  Runtime could cache the bypass state before calling validator.  But it would seem cleanest overall if it could just ask for validation only without side effects.

llvm-beanz: Caching the bypass state before calling the validator is less clean than calling the validator handling a potential error flag then needing to cache the bypass state anyways and call the validator without the flag?

It seems like whatever you need to do in the runtime to make this work with existing validators needs to be done regardless of whether or not we add a flag to the validator.

amarpMSFT: For existing validators we'd just use the default behavior which is the validator gives us a copy of the shader that's hashed.

Much cleaner if we can first just try to validate with the new flag, which will avoid making a copy of the shader.  If the call fails E_INVALIDARG, since the new flag isn't recognized, we just call again without the flag and live with the fact that the validator made an unneeded copy of the shader.

amarpMSFT: My earlier reply was exploring what we'd have to do if we used the in-place flag that you pointed out.  What we'd really do is above.  Easy way to avoid wasted shader copy if possible.  If the request isn't reasonable, then what I list as the fallback just has to be the only path.

llvm-beanz: I think that if the fallback path needs to be implemented anyways, we should start there and only add the extra flag once we can measure the performance.

That allows us to understand the impact and make an informed decision about prioritization. Every hour we spend working on DXC is an hour we don't spend on Clang, and every new feature we add to DXC is a new feature we have to figure out how to support and/or deprecate as we move forward.

proposals/infra/INF-0004-validator-hashing.md
tex3d:
```suggestion
Propose adding `DxcValidatorFlags_SkipHash = 8` flag to `dxcapi.h`.  This will indicate to the validator that container hashing is not requested, avoiding an extra copy of the shader and the hashing operation, which is not needed for runtime or sdklayers use cases.
```

tex3d (approved):
Looks good with a couple minor comments to resolve.  Don't have to move section under Detailed Design as I suggested if you don't want to.

proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -123,42 +124,140 @@ should be that all shaders are validated and hashed using what is today the
...
+Apps can configure how the runtime validates bytecode passed to it. 
+
+The validation referred to here is implementation exposed by `dxil.dll` 
+or `dxcompiler.dll`, and also embedded in `d3d12sdklayers.dll`.  
+This is the same validation used when the compiler endorses bytecode at 
+compile time by applying a hash.  So it would typically be 
+redundant to validate again, but it can be useful to validate shaders that have 
+the `BYPASS` or `PREVIEW_BYPASS` hash, or have the option to force validation. 
+
+If there is bytecode to validate, the runtime and/or debug layer attempt to
```
damyanp, line 195: > the runtime and/or debug layer

Just to clarify here (because I was under the impression that only the debug layer was ever going to do validation).  Is the proposal here that somethings the runtime will do validation (by calling into some external DLL) but at other times the debug layer will do validation (using a version statically linked with it)?

Or something else?

proposals/infra/INF-0004-validator-hashing.md
tex3d: > I'm unsure that the cost of calculating the hash for something that happens only in the debug layer is actually worth the extra support complexity of a new user exposed interface option.

Note: this will be called from normal runtime, not just from debug layer, when using `D3D12_BYTECODE_FLAG_VALIDATE_WHEN_HASH_BYPASSED` or `D3D12_BYTECODE_FLAG_VALIDATE_ALL_BYTECODE` flags.

proposals/infra/INF-0004-validator-hashing.md
amarpMSFT: The flags let the user decide what they want.   All the details are there, but brief summary:
- by default only debug layer validates only BYPASS/BYPASS_PREVIEW shaders, printing messages and not failing. The debug layer doesn't alter pass/fail behavior by design (to the extent possible)
- the app can choose to have the runtime validate as well, with or without debug layer.  When the runtime is validating, it will fail (e.g. state object creation fail) when it sees an error.  So an app could choose this mode if it wants just an extra layer of checking on any shader it gets in the field in retail (e.g. without debug layer)
- the app can choose for the validation to apply to all shaders, even hashed ones, if it wants.   Maybe they want to try a new validator.
- The app can choose where the validator implementation comes from, incl from debug layer (when enabled) or dxil.dll/dxcompiler.dll whichever is present.  The search order can be tweaked.
- The app can choose to skip validation if an implementation isn't available, or have validation fail if an implementation isn't available
- If the debug layer and runtime are both doing validation (runtime for pass/fail, debug layer for printing messages), they will always select the same validator implementation so they are consistent.   This pattern is similar to how a huge amount of existing runtime+debug layer validation works - they are guaranteed to be consistent with each other.

llvm-beanz (changes requested):
I'm not comfortable with the scope of this change. I think this is a much more substantial expansion of the change than we should be taking as a PR to a previously accepted proposal.

My central concern here is that this is an expansion of when and how the DIXL validator is used by the runtime. The expansion of using the DXIL validator beyond just the debug layers raises concern to me because of the DXIL validator's uncertain future and known security weaknesses.

At a minimum I think we need to have some discussions to more clearly understand the motivations for these expansions so that we can appropriately prioritize and plan for support, maintenance, and associated risk.

proposals/infra/INF-0004-validator-hashing.md
damyanp: I can see that the proposal adds a huge amount of flexibility.  I'm unsure, though, why we're planning on doing any of that and why it is necessary?

amarpMSFT: This is defining what was originally proposed, which is that there would be controls on when to validate.

If we're sanctioning the workflow of bypassing the validator during author time, the least we should do is let apps validate at runtime.  There isn't any extra burden on HLSL team here since this is validation code the app can also run themselves manually.  But I think it is of great utility to make it straightforward to perform the validation in all the paths and engine ends up feeding shaders to D3D by simply having the runtime do it when it sees a shader (vs the app having to locate all the paths where they might be sourcing shaders from.

amarpMSFT:
> I'm not comfortable with the scope of this change. I think this is a much more substantial expansion of the change than we should be taking as a PR to a previously accepted proposal.
> 
> My central concern here is that this is an expansion of when and how the DIXL validator is used by the runtime. The expansion of using the DXIL validator beyond just the debug layers raises concern to me because of the DXIL validator's uncertain future and known security weaknesses.
> 
> At a minimum I think we need to have some discussions to more clearly understand the motivations for these expansions so that we can appropriately prioritize and plan for support, maintenance, and associated risk.

This isn't attempting to change the proposal, but rather concretely define what the accepted proposal mentioned, which is there would be runtime control on validation.  Note that the runtime will not have a built in validator, and by default it doesn't validate.  An app has to opt in to the runtime validating, because they want to see the invalid state creation fail.  And if they opt in, they have to provide a validator implementation.

proposals/infra/INF-0004-validator-hashing.md
amarpMSFT: I can cut this flag from consideration - would you prefer that I just delete the section from the spec, or leave it in but labelled with "This section is cut until D3D team can prove that avoiding an unnecessary shader copy really matters."

proposals/infra/INF-0004-validator-hashing.md
amarpMSFT: Per offline discussion, removed runtime validation support.

proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -210,21 +196,22 @@ runtime.
...
 
-If there is bytecode to validate, the runtime and/or debug layer attempt to 
-use a validator implementation from one of these dlls. Configuration flags 
-determine the order of preference for implementation and whether it is ok to 
-skip validation if an implementation is not found.
+If the debug layer finds errors when running the bytecode validator it only 
+reports messages, but doesn't fail shader creation.
 
 The following flags in `d3d12.h` show the validation control options.  
 `ID3D12Device` exposes `ID3D12BytecodeOptions` to set the flags, shown
```
damyanp, line 217: Should this configuration be set off the debug interface now, rather than the device?

proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -251,10 +235,8 @@ Flag | Definition
...
-`D3D12_BYTECODE_FLAG_PREFER_EXTERNAL_VALIDATOR_IMPLEMENTATION` | With this flag the order the debug layer and runtime search for validator implementations is: `dxil.dll`, `dxcompiler.dll`, finally `d3d12sdklayers.dll` (if the debug layer is enabled).  In the absence of this flag (default), the validator implementation search is this order: `d3d12sdklayers.dll` (if the debug layer is present), `dxil.dll`, `dxcompiler.dll`.  Notice that if the debug layer isn't enabled, this flag has no effect on runtime validation.
-`D3D12_BYTECODE_FLAGS_DEFAULT` | By default validation is only done on bytecode with the `BYPASS` or `PREVIEW_BYPASS` hash, only from the debug layer (so no failure on error).  Equivalent to the flags: `D3D12_BYTECODE_FLAG_VALIDATE_WHEN_HASH_BYPASSED` and `D3D12_BYTECODE_FLAG_VALIDATE_ONLY_IF_DEBUG_LAYER_ENABLED_NO_FAILING` |
+`D3D12_BYTECODE_FLAG_PREFER_EXTERNAL_VALIDATOR_IMPLEMENTATION` | With this flag the order the debug layer search for validator implementations is: `dxil.dll`, `dxcompiler.dll`, finally it's own implementation in `d3d12sdklayers.dll`.  In the absence of this flag (default), the debug layer just uses its own implementation.
```
damyanp, line 238: This makes me uneasy.  Anything involving searching for executable code and blindly executing it seems like it is ripe for exploitation.

I guess if it is only enabled for developer machines, or if the debug layer has been loaded, then it's less of an issue.

I don't understand the motivation for this feature at all.

proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -195,6 +163,93 @@ multiple-year-old preview compiler for generating final compiled shaders for a
...
+The validation mode can be configured via ID3D12DebugDevice1::SetDebugParameter()`:
+
+```C++
+typedef enum D3D12_DEBUG_DEVICE_PARAMETER_TYPE
+{
+    ...
+    D3D12_DEBUG_DEVICE_PARAMETER_BYTECODE_VALIDATION_MODE   
+} D3D12_DEBUG_DEVICE_PARAMETER_TYPE;
+
+typedef enum D3D12_BYTECODE_VALIDATE_MODE
```
damyanp, line 221: For other values:

* `D3D12_DEBUG_DEVICE_PARAMETER_FEATURE_FLAGS` --> `D3D12_DEBUG_FEATURE`
* `D3D12_DEBUG_DEVICE_PARAMETER_GPU_BASED_VALIDATION_SETTINGS` --> `D3D12_DEBUG_DEVICE_GPU_BASED_VALIDATION_SETTINGS`
* `D3D12_DEBUG_DEVICE_PARAMETER_GPU_SLOWDOWN_PERFORMANCE_FACTOR` --> `D3D12_DEBUG_DEVICE_GPU_SLOWDOWN_PERFORMANCE_FACTOR`

Following this pattern, I think the enum should be called:

`D3D12_DEBUG_DEVICE_BYTECODE_VALIDATE_MODE` or `D3D12_DEBUG_BYTECODE_VALIDATE_MODE`

proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -195,6 +163,93 @@ multiple-year-old preview compiler for generating final compiled shaders for a
...
+Apps can configure how the debug layer runtime validates bytecode passed to it. 
+
+The validation is the same code the compiler uses when it endorses bytecode at 
+compile time by applying a hash.  A copy of this validation code is in 
+the debug layer for it to use.  So it would typically be 
+redundant to validate again, but it can be useful to validate shaders that have 
+the `BYPASS` or `PREVIEW_BYPASS` hash, or have the option to force validation. 
+
+If the debug layer finds errors when running the bytecode validator it only 
+reports messages, but doesn't fail shader creation.
```
damyanp, line 210: Checking here that we don't want to enable a mode where shader creation fails if it fails validation?

proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -195,6 +163,93 @@ multiple-year-old preview compiler for generating final compiled shaders for a
...
+
+> Note: an important behavior change here is that shaders with zero'd hash data
+> will always be treated as invalid by the runtime and rejected.
+> 
+> An exception is for a period of time already existing preview shaders that used 
+> zero'd hash data will be treated like having the `PREVIEW_BYPASS` hash in that
+> preview runtimes will allow them to be used, at least until those preview shaders
+> and runtimes are no longer supported.  In contrast to `PREVIEW_BYPASS` shaders,
+> these legacy zero'd has shaders will not run have the validator invoked on them
+> by the runtime/debug layer since that wasn't expected of them.
```
damyanp, line 197: I think this section might have been overly edited, I'm not exactly sure what some of it is trying to say.  An attempt at clarifying:

```suggestion
> We have an exception to allow for legacy zero'd hash shaders. These were built
> using earlier previews and so they're treated as if they had the `PREVIEW_BYPASS` hash set,
> except that the validator will not be invoked on them since this wasn't expected at the time
> they were generated. This exception will be removed once runtimes that support these preview 
> shaders are no longer supported.
```

damyanp (changes requested):
Looking good - I have a couple of remaining suggestions / questions to be addressed.

proposals/infra/INF-0004-validator-hashing.md
amarpMSFT: This is what the runtime validation option I had was for. The debug layer / runtime interaction is typically it is the runtime that does the failing, and the debug layer's role is just an observer giving commentary.

Gets messy to have to make sure all codepaths work cleanly if the debug layer suddenly starts failing on behalf of the runtime.

I'd be fine bringing back the option for the runtime to fail, as long as debug layer is on.

If we bring this back, I'd switch the enum back to a flags field, so the fail on error can be layered on.

Jesse's argument is an app can write 20 lines of code if they want, to intercept the debug errors and fail if they want.  Certainly an unusual burden to put on them.

damyanp: Or I guess they could just configure the debug layer to break on that error message as well.

I thought I'd run into cases where enabling the debug layer would cause some API calls to fail that didn't previously.

proposals/infra/INF-0004-validator-hashing.md
amarpMSFT: Fixed

damyanp (approved):

proposals/infra/INF-0004-validator-hashing.md
amarpMSFT: Yes, we try to avoid situations where having the debug layer on changes behavior.  So I started here by proposing that you could ask for the runtime to fail independent of whether the debug layer is on or not.    But based on everyone's hesitation against having the runtime to validation backed off.

Yes, folks can configure the debug layer to break on the error.  Not quite as easy as it seems because there are tons of different codepaths where validation will be invoked from, and we'd have to try to guarantee that they always use the same error ID so folks can break on all possible instances of the failure.  There are some code patterns in the debug layer where there are error ID name specific to a shader stage, for instance, and those code paths would have to make an exception for this particular error.

llvm-beanz (approved):

proposals/infra/INF-0004-validator-hashing.md
amarpMSFT: I will try to do this when I implement.  If when the dust settles I'm confident it's just one ID to break on, I can update this spec to list that ID.

PR #292, opened 2024-07-31
Remove proposals that have been copied to wg-hlsl

damyanp:
This change removes the proposals that were copied to https://github.com/llvm/wg-hlsl as part of llvm/wg-hlsl#1.


coopp (approved):
NICE!  Cleanups!

damyanp:
Some possible things I could have done, but chose not to:

1. Replace the content of these files with a reference to the new location. I deleted them instead - figured there probably weren't that many old links around that would benefit from this.
2. Renumber the current INF-0004 proposal to INF-0001. Not 100% sure there's much reason to do this, plus I'd need to check if there are any stale issues that need updating after this.

llvm-beanz (approved):

PR #298, opened 2024-08-05
Add scalar layout compatability trait

llvm-beanz:
This adds language for determining if two types are _scalar layout compatible_ by defining a _scalarized representation_ which is a sequence of types that can then be compared.

The scalar representations of types will be used in several contexts throughout HLSL including:
* "Flattened" type casting
* Scalarized initiailization lists
* Defining additional type traits (homogeneous types)

### DIFF
--- a/specs/language/basic.tex
+++ b/specs/language/basic.tex
@@ -289,6 +289,37 @@ completed. The \texttt{void} type is used to signify the return type of a
 function that returns no value. Any expression can be explicitly converted to
 \texttt{void}.
 
+\Sub{Scalarized Type Compatability}{Basic.types.scalarized}
+
+\p All types \texttt{T} have a \textit{scalarized representation}, \(SR(T)\),
+which is a list of one or more types representing each scalar element of
+\texttt{T}.
+
+\p Scalarized representations are determined as follows:
+\begin{itemize}
+\item The scalarized representation of an array \texttt{T[n]} is \(SR(T_0), ..
+SR(T_n)\).
+
+\item The scalarized representation of a vector \texttt{vector<T,n>} is \(SR(T_0),
+.. SR(T_n)\).
+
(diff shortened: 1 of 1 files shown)


damyanp (approved):

specs/language/basic.tex
```diff
@@ -289,6 +289,37 @@
...
+\p All types \texttt{T} have a \textit{scalarized representation}, \(SR(T)\),
+which is a list of one or more types representing each scalar element of
+\texttt{T}.
+
+\p Scalarized representations are determined as follows:
+\begin{itemize}
+\item The scalarized representation of an array \texttt{T[n]} is \(SR(T_0), ..
+SR(T_n)\).
+
+\item The scalarized representation of a vector \texttt{vector<T,n>} is \(SR(T_0),
```
python3kgae, line 303: How about matrix type?

llvm-beanz: I haven't actually written any spec language anywhere for matrix types, but I guess I should.

specs/language/basic.tex
```diff
@@ -289,6 +293,40 @@
...
+
+\p Scalarized representations are determined as follows:
+\begin{itemize}
+\item The scalarized representation of an array \texttt{T[n]} is \(SR(T_0), ..
+SR(T_n)\).
+
+\item The scalarized representation of a vector \texttt{vector<T,n>} is \(T_0,
+.. T_n\).
+
+\item The scalarized representation of a matrix \texttt{matrix<T,n, m>} is
```
python3kgae, line 310: Is this always in row major order like in dxc or will have different order for row major and column major?

llvm-beanz: It doesn't matter for this because it is just the types of the elements. Whether you traverse row or column major you still get the same number of elements and the same one type repeated n * m times.

python3kgae (approved):

PR #300, opened 2024-08-12
Relocate cmp values for extended texture compares

pow2clk:
The comparison value was placed just after the locations in the implementation of SampleCmpBias and SampleCmpGrad. This moves the param in the spec to match.


python3kgae (approved):

coopp (approved):
Looks good.  I like fixes like theses.  Specs matching code is awesome!  :)

llvm-beanz:
@cassiebeckley, I believe this corrects the issue you encountered. Can you please take a look?

damyanp (approved):

pow2clk:
I'd prefer to wait to confirm that this addresses @cassiebeckley's concerns, but I don't want to linger too long either.

PR #301, opened 2024-08-12
Clarifications and alignment of vector truncations

llvm-beanz:
Previously the spec had allowed trunction conversions to produce cxvalues which enabled them to be used in output arguments. This is inconsistent with the DXC and FXC implementation, which disallows binding trunctions to output arguments.

The spec also previously did not allow trunction to scalar values. Initially that was excluded due to discussions around truncations and extensions of output arguments. Since DXC and FXC both disallow trunction in output arguments, it seems safe to support truncation to scalar here with a cooresponding warning.

This will allow the spec to more consistently align with the reference implementations.

This has no change to overload resolution except that a vector argument can be truncated to a scalar producing a Truncation, Truncation Promotion, or Truncation Conversion implicit conversion sequence.:

### DIFF
--- a/specs/language/conversions.tex
+++ b/specs/language/conversions.tex
@@ -142,7 +142,7 @@ smaller dimensions or to scalar values by dropping trailing elements.
 \end{itemize}
 
 \p The resulting value of vector truncation is comprised of elements \( [0..y)
-\), dropping elements \( [y+1..x) \).
+\), dropping elements \( [y..x) \).
 
 \p A prvalue of type \texttt{matrix<T,x,y>} can be converted to a prvalue of
 type:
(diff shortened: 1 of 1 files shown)


coopp (approved):
Looks good.. and pretty interesting too.  Didn't know truncation was performed here.

damyanp:
> can be truncated to a scalar producing a Trunctaion

typo in description

specs/language/conversions.tex
```diff
@@ -132,18 +132,36 @@
...
+smaller dimensions or to scalar values by dropping trailing elements.
+
+\p A prvalue of type \texttt{vector<T,x>} can be converted to a prvalue of type:
+\begin{itemize}
+\item \texttt{vector<T,y>} only if \( y < x \), or
+\item type \texttt{T}
+\end{itemize}
+
+\p The resulting value of vector truncation is comprised of elements \( [0..y)
+\), dropping elements \( [y+1..x) \).
```
damyanp, line 145: Since `[0..y]` doesn't include element `y`, should we say we're dropping `[y..x]`?

damyanp (approved):

PR #302, opened 2024-08-14
Added some details based on the runtime implementation of hash bypass.

amarpMSFT:
- The debug layer message ID on bytecode validation messages is D3D12_MESSAGE_ID_BYTECODE_VALIDATION_ERROR
- The app can query the runtime for support for the bypass hash (if the app doesn't know what runtime it's running on)


proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -206,9 +206,11 @@ redundant to validate again, but it can be useful to validate shaders that have
 the `BYPASS` or `PREVIEW_BYPASS` hash, or have the option to force validation. 
 
 If the debug layer finds errors when running the bytecode validator it only 
-reports messages, but doesn't fail shader creation.
+reports validator error strings, but doesn't fail shader creation.  Developers can 
+choose to to break on `D3D12_MESSAGE_ID_BYTECODE_VALIDATION_ERROR` to catch these specific
```
coopp, line 210: nit:  Extra 'to'?

coopp (approved):
Neat!  I like that this can be queried for support.

proposals/infra/INF-0004-validator-hashing.md
```diff
@@ -249,6 +251,39 @@ such as creating shaders where the options would apply.  Apps must do their
...
+        // D3D12_FEATURE_BYTECODE_BYPASS_HASH_SUPPORTED
+        // and fails the call.
+    }
+    else
+    {
+        // Bypass hash supported.
+
+        // Don't need to bother looking at the member 
+        // bypassHashSupport.Supported, since it is always true if
+        // the runtime succeeded the CheckFeatureSupport() call at all
```
damyanp, line 282: If that's the case, why have the member at all?

If there's some reason to have it, then I don't think we should encourage developers to ignore it.  That seems like a possible future compat headache.

damyanp (approved):
A couple of suggestions.

proposals/infra/INF-0004-validator-hashing.md
amarpMSFT: It's just how CheckFeatureSupportI() works, fills out a struct whose contents are based on the feature being queried.  The covention is for all caps queries, there's a struct definition for what the query is.  For simple ones, it might just be a bool, and we didn't just special case that to empty structs.  

For the hash bypass if the runtime understands the query it'll always set the member to true.  Old runtimes won't recognize the query at all, so it just happens you can rely on the call failing.

damyanp: Any reason the struct can't be empty?

Is is truly always going to be the case that from now on there'll be no release of D3D that doesn't support BYPASS?

For example, we have versions of D3D (and I think Agility SDK versions) that will always reject shaders without a hash, regardless of what developer mode is set to.

amarpMSFT: It could probably be empty, is it really important though?

>Is it truly always going to be the case that from now on there'll be no release of D3D that doesn't support BYPASS?
Yes, else we'd regress apps.

>For example, we have versions of D3D (and I think Agility SDK versions) that will always reject shaders without a hash, regardless of what developer mode is set to.

This would be how non preview runtimes work, and continues to be the case even for BYPASS_PREVIEW.  You'd have to enable experimental shader models, which would fail on non preview runtimes.

For the BYPASS hash, that's specifically for retail scenarios, so we have to keep supporting it.  Apps that are willing to at least use the latest AgilitySDK can be guaranteed forever that they can just use the BYPASS hash without even checking the cap at all.  Even for newer OSs that come along and will overtake the AgilitySDK the app shipped with.

damyanp: > It could probably be empty, is it really important though?

From an API design perspective, having a struct member that's documented as "we're pretty sure this will always be true, so you don't need to both checking the value of it" doesn't seem great to me.  If it isn't there at all then there's no confusion or strangeness.

amarpMSFT: Discussed with D3D team, and consensus is to not break precedent where we have other bool caps that are always true.  It happens that for those they happened to sit alongside other caps.  So you can't just rely on CheckFeatureSupport() pass/fail to know.   

D3D agreed we should just make the recommended calling pattern that BYPASS is supported if the call to CheckFeatureSupport() succeeded and the bool is true.   While that seems to give the false impression the bool might some day change, it isn't too big a deal, since apps that really care would have used new AgilitySDK and thus not even need to check the cap at all anyhow.

PR #311, opened 2024-08-30
Revise Usual Arithmetic Conversions

### LINKED ISSUE #308: Document Usual Arithmetic Conversions
[Language](https://github.com/microsoft/hlsl-specs/tree/main/specs/language)

The language specification does not document how to convert arguments for builtin operator expressions.

**Additional context**
See [C++ Usual Arithmetic Conversions](https://timsong-cpp.github.io/cppwp/n4868/expr.arith.conv)

llvm-beanz:
There are a few discrepancies between what I originally wrote here and what DXC does. These changes adjust to the actual behavior and more closely align with implicit conversion sequences.

Fixes #308

### DIFF
--- a/specs/language/expressions.tex
+++ b/specs/language/expressions.tex
@@ -27,14 +27,17 @@ conversions} are:
 \begin{itemize}
   \item If either operand is of scoped enumeration type no conversion is
   performed, and the expression is ill-formed if the types do not match.
-  \item If either operand is a \texttt{vector<T,X>}, vector extension is
-  performed with the following rules:
+  \item If either operand is a \texttt{vector<T,X>}, vector truncation or scalar
+  extension is performed with the following rules:
   \begin{itemize}
-    \item If both vectors are of the same length, no extension is required.
+    \item If both vectors are of the same length, no dimension conversion is
+    required.
     \item If one operand is a vector and the other operand is a scalar, the
-    scalar is extended to a vector via a Splat conversion (\ref{Conv.vsplat}).
+    scalar is extended to a vector via a Splat conversion (\ref{Conv.vsplat}) to
+    match the length of the vector.
     \item Otherwise, if both operands are vectors of different lengths, the
-    expression is ill-formed.
+    vector of longer length is truncated to match the length of the shorter
+    vector (\ref{Conv.vtrunc}).
   \end{itemize}
   \item If either operand is of type \texttt{double} or \texttt{vector<double,
   X>}, the other operator shall be converted to match.


specs/language/expressions.tex
```diff
@@ -27,14 +27,17 @@
...
-    \item If both vectors are of the same length, no extension is required.
+    \item If both vectors are of the same length, no dimension conversion is
+    required.
     \item If one operand is a vector and the other operand is a scalar, the
-    scalar is extended to a vector via a Splat conversion (\ref{Conv.vsplat}).
+    scalar is extended to a vector via a Splat conversion (\ref{Conv.vsplat}) to
+    match the length of the vector.
     \item Otherwise, if both operands are vectors of different lengths, the
-    expression is ill-formed.
+    vector of longer length is truncated to match the length of the shorter
```
bfavela, line 39: Why would you truncate to the shorter length instead of extend to the longer? Won't truncation lose information?

damyanp: Extension would require making up information though...

specs/language/expressions.tex
```diff
@@ -27,14 +27,17 @@
...
+    vector (\ref{Conv.vtrunc}).
```
damyanp, line 40: Checking understanding here:

```hlsl
float a;
float1 b;
float4 c;

a + b; // -> float1
a + c; // -> float4
b + c; // -> float1
```

llvm-beanz: Ooof.... these are some interesting cases, and they reveal some _really_ odd behaviors in DXC:

```hlsl
float a;
float1 b;
float4 c;

a + b; // -> float (the difference between float and float1 is generally negligible)
a + c; // -> float4
b + c; // -> float4 ? (This is nuts)
```

In that last case `float1 + float4` DXC generates something like this:

```
BinaryOperator 'float4':'vector<float, 4>' '+'
|-ImplicitCastExpr 'vector<float, 4>':'vector<float, 4>' <HLSLVectorSplat>
| `-ImplicitCastExpr 'float' <HLSLVectorToScalarCast>
|   `-ImplicitCastExpr 'float1':'vector<float, 1>' <LValueToRValue>
|     `-DeclRefExpr 'float1':'vector<float, 1>' lvalue Var 'b' 'float1':'vector<float, 1>'
`-ImplicitCastExpr 'float4':'vector<float, 4>' <LValueToRValue>
    `-DeclRefExpr 'float4':'vector<float, 4>' lvalue Var 'c' 'float4':'vector<float, 4>'
```

So the `float1` is getting truncated to a `float` _then_ extended out to a `float4`. That seems a bit bonkers to me, but it is unique to single element vectors. If we need to support that, we could add a rule converting `vector<T, 1>` to `T`.

damyanp: Checking I'm following right - my examples were correct according to this spec, but your comment lists the actual behavior in DXC?

Would the only way to actually observe the `b+c` case be that DXC and Clang would pick different overloads if it had to choose between a `float1` and `float4` one?  So maybe possible to contrive a case but it'd be pretty unlucky to run into it by accident?

llvm-beanz: Yes, that's correct. Your reading matches the spec, DXC does something different and odd.

I can see an argument that `a + b` should result in `float`, but `b + c` turning into `float4` feels really wrong to me. HLSL does not allow vector extension because it requires you to make things up. We do allow scalar->vector extending by repeating the value, but it just feels _really_ wrong to allow `vector<T, 1>` to extend to `vector<T, N`> as if it were a scalar extension...

damyanp: `a` is a scalar, and so gets extended by splatting, so this is equivalent to `float4 c = b + float4(a,a,a,a);`

IMO it's a bad thing because it means that `float1` is special is some cases.

Compare: https://godbolt.org/z/xKTWT1dff to https://godbolt.org/z/36K8Me3nh

In the first one it triggers the special `float1` case so you get a `float4` back.  In the second one the `float1` case isn't special and you get a `float1` back.

damyanp (approved):

specs/language/expressions.tex
llvm-beanz: @tex3d, do you have thoughts on how important the `float1`->`float4` conversion is to support?

We could add a new bullet before the line "\If either operand is a \texttt{vector<T,X>}" that says something like:
> If either operand is a `vector<T, 1>` it is implicitly converted to `T`, and the following rules apply as if the operand were of type `T`.

bfavela: If you're asking to find a shader that currently depends on the splatting behavior like Damyan described, it won't be that hard. I know I've seen it, numerous times.

But my question isn't regarding what the behavior is currently - I'm asking what we think the behavior SHOULD be.

Writing code that implicitly uses different sized types is really common, even in C++. And, AFAIK, C++ throws a warning, but will automatically extend the value for you - unsigned integer uses zero extension, signed integer uses signed extension and floats will upconvert to the larger type. Should we mimic that behavior or try to match what DXC is doing (Damyan's example shows that it's all over the place currently, so probably not).

llvm-beanz: My feeling is that we shouldn't have a special case for `vector<T,1>`, we should treat it like any other vector. HLSL does not allow extending a `float2` to a `float4` implicitly, so why should it extend a `float1` to a `float4` implicitly?

I suspect the most likely answer to why we should support this for `float1` is because if we don't a lot of existing code will break. Which probably means we should support it even if it makes me sad to special case single element vectors.

tex3d: I'm a bit torn on this one.  It feels like this partly depends on the usefulness of treating float1 strictly as a vector, different than float.  I also don't know where precedent lands here wrt. other languages with vectors, especially whether languages that enforce strict separation provide an important benefit, or if it's found to be more of an annoying rule with no upside to code authors.

Utilizing the difference for overloading only seems to introduce confusion.  For example, both FXC and DXC allow you to overload a function for both types (like `float` and `float1`), but will only unambiguously resolve the call when using a `float1`.  If you try to call the function with a scalar, it will consider the call ambiguous and emit an error.  This doesn't seem like a useful scenario to even support, since I don't think it's even possible to call the scalar overload in this case.

Wouldn't it be nice to make `float1` (and `vector<float, 1>`) equivalent to (or just alias to) `float` somehow?

llvm-beanz: I might play with that idea and get back to you... We could have a template specialization for `vector<T, 1>` resolve to `T`, and we could add some logic to our truncation casting implementation to enforce generating scalars instead of single element vectors.

There would be a breaking language change here because in DXC `float1` and `float` are not the same type, and if we make them the same type it could break existing code (admittedly only sadistic code).

bfavela (approved):

tex3d (approved):
Ok for now with follow-up of #319 for the vec1/scalar issues.

specs/language/expressions.tex
llvm-beanz: Marking as resolved to follow up in the filed issue.

PR #312, opened 2024-08-30
Rework some wording around out and inout parameters

llvm-beanz:
This somewhat future proofs some aspects of the language here by saying that [in]out parmaeter initialization is copy-initialization, but the writeback assignment is _assignment_. That allows for clarity around overloaded operators so that the argument life becomse something like this:

```
auto &ArgTmp = (Arg); // Evaluate the argument expr saving it.
T Param = ArgTmp;     // Copy-initialize Param with Arg's result.
call(Param);          // Call the function.
ArgTmp = Param;       // Assign back the Param result with =.
```

### DIFF
--- a/specs/language/expressions.tex
+++ b/specs/language/expressions.tex
@@ -188,7 +188,8 @@ declared as a member of.
 
 \p Parameters are either \textit{input parameters}, \textit{output parameters},
 or \textit{input/output parameters} as denoted in the called function's
-declaration (\ref{Decl.Function}).
+declaration (\ref{Decl.Function}). For all types of parameters the argument
+expressions are evaluated before the function call occurs.
 
 \p \textit{Input parameters} are passed by-value into a function. If an argument
 to an \textit{input parameter} is of constant-sized array type, the array is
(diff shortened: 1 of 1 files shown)


coopp (approved):
Looks good to me.

specs/language/expressions.tex
```diff
@@ -204,9 +205,11 @@
 undefined behavior to not explicitly initialize an \textit{output parameter}
 before returning from the function in which it is defined. The cxvalue created
 from an argument to an \textit{input/output parameter} is initialized through
-copy-initialization from the lvalue argument expression. In both cases, the
-cxvalue shall have the type of the parameter and the argument can be converted
-to that type through implicit or explicit conversion.
+copy-initialization from the lvalue argument expression, overload resolution may
+occur on argument initialization as if the expression \texttt{T Param = Arg}
+were evaluated. In both cases, the cxvalue shall have the type of the parameter
```
bfavela, line 210: This is a bit of a malformed sentence. Either it's a run-on or missing the word "if" before "The cxvalue created".

Also small nit: I think it should be "was evaluated", not "were evaluated", as "expression" is singular.

specs/language/expressions.tex
```diff
@@ -219,11 +222,14 @@
 \texttt{T[N]}.
 
 \p On expiration of the cxvalue, the value is assigned back to the argument
-lvalue expression following an inverted conversion if applicable. The argument
-expression must be of a type or able to convert to a type that has defined
-copy-initialization to and from the parameter type. The lifetime of the cxvalue
-begins at argument expression evaluation, and ends after the function returns. A
-cxvalue argument is passed by-address to the caller.
+lvalue expression using a resolved assignment expression as if the expression
+\texttt{Arg = Param} were written\footnote{The argument expression is not
```
bfavela, line 226: Is a space character needed between 'written' and '\footnote' here?

specs/language/expressions.tex
damyanp: I wonder if a `;` instead of a `comma` on line 208 might also fix it?  Or maybe it should just be two sentences.

damyanp (approved):

specs/language/expressions.tex
llvm-beanz: Without a space the footnote mark is right next to the end of the word, and it is still correctly parsed.
Screenshot 2024-09-04 at 6 59 10 PM

bfavela (approved):

PR #313, opened 2024-08-31
Cleanup wording on ordering of conversion ranks

llvm-beanz:
This wording went through a few iterations and it ended up in a silightly confused state. This just tries to simplify and clarify the ordering of the ranks.

### DIFF
--- a/specs/language/overloading.tex
+++ b/specs/language/overloading.tex
@@ -325,7 +325,7 @@ sequence with a \textbf{Conversion} conversion, the conversion is ranked as
 rank of each conversion.
 
 \p Conversion sequence ranks are ordered from better to worse as:
-\begin{itemize}
+\begin{enumerate}
   \item \textbf{Exact Match}
   \item \textbf{Extension}
   \item \textbf{Promotion}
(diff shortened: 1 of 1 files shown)


specs/language/overloading.tex
```diff
@@ -324,12 +324,18 @@
...
 rank of each conversion.
 
-\p Conversion sequence ranks are ordered such that \textbf{Exact
-Match} rank is better than \textbf{Exact Match Extension} rank, which is better
-than \textbf{Promotion} rank, which is better than \textbf{Extension} rank,
-which is better than \textbf{Conversion} rank, which is better than
-\textbf{Truncation} rank. The rank of a conversion sequence is the rank of the
-worst ranked conversion in the sequence.
+\p Conversion sequence ranks are ordered from better to worse as:
+\begin{itemize}
```
bfavela, line 328: Does "itemize" include a numbered list? Not a big deal either way, just curious. This is WAY more readable, though!

bfavela (approved):

coopp (approved):
Looks good to me

damyanp (approved):

specs/language/overloading.tex
llvm-beanz: The current formatting for `itemize` is:
Screenshot 2024-09-04 at 7 01 15 PM

I could instead use an `enumerate` block if you think numbering would be more clear.

bfavela: Really minor, but `enumerate` might be better, especially when you're dealing with an ordered list to begin with.

PR #315, opened 2024-09-10
[0024] Add Opacity Micromaps proposal

damyanp:
The Opacity Micromaps (OMM) feature allows DirectX Ray Tracing developers to allow ray-triangle hits to be trivially classified as miss or hit without having to invoke an any-hit shader.

This spec is narrowly focused on the HLSL compiler related aspects required to implement and test the feature from the compiler perspective.

### DIFF
--- a/proposals/NNNN-opacity-micromaps.md
+++ b/proposals/NNNN-opacity-micromaps.md
@@ -4,7 +4,7 @@
 
 ## Instructions
 
-* Proposal: [0023](0023-opacity-micromaps.md)
+* Proposal: [NNNN](NNNN-opacity-micromaps.md)
 * Author(s): [Tex Riddell](https://github.com/tex3d)
 * Sponsor: [Tex Riddell](https://github.com/tex3d)
 * Status: **Under Review**
--- a/proposals/NNNN-opacity-micromaps.md
+++ b/proposals/NNNN-opacity-micromaps.md
@@ -129,6 +129,14 @@ enum class RayFlag : uint32_t {
 };
 ```
 
+#### SPIR-V
+
+This change is comaptible with the
+[SPV_EXT_opactity_micromap](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_opacity_micromap.asciidoc)
+extension, where the new flag is `ForceOpacityMicromap2StateEXT`. It also,
+coincidentally, has the same number value as `ForceOMM2State`.
+
+
 ### Diagnostic Changes
 
 A warning diagnostic (default: warning) will be introduced and emitted when
--- a/proposals/0010-vk-buffer-ref.md
+++ b/proposals/0010-vk-buffer-ref.md
@@ -197,7 +197,7 @@ struct TestPushConstant_t
 
 float4 MainPs(void) : SV_Target0
 {
-      block_p g_p(g_PushConstants.root);
+      block_p g_p = block_p(g_PushConstants.root);
       g_p = g_p.Get().next;
       if ((uint64_t)g_pi == 0) // Null pointer test
           return float4(0.0,0.0,0.0,0.0);
--- a/proposals/0023-cxx11-base.md
+++ b/proposals/0023-cxx11-base.md
@@ -1,0 +1,56 @@
+<!-- {% raw %} -->
+
+# Adopt C++11 Base
+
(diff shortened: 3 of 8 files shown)


amarpMSFT (approved):

proposals/NNNN-opacity-micromaps.md
```diff
@@ -0,0 +1,376 @@
...
+In HLSL under DXC, these are defined as static const uint values:
+
+```hlsl
+static const uint RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS = 0x400;
+static const uint RAY_FLAG_FORCE_OMM_2_STATE = 0x400;
+```
+
+> Note: the fact that these flags have the same value is only a coincidence.
+
+### Interchange Format Additions
```
llvm-beanz, line 99: Is this feature compatible with the SPIR-V extension?
https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_opacity_micromap.asciidoc

I _think_ they are and we could implement this both for DX and SPIR-V, but we should confirm and make sure that is reflected in the spec.

damyanp: I can't see any reason that we wouldn't be able to implement this for SPIR-V as well.

@tex3d, can you double check please?

tex3d: From what I can tell, this should easily map to SPIR-V as well.

damyanp: @llvm-beanz: updated with a note about this, a few lines down from this conversation.

proposals/NNNN-opacity-micromaps.md
```diff
@@ -0,0 +1,384 @@
...
+Opacity Micromaps are globally not set, not allowed, or not supported.
+If set, this flag will be present in the returned flags of the
+[RayFlags()][rayflags] intrinsic function.
+
+In HLSL under DXC, these are defined as static const uint values:
+
+```hlsl
+static const uint RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS = 0x400;
+static const uint RAY_FLAG_FORCE_OMM_2_STATE = 0x400;
+```
```
amarpMSFT, line 95: The equivalent of PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS is needed for inline raytracing as well (e.g. RayQuery), which isn't affected by RAYTRACING_PIPELINE_FLAG*.

A proposed solution is a new template parameter to the RayQuery object.  Pasting the text already proposed for the full DXR spec for this:

### RayQuery flags

RayQuery flags are optionally specified as a second template parameter to [RayQuery](#rayquery) objects to enable static specialization - runtime behavior of [RayQuery::TraceRayInline()](#rayquery-tracerayinline) takes on the OR of dynamic and static/template ray flags as well as rayquery-specific flags.

```C++
enum RAYQUERY_FLAG : uint
{
    RAYQUERY_FLAG_NONE = 0x00, // default
    RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS = 0x01,
};
```

Value                               | Definition
---------                           | ----------
`RAYQUERY_FLAG_NONE` | No options selected.
`RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS` | If this flag is specified, the rayquery object supports Opacity Micromaps. If a triangle with an OMM is encountered during traversal with this flag cleared, behavior is undefined. This flag should not be set if there are no OMMs present, since it may incur a small penalty on traversal performance overall.  This is the rayquery equivalent for [TraceRayInline()](#rayquery-tracerayinline) of what the [D3D12_RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS](#d3d12_raytracing_pipeline_flags) flag accomplishes for non-inline [TraceRay()](#traceray).
`D3D12_RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS` | If this flag is specified, the pipeline supports Opacity Micromaps. This flag is part of the [Opacity Micromaps](#opacity-micromaps) feature.

damyanp: Is ok to just get the proposal in and then do a follow-up PR to make changes to it please?

llvm-beanz (approved):
LGTM

coopp (approved):
Looks good

PR #321, opened 2024-09-18
[Lang] Revisit memory model

llvm-beanz:
This change cleans up some of the wording in the memory model and clearly defines the terms _byte_, _memory location_, _memory access_, and _memory operation_.

These terms will be useful in writing the Classes chapter where the layout of objects needs to be defined.

### DIFF
--- a/specs/language/introduction.tex
+++ b/specs/language/introduction.tex
@@ -273,21 +273,52 @@ operations if the optimization alters the behavior of the program.
 
 \Sec{\acrshort{hlsl} Memory Models}{Intro.Memory}
 
-\p Memory accesses for \gls{sm} 5.0 and earlier operate on 128-bit slots aligned
-on 128-bit boundaries. This optimized for the common case in early shaders where
-data being processed on the GPU was usually 4-element vectors of 32-bit data
-types.
+\p The fundamental storage unit in HLSL is a \textit{byte}, which is comprised
+of 8 \textit{bits}. Each \textit{bit} stores a single value 0 or 1. Each byte
+has a unique \textit{memory location}, alternatively called an \textit{address}.
 
-\p On modern hardware memory access restrictions are loosened, and reads of
-32-bit multiples are supported starting with \gls{sm} 5.1 and reads of 16-bit
-multiples are supported with \gls{sm} 6.0. \gls{sm} features are fully
-documented in the \gls{dx} Specifications, and this document will not attempt to
-elaborate further.
+\p An object that is not of an intangible type will have an associated set of
+memory locations containing one or more location. For a bit-field the location
+will represent a maximal sequence of adjacent bit-fields all having nonzero
+width.
+
+\p Each read or write to a memory location is called a \textit{memory access}.
+Operations that perform memory accesses are called \textit{memory operation}. A
+memory operation may operate on one or more memory locations. A memory operation
+must not alter memory at a location not contained in the memory location set it
+is operating on\footnote{Two subtle notes here: (1) A bit-field's memory location
+includes adjacent bit-fields, so reads and writes to bit-fields are expected to
+read and write adjacent memory if they're within the same set of locations, (2)
+padding bits inside a structure are included in the memory location of the
+structure. Reads and writes of uninitialized memory may be undefined, but a
+write is allowed to stomp over padding.}.
+
+\p Two sets of memory locations, \texttt{A} and \texttt{B}, are said to
+\textit{overlap} each other if some memory location in \texttt{A} is also in
+\texttt{B} (\(A \cap B \neq \emptyset\)). Two objects are said to
+\textit{alias}, if their memory locations overlap.
 
 \Sub{Memory Spaces}{Intro.Memory.Spaces}
 
 \p \acrshort{hlsl} programs manipulate data stored in four distinct memory
-spaces: thread, threadgroup, device and constant.
+spaces: thread, threadgroup, device and constant. Memory spaces are logical
+abstractions over physical memory. Each memory space has a defined \textit{line
+width}, which specifies the minimum readable and writable size, and a
+\textit{minimum alignment}, which defines the smallest addressable increment of
+the memory space. The two values need not be the same, although they may be.
+
+\begin{note}
+  \p Memory accesses for many resource types in \gls{dx} operate on 128-bit
+  slots aligned on 128-bit boundaries. In the terms of this specification it
+  would be said that those memory spaces have a 128-byte \textit{line width},
+  and a 128-byte \textit{minimum alignment}.
+\end{note}
+
+\p A memory location in any space may overlap with another memory location in
+the same space. A memory location in thread or threadgroup memory may not
+overlap with memory locations in any other memory spaces. It is implementation
+defined if memory locations in other memory spaces alias with memory locations
+in different spaces.
 
 \SubSub{Thread Memory}{Intro.Memory.Spaces.Thread}
 
@@ -319,3 +350,9 @@ the host may have additional synchronization concerns for host visibility.
 \gls{lane}s executing on the device. Constant memory is read-only, and an
 implementation can assume that constant memory is immutable and cannot change
 during execution.
+
+\SubSub{Constant Memory}{Intro.Memory.Spaces.Overlap}
+
+\p The \textbf{Thread} and \textbf{Thread Group} memory spacees may not overlap
+with any other memory space. All addresses in either memory space are implied to
+not alias any address in any other memory space.


specs/language/introduction.tex
```diff
@@ -273,21 +273,52 @@
...
-\p On modern hardware memory access restrictions are loosened, and reads of
-32-bit multiples are supported starting with \gls{sm} 5.1 and reads of 16-bit
-multiples are supported with \gls{sm} 6.0. \gls{sm} features are fully
-documented in the \gls{dx} Specifications, and this document will not attempt to
-elaborate further.
+\p The fundamental storage unit in HLSL is a \textit{byte}, which is comprised
+of 8 \textit{bits}. Each \textit{bit} stores a single value 0 or 1. Each byte
+has a unique \textit{memory location}, alternatively called an \textit{address}.
+
+\p An object that is not of an intangible type will have an associated set of
```
damyanp, line 280: Can we avoid the double negative somehow?

```suggestion
\p An object that is of a tangible type will have an associated set of
```

damyanp: Can this ref the section on intangible types (if there is one)?

specs/language/introduction.tex
```diff
@@ -273,21 +273,52 @@
...
+memory locations containing one or more location. For a bit-field the location
+will represent a maximal sequence of adjacent bit-fields all having nonzero
+width.
+
+\p Each read or write to a memory location is called a \textit{memory access}.
+Operations that perform memory accesses are called \textit{memory operation}. A
```
damyanp, line 286:
```suggestion
Operations that perform memory accesses are called \textit{memory operations}. A
```

specs/language/introduction.tex
```diff
@@ -273,21 +273,52 @@
...
+memory operation may operate on one or more memory locations. A memory operation
+must not alter memory at a location not contained in the memory location set it
+is operating on\footnote{Two subtle notes here: (1) A bit-field's memory location
```
damyanp, line 289:
```suggestion
must not alter memory at a location not contained in the set of memory locations it
is operating on\footnote{Two subtle notes here: (1) A bit-field's memory location
```

I was about to ask for a definition of "memory location set", but wondered if maybe rewording would help avoid it?  I don't know why, but the existing language tripped me up a bit. Now I know what it means I have less trouble parsing it.

specs/language/introduction.tex
```diff
@@ -273,21 +273,52 @@
...
+includes adjacent bit-fields, so reads and writes to bit-fields are expected to
+read and write adjacent memory if they're within the same set of locations, (2)
+padding bits inside a structure are included in the memory location of the
+structure. Reads and writes of uninitialized memory may be undefined, but a
+write is allowed to stomp over padding.}.
```
damyanp, line 294: >  Reads and writes of uninitialized memory may be undefined, but a write is allowed to stomp over padding.

Does "undefined here" mean "undefined behavior"?  I'm also wondering about the "may" - is this a decision that hasn't been made yet? Or are there some cases where it is defined?

I'm also intrigued why writing to uninitialized memory could be undefined behavior, if that's what this means. Surely that's how you initialize the memory!

specs/language/introduction.tex
```diff
@@ -273,21 +273,52 @@
...
+abstractions over physical memory. Each memory space has a defined \textit{line
+width}, which specifies the minimum readable and writable size, and a
+\textit{minimum alignment}, which defines the smallest addressable increment of
+the memory space. The two values need not be the same, although they may be.
+
+\begin{note}
+  \p Memory accesses for many resource types in \gls{dx} operate on 128-bit
+  slots aligned on 128-bit boundaries. In the terms of this specification it
+  would be said that those memory spaces have a 128-byte \textit{line width},
+  and a 128-byte \textit{minimum alignment}.
```
damyanp, line 314: Is "128-byte" correct here?

specs/language/introduction.tex
```diff
@@ -319,3 +350,9 @@
 \gls{lane}s executing on the device. Constant memory is read-only, and an
 implementation can assume that constant memory is immutable and cannot change
 during execution.
+
+\SubSub{Constant Memory}{Intro.Memory.Spaces.Overlap}
+
+\p The \textbf{Thread} and \textbf{Thread Group} memory spacees may not overlap
```
damyanp, line 356:
```suggestion
\p The \textbf{Thread} and \textbf{Thread Group} memory spaces may not overlap
```

specs/language/introduction.tex
llvm-beanz: Doh! No, it is 128-bit...

specs/language/introduction.tex
llvm-beanz: I ended up removing this paragraph from this PR. I didn't actually define _object_ yet, so I think this should go in a section about the object model, not the memory model.

specs/language/introduction.tex
llvm-beanz: Reading and writing _of_ uninitialized memory (not to) is undefined. Basically if you allocate an `int` but don't set it, reading that value is undefined, writing the undefined value you read to memory is undefined.

I probably should say it `is` undefined, rather than `may be`.

specs/language/introduction.tex
```diff
@@ -273,21 +273,46 @@
...
+of 8 \textit{bits}. Each \textit{bit} stores a single value 0 or 1. Each byte
+has a unique \textit{memory location}, alternatively called an \textit{address}.
+
+\p Each read or write to a memory location is called a \textit{memory access}.
+Operations that perform memory accesses are called \textit{memory operations}. A
+memory operation may operate on one or more memory locations. A memory operation
+must not alter memory at a location not contained in the set of memory locations it
+is operating on\footnote{Two subtle notes here: (1) A bit-field's memory location
+includes adjacent bit-fields, so reads and writes to bit-fields are expected to
+read and write adjacent memory if they're within the same set of locations, (2)
```
spall, line 286: To me the usage of 'same set of locations' is a bit ambiguous. After a few reads I assume it means the adjacent bit-fields. Maybe something like:
```suggestion
read and write adjacent memory if they're within the adjacent bitfields' memory locations, (2)
```

spall: > To me the usage of 'same set of locations' is a bit ambiguous. After a few reads I assume it means the adjacent bit-fields. Maybe something like:

Maybe a caveat is needed to specify the adjacent bitfields' memory locations, do not ALSO include their adjacent bitfields.

specs/language/introduction.tex
```diff
@@ -273,21 +273,46 @@
...
+the memory space. The two values need not be the same, although they may be.
+
+\begin{note}
+  \p Memory accesses for many resource types in \gls{dx} operate on 128-bit
+  slots aligned on 128-bit boundaries. In the terms of this specification it
+  would be said that those memory spaces have a 128-bit \textit{line width},
+  and a 128-bit \textit{minimum alignment}.
+\end{note}
+
+\p A memory location in any space may overlap with another memory location in
```
spall, line 311: Does this mean you have say a MAU of 4 bytes, but you can have a memory location that accesses at byte 0 and a different memory location that accesses at byte 2?

specs/language/introduction.tex
```diff
@@ -273,21 +273,46 @@
...
+the same space. A memory location in thread or threadgroup memory may not
+overlap with memory locations in any other memory spaces. It is implementation
+defined if memory locations in other memory spaces alias with memory locations
```
spall, line 314: This sentence here makes me wonder if I'm interpreting the first sentence in this paragraph correctly.

specs/language/introduction.tex
```diff
@@ -319,3 +344,9 @@
 \gls{lane}s executing on the device. Constant memory is read-only, and an
 implementation can assume that constant memory is immutable and cannot change
 during execution.
+
+\SubSub{Constant Memory}{Intro.Memory.Spaces.Overlap}
+
+\p The \textbf{Thread} and \textbf{Thread Group} memory spaces may not overlap
```
spall, line 350: Is paragraph just stating part of what was said in the paragraph starting on line 311? 
Nit: I was initially confused by the use of both 'memory location' and 'address'; I had just glazed over the fact that they are defined to be the same thing. It might be nice to use one or the other.

specs/language/introduction.tex
```diff
@@ -273,21 +273,46 @@
...
-elaborate further.
+\p The fundamental storage unit in HLSL is a \textit{byte}, which is comprised
+of 8 \textit{bits}. Each \textit{bit} stores a single value 0 or 1. Each byte
+has a unique \textit{memory location}, alternatively called an \textit{address}.
+
+\p Each read or write to a memory location is called a \textit{memory access}.
+Operations that perform memory accesses are called \textit{memory operations}. A
+memory operation may operate on one or more memory locations. A memory operation
+must not alter memory at a location not contained in the set of memory locations it
+is operating on\footnote{Two subtle notes here: (1) A bit-field's memory location
```
bharadwajy, line 284: Defining the commonly understood term `memory access granularity` and specifying the access granularity to be a `byte` along with subsequent usage of the term may be a better option than using "bit-fields".

llvm-beanz: Bit-fields are a specific language structure that has unique properties in the memory model because of their unique packing behavior.

specs/language/introduction.tex
```diff
@@ -273,21 +273,82 @@
...
+width}, which specifies the minimum readable and writable size, and a
+\textit{minimum alignment}, which defines the smallest addressable increment of
+the memory space. The two values need not be the same, although they may be.
+
+\begin{note}
+  \p Memory accesses for many resource types in \gls{dx} operate on 128-bit
+  slots aligned on 128-bit boundaries. In the terms of this specification it
+  would be said that those memory spaces have a 128-bit \textit{line width},
+  and a 128-bit \textit{minimum alignment}.
+\end{note}
```
bogner, line 339: There are a few constraints around memory accesses in HLSL and DXIL that you're trying to abstract over here, but I'm not sure the "line width" idea captures them effectively. In some sense it might seem nice to boil down some similar rules into a simple concept, but it's worth noting why the rules are what they are and how they might change.

1. "Legacy" cbuffer and tbuffer layout. That is, the only cbuffer layout. Here, we have a constraint that came from 16-byte DXBC registers. The [cbufferLoadLegacy] docs call this a "row" in a comment, but I don't know that there's ever been any official terminology. Here, the rules on how big a single object or element of an array can be (128 bits) come from the packing rules, and it would arguably better to just write a section on those rules akin to the notes in [maraneshi's layout visualizer] rather than try to discuss this as a general rule about access size.

[maraneshi's layout visualizer]: https://maraneshi.github.io/HLSL-ConstantBufferLayoutVisualizer/
[cbufferLoadLegacy]: https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/DXIL.rst#cbufferloadlegacy

2. Data access via TypedBuffer. This is presumably where the "line width" idea comes from, but a lot of its complexity is unnecessary if we disallow "types that happen to fit" as type arguments to `Buffer<>`. Here, we have accesses to typed buffers and textures, and the operation that accesses them operates on a 4-element contained type. A `Buffer<float>` is really a `Buffer<float4>` that we only use one element from.
   
   This gets a bit confusing for 64 bit types. Notably, `Buffer<double4>` is not valid HLSL. However, this is really an implementation detail leaking through since the storage actually splits doubles up into int32 parts. So it's probably better to just think of `Buffer<double2>` as syntactic sugar for the casts and just call this kind of memory access what it is - access into a container of 4 at-most 32-bit values.

3. Vectors of more than 4 elements don't exist in HLSL. This is simply due to the fact that there's a fixed set of vector types and no way for a user to create their own. It isn't a meaningful rule, and in spaces like local device memory we really don't need any constraints on the language here. If it's possible to write a `double8` somehow in the future in the language and that isn't used in constant or typed buffers specifically, it's straightforward for implementations to do whatever they need to do to lower it. I don't think we want to define an artificial limit there.

So I guess TLDR I think we should simply say two separate things rather than trying to define "line width":
- Objects in Constant Buffer Memory are laid out according to the constant buffer packing rules (to be defined later). Elements of structures and arrays in this layout cannot exceed 128 bits in size.
- Memory accesses into typed buffers are defined to access 4 elements of at-most 32-bit values. (Also possibly a note about emulation 64 bit values, though this may not belong in this section)

Also note that I use "constant buffer" memory in my wording above, rather than "constant memory". We may want to keep that terminology available for if we ever do something in that space that doesn't carry the constant buffer legacy.

tex3d: I agree.

I think part of the problem is trying to view all resource accesses as if they are like native memory accesses from the shader, with only the address space placing constraints on alignment and such.  I think we can evolve constant memory and raw/structured buffer memory in this direction, but not typed/texture accesses.  For memory that goes in this direction, I don't think "line width" would be a concept we want to use/keep, and "minimum alignment" will be defined in other ways, rather than by some fixed value applied to a memory type.

Some notes:

> Elements of structures and arrays in this layout cannot exceed 128 bits in size.

I don't think I would agree with that.  First, it's a confusing use of the term "element" here.  Perhaps you had a different definition of "element" in mind than what I am interpreting here, but I struggle to think of a single definition that fits into this statement.  Plenty of elements of structures and arrays in HLSL that exceed 128-bits in size can be placed into the constant buffer.  You can declare a double4 (or array of such) in a constant buffer, which will use two rows for the vector.  It's just that structures, array elements, and any type that cannot fit within the remainder of a row will be started at the beginning of the next available row.  For some of these, that's part of the high-level packing rules, not necessarily something intrinsic to the DXIL interface.  For array elements, they must be 128-bit aligned to ensure that array indexing maps to an index in the DXIL legacy constant buffer load op without impacting the index of the component read from the result.

For legacy constant buffer load in DXIL, it's important to note that this load op doesn't mean all of the components are loaded - only the components that are extracted from the result structure need to be loaded.  It's a subtle difference, but important in certain circumstances, and mismatches the concept of "line width" as applied to constant buffers.  Think of the DXIL op as a compromise as there wasn't an easy way to express the thing that's expressed easily in DXBC asm like so: `CB0[0][0].yyyz` (only loads `y` and `z` components).

PR #325, opened 2024-09-25
[202y] Propose adding C++11-style constructors

llvm-beanz:
This proposal adds user-defined constructors and destructors to HLSL and defines how they should interact with semantics and global initialization. The intent is to align with C++11.

This proposal was originally proposed as https://github.com/microsoft/hlsl-specs/pull/5, however it didn't merge before the repository was moved from private to public, and it got stranded in GitHub's tooling and closed.

### DIFF
--- a/proposals/NNNN-constructors.md
+++ b/proposals/NNNN-constructors.md
@@ -77,7 +77,7 @@ default constructor executes.
 
 For example:
 
-```c++   
+```c++
   struct VSIn {
     uint ID : SV_VertexID;
     float F;
@@ -104,6 +104,12 @@ constructors. Default initialization is incompatible with HLSL annotations, and
 can only be used on variable declarations that either use C++ attributes for
 annotations, or have no annotation (see:
 [proposal for C++ Attributes](0002-cxx-attributes.md)).
+
+HLSL will adopt C++ rules for implicitly defining special member functions as
+defined in the C++ 11 specification under
+[`[special]`](https://timsong-cpp.github.io/cppwp/n3337/#special). This will
+include a default constructor, destructor, copy constructor, move constructor,
+and copy assignment operator, and move assignment operator.
 
 For example:
 
--- a/proposals/NNNN-constructors.md
+++ b/proposals/NNNN-constructors.md
@@ -1,10 +1,10 @@
 # Constructors
 
-* Proposal: [0007](0007-constructors.md)
+* Proposal: [NNNN](NNNN-constructors.md)
 * Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
 * Sponsor: [Chris Bieneman](https://github.com/llvm-beanz)
 * Status: **Under Consideration**
-* Planned Version: 202x
+* Planned Version: 202y
 
 ## Introduction
 
--- a/proposals/0007-constructors.md
+++ b/proposals/0007-constructors.md
@@ -1,0 +1,94 @@
+# Constructors
+
+* Proposal: [0007](0007-constructors.md)
+* Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
+* Sponsor: [Chris Bieneman](https://github.com/llvm-beanz)
+* Status: **Under Consideration**
+* Planned Version: 202x
+
+## Introduction
+
+As user defined data structures get more common and complex, the absence of constructors and destructors is more and more awkward. This proposal seeks to bring the full features of C++ initialization and destruction into HLSL.
+
+## Motivation
+
+Constructors are incredibly useful for ensuring data initialization and implementing common patterns like Resource-Acquisition-Is-Initialization (RAII).
+
+## Proposed solution
+
+Introduce the following features to HLSL:
+* C++ Constructors and Destructors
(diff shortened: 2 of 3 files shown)


Keenuts:
Hello!

As I was working on input/output/global/private variable using address spaces (to convert this to SPIR-V storage classes), I reached a bit of a question:

Do you have a plan on how to handle ctor/dtor overload resolution depending on `this` address space? Or can we safely assume the `address_space` clang attribute will always be forbidden in HLSL for Ctor/Dtor overloading? (Might be useful to keep it for variable declaration for inline SPIR-V)

Example:

```cpp
struct S {
    S(int value) : member(value) { }
    /* implicit: S(int value) __attribute__((address_space(10))) : member(value) { } */
    int member;
};

static S s_global(23); // AS=10, SC=Private

[numthreads(1, 1, 1)]
void main() {
    S s_local(24) /* AS=0, SC=Function */;
}
```

Today, OpenCL deals with this with an `addrspacecast` before passing `this`, and always calls the default address-space constructor.
If we go this way for our backend, we'll have to legalize the code to fixup all pointers storage classes (since s_global would be Private and not Function).

In my initial plan, I wanted to instantiate one version of the ctor for each used address-space. But this assumes the `address_space` Clang attribute is forbidden and that the user can not define 2 constructors depending on `this`' storage class.

llvm-beanz:
This problem is bigger than just constructors. The same problem occurs for any non-static member function. HLSL 2021 does not allow overloading functions based on the implicit object parameter (constness or address space). That likely does need to change in the future. That's out of the scope of this PR.

WRT your plans for upstream LLVM, I think that we likely need to do what OpenCL does in order to maintain source compatibility for HLSL 202x and earlier. We could consider changing this for 202y, but I don't want to rule out the possibility of supporting overloading by implicit object address space especially as we look to future C++ features like deducing-this, where we may want to allow users a lot more flexibility.

proposals/NNNN-constructors.md
```diff
@@ -0,0 +1,126 @@
...
+  }
+```
+
+Currently HLSL requires that all members of a user defined data type passed into
+an entry function must have semantic annotations to initialize the structure.
+With the addition of constructors and default value initialization, HLSL will
+instead require that every field either have a semantic annotation, or a default
+initialization through either a default constructor or a default initializer.
+
+HLSL will also introduce default initialization and defaulted and deleted
```
V-FEXrt, line 102: Can you add to the spec any commentary on the inclusion/exclusion of implicit constructors?

I'm assuming we wouldn't want to have any implicit constructors in HLSL but the presence of `delete`d constructors implies we'd need some implicit compiler constructor to delete

V-FEXrt: Right got it, forgot that we still need to contend with the ctors people were previously depending on

proposals/NNNN-constructors.md
```diff
@@ -0,0 +1,126 @@
...
+Each entry function will call the global constructors before executing any
+further instructions and the global destructor immediately before returning.
+
+In library shaders, the entry translation unit's global constructors are
+executed after the linked library global constructors, and the entry's global
+destructors are executed before linked library global destructors.
+
+Non-trivial global constructors **will not** be stripped from library shaders.
+Meaning, an unused global with a non-trivial constructor will be linked into the
+final linked shader binary. This conforms to the C++ standard definitions for
```
farzonl, line 50: Other than breaking with the ISO C++ standard, is there a reason we would not strip unused global constructors? Are we opening ourselves up to customers writing a post-link optimizer to getting rid of these constructors?

proposals/NNNN-constructors.md
```diff
@@ -0,0 +1,126 @@
...
+Constructor initialization of `groupshared` objects will be followed by a memory
+barrier and group sync. This comes with inherent performance costs, so shader
+authors should use constructors with `groupshared` objects sparingly.
+
+A `groupshared` user defined object that does not have a constructor, or has a
+constructor that does not initialize any of its fields, will not incur the
+memory barrier.
+
+Due to the inherent performance drawbacks of `groupshared` construction,
+warnings will be issued by default for any code that introduces a `groupshared`
```
farzonl, line 68: This block makes it seem like we don't want to support groupshared constructors. If thats the case why a warning and not an error? 

If I'm off base maybe we should enumerate supported vs unsupported group shared cases?

proposals/NNNN-constructors.md
llvm-beanz: If you want a global registration to occur on every shader run always, the simple approach would be a global constructor that doesn't get stripped.

If instead you have something that you only want to happen when it is used, the standard C++ approach is a singleton pattern.

Having the compiler not strip unused constructors enables the global-registration use case, but doesn't prevent a user from writing code that depends on use for initialization, it is just a different pattern.

proposals/NNNN-constructors.md
llvm-beanz: I think we want to support them. They can be super useful. We just want to also make sure users understand the cost of them.

farzonl (approved):

V-FEXrt (approved):

PR #329, opened 2024-10-01
Start working on HLSL initialization lists

llvm-beanz:
This one is a bit gnarly because it involves a lot of depth-first traversing of declarations, types, and initialization expresssions.

Questions and suggestions on how to clarify are greatly appreciated!

### DIFF
--- a/specs/language/declarations.tex
+++ b/specs/language/declarations.tex
@@ -87,7 +87,7 @@ subobjects in declaration order.
 depth-first traversal of the subobjects of an object following the defined
 subobject ordering.
 
-\p Each \textit{initializer-list} is comprised of zero or more
+\p Each \textit{braced initializer list} is comprised of zero or more
 \textit{initializer-clause} expressions, which in turn may be another
 initializer-list or an \textit{assignment-expression}. Each
 assignment-expression is an object, which may be a scalar or aggregate type. A
@@ -101,9 +101,9 @@ in the flattened initializer sequence which can be implicitly converted to the
 element's type.
 
 \p An initializer-list is invalid if the flattened initializer sequence contains
-less elements than the target object's flattened ordering, or if any initializer
-\(I_n\) cannot be implicitly converted to the corresponding element \(E_n\)'s
-type.
+more or fewer elements than the target object's flattened ordering, or if any
+initializer \(I_n\) cannot be implicitly converted to the corresponding element
+\(E_n\)'s type.
 
 \Sec{Function Definitions}{Decl.Function}
 \Sec{Attributes}{Decl.Attr}
--- a/specs/language/declarations.tex
+++ b/specs/language/declarations.tex
@@ -46,6 +46,65 @@
 
 \Sec{Declarators}{Decl.Decl}
 \Sec{Initializers}{Decl.Init}
+
+\p The process of initialization described in this section applies to all
+initializers regardless of the context.
+
+\begin{grammar}
+  \define{initializer}\br
+  brace-or-equal-initializer\br
+  \terminal{(} expression-list \terminal{)}\br
+
+  \define{brace-or-equal-initializer}\br
+  \terminal{=} initializer-clause\br
+  braced-init-list\br
+
+  \define{initializer-clause}\br
+  assignment-expression\br
+  braced-init-list\br
+
+  \define{braced-init-list}\br
+  \terminal{\{} initializer-list \opt{\terminal{,}} \terminal{\}}\br
+  \terminal{\{} \terminal{\}}\br
+
+  \define{initializer-list}\br
+  initializer-clause\br
+  initializer-list \terminal{,} initializer-clause\br
+\end{grammar}
+
+\Sub{Aggregate Initialization}{Decl.Init.Agg}
+
+\p An \textit{aggregate} is a vector, matrix, array, or class.
+
+\p The subobjects of an aggregate have a defined order. For vectors and arrays
+the order is increasing subscript order. For matrices it is increasing subscript
+order with the with the subscript nesting such that in the notation
+\texttt{Mat[M][N]}, the ordering is \(Mat[0][0]...Mat[0][N]...
+Mat[M][0]...Mat[M][N]\). For classes the order is base class, followed by member
+subobjects in declaration order.
+
+\p A \textit{flattened ordering} of subobjects can be produced by performing a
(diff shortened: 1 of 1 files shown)


specs/language/declarations.tex
```diff
@@ -46,6 +46,65 @@
...
+traversal over each assignment-expression in an initializer-list and performing
+a depth-first traversal accessing each subobject of the assignment-expression.
+
+\p An initializer-list is a valid initializer if for each element \(E_n\) in the
+target object's flattened ordering there is a corresponding initializer \(I_n\)
+in the flattened initializer sequence which can be implicitly converted to the
+element's type.
+
+\p An initializer-list is invalid if the flattened initializer sequence contains
+less elements than the target object's flattened ordering, or if any initializer
```
pow2clk, line 104: nit: fewer elements

pow2clk (commented):
I understand the depth-first traversal well enough. I think an example would be helpful to some, but perhaps that doesn't fit in this context?

specs/language/declarations.tex
```diff
@@ -46,6 +46,65 @@
...
+\textit{flattened initializer sequence} is constructed by a depth-first
+traversal over each assignment-expression in an initializer-list and performing
+a depth-first traversal accessing each subobject of the assignment-expression.
+
+\p An initializer-list is a valid initializer if for each element \(E_n\) in the
+target object's flattened ordering there is a corresponding initializer \(I_n\)
+in the flattened initializer sequence which can be implicitly converted to the
+element's type.
+
+\p An initializer-list is invalid if the flattened initializer sequence contains
```
spall, line 103: An initializer-list is valid if too many elements are provided?

specs/language/declarations.tex
```diff
@@ -46,6 +46,65 @@
...
+Mat[M][0]...Mat[M][N]\). For classes the order is base class, followed by member
+subobjects in declaration order.
+
+\p A \textit{flattened ordering} of subobjects can be produced by performing a
+depth-first traversal of the subobjects of an object following the defined
+subobject ordering.
+
+\p Each \textit{initializer-list} is comprised of zero or more
+\textit{initializer-clause} expressions, which in turn may be another
+initializer-list or an \textit{assignment-expression}. Each
```
spall, line 92: I haven't seen the whole grammar so maybe assignment-expression isn't used elsewhere, but it seems natural it would mean A=B (where = is used visually). I wonder if there is a better term for this. Like 'initializer-value'. 
Also, maybe this is visualized with '=', then disregard my comment. But maybe it would be good to have a definition of 'assignment-expression' included in the provided grammar to clarify.

specs/language/declarations.tex
```diff
@@ -46,6 +46,65 @@
...
+order with the with the subscript nesting such that in the notation
+\texttt{Mat[M][N]}, the ordering is \(Mat[0][0]...Mat[0][N]...
+Mat[M][0]...Mat[M][N]\). For classes the order is base class, followed by member
+subobjects in declaration order.
+
+\p A \textit{flattened ordering} of subobjects can be produced by performing a
+depth-first traversal of the subobjects of an object following the defined
+subobject ordering.
+
+\p Each \textit{initializer-list} is comprised of zero or more
```
spall, line 90: I mentioned this offline, I think 'initializer-list' is the wrong term here since it isn't defined in the grammar to be allowed to be empty.

specs/language/declarations.tex
llvm-beanz: Updated language in the latest version

specs/language/declarations.tex
llvm-beanz: The most recent update here tried to clarify the language and not use the grammar tokens since the C++ grammar is a bit bonkers to follow.

specs/language/declarations.tex
```diff
@@ -46,6 +46,66 @@
...
+\p Each \textit{braced initializer list} is comprised of zero or more
+\textit{initializer-clause} expressions, which is either another braced
+initializer list or an expression which generates a value that either is or can
+be implicitly converted to an rvalue. Each assignment-expression is an object,
+which may be a scalar or aggregate type. A \textit{flattened initializer
+sequence} is constructed by a depth-first traversal over each
+assignment-expression in an initializer-list and performing a depth-first
+traversal accessing each subobject of the assignment-expression.
+
+\p An initializer-list is a valid initializer if for each element \(E_n\) in the
```
V-FEXrt, line 99: Are we going to protect against cases where the author makes an unintentional (but correct from flat-ordering) error? The specific example I can think of is something like

```C++
struct Foo {
    int2 bar;
    int2 baz;
};

Foo myFoo = {{1, 2, 3}, 4};
```

Both `Foo` and `myFoo`'s init list have the flattened ordering of `<[int, int, int int>`  but there is a disagreement in what the subordering between the struct and what the user wrote which at the very least could lead to significant debugging confusion

V-FEXrt: oooh it looks like that's fully WAI per https://github.com/llvm/wg-hlsl/blob/main/proposals/0013-initializer-lists.md

V-FEXrt: My inclination is that anything that isn't fully flat or matching the real nested structure should likely be a warning in a potential future static analyzer

llvm-beanz: Yea, the current behavior in HLSL is a bit special. I tend to agree that we should have diagnostics, although generally language specifications don't dictate compiler warnings, that's left to the implementation.

For this case I'm not sure about the cost tradeoff between implementing diagnostics or just fixing the language as I suggest in [this proposal](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0005-strict-initializer-lists.md).

Part of the implementation plan is to support parsing the existing behavior into an AST that would be valid following C/C++ rules (where the braces matter), so that we could easily implement a rewriter by just printing off the AST. That would make it much easier to move to strict structure enforcement and minimize how much hacking we need to do in Clang's IR generation to support the old behavior.

V-FEXrt: > although generally language specifications don't dictate compiler warnings, that's left to the implementation

Yeah sorry. This was more a stream of thoughts than any actionable feedback. 

> or just fixing the language

ack. Yeah if we can just fix it then we should 100% do that

bogner (approved):

V-FEXrt (approved):

spall (approved):

PR #343, opened 2024-10-17
[202x] Propose adding vk::SampledTexture* types

cassiebeckley:
This proposal adds a set of `vk::SampledTexture*` types to represent SPIR-V's sampled image types. The intent is to replace and deprecate the `[[vk::combinedImageSampler]]` annotation.

I have created this with the extension template with the intent of using the fast-track process, as this is a platform-specific feature that is already supported by all Vulkan hardware and drivers, and it does not change any existing language behavior.

### DIFF
--- a/proposals/NNNN-vk-sampled-texture.md
+++ b/proposals/NNNN-vk-sampled-texture.md
@@ -36,7 +36,7 @@ sampler state argument.
 
 Consider this example pixel shader which uses a texture:
 
-```
+```hlsl
 Texture2D tex0 : register(t0);
 SamplerState s : register(s0);
 
@@ -48,7 +48,7 @@ float4 main(float2 uv: TEXCOORD) : SV_Target {
 Using the proposed types, this could be rewritten to use a combined image
 sampler:
 
-```
+```hlsl
 vk::SampledTexture2D tex0 : register(t0);
 
 float4 main(float2 uv: TEXCOORD) : SV_Target {
@@ -59,7 +59,7 @@ float4 main(float2 uv: TEXCOORD) : SV_Target {
 This is simpler and a more accurate representation of the underlying interface
 than using the existing annotation, which looks like:
 
-```
+```hlsl
 [[vk::combinedImageSampler]]
 Texture2D tex0 : register(t0);
 [[vk::combinedImageSampler]]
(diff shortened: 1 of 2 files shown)


proposals/NNNN-vk-sampled-texture.md
```diff
@@ -0,0 +1,174 @@
...
+Vulkan's combined image sampler types (`OpTypeSampledImage`).
+
+## Motivation
+
+The [existing annotation](https://github.com/microsoft/DirectXShaderCompiler/wiki/Vulkan-combined-image-sampler-type)
+for combined image samplers (`[[vk::combinedImageSampler]]`) was designed with
+the explicit goal of avoiding defining new HLSL types and functions. This was
+intended to minimize the number of changes to the frontend needed to implement
+the feature. However, it is verbose and confusing for users, and we do not
+intend to implement it in upstream Clang.
```
s-perron, line 28: I would mention that the implementation requires an transformation to run that pairs up textures and samplers, but it cannot work in all cases. This can lead to very subtle errors when, for example, the same sampler is used with two different textures.

s-perron (commented):
This generally look good to me. The only other detail is to consider how this will be implemented in clang. We might need a new attribute for the resource handle. See https://github.com/llvm/wg-hlsl/pull/76.

s-perron (approved):

proposals/NNNN-vk-sampled-texture.md
```diff
@@ -0,0 +1,184 @@
+<!-- {% raw %} -->
+
+# Vulkan Sampled Textures
+
+* Proposal: [NNNN](NNNN-vk-sampled-texture.md)
+* Author(s): [Cassandra Beckley](https://github.com/cassiebeckley)
+* Sponsor: [Steven Perron](https://github.com/s-perron) and TBD
+* Status: **Under Consideration**
+* Required Version: HLSL 202x
```
llvm-beanz, line 9: Per our discussion:

Marking the attribute that you want to remove deprecated in HLSL 2021, then remove it in HLSL 202x seems like a good path forward.

For the new types, we can use a `__has_feature` check in the preprocessor and make it available in all language modes.

s-perron (approved):

llvm-beanz (approved):

PR #358, opened 2024-12-16
Specification language to describe flat conversions and aggregate splats.

spall:
Specification language to describe flat conversions and aggregate splats.

### DIFF
--- a/specs/language/conversions.tex
+++ b/specs/language/conversions.tex
@@ -118,6 +118,19 @@ prvalue of arithmetic or unscoped enumeration type can be converted to a prvalue
 of boolean type. A zero value is converted to \texttt{false}; all other values
 are converted to \texttt{true}.
 
+\Sec{Flat conversion}{Conv.flat}
+
+\p There is a flattened order of subobjects for an aggregate, \texttt{A},
+\texttt{A$^\prime$}, as described in section \ref{Decl.Init.Agg}.  An aggregate,
+\texttt{A}, can be flat converted to an aggregate, \texttt{B}, whose flattened
+form is \texttt{B$^\prime$}, if the following conditions are true: 
+\begin{itemize}
+\item The length of \texttt{A$^\prime$}, \texttt{N}, is greater than or equal to the
+length of \texttt{B$^\prime$}, \texttt{M}.
+\item For every index \texttt{I} from \texttt{0} to \texttt{M-1}, there exists an
+implicit conversion from \texttt{A$^\prime$[I]} to \texttt{B$^\prime$[I]}.
+\end{itemize}
+
 \Sec{Vector splat conversion}{Conv.vsplat}
 
 \p A glvalue of type \texttt{T} can be converted to a cxvalue of type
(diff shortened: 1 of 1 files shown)


spall:
> @microsoft-github-policy-service agree company="Microsoft"

@microsoft-github-policy-service agree company="Microsoft"

specs/language/conversions.tex
```diff
@@ -130,6 +143,42 @@
...
 value replicated into each element of the destination.
 
+\p A glvalue of type \texttt{vector<T,1>} can be converted to a cxvalue of type
+\texttt{vector<T,x>} or a prvalue of type \texttt{vector<T,1>} can be converted
+to a prvalue of type \texttt{vector<T,x>}. The destination value is the value in
+the source vector replicated into each element of the destination.
+
+\Sec{Array and Class splat conversion}{Conv.asplat}
+
+\p A glvalue of type \texttt{T} can be converted to a cxvalue of type
```
llvm-beanz, line 153: I believe that this conversion can only produce a prvalue, not a cxvalue. The only place cxvalues are created is for `inout` parameters, and I believe this code is invalid:

```hlsl
struct S {
    float V[4];
};

void fn(inout float4 F) {}

export void call() {
    S V = {1.xxxx};
    fn((float4)V);
    fn(V);
}
```
see: https://godbolt.org/z/Ms4ocrGMf

specs/language/conversions.tex
```diff
@@ -130,6 +143,42 @@
...
+flattened \texttt{S}.
+
+\p A glvalue of type \texttt{vector<T,1>} can be converted to a cxvalue of type
+\texttt{struct S} or a prvalue of type \texttt{vector<T,1>} can be converted to a
+prvalue of type \texttt{struct S}, if there are valid conversions from
+\texttt{T} to each \texttt{U} in flattened \texttt{S}. The destination value is
+the value in the source vector replicated into each element of the flattened
+\texttt{S}.
+
+\p A glvalue of type \texttt{T} can be converted to a cxvalue of type
```
llvm-beanz, line 168: There is an odd case for vector<->array conversions where we can create a cxvalue, but only from an explicit cast:

```hlsl
void fn(inout float4 F) {}

export void call() {
    float V[4] = {1.xxxx};
    fn((float4)V); // valid
    fn(V); // invalid
}
```
see: https://godbolt.org/z/c3YcGWPex

specs/language/conversions.tex
```diff
@@ -118,6 +118,19 @@
 of boolean type. A zero value is converted to \texttt{false}; all other values
 are converted to \texttt{true}.
 
+\Sec{Flat conversion}{Conv.flat}
+
+\p There is a flattened order of subobjects for an aggregate, \texttt{A},
```
llvm-beanz, line 123: I think this language is fine for now, but we should probably have some common definitions and unify this with the language in [*Decl.Init.Agg*](https://microsoft.github.io/hlsl-specs/specs/hlsl.html#Decl.Init.Agg).

I filed #372 to track that separately.

specs/language/conversions.tex
```diff
@@ -130,6 +143,42 @@
...
+\texttt{T} to each \texttt{U} in flattened \texttt{S}. The destination value is
+the value in the source vector replicated into each element of the flattened
+\texttt{S}.
+
+\p A glvalue of type \texttt{T} can be converted to a cxvalue of type
+\texttt{U[]} or a prvalue of type \texttt{T} can be converted to a
+prvalue of type \texttt{U[]}, if there are valid conversions from
+\texttt{T} to each \texttt{V} in flattened \texttt{U}. The destination value
+is the source value \texttt{T} replicated into each element of the flattened
+\texttt{U[]}.
```
llvm-beanz, line 173: `U[]` is not the same type as `U[N]`, so we should be clear we're talking about sized arrays here.

llvm-beanz (commented):
Please doublecheck which conversions are allowed to produce cxvalues vs which conversions can only produce prvalues.

specs/language/conversions.tex
spall: This currently fails in the clang version.

specs/language/conversions.tex
spall: Should this be changed to 'Elementwise conversion'?

specs/language/conversions.tex
spall: Issue to track this: https://github.com/llvm/llvm-project/issues/131938

specs/language/conversions.tex
```diff
@@ -130,6 +155,37 @@
...
+\p A prvalue of type \texttt{vector<T,1>} can be converted
+to a prvalue of type \texttt{vector<T,x>}. The destination value is the value in
+the source vector replicated into each element of the destination.
+
+\Sec{Array and Class splat conversion}{Conv.asplat}
+
+\p A prvalue of type \texttt{T} can be converted to a
+prvalue of type \texttt{struct S}, if there are valid conversions from
+\texttt{T} to each \texttt{U} in flattened \texttt{S}, see section
+\ref{Decl.init.Agg} for description of a flattened ordering. The destination
```
V-FEXrt, line 167: `\ref{Decl.init.Agg}` is resolving to `???` in the rendered document

V-FEXrt (commented):
LGTM

Will leave Chris to approve

specs/language/conversions.tex
spall: Could be that section isn't defined or their is a problem in the build process; You really have to run pdflatex like 2+ times. I'll check which.

spall: I confirmed it is a problem with the build process not the .tex file itself.

V-FEXrt (approved):

specs/language/conversions.tex
```diff
@@ -162,14 +218,14 @@
 
 \Sec{Component-wise conversions}{Conv.cwise}
 
-\p A glvalue of type \texttt{vector<T,x>} can be converted to a cxvalue of type
+\p A glvalue of type \texttt{vector<T,x>} can be converted to a prvalue of type
```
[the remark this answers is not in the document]
spall: It looks like this edit and the one below were made in error.

llvm-beanz (approved):

PR #361, opened 2025-01-03
Rework hlsl-vector-type into two specs

pow2clk:
This splits the spec into two. dxil-vectors concerns the addition of vectors to DXIL only. hlsl-long-vector-type relates to the addition of long vectors in the HLSL language and also for select DXIL intrinsics.

Throughout, this adds additional details concerning testing and support. It makes a few alterations to the originally proposed behavior particularly concerning the loading and storing of long vectors whether from/to raw buffers or groupshared variables. The latter intrinsics were dropped entirely in favor of existing assignment operations being lowered to appropriate operations.
Long vectors are allowed in structs and non-entry function signatures and disallowed in shader signatures, cbuffers/tbuffers, and as elements of non-raw buffers.

Note that the use of 6.9 is a placeholder for the release vehicle for this feature.

### DIFF
--- a/proposals/0026-hlsl-long-vector-type.md
+++ b/proposals/0026-hlsl-long-vector-type.md
@@ -1,11 +1,11 @@
 <!-- {% raw %} -->
 
-* Proposal: [0026-HLSL-Vectors](0026-hlsl-vector-type.md)
+* Proposal: [0026-HLSL-Vectors](0026-hlsl-long-vector-type.md)
 * Author(s): [Anupama Chandrasekhar](https://github.com/anupamachandra)
 * Sponsor: [Damyan Pepper](https://github.com/damyanp)
 * Status: **Under Consideration**
 
-# HLSL Vectors
+# HLSL Long Vectors
 
 ## Introduction
 
--- a/proposals/0026-hlsl-long-vector-type.md
+++ b/proposals/0026-hlsl-long-vector-type.md
@@ -153,7 +153,7 @@ Of the above list, the following will produce the appropriate unary, binary, or 
 
 #### Disallowed vector intrinsics
 
-* Only applicable to for shorter vectors: AddUint64, asdouble, asfloat, asfloat16, asint, asint16, asuint, asuint16, D3DCOLORtoUBYTE4, cross, distance, dst, faceforward, length, normalize, reflect, refract, NonUniformResourceIndex
+* Only applicable to shorter vectors: AddUint64, asdouble, asfloat, asfloat16, asint, asint16, asuint, asuint16, D3DCOLORtoUBYTE4, cross, distance, dst, faceforward, length, normalize, reflect, refract, NonUniformResourceIndex
 * Only useful for disallowed variables: EvaluateAttributeAtSample, EvaluateAttributeCentroid, EvaluateAttributeSnapped, GetAttributeAtVertex
 
 ### Debug Support
--- a/proposals/0026-hlsl-long-vector-type.md
+++ b/proposals/0026-hlsl-long-vector-type.md
@@ -27,7 +27,7 @@ To take advantage of specialized hardware that can accelerate longer vector oper
 
 ## Proposed solution
 
-Enable vectors of length between 5 and 128 inclusive in HLSL using existing template-based vector declarations.
+Enable vectors of length between 5 and 1024 inclusive in HLSL using existing template-based vector declarations.
 Such vectors will hereafter be referred to as "long vectors".
 These will be supported for all elementwise intrinsics that take variable-length vector parameters.
 For certain operations, these vectors will be represented as native vectors using
@@ -304,7 +304,7 @@ The subscript access is flexible and generic and makes other accessors redundant
 ## Open Issues
 
 * Q: Is there a limit on the Number of Components in a vector?
-  * A: 128. It's big enough for some known uses.
+  * A: 1024. It's big enough for known uses.
 There aren't concrete reasons to restrict the vector length.
 Having a limit facilitates testing and sets expectations for both hardware and software developers.
 
--- a/proposals/0026-hlsl-long-vector-type.md
+++ b/proposals/0026-hlsl-long-vector-type.md
@@ -1,65 +1,81 @@
 <!-- {% raw %} -->
 
-* Proposal: [0026-HLSL-Vectors](0026-hlsl-long-vector-type.md)
-* Author(s): [Anupama Chandrasekhar](https://github.com/anupamachandra)
-* Sponsor: [Damyan Pepper](https://github.com/damyanp)
-* Status: **Under Consideration**
-
 # HLSL Long Vectors
+
+* Proposal: [0026-HLSL-Vectors](0026-hlsl-vector-type.md)
+* Author(s): [Anupama Chandrasekhar](https://github.com/anupamachandra), [Greg Roth](https://github.com/pow2clk)
+* Sponsor: [Greg Roth](https://github.com/pow2clk)
+* Status: **Under Consideration**
 
 ## Introduction
 
-HLSL has supported vectors in a limited capacity (int3, float4, etc.), and these are scalarized in DXIL; small vectors while useful in a traditional graphics context do not scale well with the evolution on HLSL as a more general purpose language targetting Graphics and Compute. Notably, with the ubiquitous adoption of machine learning techniques which often get expressed as vector-matrix operations, there is a need for supporting larger vector sizes in HLSL and preserving these vector objects at the DXIL level to take advantage of specialized hardware that can accelerate vector operations.
+HLSL has supported vectors in a limited capacity (int3, float4, etc.).
+These are scalarized in DXIL.
+While they are useful in a traditional graphics context,
+ small vectors do not scale well with the evolution of HLSL as a more general purpose language targeting Graphics and Compute.
+Notably, the adoption of machine learning techniques expressed as vector-matrix operations require larger vector sizes to be representable in HLSL.
+To take advantage of specialized hardware that can accelerate vector operations,
+ these and other vector objects need to be preserved at the DXIL level.
 
 ## Proposed solution
 
-Enable vectors of longer length in HLSL and preserve the vector type in DXIL.
+Enable vectors of length greater than 4 in HLSL using existing template-based vector declarations.
+Preserve the vector type in DXIL.
 
 ## Detailed design
 
-### HLSL vectors `vector<T, N>`
+### HLSL vectors
 
-Currently HLSL allows `vector<T, N> name;` where `T` is any [scalar](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-scalar) type and `N`, number of
-components, is a positive integer less than or equal to 4. See current definition [here](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-vector). 
-This proposal extends this support to longer vectors (beyond 4). 
+Currently HLSL allows declaring vectors using a templated representation:
+
+```hlsl
+vector<T, N> name;
+```
+
+`T` is any [scalar](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-scalar) type.
+`N` is the number of components and must be an integer between 1 and 4 inclusive.
+See the vector definition [documentation](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-vector) for more details.
+This proposal adds support for vectors of length greater than 4.
 
 The default behavior of HLSL vectors is preserved for backward compatibility, meaning, skipping the last parameter `N`
 defaults to 4-component vectors and the use `vector name;` declares a 4-component float vector, etc. More examples
 [here](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-vector).
+Declarations of vectors longer than 4 require the use of the template declaration.
+Unlike vector sizes between 1 and 4, no shorthand declarations are provided.
 
 The new vectors will be supported in all shader stages including Node shaders. There are no control flow or wave
-uniformity requirements, but implementations may specify best practices in certain uses for optimal performance. 
+uniformity requirements, but implementations may specify best practices in certain uses for optimal performance.
 
-**Restrictions on the uses of vectors with N > 4** 
+Restrictions on the uses of vectors with N > 4:
 
 * Vectors with length greater than 4 are not permitted inside a `struct`.
 * Vectors with length greater than 4 are not permitted as shader input/output parameters.
 
-**Constructing vectors**
+#### Constructing vectors
 
-HLSL vectors can be constructed through initializer lists and constructor syntax initializing or by assignment.
+HLSL vectors can be constructed through initializer lists, constructor syntax initialization, or by assignment.
 
 Examples:
 
-``` 
-vector<uint, 5> vecA = {1, 2, 3, 4, 5}; 
+``` hlsl
+vector<uint, 5> vecA = {1, 2, 3, 4, 5};
 vector<uint, 6> vecB = vector<uint, 6>(6, 7, 8, 9, 0, 0);
 uint4 initval = {0, 0, 0, 0};
 vector<uint, 8> vecC = {uint2(coord.xy), vecB};
 vector<uint, 6> vecD = vecB;
 ```
 
-**Load and Store vectors from Buffers/Arrays**
+#### Load and Store vectors from Buffers/Arrays
 
-For loading and storing N-dimensional vectors from ByteAddressBuffers we use the `LoadN` and `StoreN` methods, extending
-the existing Load/Store, Load2/Store2, Load3/Store3 and Load4/Store4 methods.
+For loading and storing N-dimensional vectors from ByteAddressBuffers we use the templated load and store methods
+by providing a vector type of the required size as the template parameter.
 
-``` 
+```hlsl
 // Load/Store from [RW]ByteAddressBuffers
 RWByteAddressBuffer myBuffer;
 
-vector<uint, N> val = myBuffer.LoadN(uint StartOffsetInBytes); 
-myBuffer.StoreN<T>(uint StartoffsetInBytes, vector<T, N> stVec);
+vector<T, N> val = myBuffer.Load< vector<T, N> >(uint StartOffsetInBytes);
+myBuffer.Store< vector<T, N> >(uint StartoffsetInBytes, vector<T, N> stVec);
 
 // Load/Store from groupshared arrays
 groupshared T inputArray[512];
@@ -69,25 +85,32 @@ Load(vector<T,N> ldVec, groupshared inputArray, uint offsetInBytes);
 Store(vector<T,N> stVec, groupshared outputArray, uint offsetInBytes);
 ```
 
-**Operations on vectors** 
+#### Operations on vectors
 
-Support all HLSL intrinsics that are important as activation functions: fma, exp, log, tanh, atan, min, max, clamp, and
-step. Eventually support all HLSL operators and math intrinsics that are currently enabled for vectors.
+Support all HLSL intrinsics that are important as activation functions:
+
+* fma
+* exp
+* log
+* tanh
+* atan
+* min
+* max
+* clamp
+* step
+
+Eventually support all HLSL operators and math intrinsics that are currently enabled for vectors.
 
 Refer to the HLSL spec for an exhaustive list of [Operators](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators) and [Intrinsics](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-intrinsic-functions).
 
-Note: Additionally any mathematical operations missing from the above list but needed as activation functions for neural
-network computations will be added.
-
 ### Debug Support
-First class debug support for HLSL vectors. Emit `llvm.dbg.declare` and `llvm.dbg.value` intrinsics that can be used by tools for better debugging experience. Open Issue: Handle DXIL scalarized and vector paths. 
 
+First class debug support for HLSL vectors. Emit `llvm.dbg.declare` and `llvm.dbg.value` intrinsics that can be used by tools for better debugging experience. Open Issue: Handle DXIL scalarized and vector paths.
 
 ### Diagnostic Changes
 
 * Additional error messages for illegal or unsupported use of arbitrary length vectors.
 * Remove current bound checks (N <= 4) for vector size in supported cases, both HLSL and DXIL.
-
 
 ### Validation Changes
 
@@ -106,7 +129,6 @@ Open Issue: Can implementations support vector DXIL?
 
 ### Minimum Support Set
 
-
 ## Testing
 
 * How will correct codegen for DXIL/SPIRV be tested?
@@ -117,30 +139,37 @@ Open Issue: Can implementations support vector DXIL?
 * How will the execution results be tested?
 * A: *HLK tests*
 
-
 ## Alternatives considered
 
-Our original proposal introduced an opaque Cooperative Vector type to HLSL to limit the scope of the feature to small
-neural network evaluation and also contain the scope for testing. But aligning with the long term roadmap of HLSL to
-enable generic vectors, it makes sense to not introduce a new datatype but use HLSL vectors, even if the initial
-implementation only exposes partial functionality.
+The original proposal introduced an opaque type to HLSL that could represent longer vectors.
+This would have been used only for cooperative vector operations.
+This would have limited the scope of the feature to small neural network evaluation and also contain the scope for testing some.
+
+Representing vectors used in neural networks as LLVM vectors also allows leveraging existing optimizations.
+This direction also aligns with the long term roadmap of HLSL to enable generic vectors.
+Since the new data type would have required extensive testing as well,
+the testing burden saved may not have been substantial.
+Since these vectors are to be added eventually anyway, the testing serves multiple purposes.
+It makes sense to not introduce a new datatype but use HLSL vectors,
+even if the initial implementation only exposes partial functionality.
 
 ## Open Issues
+
 * Q: Is there a limit on the Number of Components in a vector?
-* A: Chose a number based on precedents set by other languages. Support atleast 128.
+  * A: Chose a number based on precedents set by other languages. Support atleast 128.
 * Q: Usage restrictions
-* A: *General vectors (N > 4) are not permitted inside structs.*
+  * A: General vectors (N > 4) are not permitted inside structs.
 * Q: Does this have implications for existing HLSL source code compatibility?
-* A: *No, existing HLSL code is unaffected by this change.*
-* A: *Change the default N = 4 for vectors? Will affect existing shaders.*
+  * A: No, existing HLSL code is unaffected by this change.
+* Q: Should this change the default N = 4 for vectors?
+  * A: No. While the default size of 4 is less intuitive in a world of larger vectors, existing code depends on this default, so it remains unchanged.
 * Q: How will SPIRV be supported?
-* A: 
-* Q: When do HLSL vectors remain as vectors and when do they get scalarized in DXIL?
-* A: 
+  * A: TBD
+* Q: Under what conditions do HLSL vectors remain as vectors and when do they get scalarized in DXIL?
+  * A: UNRESOLVED
 * Q: Can all implementations support vector DXIL?
-* A: Feature check?
+  * A: Feature check?
 
 ## Acknowledgments
-
 
 <!-- {% endraw %} -->
--- a/proposals/0026-hlsl-long-vector-type.md
+++ b/proposals/0026-hlsl-long-vector-type.md
@@ -27,7 +27,7 @@ To take advantage of specialized hardware that can accelerate longer vector oper
 
 ## Proposed solution
 
-Enable vectors of length between 4 and 128 inclusive in HLSL using existing template-based vector declarations.
+Enable vectors of length between 5 and 128 inclusive in HLSL using existing template-based vector declarations.
 Such vectors will hereafter be referred to as "long vectors".
 These will be supported for all elementwise intrinsics that take variable-length vector parameters.
 For certain operations, these vectors will be represented as native vectors using
@@ -44,10 +44,11 @@ vector<T, N> name;
 ```
 
 `T` is any [scalar](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-scalar) type.
-`N` is the number of components and must be an integer between 1 and 4 inclusive.
+`N` is the number of components and must be a constant integer expression between 1 and 4 inclusive.
 See the vector definition [documentation](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-vector) for more details.
 This proposal adds support for long vectors of length greater than 4 by
- allowing `N` to be greater than 4 where previously such a declaration would produce an error.
+ allowing `N` to be a constant integer expression greater than 4
+ where previously such a declaration would produce an error.
 
 The default behavior of HLSL vectors is preserved for backward compatibility, meaning, skipping the last parameter `N`
 defaults to 4-component vectors and the use `vector name;` declares a 4-component float vector, etc. More examples
@@ -66,6 +67,7 @@ Long vectors can be:
 * Parameters and return types of non-entry functions.
 * Stored in groupshared memory.
 * Static global variables.
+* Local function scoped variables.
 
 Long vectors are not permitted in:
 
@@ -82,7 +84,8 @@ implementations may specify best practices in certain uses for optimal performan
 #### Constructing vectors
 
 HLSL vectors can be constructed through initializer lists, constructor syntax initialization, or by assignment.
-Vectors can be initialized and assigned from various casting operations including scalars and arrays.
+Vectors can be initialized and assigned from various casting operations including scalars, arrays, and initialization lists.
+Initialization of vectors from vectors or initialization lists with fewer elements than the assigned vector are not allowed.
 Long vectors will maintain equivalent casting abilities.
 
 Examples:
@@ -314,12 +317,9 @@ Having a limit facilitates testing and sets expectations for both hardware and s
 * Q: Should this change the default N = 4 for vectors?
   * A: No. While the default size of 4 is less intuitive in a world of larger vectors, existing code depends on this default, so it remains unchanged.
 * Q: How will SPIR-V be supported?
-  * A: TBD
+  * A: TBD. SPIR-V could be represented as an array of elements, scalarized to use scalars, or a new vector type.
 * Q: should swizzle accessors be allowed for long vectors?
   * A: No. It doesn't make sense since they can't be used to access all elements
        and there's no way to create enough swizzle members to accommodate the longest allowed vector.
-* Q: How should scalar groupshared arrays be loaded/stored into/out of long vectors.
-  * A: After some consideration, we opted not to include explicit Load/Store operations for this function.
-       There are at least a couple ways this could be resolved, and the preferred solution is outside the scope.
 
 <!-- {% endraw %} -->
--- a/proposals/0026-hlsl-long-vector-type.md
+++ b/proposals/0026-hlsl-long-vector-type.md
@@ -21,15 +21,16 @@ This feature adds the ability to load, store, and perform select operations on H
 
 The adoption of machine learning techniques expressed as vector-matrix operations
  require larger vector sizes to be representable in HLSL.
-To take advantage of specialized hardware that can accelerate vector operations,
- these and other vector objects need to be preserved at the DXIL level.
+To take advantage of specialized hardware that can accelerate longer vector operations,
+ these vectors need to be preserved in the exchange format as well.
 
 ## Proposed solution
 
 Enable vectors of length between 4 and 128 inclusive in HLSL using existing template-based vector declarations.
 Such vectors will hereafter be referred to as "long vectors".
 These will be supported for all elementwise intrinsics that take variable-length vector parameters.
-For certain operations, these vectors will be represented as native vectors using [dxil vectors](NNNN-dxil-vectors.md).
+For certain operations, these vectors will be represented as native vectors using
+ [Dxil vectors](NNNN-dxil-vectors.md) and equivalent SPIR-V representations.
 
 ## Detailed design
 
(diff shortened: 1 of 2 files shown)


proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,278 @@
...
+
+## Introduction
+
+HLSL has always supported vectors of as many as four elements of different element types (int3, float4, etc.).
+These are useful in a traditional graphics context for representation and manipulation of
+ geometry and color information.
+The evolution of HLSL as a more general purpose language targeting Graphics and Compute
+ greatly benefit from longer vectors to fully represent these operations rather than to try to
+ break them down into smaller constituent vectors.
+This feature adds the ability to declare and use native HLSL vectors longer than four elements.
```
farzonl, line 18: I don't understand the distinction between declaration and use.  I think this line should just be simplified to add native support like so:
```suggestion
This feature introduces native support for vectors exceeding four elements.
```

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,278 @@
...
+# HLSL Long Vectors
+
+* Proposal: [0026-HLSL-Vectors](0026-hlsl-vector-type.md)
+* Author(s): [Anupama Chandrasekhar](https://github.com/anupamachandra), [Greg Roth](https://github.com/pow2clk)
+* Sponsor: [Greg Roth](https://github.com/pow2clk)
+* Status: **Under Consideration**
+
+## Introduction
+
+HLSL has always supported vectors of as many as four elements of different element types (int3, float4, etc.).
```
farzonl, line 12: Always doesn't indicate enough that this is past behavior. and `as many as` just read awkward to me. Also replaced the repeat of the word element with `scalar`.
```suggestion
HLSL has traditionally supported vectors with up to four elements of different scalar types (int3, float4, etc.).
```

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,278 @@
...
+These are useful in a traditional graphics context for representation and manipulation of
```
farzonl, line 13: If you use the suggestion in line 12 remove the word `traditional` in line 13.

proposals/NNNN-dxil-vectors.md
```diff
@@ -0,0 +1,183 @@
...
+* Unary
+  * Exp
+  * Htan
+  * Atan
+  * Log
+* Binary
+  * FMin
+  * FMax
+* Tertiary
+  * Fma
```
jenatali, line 124: Just to point out, DXIL has both `fma` and `fmad`. The valid overload set for `fma` is fp64, where the valid overload set for `fmad` is all floating point types. I assume that `fma` in this list is intended to refer to both, but it would be good to be explicit.

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,278 @@
...
+* Q: Should this change the default N = 4 for vectors?
+  * A: No. While the default size of 4 is less intuitive in a world of larger vectors, existing code depends on this default, so it remains unchanged.
+* Q: How will SPIR-V be supported?
+  * A: TBD
+* Q: should swizzle accessors be allowed for long vectors?
+  * A: No. It doesn't make sense since they can't be used to access all elements
+       and there's no way to create enough swizzle members to accommodate the longest allowed vector.
+* Q: How should scalar groupshared arrays be loaded/stored into/out of long vectors.
+  * A: After some consideration, we opted not to include explicit Load/Store operations for this function.
+       There are at least a couple ways this could be resolved, and the preferred solution is outside the scope.
```
damyanp, line 276: Does this mean that long vectors can marked as groupshared and it just works? Wondering if this Q&A belongs in the DXIL spec?

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,278 @@
...
+ the element type and number of elements (e.g. float2, double4) are allowed for long vectors.
+
+The new vectors will be supported in all shader stages including Node shaders. There are no control flow or wave
+uniformity requirements, but implementations may specify best practices in certain uses for optimal performance.
+
+Long vectors are not permitted in:
+
+* Resource types other than ByteAddressBuffer or StructuredBuffer.
+* Any element of the shader's signature including entry function parameters and return types.
+* Cbuffers or tbuffers.
```
damyanp, line 64: Can I construct a long vector from an array stored in a cbuffer?

proposals/NNNN-dxil-vectors.md
```diff
@@ -0,0 +1,183 @@
...
+#### Elementwise intrinsics
+
+A selection of elementwise intrinsics are given additional native vector forms.
+Elementwise intrinsics are those that perform their calculations irrespective of the location of the element
+ in the vector or matrix arguments except insofar as that position corresponds to those of the other elements
+ that might be used in the individual element calculations.
+An elementwise intrinsic `foo` that takes scalar or vector arguments could theoretically implement its vector version using a simple loop and the scalar intrinsic variant.
+
+```c++
+vec<TYPE, NUM> foo(vec<TYPE, NUM> a, vec<TYPE, NUM> b) {
```
damyanp, line 93: Is there significance to `vec` rather than `vector` here?

proposals/NNNN-dxil-vectors.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Potential Changes to DXIL Consumers
+
+As this removes no existing DXIL features, the former representation of vectors is still valid.
+However, DXIL consumers may expect native vectors where they are supported and may misinterpret
+ vectors scalarized into arrays as being native arrays.
+This is unlikely to produce any faulty results, but may miss some optimizations.
+
+As DXIL with native vectors might be linked to create a DXIL shader without that support,
+ some additional scalarization might be necessary when linking in such cases.
```
damyanp, line 137: Do we support linking code built for different shader models?  I'd have thought we'd require the SM's to match when linking!

proposals/NNNN-dxil-vectors.md
```diff
@@ -0,0 +1,183 @@
...
+Individual vectors will now be represented as a single native vector and arrays of vectors will remain
+ as arrays of native vectors, though multi-dimensional arrays will still be flattened to one dimension.
+
+Scalarization of these vectors will continue to be done for uses that don't support native vectors,
+ but it will be done using extractelement instructions from the native vectors
+ instead of loads from the scalarized array representation.
+
+Single-element vectors are not valid in DXIL.
+At the language level, they may be supported for corresponding intrinsic overloads,
+  but such vectors should be represented as scalars in the final DXIL output.
```
damyanp, line 51: It'd be good to understand the rationale for this.  Is this just about reducing testing / implementation cost, or is there something more fundamental I'm missing?

If I'm following correctly, this would make it pretty hard to implement something like specialization constants in DXIL where the constant controlled the size of the vector.

damyanp (approved):

pow2clk (commented):
Update shortly.

proposals/0026-hlsl-long-vector-type.md
pow2clk: This comment went to the wrong suggestion. Oops

proposals/0026-hlsl-long-vector-type.md
pow2clk: Declaration without use in intrinsics isn't terribly useful of course, but there is a distinction between them. I drew the distinction because both declaration and use are limited in certain ways. The rest of the spec addresses them individually and this set that up. 

I don't feel too strongly about it for the introduction though.

llvm-beanz: It would be really odd to have the language allow you to declare something that you then can't use. I might instead phrase this something like:

> This feature allows declaring objects of vector type with more than four elements, and defines how those objects may be used.

proposals/0026-hlsl-long-vector-type.md
llvm-beanz: You should be able to construct a long vector from an array regardless of where the array is stored. HLSL's aggregate initialization rules would allow this so it seems like we shouldn't special case long vectors.

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,278 @@
...
+* tanh
+* atan
+* min
+* max
+* clamp
+* step
+
+#### Disallowed vector intrinsics
+
+* Only applicable to for shorter vectors: AddUint64, asdouble, asfloat, asfloat16, asint, asint16, asuint, asuint16, D3DCOLORtoUBYTE4, cross, distance, dst, faceforward, length, normalize, reflect, refract, NonUniformResourceIndex
```
llvm-beanz, line 149: What is the reasoning for disallowing the conversion functions?

proposals/NNNN-dxil-vectors.md
jenatali: Makes sense. I was surprised at the tiny scope of opcodes listed here, but I assumed that the primary use case is not for doubles, so listing only `fma` without `fmad` confused me.

proposals/NNNN-dxil-vectors.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Introduction
+
+While DXIL is intended and able to support language vectors,
+ those vectors must be broken up into individual scalars to be valid DXIL.
+This feature introduces the ability to represent native vectors in DXIL for some uses.
+
+## Motivation
+
+While the original shape of the vectors may be reconstructed from their scalarized form,
```
llvm-beanz, line 21: I might phrase this differently. I think there are two big motivations for vectorizing DXIL.

> Many GPUs support vector operations, and with DXIL today they must rely on auto-vectorization to re-materialize vectors late in the compiler. Auto-vectorization is a notoriously difficult compiler optimization that frequently fails generating sub-optimal code. Conversely scalarization is a trivial compiler transformation that never fails. Allowing DXIL to retain vectors as they appeared in source allows hardware that can utilize vector optimizations to do so more easily without penalizing hardware that requires scalarization.

-and-

> One common complaint about DXIL is the size of compiled programs. Vector operations can express in a single instruction operations that would historically take _N_ instructions in scalar DXIL. For this reason allowing vector preservation in DXIL may result in reduced file size for compiled DXIL programs that utilize vectors.

proposals/NNNN-dxil-vectors.md
```diff
@@ -0,0 +1,183 @@
...
+ it requires additional work of the DXIL consumer and results in larger DXIL binary sizes.
+Although it has never been allowed in DXIL, the LLVM IR that DXIL is based on can represent native vectors.
+By allowing these native vector types in DXIL, the size of generated DXIL can be reduced and
+ new opportunities for expanding vector capabilities in DXIL ar introduced.
+
+## Proposed solution
+
+Native vectors are allowed in DXIL version 1.9 or greater.
+These can be stored in allocas, static globals, and groupshared variables.
```
llvm-beanz, line 30: Importantly we can have SSA values of vector type.

proposals/0026-hlsl-long-vector-type.md
damyanp: Maybe I'm missing the context or the significance of not including explicit Load/Store operations here, or just plain misunderstanding this Q/A altogether.  The "preferred solution is outside the scope" seems to suggest that there's not going to be a way to use long vectors in groupshared.

proposals/NNNN-dxil-vectors.md
```diff
@@ -0,0 +1,191 @@
...
+ those vectors must be broken up into individual scalars to be valid DXIL.
+This feature introduces the ability to represent native vectors in DXIL for some uses.
+
+## Motivation
+
+While the original shape of the vectors may be reconstructed from their scalarized form,
+ it requires additional work of the DXIL consumer and results in larger DXIL binary sizes.
+Although it has never been allowed in DXIL, the LLVM IR that DXIL is based on can represent native vectors.
+By allowing these native vector types in DXIL, the size of generated DXIL can be reduced and
+ new opportunities for expanding vector capabilities in DXIL ar introduced.
```
inbelic, line 25:
```suggestion
 new opportunities for expanding vector capabilities in DXIL are introduced.
```

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,289 @@
...
+Unlike vector sizes between 1 and 4, no shorthand declarations that concatenate
+ the element type and number of elements (e.g. float2, double4) are allowed for long vectors.
+
+The new vectors will be supported in all shader stages including Node shaders. There are no control flow or wave
+uniformity requirements, but implementations may specify best practices in certain uses for optimal performance.
+
+Long vectors can be:
+
+* Elements of arrays, structs, StructuredBuffers, and ByteAddressBuffers.
+* Parameters and return types of non-etry functions.
```
V-FEXrt, line 63: nit: typo
```suggestion
* Parameters and return types of non-entry functions.
```

tex3d (commented):
Some general high-level points:

What gates this feature in HLSL?  Is it part of a future HLSL language version?  Just Shader Model 6.9 (which would seem odd considering this is basically a language change)?  I don't see why we can't map long vectors to legacy scalarized shader models, with some requiring a bit more work, such as native vector load/store DXIL operations, vector reductions (dot) to expansions, and so on, but these are still mappable.

If not by language version, how do we enforce the limitations? Globally by shader model?  Typically, shader model limitations are applied according to what's used, not what's present, so do we plan to do the same here (and how?), or use the shader model to change the language accepted everywhere, which is odd?

This doesn't address targeting SPIR-V from HLSL at all.

I still don't see how we will be able to support specialization constant sized vectors in HLSL or any equivalent feature in DXIL using this approach.

Many of the DXIL details I believe belong only in the DXIL spec.  Some testing details in the DXIL spec I think belong only in the HLSL spec.

The limits we define should not be the temporary ones for implementation purposes, which could be discussed in some other section or doc perhaps.  Limits on intrinsics with native vector DXIL overloads and limiting use in cbuffer/tbuffer could fall under this category.

No mention about vector and matrix swizzles lengths deliberately not being extended to allow long vector results.

I have some concern about 64-bit component types and in particular `double` in long vectors, since there are a number of special cases for these.  They might need additional testing scenarios, or to be added to temporary exclusions for preview.

proposals/NNNN-dxil-vectors.md
tex3d: Yes, I agree this is an implementation detail that doesn't belong in this spec.  We might want an infra spec to track details like this.

proposals/NNNN-dxil-vectors.md
```diff
@@ -0,0 +1,183 @@
...
+As this removes no existing DXIL features, the former representation of vectors is still valid.
+However, DXIL consumers may expect native vectors where they are supported and may misinterpret
+ vectors scalarized into arrays as being native arrays.
+This is unlikely to produce any faulty results, but may miss some optimizations.
+
+As DXIL with native vectors might be linked to create a DXIL shader without that support,
+ some additional scalarization might be necessary when linking in such cases.
+
+This feature involves no changes to previous shader models and any DXIL produced for earlier versions
+  should continue to behave exactly as before.
```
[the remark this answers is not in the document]
tex3d: Note: this is an implementation detail, so it probably belongs in an infra spec if anywhere.  It sort of goes without saying that the definition of prior DXIL versions remains the same, and that the compiler should continue to produce correct code for them.  Whether it is able to produce identical code is another issue, which might not be fully possible, as many types of changes can impact exact IR results in benign ways.

llvm-beanz:
> Some general high-level points:
> 
> What gates this feature in HLSL? Is it part of a future HLSL language version? Just Shader Model 6.9 (which would seem odd considering this is basically a language change)? I don't see why we can't map long vectors to legacy scalarized shader models, with some requiring a bit more work, such as native vector load/store DXIL operations, vector reductions (dot) to expansions, and so on, but these are still mappable.
> 
> If not by language version, how do we enforce the limitations? Globally by shader model? Typically, shader model limitations are applied according to what's used, not what's present, so do we plan to do the same here (and how?), or use the shader model to change the language accepted everywhere, which is odd?

This is definitely a discussion we should explore more. From a purity perspective my gut is that long vectors is a language feature for HLSL 202x. The downside to that is that anyone using SM 6.9 features for long vectors will need to update their codebase to HLSL 202x. This might cause us to reconsider the scoping of 202x to support calling it "done" sooner than originally planned.

If we aren't okay with this feature requiring updating to 202x, it probably ins't the worst thing to make it available in older language modes, but we may want to consider making it an opt-in feature. I haven't thought about this extensively but this is definitely something we should discuss.

> I still don't see how we will be able to support specialization constant sized vectors in HLSL or any equivalent feature in DXIL using this approach.

I don't think we should be thinking about specialization constants as part of this feature. HLSL does not have support for specialization constants. There is a Vulkan extension, but the Vulkan extension is not core to the language and it is not really the direction that the language-based solution would take.

I realize that this feedback may seem counter to other feedback I've provided about needing a comprehensive plan for a feature. The distinction I see here is that designing specialization constant support for HLSL and fixing the existing language features that currently don't support specialization constants and should (like a bunch of the attributes, and operations where we require immediate values), is a _much_ larger feature and will require significant investment and planning. Since we don't have a plan for how to support specialization constants in HLSL as a proper language feature, we shouldn't use that to roadblock or derail the design of this feature.

If we make a mistake here and we need to adjust or redesign things when we add specialization constants in the future, we can cross that road when we get to it. We are all human, making design mistakes is something that is going to happen, we can't let the possibility of a mistake prevent us from making forward progress.

proposals/NNNN-dxil-vectors.md
```diff
@@ -0,0 +1,183 @@
...
+
+While DXIL is intended and able to support language vectors,
+ those vectors must be broken up into individual scalars to be valid DXIL.
+This feature introduces the ability to represent native vectors in DXIL for some uses.
+
+## Motivation
+
+While the original shape of the vectors may be reconstructed from their scalarized form,
+ it requires additional work of the DXIL consumer and results in larger DXIL binary sizes.
+Although it has never been allowed in DXIL, the LLVM IR that DXIL is based on can represent native vectors.
```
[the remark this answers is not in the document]
pow2clk: I've incorporated your suggested language.

proposals/NNNN-dxil-vectors.md
```diff
@@ -0,0 +1,183 @@
...
+
+In their alloca and variable representations, vectors in DXIL will always be represented as vectors.
+Previously individual vectors would get scalarized into scalar arrays and arrays of vectors would be flattened
+ into a one-dimensional scalar array with indexing to reflect the original intents.
+Individual vectors will now be represented as a single native vector and arrays of vectors will remain
+ as arrays of native vectors, though multi-dimensional arrays will still be flattened to one dimension.
+
+Scalarization of these vectors will continue to be done for uses that don't support native vectors,
+ but it will be done using extractelement instructions from the native vectors
+ instead of loads from the scalarized array representation.
```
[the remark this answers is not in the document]
pow2clk: I agree this should be mentioned. I've added it to its own design subsection

proposals/NNNN-dxil-vectors.md
pow2clk: This list is getting removed as we've determined that scalarized versus native elementwise intrinsics are an implementation detail.

pow2clk (commented):
Thanks all! I've responded inline and a commit is forthcoming.

proposals/NNNN-dxil-vectors.md
jenatali: As long as there's sufficient test content that all vectorized intrinsics are covered, I'm fine without an exhaustive list in the spec.

pow2clk: The exhaustive list is now the "allowed elementwise list" that was above this removed section.

proposals/0026-hlsl-long-vector-type.md
pow2clk: Sorry I didn't address this. Some of these comments didn't show up in my review of the files. I think you're referring to the as<Something> functions? There is some variability there. Some of them map to a simple bitcast. Those are fine. Some of them take multiple parameters representing low and hi bits that don't map as neatly.

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -1,146 +1,278 @@
...
+* tanh
+* atan
+* min
+* max
+* clamp
+* step
+
+#### Disallowed vector intrinsics
+
+* Only applicable to for shorter vectors: AddUint64, asdouble, asfloat, asfloat16, asint, asint16, asuint, asuint16, D3DCOLORtoUBYTE4, cross, distance, dst, faceforward, length, normalize, reflect, refract, NonUniformResourceIndex
```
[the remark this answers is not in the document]
pow2clk: I don't know why these comments didn't show up in my pass of the file diff. As I mentioned above in comment https://github.com/microsoft/hlsl-specs/pull/361#discussion_r1915504468, I agree some of these should be supported.

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -54,29 +55,52 @@ Declarations of long vectors require the use of the template declaration.
...
+
 The new vectors will be supported in all shader stages including Node shaders. There are no control flow or wave
 uniformity requirements, but implementations may specify best practices in certain uses for optimal performance.
 
+Long vectors can be:
+
+* Elements of arrays, structs, StructuredBuffers, and ByteAddressBuffers.
+* Parameters and return types of non-entry functions.
+* Stored in groupshared memory.
+* Static global varaibles.
```
damyanp, line 68: typo

```suggestion
* Static global variables.
```

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -54,29 +55,52 @@ Declarations of long vectors require the use of the template declaration.
...
+vector<float, 5> CastArr = (vector<float, 5>)arr;
+vector<float, 6> ArrScal = {arr, 7.9};
+vector<float, 10> ArrArr = {arr, arr};
+vector<float, 15> Scal = 4.2;
 ```
 
+float4 main(uint size: S) : SV_Target {
+   return (float4)arr;
+vector<uint, 8> vecC = {uint2(coord.xy), vecB};
+
```
damyanp, line 103: This is outside the ` ``` ` so doesn't format correctly.  Also looks like the `main` function is missing a close brace.  Is the `vecC` part means to be inside the function?

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -105,8 +129,9 @@ myBuffer.Store(elementIndex, val);
 
 #### Accessing elements of long vectors
 
-Long vectors support the existing vector subscript operators to return the scalar element values.
-They do not support swizzle operations as they are limited to only the first four elements.
+Long vectors support the existing vector subscript operators `[]` to access the scalar element values.
+They do not support any swizzle operations.
+Swizzle operations are limited to the first four elements and the accessors are named according to the graphics domain.
```
damyanp, line 134: These two sentences seem contradictory now.  I think the original version was clearer - swizzle isn't supported, because of reasons.  Is the intent to change this to say that they're supported now for just the first four elements?

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -129,61 +154,60 @@ Refer to the HLSL spec for an exhaustive list of [Operators](https://learn.micro
...
 Use of long vectors in unsupported intrinsics should produce validation errors.
 
-## Runtime Additions
-
-Support for Long vectors requires dxil vector support as defined in [the specification](NNNN-dxil-vectors.md).
+### Device Capability
 
-Use of long vectors in a shader should be indicated in DXIL with the corresponding
- shader model version and shader feature flag.
+Devices that support Shader Model 6.9 will be required to fully support this feature.
```
damyanp, line 210: In the interchange format section above we have:

> Long vectors can be represented in DXIL, SPIR-V or other interchange formats as scalarized elements or native vectors.
Representation of native vectors in DXIL depends on [dxil vectors](NNNN-dxil-vectors.md).

This seems to imply that long vectors can be supported in existing shader models.  I think it's only native DXIL vectors feature that actually requires SM 6.9?

proposals/0026-hlsl-long-vector-type.md
damyanp: Ok. I think that this Q/A is a bit confusing as written now.  If I understand correctly, it's saying that we're not going to tackle solving the problem of people wanting buffer-like access to groupshared memory. So in the same way that we don't support explicit Load/Store operations for other data types, we're not going to add new ones for this.  Marking a long vector as groupshared is still expected to work.

proposals/NNNN-dxil-vectors.md
damyanp: I guess this is something we'll have to tackle when we start thinking about adding native specialization constants to HLSL.

proposals/0026-hlsl-long-vector-type.md
pow2clk: Left over margin notes. I've removed them.

proposals/0026-hlsl-long-vector-type.md
pow2clk: I don't think they are. The first says that long vectors don't support swizzle operations. The second describes what they are to try to explain why. I'm not sure the rationale needs to be here though. I can put it in alternatives considered.

proposals/0026-hlsl-long-vector-type.md
pow2clk: That wasn't what I intended to imply with that statement, rather that implementations could choose either approach as we intend to do at least temporarily. However, it is true that this is possible. One of the major remaining questions is if we should.

proposals/0026-hlsl-long-vector-type.md
pow2clk: What you've said is accurate, so I don't know why you think it's confusing. I'm happy to take another run at it, but I think it's succinct and sufficient. It doesn't go into detail because I don't think we need to. It's a feature for another day. I could create an issue for it that could contain all the details if that would help.

damyanp: I was only able to reach the understanding by this conversation.

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,323 @@
...
+defaults to 4-component vectors and the use `vector name;` declares a 4-component float vector, etc. More examples
+[here](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-vector).
+Declarations of long vectors require the use of the template declaration.
+Unlike vector sizes between 1 and 4, no shorthand declarations that concatenate
+ the element type and number of elements (e.g. float2, double4) are allowed for long vectors.
+
+#### Allowed Usage
+
+The new vectors will be supported in all shader stages including Node shaders. There are no control flow or wave
+uniformity requirements, but implementations may specify best practices in certain uses for optimal performance.
```
anupamachandra, line 62: Is this line needed for the long vector spec? It's probably implicitly assumed?

proposals/0026-hlsl-long-vector-type.md
damyanp: Before I resolve this conversation: do we have something written down that is tracking this remaining question?

pow2clk: I've created #371

proposals/0026-hlsl-long-vector-type.md
pow2clk: Which part do you take to be implicitly assumed? I don't think support across all shader stages should be assumed. The control/wave uniformity requirements touch more on the followups and could probably be removed. I'm inclined to leave the implementations encouraging best practices language as some platforms might end up scalarizing vector operations at some point or another which won't lead to the best performance.

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,325 @@
...
+## Motivation
+
+The adoption of machine learning techniques expressed as vector-matrix operations
+ require larger vector sizes to be representable in HLSL.
+To take advantage of specialized hardware that can accelerate longer vector operations,
+ these vectors need to be preserved in the exchange format as well.
+
+## Proposed solution
+
+Enable vectors of length between 4 and 128 inclusive in HLSL using existing template-based vector declarations.
```
dneto0, line 30: if the range is inclusive, then that 4 should be 5.

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,325 @@
...
+```hlsl
+vector<T, N> name;
+```
+
+`T` is any [scalar](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-scalar) type.
+`N` is the number of components and must be an integer between 1 and 4 inclusive.
+See the vector definition [documentation](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-vector) for more details.
+This proposal adds support for long vectors of length greater than 4 by
+ allowing `N` to be greater than 4 where previously such a declaration would produce an error.
+
```
dneto0, line 51: Somewhat related to what @tex3d asked, but more broad:  the spec needs to say how N can be written.
* Can it be `N` where earlier there is a declaration `const uint N = 15;`
* Can it be `N` where `N` is the parameter to a function, and the vector type declaration is  inside the function.

I guess it has to be statically computed.  (I'm not sure what the HLSL term for it.)

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,325 @@
...
+The new vectors will be supported in all shader stages including Node shaders.
+
+Long vectors can be:
+
+* Elements of arrays, structs, StructuredBuffers, and ByteAddressBuffers.
+* Parameters and return types of non-entry functions.
+* Stored in groupshared memory.
+* Static global variables.
+
```
dneto0, line 69: I assume the type of a local (function scoped) variable can have long vector type.  That's not mentioned. The examples below show cases like these.

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,325 @@
...
+#### Constructing vectors
+
+HLSL vectors can be constructed through initializer lists, constructor syntax initialization, or by assignment.
+Vectors can be initialized and assigned from various casting operations including scalars and arrays.
+Long vectors will maintain equivalent casting abilities.
+
+Examples:
+
+```hlsl
+vector<uint, 5> InitList = {1, 2, 3, 4, 5};
```
dneto0, line 91: What happens if not all elements are listed explicitly.  Are the rest zero-initialized?

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -1,146 +1,278 @@
...
 vector<uint, 6> vecD = vecB;
 ```
 
-**Load and Store vectors from Buffers/Arrays**
+#### Vectors in Raw Buffers
 
-For loading and storing N-dimensional vectors from ByteAddressBuffers we use the `LoadN` and `StoreN` methods, extending
-the existing Load/Store, Load2/Store2, Load3/Store3 and Load4/Store4 methods.
+N-element vectors are loaded and stored from ByteAddressBuffers using the templated load and store methods
+with a vector type of the required size as the template parameter and byte offset parameters.
```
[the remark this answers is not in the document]
dneto0: You could say that a long vector <T,N> has the same alignment requirements as an array of N elements of type T.
Except that may be different from vector<float,4> which might have 16byte alignment.  With scalar alignment they would be the same. (Sorry, I don't remember if modern HLSL assumes scalar alignment for vectors.)

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,325 @@
...
+
+```
+
+StructuredBuffers with N-element vectors are declared using the template syntax
+ with a long vector type as the template parameter.
+N-element vectors are loaded and stored from ByteAddressBuffers using the templated load and store methods
+with the element index parameters.
+
+```hlsl
+RWStructuredBuffer< vector<T, N> > myBuffer;
```
dneto0, line 122: Is the space between `> >` necessary, as per C++03 style parsing?  Might be good to call out.
I hope no space is required.

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,325 @@
...
+* Type Conversions: asdouble, asfloat, asfloat16, asint, asint16, asuint, asuint16
+
+#### Disallowed vector intrinsics
+
+* Only applicable to shorter vectors: AddUint64, D3DCOLORtoUBYTE4, cross, distance, dst, faceforward, length, normalize, reflect, refract, NonUniformResourceIndex
+* Only useful for disallowed variables: EvaluateAttributeAtSample, EvaluateAttributeCentroid, EvaluateAttributeSnapped, GetAttributeAtVertex
+
+### Interchange Format Additions
+
+Long vectors can be represented in DXIL, SPIR-V or other interchange formats as scalarized elements or native vectors.
```
dneto0, line 163: This section should be informative.

A good SPIR-V representation could be as arrays of elements. So it's still a single SSA value.

dneto0 (commented):
Thanks for the spec proposal.

proposals/0026-hlsl-long-vector-type.md
llvm-beanz: HLSL 2021 templates are C++98/03-era, so the space is required.

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,325 @@
...
+* A work graph record.
+
+While this describes where long vecgtors can be used and later sections will describe how,
+implementations may specify best practices in certain uses for optimal performance.
+
+#### Constructing vectors
+
+HLSL vectors can be constructed through initializer lists, constructor syntax initialization, or by assignment.
+Vectors can be initialized and assigned from various casting operations including scalars and arrays.
+Long vectors will maintain equivalent casting abilities.
```
anupamachandra, line 86: It would be nice to have a way to initialize all elements of the vector with one value.

proposals/NNNN-dxil-vectors.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+Here and hereafter, `NUM` is the number of elements in the loaded vector, `TYPE` is the element type name,
+ and `TY` is the corresponding abbreviated type name (e.g. `i64`, `f32`).
+
+#### Vector access
+
+Dynamic access to vectors were previously converted to array accesses.
+Native vectors can be accessed using `extractelement`, `insertelement`, or `getelementptr` operations.
+Previously usage of `extractelement` and `insertelement` in DXIL didn't allow dynamic index parameters.
```
anupamachandra, line 98: Just confirming that this spec does allow dynamic index parameters in EIE adn IEI?

proposals/NNNN-dxil-vectors.md
```diff
@@ -0,0 +1,190 @@
...
+
+```asm
+ <[NUM] x [TYPE]> @dx.op.unary.v[NUM][TY](i32 opcode, <[NUM] x [TYPE]> operand1)
+ <[NUM] x [TYPE]> @dx.op.binary.v[NUM][[TY]](i32 opcode, <[NUM] x [TYPE]> operand1, <[NUM] x [TYPE]> operand2)
+ <[NUM] x [TYPE]> @dx.op.tertiary.v[NUM][TY](i32 opcode, <[NUM] x [TYPE]> operand1, <[NUM] x [TYPE]> operand2, <[NUM] x [TYPE]> operand3)
+```
+
+The scalarized variants of these DXIL intrinsics will remain unchanged and can be used in conjunction
+ with the vector variants.
+This means that the same language-level vector could be used in scalarized operations and native vector operations
```
anupamachandra, line 131: In what situations are we likely to see scalarization as opposed to vector results?

anupamachandra: Also, wanted to confirm that DXIL vectorization is not limited to HLSL *long* vectors but even applies to the original ones.

pow2clk: The length of the native vectors isn't part of this spec, so native vectors can be any length. I tried to communicate that without calling out long vectors explicitly to avoid a codependency. I think if this spec were viewed outside the context of a PR where long vectors is included, that might be clearer.

As to what situations we will see scalarizations, if that's useful for a preview of this feature, we can definitely work that out what can be expected as we have no strong opinions except to enable a preview. At full release of this feature, the scalar versions are not going away and the vector versions will apply to a broad selection of opcodes. This means that DXIL consumers will ultimately be expected to handle vectorized or scalarized versions of any relevant operations.

proposals/0026-hlsl-long-vector-type.md
pow2clk: Done.

proposals/0026-hlsl-long-vector-type.md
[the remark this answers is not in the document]
pow2clk: Tex and I will discuss this and I'll update the spec accordingly.

proposals/0026-hlsl-long-vector-type.md
pow2clk: I don't think it's useful to go into detail about something that someone wanted to tack onto this spec, but isn't related to it. Perhaps the best approach is to just remove it.

proposals/0026-hlsl-long-vector-type.md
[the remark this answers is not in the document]
tex3d: Alignment for vectors in HLSL is purely based on the alignment of the scalar element type.  Constant buffer layout ("legacy") is a special case where a vector that would cross a 16-byte aligned boundary is started at the next 16-byte aligned location instead.

Array elements have the same natural alignment based on scalar element type as vectors.  However, they also have a special rule for constant buffer layout ("legacy") where each array element must start at the beginning of a 16-byte aligned boundary.  There are also special rules for signature inputs/outputs for arrays, but these can't be described by a concept of alignment.

The "legacy" constant layout rules cannot be described by the concept of type alignments.

Changing the natural alignment rules for a vector depending on whether it has > 4 elements seems weird.

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,325 @@
...
+## Motivation
+
+The adoption of machine learning techniques expressed as vector-matrix operations
+ require larger vector sizes to be representable in HLSL.
+To take advantage of specialized hardware that can accelerate longer vector operations,
+ these vectors need to be preserved in the exchange format as well.
+
+## Proposed solution
+
+Enable vectors of length between 5 and 128 inclusive in HLSL using existing template-based vector declarations.
```
damyanp, line 30:
```suggestion
Enable vectors of length between 5 and 1024 inclusive in HLSL using existing template-based vector declarations.
```

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -0,0 +1,325 @@
...
+Swizzle operations were not supported because they are limited to the first four elements.
+The names of the accessors (xyzw or rgba) are named according to the expected content of
+ those vectors in a graphics context.
+Since that intretation does not apply to longer vectors, it could be confusing.
+The subscript access is flexible and generic and makes other accessors redundant.
+
+## Open Issues
+
+* Q: Is there a limit on the Number of Components in a vector?
+  * A: 128. It's big enough for some known uses.
```
damyanp, line 307:
```suggestion
  * A: 1024. It's big enough for some known uses.
```

proposals/0026-hlsl-long-vector-type.md
pow2clk: Done

proposals/0026-hlsl-long-vector-type.md
pow2clk: Done

llvm-beanz (approved):

PR #373, opened 2025-01-24
Update INF-0004-validator-hashing.md with applicable AgilitySDK version

amarpMSFT:
Now that we know which AgilitySDK version will support hash bypass, the "TBD" in the spec can be replaced with the actual version. 

Also consider whether to move this out of "proposal" state, perhaps as soon as that AgilitySDK becomes available?  Actually might need to stay in proposal state until any DXC side changes are finished, such as publishing dxil.dll?

### DIFF
--- a/proposals/infra/INF-0004-validator-hashing.md
+++ b/proposals/infra/INF-0004-validator-hashing.md
@@ -167,7 +167,7 @@ even less likely.
 
 ### D3D Runtime Behavior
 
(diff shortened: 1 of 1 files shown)


damyanp (approved):

damyanp:
@bogner / @tex3d / @pow2clk - we just need one more review.  Please feel free to squash&merge this when you're ready.

llvm-beanz (approved):

Issue #360, opened 2024-12-18
Back out unnecessary OMM flag validation and SM 6.9 dependency


tex3d:
**Which proposal does this relate to?**
https://github.com/microsoft/hlsl-specs/blob/dc68b908fb2521d729337fe270eddab0878ec6c4/proposals/0024-opacity-micromaps.md?plain=1#L114-L117

**Describe the issue or outstanding question.**
Before this proposal, unknown RayFlags had no diagnostics, masking, or validation.  If an unknown flag was used, the DXR spec says that constitutes "undefined behavior", and that any unrecognized flag is not guaranteed to be available when reading the RayFlag state back.

This proposal adds error diagnostics for static use of new flags when `SM < 6.9`, warnings for use of the enum values, masking the RayFlags when generating the DXIL op to prevent use of any unknown flags for the current shader model, and tests for such.

After discussions, it is felt that the proposed change is overly restrictive and may add an unnecessary mask operation.  Without this restriction, the new flags could be used even with a prior shader model on a device that supports OMM, without any real dependency on SM 6.9.

This has implications for the proposal to use availability diagnostics on the new flag values as well.

We propose to remove these extra diagnostics, masking, validation, and testing from the proposal.  The new flags will be usable on prior shader models as long as the runtime and device recognize the flags, even if OMM is unsupported, with undefined behavior otherwise.  Device support for OMM or support for SM 6.9 is sufficient to guarantee that a device will accept the new flags without undefined behavior, without requiring that the shader is compiled to SM 6.9.

However, for RayQuery, it's [proposed](https://github.com/microsoft/hlsl-specs/pull/339) that a new template argument be added, which will require a new parameter to a new version of the AllocateRayQuery DXIL op.  This will still require a new version of DXIL, and thus require SM 6.9 to use.

llvm-beanz:
A user on the DirectX Discord's `#hlsl` recently said:

>in any case HLSL will accept a lot of nonsense that does nothing, so you do need to make sure you understand what exactly you're writing

This is the state of our language today.

From the brief discussion we had during triage today, IMO there are places we could relax from what we put in the spec (dynamic enforcement might be overkill), but we should not fully remove this validation.

Just because something unsupported works if you hold it just right, doesn't mean it is good design, or that it is ergonomic for our users. I believe we should strive to make HLSL less magic, and less unintuitive, and that means making the language and compiler stricter about what we accept.

[...]

PR #374, opened 2025-01-28
[OMM] Remove validation masking, update spec

bob80905:
No validation mask is needed for the new OMM flags. Source level validation is still needed to ensure the flags are used correctly, but the validator will not need to run the flags through a mask to validate them. This PR updates the spec to reflect this change to the validator, and proposes some new diagnostics with the new `RAYQUERY_FLAG::RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS` flag.

Fixes https://github.com/microsoft/hlsl-specs/issues/375
Fixes https://github.com/microsoft/hlsl-specs/issues/360

PR #377, opened 2025-01-30
[OMM] Remove validation masking, update spec

bob80905:
No validation mask is needed for the new OMM flags. Source level validation is still needed to ensure the flags are used correctly, but the validator will not need to run the flags through a mask to validate them. Additionally, this PR removes any mention of flag value validation, except for one specific case where a RayQuery ray flag template argument necessitates the RayQuery ray query flag template argument's value. 
This PR updates the spec to reflect new changes to the validator, and proposes some new diagnostics with the new RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS flag. It defines a new DXIL OP, `AllocateRayQuery2`, which will be generated when RayQuery is given a 2nd, optional, ray query template flag argument in HLSL. 

This PR is a transfer from https://github.com/microsoft/hlsl-specs/pull/374

Fixes https://github.com/microsoft/hlsl-specs/issues/375
Fixes https://github.com/microsoft/hlsl-specs/issues/360

### DIFF
--- a/proposals/0024-opacity-micromaps.md
+++ b/proposals/0024-opacity-micromaps.md
@@ -118,9 +118,10 @@ static const uint RAY_FLAG_FORCE_OMM_2_STATE = 0x400;
 
 ### Interchange Format Additions
 
-This adds no DXIL operations or metadata, it only adds a new flag value that
-may be used with existing DXIL operation parameters, returned by a DXIL
-operation.
+This adds a new DXIL operation, which will be a new version of AllocateRayQuery.
+When the flag exists, this new version of AllocateRayQuery will be generated. If
+the hardware supports OMM and `RAYQUERY_FLAG_NONE` is set, then the old version
+of AllocateRayQuery will be used instead.
 
 The DXIL operations which either accept or return `RayFlags`, and therefore may
 accept or return the new `RAY_FLAG_FORCE_OMM_2_STATE` are the following (along
@@ -131,21 +132,18 @@ with brief descriptions):
 - `RayQuery_TraceRayInline` - Trace a ray (with ray flags OR'd with the
   RayQuery's constant ray flags)
 
-When lowering to DXIL intrinsics, we will mask the flags using the legal mask
-for the shader target.  This will prevent undefined behavior if invalid flags
-were specified and either warnings were ignored, or the flags were not a known
-constant value during semantic analysis.
-
-In `DxilConstants.h`, the `RayFlag::ForceOMM2State` flag is added.
-Propose adding ValidMask values for diagnostics and validation.
+In `DxilConstants.h`, the `RayFlag::ForceOMM2State` flag is added. The new 
+`RAYQUERY_FLAG ` enum is added too, with values defined above. The whole enum
+declaration is accompanied by an availability attribute, which will be set to
+shader model 6.9. The availability attribute will be responsible for ensuring 
+that none of the enum values listed out in the enum are used when the target
+shader model is less than 6.9. 
 
 ```cpp
 // Corresponds to RAY_FLAG_* in HLSL
 enum class RayFlag : uint32_t {
   ...
-  ForceOMM2State = 0x400, // Force 2-state in Opacity Micromaps
-  ValidMask_1_8 = 0x3ff, // valid mask up through DXIL 1.8
-  ValidMask = 0x7ff, // current valid mask
+  ForceOMM2State = 0x400 // Force 2-state in Opacity Micromaps
 };
 ```
 
@@ -170,9 +168,10 @@ As an implementation detail, an attribute may be used on the new flag
 definitions, such as an existing Clang availability attribute or a new custom
 HLSL-specific attribute.
 
-AST traversal from entry points will traverse DeclRefs and initializers to
-detect the use of the new ray flag.  AST traversal will be added for subobject
-declarations on library targets to detect any use of the new pipeline flag.
+AST traversal from entry points will traverse DeclRefs, DeclRefExprs, and 
+initializers to detect the use of any new ray flags.  AST traversal will be 
+added for subobject declarations on library targets to detect any use of the
+new pipeline flag.
 
 This will have implications for uses of these flags outside of the intended
 targets.  Since they are just uint values, it's technically legal to refer to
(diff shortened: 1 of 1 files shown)


proposals/0024-opacity-micromaps.md
```diff
@@ -78,10 +78,13 @@ doesn't apply for inline raytracing however. `RayQuery` objects are
...
 
-`RayQuery<RAY_FLAGS, RAYQUERY_FLAGS>`
+```
+template<uint StaticRayFlags, uint RayQueryFlags = RAYQUERY_FLAG_NONE>
+class RayQuery;
+```
 
 ```hlsl
-enum RAYQUERY_FLAG : uint
+enum class RAYQUERY_FLAG : uint32_t
```
tex3d, line 87: Should probably back-out this changed line.

This is HSL-equivalent code, which doesn't actually have a defined enum type, so it represents a kind of virtual ("as-if") enum.  The defined values are not scoped under an enum, since they are actually `static const uint` values, so enum "class" is inappropriate, as that would change the scope of the values.

Also, since this is HLSL, `uint` is fine as the storage type as well.

proposals/0024-opacity-micromaps.md
```diff
@@ -200,46 +217,79 @@ In addition to the AST traversal detecting any explicit use of the new flag,
 the same DefaultError warning diagnostic will be added to detect when the
 new ray flag is used at the `RayQuery` template argument, `TraceRay()`, or
 `RayQuery::TraceRayInline()` (when it is immediate).
```
tex3d, line 219: The preceding paragraph talks about checking immediate constant RayFlags values on the method calls.  We decided not to do this and only use the availability check mechanism.

Two paragraphs earlier (I can't comment directly on unchanged areas), it talks about looking for the `RaytracingPipelineFlags::AllowOpacityMicromaps` in the flags for the Raytracing pipeline config1 subobject.  This is on top of checking availability attributes on expressions used.  We should remove this paragraph as well, since we decided not to add this checking.

proposals/0024-opacity-micromaps.md
```diff
@@ -200,46 +217,79 @@ In addition to the AST traversal detecting any explicit use of the new flag,
...
-This can make use of the new `ValidMask*` values.
 
 Proposed DefaultError warning diagnostic:
 
-- `"%select{RaytracingPipelineFlags|RayFlags}0 (0x%1) includes unsupported bits for shader model %2; valid mask: 0x%3"`.
+- `"%select{RaytracingPipelineFlags|RayFlags}0 (0x%1) includes unsupported bits for shader model %2"`.
   This new warning will have a different warning group, such as
   `hlsl-availability`.
```
tex3d, line 225: This pertains to the checks being removed above, so it should also be removed.

proposals/0024-opacity-micromaps.md
```diff
@@ -112,15 +115,29 @@ In HLSL under DXC, these are defined as static const uint values:
...
+This is a new version of `AllocateRayQuery` that has a `RayQueryFlags`
+parameter corresponding to the new template argument in HLSL.
+When the `RayQueryFlags` template argument is non-zero, this new
+`AllocateRayQuery2` DXIL op is used, otherwise the current `AllocateRayQuery`
+DXIL op is used.
+
+Currently, the `AllocateRayQuery` DXIL Op has the following function signature:
+`uint [[hidden]] AllocateRayQuery(in uint flags);`
+The new DXIL Op, `AllocateRayQuery2`, will instead have this signature:
+`uint [[hidden]] AllocateRayQuery2(in uint flags, in uint ray_query_flags);`
```
tex3d, line 140: These are HLSL intrinsic signatures, and they are expressed inline instead of in a code block.  I meant that we need the DXIL signature for the `dx.op.allocateRayQuery2` operation.  Something like:

```llvm
; Function Attrs: nounwind
declare i32 @dx.op.allocateRayQuery2(i32 OpCode, i32 constRayFlags, i32 RayQueryFlags)
```

proposals/0024-opacity-micromaps.md
```diff
@@ -164,15 +181,15 @@ a reachable use of one of the new flags is encountered.
...
-detect the use of the new ray flag.  AST traversal will be added for subobject
-declarations on library targets to detect any use of the new pipeline flag.
+HLSL-specific attribute. Specifically, of the new flags introduced,
+`RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS`, 
+`RAY_FLAG_FORCE_OMM_2_STATE`, and `RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS`
+will have an availability attribute, which restricts their usage to
+shader model 6.9. `RAYQUERY_FLAG_NONE` will be left unrestricted.
 
 This will have implications for uses of these flags outside of the intended
 targets.  Since they are just uint values, it's technically legal to refer to
```
tex3d, line 195: I think we can use `DefaultError` here now, since this will be the only diagnostic used to catch use of these flags in an earlier shader model.

proposals/0024-opacity-micromaps.md
```diff
@@ -200,46 +217,79 @@ In addition to the AST traversal detecting any explicit use of the new flag,
...
+Validation will be added to check the `RayFlags` parameters, and, if 
+applicable, the `RayQueryFlags` parameters, for each applicable DXIL operation,
+including the new `AllocateRayQuery2` operation. An error will be emitted if 
+the flags are constant, the new flag is used, and the shader model 
+is less than 6.9.
 
 Proposed validation error diagnostic:
 
-- `"RayFlags used in '%0' specifies unknown flags (0x%1) for shader model %2; valid mask: 0x%3"`
+- `"RayFlags used in '%0' specifies unknown flags (0x%1) for shader model %2"`
```
tex3d, line 264: We shouldn't be doing this validation of the RayFlags values.  Plus, `AllocateRayQuery2` should be defined to only be valid for SM 6.9+, so a separate check should catch any usage of that intrinsic.

proposals/0024-opacity-micromaps.md
```diff
@@ -200,46 +217,79 @@ In addition to the AST traversal detecting any explicit use of the new flag,
...
-Proposed validation error diagnostic:
+Proposed validation error diagnostics:
 
 - `"ConstRayFlags argument of AllocateRayQuery '%0' must be constant"`
+- `"ConstRayFlags and RayQueryFlags arguments of AllocateRayQuery2 '%0' must be constant"`
+
+Finally, validation will be added for the `AllocateRayQuery2` DXIL operation
+to ensure that when `RAY_FLAG_FORCE_OMM_2_STATE` is set on the ConstRayFlags
+argument, the `RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS` is also set on the
+RayQueryFlags argument.
```
tex3d, line 277: Just a reminder to synchronize argument names referred to here with whatever's used in the signature above, and use the same for the hctdb.py argument names.  It turns out that `ConstRayFlags` should be `constRayFlags`, as it's currently defined for `AllocateRayQuery`.

tex3d (changes requested):
Some more updates needed.

proposals/0024-opacity-micromaps.md
```diff
@@ -193,37 +194,19 @@ shader model 6.9. `RAYQUERY_FLAG_NONE` will be left unrestricted.
...
-When compiling a library with the
-`RaytracingPipelineFlags::AllowOpacityMicromaps` flag set in a
-[Raytracing pipeline config1][pipeline-config] subobject,
-a new DefaultError warning diagnostic will be added if the shader model is less
-than 6.9. This will detect the case where the flag is not spelled out and
-caught by AST traversal.
-
 Current HLSL diagnostics in DXC do not verify `RayFlags` values in any context.
 `TraceRay()` and `RayQuery::TraceRayInline()` accept non-immediate values, but
 the `RayFlags` provided as a template argument to `RayQuery` must be immediate.
```
tex3d, line 208: This paragraph was intended to provide context for the other checks when flags were immediate.  However, it seems a little out of place now.  I mentioned you could probably remove it in our conversation yesterday.

proposals/0024-opacity-micromaps.md
```diff
@@ -253,32 +236,22 @@ specified.
 Three DXIL operations accept `RayFlags` as input, but only one requires this
 input to be immediate: `AllocateRayQuery`.
```
tex3d, line 233: This isn't accurate anymore.  There are two trace ops and two allocate ops (for which this must be immediate) now.

proposals/0024-opacity-micromaps.md
```diff
@@ -134,10 +134,11 @@ When the `RayQueryFlags` template argument is non-zero, this new
...
 DXIL op is used.
 
-Currently, the `AllocateRayQuery` DXIL Op has the following function signature:
-`uint [[hidden]] AllocateRayQuery(in uint flags);`
-The new DXIL Op, `AllocateRayQuery2`, will instead have this signature:
-`uint [[hidden]] AllocateRayQuery2(in uint flags, in uint ray_query_flags);`
+The new DXIL Op, `AllocateRayQuery2`, will have this signature:
+```DXIL
+; Function Attrs:
+nounwind declare i32 @dx.op.allocateRayQuery2(i32 OpCode, i32 constRayFlags, i32 RayQueryFlags)
```
tex3d, line 140: nounwind got pushed to the wrong line.
```suggestion
; Function Attrs: nounwind
declare i32 @dx.op.allocateRayQuery2(i32 OpCode, i32 constRayFlags, i32 RayQueryFlags)
```

tex3d (approved):

damyanp:
@llvm-beanz - could you have a look at this please?  Change LGTM, but I suspect it'd be good to get more experienced eyes over changes to validation like this.

proposals/0024-opacity-micromaps.md
```diff
@@ -315,21 +336,24 @@ See [Opacity Micromaps][dxr-omm] in the Raytracing spec for details.
 ### Diagnostics
 
 - Check availability-based diagnostics for each flag, including recursing
-  through DeclRefs and their initializers.
+  through DeclRefs, DeclExprRefs, and their initializers.
```
llvm-beanz, line 339: Pretty sure I understood what you meant, but I think this is more accurate.
```suggestion
  through DeclExprRefs to Decls and their initializers.
```

llvm-beanz (approved):
One small suggestion, but looks good.

proposals/0024-opacity-micromaps.md
bob80905: I think the LLVM type is actually DeclRefExpr, I'll use that instead.

PR #378, opened 2025-01-31
[0027] Shader Execution Reordering: rename "ReorderThread" to "MaybeReorderThread"

damyanp:
As noted by @Jasper-Bekkers in [this comment](https://github.com/microsoft/hlsl-specs/pull/277#issuecomment-2616782435), "MaybeReorderThread" may a better name to make it clear that some implementations may not actually reorder threads at this point.

### DIFF
--- a/proposals/0027-shader-execution-reordering.md
+++ b/proposals/0027-shader-execution-reordering.md
@@ -10,11 +10,11 @@ Michael Haidl, Simon Moll, Martin Stich
 
 ## Introduction
 
-This proposal introduces `ReorderThread`, a built-in function for raygeneration shaders to
+This proposal introduces `TryReorderThread`, a built-in function for raygeneration shaders to
 explicitly specify where and how shader execution coherence can be improved.
 Additionally, `HitObject` is introduced to decouple traversal, intersection
 testing and anyhit shading from closesthit and miss shading. Decoupling these
-shader stages gives an increase in flexibility and enables `ReorderThread` to
+shader stages gives an increase in flexibility and enables `TryReorderThread` to
 improve coherence for closesthit and miss shading, as well as subsequent operations.
 
 ## Motivation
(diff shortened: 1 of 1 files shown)


amarpMSFT:
I'd suggest TryReorderThread(), the "Try" has some precedence in other APIs IIRC.  Maybe maybe does as well though :).

damyanp:
> I'd suggest TryReorderThread(), the "Try" has some precedence in other APIs IIRC. Maybe maybe does as well though :).

So I did have a version that used Try, and I felt icky doing that one because I strongly expect a function called TryXXX to return a boolean indicating whether it did or did not do what was asked.

Jasper-Bekkers:
> I'd suggest TryReorderThread(), the "Try" has some precedence in other APIs IIRC. Maybe maybe does as well though :).

The suggestion came from https://doc.rust-lang.org/beta/std/mem/union.MaybeUninit.html - it other places Try also got suggested but it wouldn't have my preference since it would imply a slightly different behavior: "try to do this operation and return success/failure", rather then "I'll do this, or not".

rasmusnv:
Could we instead add a feature flag that indicates whether `ReorderThread` is expected to be honored or be a no-op on a particular device? Of course, a particular invocation of `ReorderThread` could be ignored, e.g., at the tail end of a dispatch as work is running out, but I'm not sure that is actionable information.

damyanp:
> Could we instead add a feature flag that indicates whether `ReorderThread` is expected to be honored or be a no-op on a particular device? Of course, a particular invocation of `ReorderThread` could be ignored, e.g., at the tail end of a dispatch as work is running out, but I'm not sure that is actionable information.

I think we should consider adding the feature flag regardless, but I don't think that changes the feedback on the name.  When authoring / reading shaders, you don't necessarily know what device it will run on so the "Maybe" clue I think still has value.

llvm-beanz (approved):

spall (approved):

PR #379, opened 2025-01-31
Require SER support on any SM 6.9 device with raytracing support.

amarpMSFT:
- Require devices that support SM 6.9 and raytracing to support SER, whether it can reorder or not.  If not, ReorderThread() is just a no-op.
- Expose D3D runtime cap indicating BOOL ShaderExecutionReorderingActuallyReorders
- Example of using SER to just TraceRay without Invoke() as a way to just get T value from the HitObject without running ClosestHit/Miss. This works on any device.

### DIFF
--- a/proposals/0027-shader-execution-reordering.md
+++ b/proposals/0027-shader-execution-reordering.md
@@ -1527,3 +1527,56 @@ Validation errors:
 - Validate that `opcode` is one of the supported opcodes in the table above.
 - Validate that `hit object` is not undef.
 - Validate that `index`, `row`, and `col` are constant and in a valid range.
+
+## Device Support
+
+Devices that support Shader Model 6.9 and raytracing must support the 
+Shader Execution Reordering HLSL methods in this spec. This doesn't mean 
+all devices must support performing thread reordering - it is valid for 
+an implementation to do nothing there.  Applications write one codebase 
+using SER, and devices that can take advantage will, and other devices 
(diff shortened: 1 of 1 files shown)


proposals/0027-shader-execution-reordering.md
```diff
@@ -1527,3 +1527,56 @@ Validation errors:
...
+of SER can be useful.
+
+For instance, suppose an app wants to trace a ray, potentially including AnyHit 
+shader invocations, and just wants the final T value without running
+the ClosestHit shader (even if it happens to exist in the HitGroup).
+
+The app can call `TraceRay()` returning a `HitObject`, call
+`GetRayTCurrent()` on the `HitObject` to get the `T` value and be done.
+Not calling `Invoke()`, skips `ClosestHit`/`Miss` invocation, and this 
+works on any device with Shader Model 6.9 support.
```
damyanp, line 1577: > Not calling `Invoke()`, skips `ClosestHit`/`Miss` invocation, and this works on any device with Shader Model 6.9 support.

Is / does this need to be mentioned in the detailed design for HitObject?  ie that all implementations need to support Invoke() never being called?

damyanp (approved):

damyanp:
@tex3d / @llvm-beanz - this spec change I think should be pretty uncontroversial?

joaosaffran (approved):
LGTM

llvm-beanz (approved):

PR #382, opened 2025-02-05
[SER] Add CommittedCustomHitKind to HitObject::FromRayQuery

rasmusnv:
There is currently no way to specify HitKind for a HitObject created from RayQuery with COMMITTED_PROCEDURAL_PRIMITIVE_HIT and custom attributes. This PR adds another parameter to the call to allow this.

### DIFF
--- a/proposals/0027-shader-execution-reordering.md
+++ b/proposals/0027-shader-execution-reordering.md
@@ -177,10 +177,10 @@ If no hit is committed in the RayQuery,
 the HitObject returned is a NOP-HitObject. A shader table record can be assigned
 separately, which in turn allows invoking a shader.
 
-An overload takes custom attributes associated with
-COMMITTED_PROCEDURAL_PRIMITIVE_HIT. It is ok to always use the overload, even
-for COMMITTED_TRIANGLE_HIT. For anything other than a procedural hit, the
-specified attributes are ignored.
+An overload takes a user-defined hit kind and custom attributes associated with
+COMMITTED_PROCEDURAL_PRIMITIVE_HIT (see `ReportHit` for details).
+It is ok to always use the overload, even for COMMITTED_TRIANGLE_HIT. For anything
+other than a procedural hit, the specified hit kind and attributes are ignored.
 
 ```C++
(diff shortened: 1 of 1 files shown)


damyanp:
@llvm-beanz / @tex3d - the looks like a straightforward spec change?

simoll:
The two FromRayQuery DXIL opcodes have a different parameter count (and it's not just the `udt` for the attributes anymore). The canonical way to implement this in DXC is to have distrinct DXIL intrinsic declarations (== opcode classes). Can we change the spec to give the DXIL opcode `HitObject_FromRayQueryWithAttrs` the declaration `@dx.op.hitObject_FromRayQueryWithAttrs.Attr`?

llvm-beanz (approved):

proposals/0027-shader-execution-reordering.md
```diff
@@ -1338,6 +1341,9 @@ Validation errors:
 - Validate the compatibility of type `AttrT`.
 - Validate that `attributes` is a valid pointer.
 
+Validation warnings:
```
tex3d, line 1344: I don't think there are any validation "warnings".  I think this should be "errors"
```suggestion
Validation errors:
```

Looking a bit more broadly in the document, I see this pattern elsewhere.  Were these intended to describe Sema diagnostics on the HLSL intrinsic?  I think that should be in a different section.

I also see things like:
```md
Validation errors:
- Validate that `opcode` equals `HitObject_MakeNop`.
```

I don't think there is any specific validation for this feature here.  There is general DXIL op checking if you mean matching expected function name based on OpCodeClass with the DxilOp argument.

None of these changes are necessary as part of this PR, but you could take the suggestion on this new one if you'd like.  I think a follow up issue to clean this up should be filed though.

tex3d (approved):

proposals/0027-shader-execution-reordering.md
tex3d: I filed issue #444 to capture the clean-up follow-up so this may be resolved and merged without fixing this local change for now.

PR #389, opened 2025-02-11
Really really this time fix pr-labeler (I swear)

llvm-beanz:
This is such a pain to deal with, but I setup a test bed in my fork and verified that this actually really works this time.

Here's the link to my test PR: https://github.com/llvm-beanz/hlsl-specs/pull/3

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -16,4 +16,4 @@ jobs:
         ref: main
     - uses: actions/labeler@v5
       with:
-          configuration-path: .github/prs-labels.yml
+          configuration-path: .github/pr-labels.yml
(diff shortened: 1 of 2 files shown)


damyanp (approved):
Fixed for realz, right?

.github/workflows/pr-labeler.yml
```diff
@@ -16,4 +16,4 @@ jobs:
         ref: main
     - uses: actions/labeler@v5
       with:
-          configuration-path: .github/prs-labels.yml
+          configuration-path: .github/pr-labels.yml
```
bfavela, line 19: Is this whitespace correct? It doesn't line up with the above lines at the same indentation (lines 15-16)

llvm-beanz: It works... I YAML isn't too picky about how much you indent sub-objects, just that they are further indented.

PR #390, opened 2025-02-14
[0029] Cooperative Vectors: add proposal

damyanp:


llvm-beanz (approved):
There are aspects of this spec that are still in flux, but we've done a bunch of iteration internally so I'm comfortable with this going in as under consideration.

tex3d (approved):
I think we can take this in "under consideration" while we have issues to track outstanding work.

Icohedron:
Just noting a few typos I caught yesterday for forgot to comment on :P

PR #393, opened 2025-02-17
Begin specifying object model

### LINKED ISSUE #372: Define object, subobject, and non-divisible sub-subobject
In the language specification under [*Decl.Init.Agg*](https://microsoft.github.io/hlsl-specs/specs/hlsl.html#Decl.Init.Agg), and in #358, we talk about objects and flattened orderings.

We should really have unified definitions earlier in the spec for "object", "subobject", "flattened ordering" and a notion of a non-divisible subobject (a subobject that does not contain subobjects).

llvm-beanz:
This change begins to define the object model for HLSL and clearly denotes aggregate and non-aggregate types. It provides base definitions for sub-object ordering and scalar elementwise ordering. It also defines non-decomposable objects and stubs out sections for constructon and access control.

HLSL does not currently support user-defined constructors or access control specifiers, however both are used implicitly and denote non-decomposible objects.

Fixes #372

### DIFF
--- a/specs/language/declarations.tex
+++ b/specs/language/declarations.tex
@@ -74,37 +74,29 @@ initializers regardless of the context.
 
 \Sub{Aggregate Initialization}{Decl.Init.Agg}
 
-\p An \textit{aggregate} is a vector, matrix, array, or class.
-
-\p The subobjects of an aggregate have a defined order. For vectors and arrays
-the order is increasing subscript order. For matrices it is increasing subscript
-order with the subscript nesting such that in the notation
-\texttt{Mat[M][N]}, the ordering is \(Mat[0][0]...Mat[0][N]...
-Mat[M][0]...Mat[M][N]\). For classes the order is base class, followed by member
-subobjects in declaration order.
-
-\p A \textit{flattened ordering} of subobjects can be produced by performing a
-depth-first traversal of the subobjects of an object following the defined
-subobject ordering.
+\p An \textit{aggregate} is a vector, matrix, array, or class that does not have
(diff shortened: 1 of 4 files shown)


specs/language/declarations.tex
```diff
@@ -74,37 +74,29 @@
...
 \p Each \textit{braced initializer list} is comprised of zero or more
 \textit{initializer-clause} expressions, which is either another braced
 initializer list or an expression which generates a value that either is or can
 be implicitly converted to an rvalue. Each assignment-expression is an object,
-which may be a scalar or aggregate type. A \textit{flattened initializer
-sequence} is constructed by a depth-first traversal over each
-assignment-expression in an initializer-list and performing a depth-first
-traversal accessing each subobject of the assignment-expression.
+which may be a scalar or aggregate type. A \textit{element-wise initializer
+sequence} is constructed by concatenating the scalar element ordering
```
farzonl, line 88: I think I see how this would work for matricies and vectors and 1 dim arrays. I'm not as clear on how it works for multi dim Arrays and Classes of classes/structs. I think the `depth-first traversal` algorithm  covered finding these nested initalizers. Can you explain what `concatenating` is doing here to cover cases like arrays of classes and classes of classes? 

Second should we define somewhere what `scalar element ordering` is?

When I think scalar i'm thinking the non container types like primitive types like float/int/bool/etc which makes me think we are excluding classes, but thats at odds with the types Aggregate Initialization covers.

llvm-beanz: Are you suggesting something other than the `(\ref{Intro.Object.SubOrdering})` that is right after the use of the term?

PR #405, opened 2025-02-22
Rough proposal collecting thoughts on uniformity

llvm-beanz:
This has been sitting around in my checkout for close to a year, and I've been meaning to finish dumping thoughts into it and post it for wider review and feedback. So here goes!

Please be gentle, it's real rough.

### DIFF
--- a/proposals/NNNN-uniformity-qualifiers.md
+++ b/proposals/NNNN-uniformity-qualifiers.md
@@ -3,8 +3,9 @@
 # Uniformity Qualifiers
 
 * Proposal: [NNNN](NNNN-uniformity-qualifiers.md)
-* Author(s): [Chris Bieneman](https://github.com/llvm-beeanz)
-* Sponsor: [Chris Bieneman](https://github.com/llvm-beeanz)
+* Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
+* Sponsor: [Chris Bieneman](https://github.com/llvm-beanz)
+
 * Status: **Under Consideration**
 
 ## Introduction
--- a/proposals/NNNN-uniformity-qualifiers.md
+++ b/proposals/NNNN-uniformity-qualifiers.md
@@ -116,8 +116,8 @@ incorrect source annotations.
 
 ### Uniformity Requirements for Functions
 
-HLSL should introduce a new set of attributes for defining the control flow
-uniformity requirements of functions. These new attributes should take the form:
+This proposal introduces a new set of attributes for defining the control flow
+uniformity requirements of functions. These new attributes take the form:
 
 ```hlsl
 [[hlsl::required_uniform(group|simd|quad)]]
--- a/proposals/NNNN-uniformity-qualifiers.md
+++ b/proposals/NNNN-uniformity-qualifiers.md
@@ -35,7 +35,7 @@ Uniformity can refer to data or control flow. If a variable has the same value
 across all threads in a scope, it is said to be _uniform_ across that scope.
 Similarly if all threads within a scope are actively executing instructions
 within a control flow block, the control flow is said to be _uniform control
-flow_.
+flow_ across that scope.
 
 ## Motivation
 
--- a/proposals/NNNN-uniformity-qualifiers.md
+++ b/proposals/NNNN-uniformity-qualifiers.md
@@ -51,7 +51,7 @@ uniformity to HLSL's type system and programming model.
 
 ### Uniformity as a Type Qualifier
 
-HLSL should introduce a new set of type qualifiers to represent the different
+This proposal introduces a new set of type qualifiers to represent the different
 scopes of uniformity:
 * `group_uniform`
 * `simd_uniform`
--- a/proposals/NNNN-uniformity-qualifiers.md
+++ b/proposals/NNNN-uniformity-qualifiers.md
@@ -1,0 +1,131 @@
+<!-- {% raw %} -->
+
+# Uniformity Qualifiers
+
+* Proposal: [NNNN](NNNN-uniformity-qualifiers.md)
+* Author(s): [Chris Bieneman](https://github.com/llvm-beeanz)
+* Sponsor: [Chris Bieneman](https://github.com/llvm-beeanz)
+* Status: **Under Consideration**
+
+## Introduction
+
+The HLSL Single Program Multiple Data (SPMD) programming model defines a program
+in terms of how it operates on a single element of data. An SPMD program may be
+executed on a traditional scalar processor or a Single Instruction Multiple Data
+(SIMD) processor. When executing on a SIMD processor the program may execute
+where a single instruction produces results for multiple threads of execution
+from the source programming model. This is sometimes referred to as Single
+Instruction Multiple Threads (SIMT).
+
+Under HLSL's execution model, groups of threads form hierarchical scopes:
+* A _dispatch_ represents the full set of threads spawned from a CPU API
+  invocation.
+* A _thread group_ represents a subset of a dispatch that can execute
+  concurrently.
+* A _wave_ represents a subset of a thread group that represents in a single
+  SIMD processor.
+* A _quad_ represents a grouping of four adjacent threads in a wave.
+
+When a shader is executing threads concurrently on one or more processing cores,
+an emergent property of _uniformity_ exists within the thread group, wave and
+quad scopes.
+
+Uniformity can refer to data or control flow. If a variable has the same value
+across all threads in a scope, it is said to be _uniform_ across that scope.
+Similarly if all threads within a scope are actively executing instructions
+within a control flow block, the control flow is said to be _uniform control
+flow_.
+
+## Motivation
+
+Uniformity of data and control flow are central concepts to SIMT execution
+models, and is required for correct execution of shader programs. Despite
+the importance of this fundamental property it is not represented in any
+explicit way in the HLSL language.
+
+This proposal, seeks to address that by introducing core concepts around
+uniformity to HLSL's type system and programming model.
+
+## Proposed solution
+
+### Uniformity as a Type Qualifier
+
+HLSL should introduce a new set of type qualifiers to represent the different
+scopes of uniformity:
+* `group_uniform`
+* `simd_uniform`
+* `quad_uniform`
+* non-uniform (default state with no associated keyword)
+
+`group_uniform` is the highest scoping of uniformity, and implies all other
+scopes. `simd_uniform` implies `quad_uniform`.
+
+A new "UniformityReduction" cast will reduce the uniform scope allowing
+conversion of one uniformity scope to another uniformity scope as long as the
+source scope has a greater uniformity scope.
+
+Any GLValue with a uniformity scope can be implicitly converted to a GLValue
+with reduced uniform scope or no uniformity scope.
+
+Any PRValue with a uniformity scope can be implicitly converted to a PRValue
+with reduced uniform scope or no uniformity scope.
+
+No implicit or explicit cast can increase uniformity scope.
+
+HLSL library functionality that produces uniform results will be updated to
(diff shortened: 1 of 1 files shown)


proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,131 @@
+<!-- {% raw %} -->
+
+# Uniformity Qualifiers
+
+* Proposal: [NNNN](NNNN-uniformity-qualifiers.md)
+* Author(s): [Chris Bieneman](https://github.com/llvm-beeanz)
+* Sponsor: [Chris Bieneman](https://github.com/llvm-beeanz)
```
damyanp, line 7:
```suggestion
* Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
* Sponsor: [Chris Bieneman](https://github.com/llvm-beanz)

```

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,131 @@
...
+Instruction Multiple Threads (SIMT).
+
+Under HLSL's execution model, groups of threads form hierarchical scopes:
+* A _dispatch_ represents the full set of threads spawned from a CPU API
+  invocation.
+* A _thread group_ represents a subset of a dispatch that can execute
+  concurrently.
+* A _wave_ represents a subset of a thread group that represents in a single
+  SIMD processor.
+* A _quad_ represents a grouping of four adjacent threads in a wave.
```
damyanp, line 27: Worth including Vulkan terminology here as well?

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,131 @@
...
+
+When a shader is executing threads concurrently on one or more processing cores,
+an emergent property of _uniformity_ exists within the thread group, wave and
+quad scopes.
+
+Uniformity can refer to data or control flow. If a variable has the same value
+across all threads in a scope, it is said to be _uniform_ across that scope.
+Similarly if all threads within a scope are actively executing instructions
+within a control flow block, the control flow is said to be _uniform control
+flow_.
```
damyanp, line 37:
```suggestion
within a control flow block, the control flow is said to be _uniform control
flow_ across that scope.
```

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Motivation
+
+Uniformity of data and control flow are central concepts to SIMT execution
+models, and is required for correct execution of shader programs. Despite
+the importance of this fundamental property it is not represented in any
+explicit way in the HLSL language.
```
damyanp, line 44: Maybe mention that this requires backend compilers to attempt to analyse for uniformity?

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,131 @@
...
+
+This proposal, seeks to address that by introducing core concepts around
+uniformity to HLSL's type system and programming model.
+
+## Proposed solution
+
+### Uniformity as a Type Qualifier
+
+HLSL should introduce a new set of type qualifiers to represent the different
+scopes of uniformity:
```
damyanp, line 54:
```suggestion
This proposal introduces a new set of type qualifiers to represent the different
scopes of uniformity:
```

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,131 @@
...
+* `group_uniform`
+* `simd_uniform`
+* `quad_uniform`
```
damyanp, line 57: It'd be good to explain how these relate to the dispatch, thread group, wave etc. scopes that were defined above.

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,131 @@
...
+* non-uniform (default state with no associated keyword)
```
damyanp, line 58: Might be nice to have a way to explicitly mark non-uniform?  Would leave open possibility of a "strict" mode where uniformity annotations are required and would allow developers to make their intent more explicit.

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,131 @@
...
+
+Uniformity qualifiers may be applied on shader inputs. When applied to an input
+the compiler will diagnose known cases where the qualifier mismatches, and it
+will trust the user in other cases. A runtime validation may be added to catch
+incorrect source annotations.
+
+### Uniformity Requirements for Functions
+
+HLSL should introduce a new set of attributes for defining the control flow
+uniformity requirements of functions. These new attributes should take the form:
```
damyanp, line 119: Wondering if we should try and frame this as what this proposal is proposing, rather than what we think someone needs to propose?

```suggestion
This proposal introduces a new set of attributes for defining the control flow
uniformity requirements of functions. These new attributes take the form:
```

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,131 @@
...
+an emergent property of _uniformity_ exists within the thread group, wave and
+quad scopes.
+
+Uniformity can refer to data or control flow. If a variable has the same value
+across all threads in a scope, it is said to be _uniform_ across that scope.
+Similarly if all threads within a scope are actively executing instructions
+within a control flow block, the control flow is said to be _uniform control
+flow_.
+
+## Motivation
```
damyanp, line 39: I'd love to know what, if anything, this proposal does to NonUniformResourceIndex.

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,131 @@
...
+models, and is required for correct execution of shader programs. Despite
+the importance of this fundamental property it is not represented in any
+explicit way in the HLSL language.
+
+This proposal, seeks to address that by introducing core concepts around
+uniformity to HLSL's type system and programming model.
+
+## Proposed solution
+
+### Uniformity as a Type Qualifier
```
damyanp, line 51: Are there alternative approachs to using type qualifiers?  I think doing so means we need to change the grammar?

Could something like:

```
group_uniform<int> myValue
```

Be made to work?

damyanp (approved):

llvm-beanz (commented):
One more note for myself here:

The `[[hlsl::required_uniform(..)]]` attribute should imply to a function that it's initial control flow uniformity is the value set, otherwise all non-entry functions should assume default uniformity as non-uniform. Entry functions should assume default `group` uniformity.

The uniformity of each control flow block should be the intersection of the uniformity of the parent block and the condition (if any) that the block be entered.

proposals/NNNN-uniformity-qualifiers.md
llvm-beanz: It should cease to exist, and I should capture that.

proposals/NNNN-uniformity-qualifiers.md
llvm-beanz: If we have a template and non-template spelling that don't use the same words we could do something similar. For example:

```c++
template<typename T>
using GroupUniform = group_uniform T
```

Would simply enable:
```hlsl
GroupUniform<int> MyInt;
```

proposals/NNNN-uniformity-qualifiers.md
llvm-beanz: I need to think on this. My intent was that we would only have a strict mode, so explicit uniformity would be required everywhere that requires uniformity, and that no annotation would mean non-uniform. With that approach I'm not sure how an explicit `non_uniform` helps, but maybe I'm missing something.

damyanp: I'm mainly wanting to avoid developers having to come up with naming or commenting conventions to non-uniformity in their code.

magcius:
Possibly related, it might also be worth talking about the whole "dynamic uniform" vs. "subgroup uniform" behavior. As dumb as it sounds, is `SV_GroupID` uniform across the group? Across the wave? Across the dispatch?

Basically, what are drivers/hardware allowed to pack together into waves? This sort of packing is more common in the draw side of things (SV_InstanceID is infamously not uniform), but so far it's been very undefined.

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,132 @@
...
+
+This proposal introduces a new set of type qualifiers to represent the different
+scopes of uniformity:
+* `group_uniform`
+* `simd_uniform`
+* `quad_uniform`
+* non-uniform (default state with no associated keyword)
```
Tobski, line 59: Is there any merit to having a `command_uniform` scope or similar indicating uniformity across all thread groups? E.g. for constants? There are, to my knowledge, some implementations that care about this (at least outside of DX12).

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,132 @@
...
+produce appropriately qualified uniform types. These functions can produce
+uniform values from non-uniform inputs. For example:
+
+```hlsl
+simd_uniform bool WaveActiveAllTrue(bool);
+quad_uniform bool QuadAny(bool);
+```
+
+Compile-time constants and Groupshared variable declarations imply
+`group_uniform` uniformity.
```
Tobski, line 86: Should constant buffer/SRV data also be automatically marked this way? They're supposed to be read-only and non-volatile for the lifetime of the shader AFAIU.

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,132 @@
...
+  if (GSVal > GSVal2) { // This control-flow can be defined as simd_uniform
+
+  }
+}
+```
+
+Uniformity qualifiers may be applied on shader inputs. When applied to an input
+the compiler will diagnose known cases where the qualifier mismatches, and it
+will trust the user in other cases. A runtime validation may be added to catch
+incorrect source annotations.
```
Tobski, line 115: It would be good to have some sort of interaction with `NonUniformResourceIndex` documented - feels like these qualifiers could just outright replace it.

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,132 @@
...
+* `simd_uniform`
+* `quad_uniform`
+* non-uniform (default state with no associated keyword)
+
+`group_uniform` is the highest scoping of uniformity, and implies all other
+scopes. `simd_uniform` implies `quad_uniform`.
+
+A new "UniformityReduction" cast will reduce the uniform scope allowing
+conversion of one uniformity scope to another uniformity scope as long as the
+source scope has a greater uniformity scope.
```
Tobski, line 66: Something i'm missing here is what happens if you did, e.g.:

```hlsl
group_uniform int a;
simd_uniform int b;
int c = a * b;
```

I would _expect_ c to automatically be treated as simd_uniform (i.e. the smaller of the scopes) as the result, but that's not spelled out here AFAICT.

Tobski: Oh wait now I look again that's what the next two sentences are saying - perhaps worth an example?

proposals/NNNN-uniformity-qualifiers.md
```diff
@@ -0,0 +1,132 @@
...
+uniformity to HLSL's type system and programming model.
+
+## Proposed solution
+
+### Uniformity as a Type Qualifier
+
+This proposal introduces a new set of type qualifiers to represent the different
+scopes of uniformity:
+* `group_uniform`
+* `simd_uniform`
```
amd-mherdeg, line 57: Given that HLSL already has wave ops, WaveSize, MatrixScope::Wave, etc., why not use wave_uniform?

PR #408, opened 2025-02-25
Report on impact of overload resolution behavior changes

llvm-beanz:
The HLSL draft specification aligns HLSL's overload resolution rules with C++ overload resolution rules. This will have an impact on source compatibility for shaders migrating from DXC to Clang.

This report analyzes the problem space, includes details about experimentation migrating an existing HLSL codebase to Clang, and suggests several possible options for mitigating the problem as well identifying a preferred approach to be implemented in Clang and reflected in the HLSL 202x language specification.

### DIFF
--- a/reports/Overload-Resolution.md
+++ b/reports/Overload-Resolution.md
@@ -218,7 +218,7 @@ conversions.
 This option is included only because it is technically feasible, not because it
 is a good idea.
 
-### Careful Curation of Overloads
+### Recommended Approach: Careful Curation of Overloads
 
 A more measured partial mitigation is to add additional overloads to builtin
 functions to resolve ambiguity in some but not all cases. Following this model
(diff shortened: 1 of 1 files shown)


reports/Overload-Resolution.md
```diff
@@ -0,0 +1,252 @@
...
+This proposal suggests that:
+* The first category of overloads should be added to HLSL as official language
+  overloads.
+* The second category of overloads should be added as HLSL 202x features to be
+  removed in HLSL 202y, and Clang should diagnose the implicit conversions and
+  warn about deprecation in 202x pending removal in 202y.
+* The third category of overloads should be left out from Clang and produce
+  errors when migrating from DXC to Clang.
+
+As the recommended approach the following issues have been filed:
```
spall, line 242: approach, the folllowing

V-FEXrt (approved):
What is the mechanical difference between a report and a proposal?

llvm-beanz:
> What is the mechanical difference between a report and a proposal?

Yea... I didn't really feel like this was a proposal since I'm not really proposing a change to the language behavior (although I guess that's slicing an odd dimension since the available overloads is part of the "standard library" definition).

We had previously (without a proposal) aligned the language specification on using a C++-like best-match algorithm, and I didn't really want to re-open previous discussions and navigate the proposal process. Maybe that was a mistake, but in that context I viewed this document more as a defect report with mitigations.

V-FEXrt:
> We had previously (without a proposal) aligned the language specification on using a C++-like best-match algorithm

Yeah, given that I see why it would be on the fence and ultimately you just have to pick a place for it. I'm happy having it as a report I just wanted to get an understanding for why it would be one versus the other.

> since the available overloads is part of the "standard library" definition

IMO the standard lib falls very clearly under the "spec of the language" but HLSL makes everything fuzzier with all the shader model specs. Do we have any code that is "standard library" but not specified by the SM?

llvm-beanz:
> Do we have any code that is "standard library" but not specified by the SM?

I think this is somewhat of a vague area that we need to work on. Ideally the "standard library" is more tied to language version than shader model. Because some of the APIs in the standard library have hardware requirements that not all targets support we'll likely need to define "required" and "optional" features in the standard library as we define it in the spec.

Shader Models also don't map cleanly to Vulkan/SPIRV, so we'll have to rectify that too.

PR #409, opened 2025-02-27
[0029] Extend ComponentType for interpretations

### LINKED ISSUE #402: [0029] Cooperative Vectors: consider using unique interpretation enums for variable vs memory
https://github.com/microsoft/hlsl-specs/pull/398#discussion_r1962890711

There are two types of interpretations in the spec currently:

1. Interpreting vectors stored in variables
2. Interpreting vectors / matrices store in memory

These both use the same enum.  This means that there are values in the enum that don't apply to the memory resources (ie the packed ones) that we'll need to do work to prohibit using.

If we used a different enum for these then we woudn't have this problem.  Having an additional enum might be confusing / annoying to work with though.

damyanp:
Previously, the spec used defined a new enum value to describe how to interpret vectors stored in memory. 
This enum was very similar to ComponentType.

This change makes it so that instead new values are added to ComponentType where necessary, and documents which values are valid in various contexts.  The D3D enum is updated to align with ComponentType.

Closes #402

### DIFF
--- a/proposals/0029-cooperative-vector.md
+++ b/proposals/0029-cooperative-vector.md
@@ -240,6 +240,17 @@ This operation returns a vector of size `NUMo` and contains elements of type
 `TYo`. The result vector does not have an interpretation parameter, its type is
 the declared type.
 
+#### Validation
+
+* **input interpretation** must be a value corresponding to one of the following
+  `ComponentType`s: `I16`, `U16`, `I32`, `U32`, `F16`, `F32`, `PackedS8x32`,
+  `PackedU8x32`, `U8`, `I8`, `F8_E4M3`, `F8_E5M2`.
+* **matrix interpretation** must be a value corresponding to one of the
+  following `ComponentType`s: `I16`, `U16`, `I32`, `U32`, `F16`, `F32`, `U8`,
+  `I8`, `F8_E4M3`, `F8_E5M2`, 
+* **bias vector interpretation** must be a value corresponding to one of the
+  following `ComponentType`s: `I16`, `U16`, `I32`, `U32`, `F16`, `F32`, `U8`,
+  `I8`, `F8_E4M3`, `F8_E5M2`, 
 
 ### Vector Outer Product
 
@@ -287,6 +298,12 @@ supported by all implementations. [CheckFeatureSupport] can be used to
 determine which combinations are supported. A list of combinations that are
 guaranteed to be supported on all implementations can be found in
 [Minimum Support Set].
+
+#### Validation
+
+* **matrix interpretation** must be a value corresponding to one of the
+  following `ComponentType`s: `I16`, `U16`, `I32`, `U32`, `F16`, `F32`, `U8`,
+  `I8`, `F8_E4M3`, `F8_E5M2`, 
 
 
 ### Reduce Sum Accumulate
--- a/proposals/0029-cooperative-vector.md
+++ b/proposals/0029-cooperative-vector.md
@@ -130,7 +130,7 @@ specification we add four operations:
 #### Syntax
  
 ``` llvm 
-declare <[NUMo] x [TYo] @dx.op.matvecmul.v[NUMo][TYo].v[NUMi][TYi](
+declare <[NUMo] x [TYo]> @dx.op.matvecmul.v[NUMo][TYo].v[NUMi][TYi](
     immarg i32        ; opcode
     <[NUMi] x [TYi]>, ; input vector
     immarg i32,       ; input interpretation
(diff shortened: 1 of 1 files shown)


proposals/0029-cooperative-vector.md
```diff
@@ -331,10 +331,13 @@ can be found in [Minimum Support Set].
 
 ### Type Interpretations
 
-The various "interpretation" arguments specify a value from the following enum:
+#### From-Register Interpretations
```
anupamachandra, line 334: So, I understand why you chose "Register" it feels too low level. So for the two types of enum we have are one that  reinterprets a type and the other that tells us wha the buffer type is, so how about `enum DXILTypeReinterpretation` and `enum DXILRawBufferElementType`

amarpMSFT: I like where Anupama is going with this, agreed "Register" seems implementation specific (immediately had my imagining some physical manifestation) - a name that more reflects the semantics would be better.   

In that sense DXILTypeReinterpretation seems to make sense.  And in the API maybe it's D3D12_LINEAR_ALGEBRA_VECTOR_REINTERPRET_DATATYPE_*  (trying out LINEAR_ALGEBRA for size, otherwise COOPERATIVE_VECTOR)?

Not sure about DXILRawBufferElementType though, that also seems specific -> maybe we'll have memory pointers some day and it isn't thought of as "RawBuffer".   Is the situation that what sits in memory is this type, and when it's accessed it's unpacked?  Like how DXGI_FORMAT works for typed buffers in D3D? If so should the name have "format" in it just like how we use DXGI_FORMAT to imply one format in memory and from shader point of view it's expanded?  For DXGI format naming we have multiple parrts in the name to indicate what's the memory layout vs what's the interpretation.  e.g. R16_FLOAT says it's float16 in memory but in the shader it's expanded to float32.  Could something like this style naming work?  Might not need to use the "R" channel convention.

damyanp: After various conversations it doesn't seem that we're making things much easier for driver implementations or for users of the D3D API calls by having separate enums.  However, it does seem that there may be some benefits from aligning with the existing ComponentType enum defined in DxilConstants.h.  The proposal has been updated to show what that would look like.

proposals/0029-cooperative-vector.md
```diff
@@ -527,19 +589,24 @@ typedef enum D3D12_FEATURE {
     D3D12_FEATURE_COOPERATIVE_VECTOR;
 };
 
+// This is designed to match the ComponentType enum values but omits data 
+// types that are not currently specified to work with this API. The names are chosen
+// to more closely match those used by HLSL developers, as opposed to the ComponentType 
+// names that align with LLVM IR.
```
damyanp, line 595: @amarpMSFT - does lining the values up to match the DXIL constants with the runtime ones make sense?  I wonder if we really want to be tieing the runtime API with a particular interchange format at all?

proposals/0029-cooperative-vector.md
```diff
@@ -331,36 +331,98 @@ can be found in [Minimum Support Set].
...
+
+Input vectors stored in registers (eg `vector<float, 16>`) are interpreted
+according to the Conversion Rules shown below.
+
+For these vectors there is a distinction between the physical type and the
+logical type. The **input interpretation** argument for these vectors describes
+how to convert from the physical to logical type. This allows elements to be
+interpreted as types not natively supported by HLSL, e.g. uint8/sint8. For
+packed interpretations, a single physical element can expand into multiple
+logical elements.
```
alsepkow, line 380: Would there ever be a reason to do multiple physical elements to a single logical element? Just curious.

proposals/0029-cooperative-vector.md
```diff
@@ -130,7 +130,7 @@ specification we add four operations:
 #### Syntax
  
 ``` llvm 
-declare <[NUMo] x [TYo] @dx.op.matvecmul.v[NUMo][TYo].v[NUMi][TYi](
+declare <[NUMo] x [TYo]> @dx.op.matvecmul.v[NUMo][TYo].v[NUMi][TYi](
```
damyanp, line 133: For each opcode explicitly list the set of valid values for each argument.

llvm-beanz (approved):
Discussed with @damyanp offline. I approve with the comment that we should explicitly list the validation rules for each opcode so that we know what enum values are valid in the arguments of the opcode.

proposals/0029-cooperative-vector.md
damyanp: This thread is moot with recent changes, so marking it resolved.

proposals/0029-cooperative-vector.md
damyanp: Maybe for 64-bit types, but the trend is for less bits-per-element!

proposals/0029-cooperative-vector.md
```diff
@@ -332,36 +349,98 @@ can be found in [Minimum Support Set].
 
 ### Type Interpretations
 
-The various "interpretation" arguments specify a value from the following enum:
+The `ComponentType` enum in `DxilConstants.h` is extended as shown below, with
```
anupamachandra, line 352: Do we want a file reference in the spec? isn't the enumeration alone sufficient?

damyanp: Learning about the location of these enums was pretty key to be understanding parts of the discussion around this, so I think it is worth including to be clear what we mean when we list enums.

proposals/0029-cooperative-vector.md
amarpMSFT: It doesn't hurt to line them up if the intended use is only with HLSL use anyhow.   If we ever ended up in a path where someone is feeding type choices of some kind from the API it would be handy if they lined up.

proposals/0029-cooperative-vector.md
```diff
@@ -528,19 +607,24 @@ typedef enum D3D12_FEATURE {
...
-    D3D12_COOPERATIVE_VECTOR_DATATYPE_UINT16       = 3,
-    D3D12_COOPERATIVE_VECTOR_DATATYPE_UINT32       = 4,
-    D3D12_COOPERATIVE_VECTOR_DATATYPE_SINT8        = 5,
-    D3D12_COOPERATIVE_VECTOR_DATATYPE_SINT16       = 6,
-    D3D12_COOPERATIVE_VECTOR_DATATYPE_SINT32       = 7,    
-    D3D12_COOPERATIVE_VECTOR_DATATYPE_SINT8_PACKED = 8,
-    D3D12_COOPERATIVE_VECTOR_DATATYPE_UINT8_PACKED = 9,
-    D3D12_COOPERATIVE_VECTOR_DATATYPE_FLOAT_E4M3   = 10,      // FP8: 1 sign bit, 4 exp bits, 3 mantissa bits
-    D3D12_COOPERATIVE_VECTOR_DATATYPE_FLOAT_E5M2   = 11       // FP8: 1 sign bit, 5 exp bits, 2 mantissa bits
+  D3D12_COOPERATIVE_VECTOR_DATATYPE_SINT16          =  2, // ComponentType::I16
```
amarpMSFT, line 616: Can we rename D3D12_COOPERATIVE_VECTOR_* to D3D12_LINEAR_ALGEBRA_*?  (can be separate PR if that's more appropriate).  My assumption is in HLSL the name space is dx::linalg?

Even for the tier.  I'd imagine if we add matrix*matrix support it would be TIER_2_0.   And if we add capabilities specific to coop vec in the future, they would just be independent caps (and/or also part of TIER_2_0).

damyanp: Let's do that in a separate PR.

amarpMSFT (approved):

PR #412, opened 2025-03-01
[0029] Remove isResultSigned and use different opcodes instead

### LINKED ISSUE #399: [0029] Remove "isResultSigned" arguments and use separate opcodes when signedness matters
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md

Currently, the matvecmul / matvecmuladd DXIL operations have a "isResultSigned" argument to indicate whether or not the return value is signed:

``` llvm 
declare <[NUMo] x [TYo] @dx.op.matvecmul.v[NUMo][TYo].v[NUMi][TYi](
    i32               ; opcode
    <[NUMi] x [TYi]>, ; input vector
    i32,              ; input interpretation
    %dx.types.Handle, ; matrix resource
    i32,              ; matrix offset
    i32,              ; matrix interpretation
    i32,              ; matrix M dimension    
    i32,              ; matrix K dimension    
    i32,              ; matrix layout
    i32,              ; matrix transpose
    i32,              ; matrix stride
    i1)               ; isResultSigned
```

Some issues with this:

* it is inconsistent with how other DXIL operations are defined where we use a different opcode for floats vs signed vs unsigned (for example, FMax, SMax, IMax)
* it is unclear what isResultSigned=1 would mean for a non-integer type, and so the argument should really be ignored.

Suggested solution:

* make float, unsigned and signed variants of these opcodes
* remove isResultSigned.

Maybe something like:

```llvm
declare <[NUMo] x [TYo]> @dx.op.matvecmul.[F|U|I].v[NUMo][TYo].v[NUMi][TYi]( ... )
```

damyanp:
In the same way that many existing DXIL operations have "F", "I" and "U" variants, this change describes using the same pattern for the matrix-vector-multiply operations.

Closes #399

### DIFF
--- a/proposals/0029-cooperative-vector.md
+++ b/proposals/0029-cooperative-vector.md
@@ -129,9 +129,8 @@ specification we add four operations:
 
 #### Opcodes
 
-Different opcode values are used to differentiate between float, signed
-integer and unsigned integer (in much the same was as `FMax`, `IMax` and `UMax`
-do.)
+Different opcode values are used to differentiate between float, signed integer
+and unsigned integer (in much the same way as `FMax`, `IMax` and `UMax` do.)
 
 > The exact numeric values are TBD.  Here the values are shown as "offset from
 > the first opcode value."
--- a/proposals/0029-cooperative-vector.md
+++ b/proposals/0029-cooperative-vector.md
@@ -127,10 +127,29 @@ specification we add four operations:
 
 ### Matrix-Vector Multiply and Multiply-Add Operations
 
+#### Opcodes
+
+Different opcode values are used to differentiate between float, unsigned
+integer and signed integer (in much the same was as `FMax`, `IMax` and `UMax`
+do.)
+
+> The exact numeric values are TBD.  Here the values are shown as "offset from
+> the first opcode value."
+
+| Opcode | Name          | Description                                                           |
+|--------|---------------|-----------------------------------------------------------------------|
+| 0      | FMatVecMul    | Matrix-Vector multiply that returns a vector of floats                |
+| 1      | FMatVecMulAdd | Matrix-Vector multiply-add that returns a vector of floats            |
+| 2      | IMatVecMul    | Matrix-Vector multiply that returns a vector of signed integers       |
+| 3      | IMatVecMulAdd | Matrix-Vector multiply-add that returns a vector of signed integers   |
+| 4      | UMatVecMul    | Matrix-Vector multiply that returns a vector of unsigned integers     |
+| 5      | UMatVecMulAdd | Matrix-Vector multiply-add that returns a vector of unsigned integers |
+
+
 #### Syntax
  
 ``` llvm 
-declare <[NUMo] x [TYo] @dx.op.matvecmul.v[NUMo][TYo].v[NUMi][TYi](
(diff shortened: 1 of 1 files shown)


proposals/0029-cooperative-vector.md
```diff
@@ -127,10 +127,29 @@ specification we add four operations:
 
 ### Matrix-Vector Multiply and Multiply-Add Operations
 
+#### Opcodes
+
+Different opcode values are used to differentiate between float, unsigned
```
anupamachandra, line 132: minor nit: swap unsigned and signed to match with the order of `Imax` and `Umax`

damyanp: like this?

anupamachandra:
LGTM!

proposals/0029-cooperative-vector.md
```diff
@@ -127,10 +127,29 @@ specification we add four operations:
 
 ### Matrix-Vector Multiply and Multiply-Add Operations
 
+#### Opcodes
+
+Different opcode values are used to differentiate between float, signed
+integer and unsigned integer (in much the same was as `FMax`, `IMax` and `UMax`
```
bob80905, line 133: nit: ways as `FMax`

bob80905 (approved):

anupamachandra (approved):

proposals/0029-cooperative-vector.md
```diff
@@ -127,10 +127,29 @@ specification we add four operations:
...
+> the first opcode value."
+
+| Opcode | Name          | Description                                                           |
+|--------|---------------|-----------------------------------------------------------------------|
+| 0      | FMatVecMul    | Matrix-Vector multiply that returns a vector of floats                |
+| 1      | FMatVecMulAdd | Matrix-Vector multiply-add that returns a vector of floats            |
+| 2      | IMatVecMul    | Matrix-Vector multiply that returns a vector of signed integers       |
+| 3      | IMatVecMulAdd | Matrix-Vector multiply-add that returns a vector of signed integers   |
+| 4      | UMatVecMul    | Matrix-Vector multiply that returns a vector of unsigned integers     |
+| 5      | UMatVecMulAdd | Matrix-Vector multiply-add that returns a vector of unsigned integers |
```
tex3d, line 145: I'm not sold on the opcode split.

I just identified another potential missing boolean property from the operation (#413) and proposed we add an immediate flags parameter (#414) instead of encoding these separate properties in bools, and instead of splitting the operation by float, signed int, and unsigned int.

Some advantages to flags, as I see it:
- This is closer to the way the [NV SPIR-V extension](https://github.khronos.org/SPIRV-Registry/extensions/NV/SPV_NV_cooperative_vector.html) uses additional optional operands for `MatrixResultSignedComponents` and `MatrixBSignedComponents`.  Although transpose is a separate boolean value in that definition, I don't see why it couldn't have used the extra optional operand in the same way.
- This would reduce the number of extra arguments to the operation
- Lowering and handling each of these complex operations is likely going to be unified, thus triplets of opcodes will need to be matched/translated to/from a separate signed/unsigned property to collapse common code paths anyway.
- We don't technically need to separate `I` and `F` operations, since that will be already identified as part of the llvm type and in the overload name.

On another note, if we do split the opcodes like this, I think it would be better to group the `(F|I|U)MatVecMul` together and the `(F|I|U)MatVecMulAdd` together, rather than interleaving these to group by return type.

PR #416, opened 2025-03-02
[0029] Use flags to indicate signedness of input/output vectors

### LINKED ISSUE #399: [0029] Remove "isResultSigned" arguments and use separate opcodes when signedness matters

### LINKED ISSUE #413: [0029] missing signed property for integer inputs required for conversions
https://github.com/microsoft/hlsl-specs/blob/3102cc58682f921b3a3570f022d1750dbabd5f05/proposals/0029-cooperative-vector.md?plain=1#L377-L380

In the above snippet under the [Conversion Rules](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md#conversion-rules) section, conversions from integers to smaller integer types with saturation and to floating point types are defined. These conversions cannot be performed correctly without knowing whether the input value is a signed or unsigned encoding.  This knowledge is not encoded into the llvm type, so we need something else to encode this property.

Here's an example:

```llvm
%inputVector = <8 x i32> ...

%result = <32 x i32> call @dx.op.matvecmul.v[32][i32].v[8][i32](
     OPCODE,
     %inputVector,
     2,               ; input interpretation - UnsignedInt8
     %matrixResource,
     0,               ; matrix offset
     2,               ; matrix interpretation - UnsignedInt8
     32,              ; matrix M dimension
     32,              ; matrix K dimension
     2,               ; matrix layout - InferencingOptimal
     0,               ; matrix transpose - false
     0,               ; matrix stride
     0);              ; isResultSigned - false
```

In this example, saturating translation to UnsignedInt8 would produce a different result for negative input values depending on the interpretation of those bits.

Perhaps the intent was to require a matching between the signed/unsigned status of the input encoding and the input interpretation to which we translate those vector components.  This isn't called out explicitly if so.  Additionally, it doesn't solve the issue of the integer to float conversion.  In that case, do we require all inputs to be a signed encoding?  This too is not called out in the spec.
 
**Additional context**

Note that the equivalent [SPV_NV_cooperative_vector](https://github.khronos.org/SPIRV-Registry/extensions/NV/SPV_NV_cooperative_vector.html) SPIR-V extension uses an additional `MatrixBSignedComponents` operand to indicate this property.

Also, see related suggestion #414 for introducing a flags field to encode this and other properties.

### LINKED ISSUE #414: [0029] Propose using a flags field instead of separate immediate properties
See the [Syntax](https://github.com/microsoft/hlsl-specs/blob/3102cc58682f921b3a3570f022d1750dbabd5f05/proposals/0029-cooperative-vector.md#syntax) section of the `0029-cooperative-vector.md` proposal:

https://github.com/microsoft/hlsl-specs/blob/3102cc58682f921b3a3570f022d1750dbabd5f05/proposals/0029-cooperative-vector.md?plain=1#L132-L163

In these operations, two immediate properties values are provided as separate `i1` parameters: `matrix transpose` and `isResultSigned`.  There is a third immediate property identified in issue #413 for whether an integer input vector is interpreted as a signed or an unsigned encoding when converting to the selected input interpretation.  While there is a proposal #399 to incorporate the `isResultSigned` into the OpCode instead of providing it as a parameter, this still leaves two boolean properties, while expanding to three times the number of opcodes for these two operations.  Due to the commonalities in these operations, I have no doubt that these three opcodes variants would be handled by common code that translates the opcode difference back into a property, as the current definition specifies.

My suggestion is to add a flags field instead of separately supplying each of these properties to the operations, and instead of expanding the OpCodes to separate float, unsigned int, and signed int results.

I think that the flags can be shared with similar flags that may be required for other operations in the future, such as for wave matrix.  For instance, if the default orientation of the matrix for this operation is equivalent to the "left" matrix in a wave matrix operation, the flag may be named or at least be equal to `LEFT_MATRIX_TRANSPOSE`, and will apply to any `M,K` dimension left matrix.

A `RESULT_UNSIGNED` may be used when the result is an unsigned result.  Side note: It seems better to assume a default of signed since it matches the default interpretation for the floating-point case as well.

For proposal #413, a flag `INPUT_VECTOR_UNSIGNED` could indicate an unsigned encoding in the input vector, instead of a new `i1` argument.

With these flags, the `i1` arguments could be removed, and it would not be necessary to expand the OpCodes for the signed/unsigned/float value of the result.

damyanp:
This is an attempt to resolve #399, #413 in the way proposed by #414.

* replace the "isResultSigned" parameter with a "flags" parameter
* defines flags for unsigned input and unsigned output vectors

Resolves #399
Resolves #413
Resolves #414


proposals/0029-cooperative-vector.md
```diff
@@ -142,7 +142,7 @@ declare <[NUMo] x [TYo] @dx.op.matvecmul.v[NUMo][TYo].v[NUMi][TYi](
     immarg i32,       ; matrix layout
     immarg i1,        ; matrix transpose
     i32,              ; matrix stride
-    immarg i1)        ; isResultSigned        <<< See #399
+    immarg i8)        ; flags
```
tex3d, line 145: I think i32 for a flags field would be more typical/appropriate, even if we don't imagine burning through those bits.  I just don't think there's any real advantage to using i8 for flags instead.

Additionally, it would help to identify the flag set in the comment here, for instance by the enum name or prefix in DxilConstants.h (`kMatVecMulFlag_*` at the moment), unless we have a better idea.

proposals/0029-cooperative-vector.md
```diff
@@ -239,6 +239,20 @@ This operation returns a vector of size `NUMo` and contains elements of type
 `TYo`. The result vector does not have an interpretation parameter, its type is
 the declared type.
 
+#### Operation Flags
+
+The operation flags are generated by OR'ing together values from these constants:
+
+```
+const uint8_t kMatVecMulFlag_UnsignedInputVector = 0x1;
+const uint8_t kMatVelMulFlag_UnsignedOutputVector = 0x2;
```
tex3d, line 248: Typo:
```suggestion
const uint8_t kMatVecMulFlag_UnsignedOutputVector = 0x2;
```

proposals/0029-cooperative-vector.md
```diff
@@ -142,7 +142,7 @@ declare <[NUMo] x [TYo] @dx.op.matvecmul.v[NUMo][TYo].v[NUMi][TYi](
     immarg i32,       ; matrix layout
     immarg i1,        ; matrix transpose
```
tex3d, line 143: I think the matrix transpose property should be specified through the operation flags as well.

proposals/0029-cooperative-vector.md
```diff
@@ -239,6 +239,20 @@ This operation returns a vector of size `NUMo` and contains elements of type
 `TYo`. The result vector does not have an interpretation parameter, its type is
 the declared type.
 
+#### Operation Flags
+
+The operation flags are generated by OR'ing together values from these constants:
+
+```
+const uint8_t kMatVecMulFlag_UnsignedInputVector = 0x1;
```
anupamachandra, line 247: Can we call this `KMatVecOpFlag*` as it covers both the Mul and MulAdd operations?

proposals/0029-cooperative-vector.md
```diff
@@ -239,6 +239,20 @@ This operation returns a vector of size `NUMo` and contains elements of type
...
+const uint8_t kMatVelMulFlag_UnsignedOutputVector = 0x2;
+```
+
+If the input or output vector is not marked as unsigned then it should be
+interpreted as signed.
+
+It is only valid to mark an input or output vector as unsigned for integral
```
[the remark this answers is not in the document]
anupamachandra: Agree with Tex's points above.

damyanp:
We've decided to go with the approach in #415 .

Issue #357, opened 2024-12-03
Document constant buffers in the language spec


hekota:
We need to add constant buffers to the HLSL language spec. 
That includes documenting the different ways to declare a constant buffer (`cbuffer` vs. `ConstantBuffer`), `cbuffer` and `packoffset` grammar, and rules on what exactly is allowed inside a `cbuffer` declaration.

DXC currently supports quite a few more or less unusual constructs such as:
- nested `cbuffer` declarations
- multiple `cbuffer` declarations with the same name
- constants defined outside of `cbuffer` context (to be placed in `$Globals` constant buffer and bound to first register `b0`)
- namespace declarations inside `cbuffer` context
- function definitions inside `cbuffer` context

The spec should clarify which of these are we going to support in Clang.

@tex3d put together an example that these various interesting cases here:
https://godbolt.org/z/E4cdx7xj8

Same shader with FXC:
https://shader-playground.timjones.io/39c0683735c0cd79b5970b5116765940

One difference between FXC and DXC in the nested `cbuffer` case. FXC binds outer `cbuffer` declaration before the nested one while DXC binds it after.

Related issues:
- llvm/llvm-project#117715
- llvm/llvm-project#124064

PR #419, opened 2025-03-04
Constant buffers language spec - initial draft

hekota:
Adds constant buffers to the language spec. Closes microsoft/hlsl-specs#357

image
image
image
image


specs/language/resources.tex
```diff
@@ -663,19 +664,237 @@
...
+\begin{itemize}
+\item \textit{scalar types}
+\item \textit{vector types}
+\item \textit{matrix types} with only one row in \textit{column\_major} storage layout
+\end{itemize}
+
+\p If they do not fit the remaining space of the row, they will placed at the
+start a new 16-byte aligned row.
+
+\p Vectors are aligned by the size of a single vector element type.
```
V-FEXrt, line 786: This doesn't account for the row wrapping right? Shouldn't it be something like:

`Vectors are aligned by the size of a single vector element type unless that alignment results in crossing the 16 byte row boundary, in which case it is aligned to the next row"

specs/language/resources.tex
```diff
@@ -663,19 +664,237 @@
...
+matrix) is aligned to a 16-byte row.
+
+\p Members of structured types used in constant buffers follow these same rules.
+Because structures are always 16-byte row-aligned, the offset layout within the
+structure is independent of the particular structure instance location.
+
+\Sub{Packoffset annotations}{Resources.cnbuf.po}
+
+\p Shader constants declared in \texttt{cbuffer} declaration group can have an
+optional \texttt{packoffset} annotation added before their \texttt{;} terminal.
```
V-FEXrt, line 806: Should we add an `If an annotation is added to one shader constant in a declaration group then it must be added to all shader constants in that group`?

specs/language/resources.tex
```diff
@@ -663,19 +664,237 @@
...
+rows of the constant buffer. Each row treated as a vector of four 32-bit
+elements.
+
+\p The \textit{packoffset-element} identifies specific location within the
+vector where \texttt{x} or \texttt{r} is the first element of the vector,
+\texttt{y} or \texttt{g} is the second one, \texttt{z} or \texttt{b} is the
+third one and \texttt{w} or \texttt{a} is the fourth, and last element of the
+vector.
+
+\p Shader constant that has a structure, array or matrix type must always be
```
V-FEXrt, line 839: I'm guessing the answer is "backwards compatibility" but is there a strong argument against something like 

```
Shader constant that has a structure, array or matrix type must always be
16-byte row-aligned. Therefore it may not specify a packoffset-element
```

My first thought on why not is that it could create "unbalanced" code like
```
  cbuffer MyConstants {
    MyStruct Foo : packoffset(c0);
    float Foo : packoffset(c1.x);
  };
```

but beyond that I'm not sure. Since its necessary to check x/y/z anyways it seems like the change would complicate implementations

In the end, I guess I'd just like to hear your thoughts on why it was picked to be this. Personally I'm torn because I prefer the language consistency of keeping it but also dislike the idea of language feature that does "nothing"

specs/language/resources.tex
```diff
@@ -663,19 +664,237 @@
...
+\p Default constant buffer declarations can have an optional
+\textit{resource-binding} annotation with a \textit{register-type} \texttt{c} or
+\texttt{C} added before their \texttt{;} terminal. This annotation does not
+define a virtual register binding but rather an index into an array of 16-byte
+rows that represents the default constant buffer, similar to the
+\texttt{packoffset} annotation.
+
+\p For example
+\begin{HLSL}
+  float4 CameraPos : register(c2);
```
V-FEXrt, line 873: Mostly asking for historical context. Why is it called a register in $Globasl but packoffset for the rest?

specs/language/resources.tex
hekota: I have no idea why it was done like that. Maybe @tex3d would know?

specs/language/resources.tex
hekota: My guess is there was just a single constant buffer `$Globals` before `cbuffer` blocks and `packoffset` were added, and `register` was an established annotation at that time. So it just got extended with a 'c' value and new meaning.
@tex3d - please correct me if I am wrong! :)

specs/language/resources.tex
```diff
@@ -663,19 +664,241 @@
...
+\p \texttt{ConstantBuffer<T>} can be defined at local scope to represent a local
+reference to a constant buffer that can be associated with a global buffer when
+assigned. The reference must be resolvable to a unique global buffer declaration
+before use.
+
+\p Constant buffers cannot be implicitly nor explicitly cast from one type to
+another. This means that local \texttt{ConstantBuffer} can only be assigned from
+\texttt{ConstantBuffer} with the same template type \texttt{T}.
+
+\Sub{Constant Buffer Layout}{Resources.cnbuf.lay}
```
llvm-beanz, line 768: We probably shouldn't describe the layout as part of the language. This is really a constraint of DXIL that SPIR-V doesn't have, and we're almost certainly going to want to change this in a future shader model.

The language you've written here is great though. We should take this language and capture it in the DXIL specification.

specs/language/resources.tex
```diff
@@ -663,19 +664,241 @@
...
+\p Variable declarations in the \texttt{cbuffer} declaration group are called \textit{shader constants}.
+
+\p Shader constants can be referenced from anywhere in the translation unit after they are declared by directly using the declaration name. This implies that all shader constants declared in a translation unit must have unique names, even though they might be declared in different \texttt{cbuffer} declaration groups.
+
+\p Variable declarations in the \texttt{cbuffer} declaration group cannot have \texttt{groupshared} or \texttt{static} variable modifiers.
+
+\p Any declaration in the \texttt{cbuffer} declaration group that is not a
+\textit{namespace-declaration} or a \textit{variable-declaration} is going to be
+processed as if it was declared outside of the group at the
+global scope.
```
llvm-beanz, line 701: I'm trying to think about the relationship we want between namespaces and cbuffers. We got some GAB feedback about namespaces that referenced https://github.com/microsoft/DirectXShaderCompiler/issues/4514, which is wanting to put cbuffers in a namespace, and we similarly have this issue you filed which seeks to support namespaces inside a cbuffer https://github.com/microsoft/DirectXShaderCompiler/issues/7322.

My instinct is that we should make cbuffers only contain variable declarations and disallow namespaces in cbuffers, and instead allow cbuffers in namespaces.

The benefit of that approach is that we don't ever have to restrict what we can put in a namespace because it is enclosed in a cbuffer.

Thoughts?

llvm-beanz (commented):
I have some thoughts about the

specs/language/resources.tex
llvm-beanz: Oh! I think I missed that those issues are worded differently but refer to the same issue.

DXC does allow namespaces in cbuffers (and they work), but not cbuffers in namespaces. I feel like this relationship is backwards.

hekota: DXC actually allows cbuffers in namespaces as long as it is not the only thing in the namespace, which is a bug we should fix: https://godbolt.org/z/GaWsPxxa7

I agree that we should make cbuffers only contain variable declarations. I'll update the spec.

specs/language/resources.tex
hekota: What about `packoffset` annotations? They are part of the language and define specific offset in the constant
buffer.

V-FEXrt (approved):

specs/language/resources.tex
```diff
@@ -663,19 +665,235 @@
 
 \Sec{Constant Buffers}{Resources.cnbuf}
 
+Constant buffer represent a resource that contains read-only constant data in a specific memory layout.
```
Icohedron, line 668:
```suggestion
Constant buffers represent resources that contain read-only constant data in a well-defined memory layout.
```

specs/language/resources.tex
```diff
@@ -663,19 +665,235 @@
...
+
+\Sub{Constant Buffer Declaration Block}{Resources.cnbuf.cb}
+
+\p Constant buffer can be declared using the \texttt{cbuffer} specifier.
```
Icohedron, line 672:
```suggestion
\p A constant buffer can be declared using the \texttt{cbuffer} specifier.
```

specs/language/resources.tex
```diff
@@ -663,19 +665,235 @@
...
+assigned. The reference must be resolvable to a unique global buffer declaration
+before use.
+
+\p Constant buffers cannot be implicitly nor explicitly cast from one type to
+another. This means that local \texttt{ConstantBuffer} can only be assigned from
+\texttt{ConstantBuffer} with the same template type \texttt{T}.
+
+\Sub{Constant Buffer Layout}{Resources.cnbuf.lay}
+
+\p Constant buffer is arranged like an array of 16-byte rows, or 4-component
```
Icohedron, line 765:
```suggestion
\p A constant buffer is arranged like an array of 16-byte rows, or 4-component
```

specs/language/resources.tex
```diff
@@ -663,19 +665,235 @@
...
+  };
+\end{HLSL}
+
+The \texttt{Pos} shader constant will be located at byte offset \texttt{24} in
+the constant buffer \texttt{MyConstants}.
+
+\Sub{Default Constant Buffer}{Resources.cnbuf.defcb}
+
+\p All variables declaration in the global scope are implicitly added to default
+constant buffer called \texttt{\$Globals}, unless they are marked
```
Icohedron, line 856:
```suggestion
\p All variable declarations in the global scope are implicitly added to a default
constant buffer called \texttt{\$Globals}, unless they are marked
```

Icohedron (approved):
Thanks, I now understand constant buffers, virtual registers, and virtual register spaces more now! 
I have a few nits, but overall I think it is well-written and easy to understand as it is.

PR #420, opened 2025-03-06
Change ReduceSumAccumulate to InterlockedAdd overload for vectors, some more details on FP8, matrix strides etc.

anupamachandra:
Covers the following feedback:

**OuterProductAccumulate/ReduceSumAccumulate**

*Resolution:*

* Agreement to make ReduceSumAccumulate & OuterProductAccumulate optional parts of the CoopVectors spec. `DONE`

* Agreement to replace ReduceSumAccumulate with regular atomics & cover in a separate spec, which will be helpful for other use cases. `Replaced by InterlockedAdd overload, per latest consensus`. To be limited to just buffer memory, at least for initial version `DONE`
* Minimum requirement for training function would still be FP16 at the API surface, but internal implementation will leverage FP32 math. `Implementation detail not covered in spec`. Rationale: Some current gen hardware only supports FP32.

* Open Questions: Course of action for odd length vectors `Open as issue if concerns remain`

**FP8 Emulation Path**

*Resolution:* 

* Agreement to make explicit in the spec that FP8 matrices must be in optimal layout, which gives the opportunity to store data in FP16. `DONE`
* Agreement that the driver and compiler just need to have the same logic for which stride to use, which will be implicit for optimal layouts. 
* Agreement: spec will only require FP16 internal precision `Internal precision is an implementation detail not covered in spec`
* Agreement that the conversion to FP8 optimal layout will be done precisely (exact expected value) `Implementation detail not covered in spec`

**Stride**

*Resolution:*
* Agreement to add an alignment requirement for stride (multiple of 16) to the spec. `NONE`

### DIFF
--- a/proposals/0029-cooperative-vector.md
+++ b/proposals/0029-cooperative-vector.md
@@ -308,8 +308,9 @@ declare void @dx.op.atomicBinOp.v[NUM][TY](
 
 #### Overview
 
-Adds new vector overloads to `@dx.op.atomicBinOp` to support accumulation of a vector component-wise atomically (with device scope) to the corresponding elements of an array in memory. See note in
-[Atomic Operations].
+Adds new vector overloads to `@dx.op.atomicBinOp` to support accumulation of a
+vector component-wise atomically (with device scope) to the corresponding
+elements of an array in memory. See note in[Atomic Operations].
 
 #### Arguments
 
@@ -318,7 +319,9 @@ type `TY`.
 
 The input vector is accumulated to the writeable raw-buffer resource specified
 by **resource handle** and offset **coordinate c0**.  The base address
-and **coordinate c0** must be 64 byte aligned.**coordinate c1** and **coordinate c2** are 0 for raw-buffers. The function doesn't return any value. 
+and **coordinate c0** must be 64 byte aligned.**coordinate c1**
+and **coordinate c2** are 0 for raw-buffers. The function doesn't return any
+value. 
 
 The **binary operation code** for *ADD* operation is `0`.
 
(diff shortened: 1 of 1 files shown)


proposals/0029-cooperative-vector.md
```diff
@@ -289,32 +289,41 @@ guaranteed to be supported on all implementations can be found in
...
 ```
 
 #### Overview
 
-Accumulates the components of a vector component-wise atomically (with device
-scope) to the corresponding elements of an array in memory. See note in
-[Atomic Operations].
+Adds new vector overloads to `@dx.op.atomicBinOp` to support accumulation of a
+vector component-wise atomically (with device scope) to the corresponding
+elements of an array in memory. See note in[Atomic Operations].
```
llvm-beanz, line 313: I don't like that we're reusing atomicBinOp for an operation that doesn't have a full scope of atomicity. I'd rather we just scalarize to a bunch of separate atomicBinOp calls for each element rather than declare an overload that takes a vector and isn't atomic across the full input.

anupamachandra: That's a fair point. We would like implementations to be able to use any vector atomics that they can support though and choose the units of atomicity. Should we keep the InterlockedAdd override, but introduce a new dxil op?

damyanp: Another opcode would be an option, but we probably need to have a conversation about the name since it isn't atomic.

I suggest that we remove this part of the change from this PR so that we can get the rest in while we discuss what to do with ReduceSumAccumulate?

damyanp:
@anupamachandra - can we close this PR?

anupamachandra:
Yep.

PR #421, opened 2025-03-07
Long vector test plan

alsepkow:
.


proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,199 @@
...
+I don't think there are any particularily interesting vector sizes to test. So I propose testing sizes of 5, 25, 100, 500, 
+
+# Phases
+Do the test work in two simple phases.
+
+1. Implement and validate (locally against WARP) the 3 test categories.
+2. HLK related work:
+    - Add a SM 6.9 HLK requirement. Includes updating the HLK requirements doc.
+    - Update mm_annotate_shader_op_arith_table.py to annotate the new test cases with HLK GUIDS and requirements
+    - Add new tests to HLK playlist
```
damyanp, line 29: Do we understand all the steps required to get these tests ingested into whatever it is that the IHVs receive?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,199 @@
...
+2. LLVM Native operation tests
+    - TBD: Where do we test these? Tex mentioned some existing tests but it sounds like we don't want to use those. And they aren't HLK tests anyways.
+    - TODO: Discuss how we want to test ops that don't boil down to a single DXIL op. Still seems like something we should test in the HLK?
+
+3. Standard loading and storing of long vectors
+    - TODO: Do we have tests that do this today?
+
+# Vector Sizes to test
+
+I don't think there are any particularily interesting vector sizes to test. So I propose testing sizes of 5, 25, 100, 500,
```
damyanp, line 20: Presumably we already have tests covering <5 cases?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,199 @@
...
+1. DXIL Intrinsic tests: Go through the operators that support vectors and make a list of the DXIL ops used.
+    - The table to reference for this is in HLOperationLower.cpp.
+    - Many, if not all, of these probably have existing test cases that just need to be modified to use vectors
+        - I think that the 'implementation' will largely be editing ShaderOpArithTable.xml, which is used to generate TAEF 'data driven' tests. Minor shader code updates and added input/output of vectors
+    - Some tests may need to be duplicated for the vector variant of the test.
+    - The tests cases in this category are any intrinsics in the below table that have a DXIL::OpCode::* in the DXIL OPCode colum
+
+2. LLVM Native operation tests
+    - TBD: Where do we test these? Tex mentioned some existing tests but it sounds like we don't want to use those. And they aren't HLK tests anyways.
+    - TODO: Discuss how we want to test ops that don't boil down to a single DXIL op. Still seems like something we should test in the HLK?
```
damyanp, line 13: > Still seems like something we should test in the HLK?

Agreed.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,199 @@
...
+# Notes
+    - The 'ExecTests' (ExecutionTest.cpp) are used to generate the HLK tests (DxilConf tests)
+
+# Open Questions
+    1. How does the HLK work for IHVs running DXIL tests given that we're undocked. Or does that not matter because the DXIL HLK tests provide raw DXIL? Just trying to understand the HLK testing lifecycle here vs what I was used to in the OS.
+    2. Do we need a new HLK requirement? If so, we probably need to get that in soon. That could also require HLK gatherer updates.
+        - For tests where we only add new test cases we do not need to update HLK test GUIDs.
+        - If we want to add a new HLK requirement, then we will probably want to create new tests which consume that requirement. Individual test cases can not have individual HLK requirements.
+        - The existing HLK reqs we have look pretty dated. "Device.Graphics.WDDM27.AdapterRender.D3D12.DXILCore.ShaderModel65.CoreRequirement". Add a requirement for SM 6.9? We're requiring Long Vectort support for SM 6.9
+    3. Do the work in two 'phases'? Add the tests. Then do the HLK side.
```
damyanp, line 53: I think we should build up a plan that involves multiple deliveries to the IHVs.  If we can identify a set of high-priority tests (eg those needed by a particular demo) we could try and get those out first to support IHVs building drivers for the preview release.

damyanp (commented):
Looks good so far.

We try and format these specs to 80-columns to make them easier to review in source form.

Are the tables of operations a list of all the things that'll need tests added or is that still being determined?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: I'm familiar with the HLK, but I do need to check in on the latest processes with CD. I suspect a few details have changed.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: Yes, assuming. I can make a point to verify that as I add new tests/test cases. Everything I had spot checked was covered.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: From some further discussion with @pow2clk on Friday we also want to test some LLVM ops. I'll add a point to this spec to capture that. But I think it would be good to also include these in the HLK.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: I was thinking about this and if IHVs are OK with it I think the easiest way is to just share stand-alone test binaries/collateral. Can you help me sync up with the right people to figure out what works for IHVs? What have we done in the past?

alsepkow:
> Looks good so far.
> 
> We try and format these specs to 80-columns to make them easier to review in source form.
> 
> Are the tables of operations a list of all the things that'll need tests added or is that still being determined?

Will update the formatting. The intrinsics in the table are all of the explicitly listed intrinsics in the [hlsl long vector](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0026-hlsl-long-vector-type.md#allowed-elementwise-vector-intrinsics). I think some operators are missing from this table. I'll look to get everything added so we have an exhaustive list.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
damyanp: Yup, we can sort that out, but I'm more interested in what and when we'll deliver these rather than the mechanism used to do it.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,243 @@
...
+     operators table below.
+     - We have added some test cases for these with the long vectors work, but we
+     will want to make sure we add HLK coverage for all of these.
+
+3. Standard loading and storing of long vectors
+     - Ensure we have some basic tests doing standard loading/storing of long vectors.
+
+# Vector Sizes to test
+
+I don't think there are any particularly interesting vector sizes to test. So I
```
damyanp, line 28: > I don't think there are any particularly interesting vector sizes to test.

I think we might be able to come up with some.

For example, I think vector<half, 7>, vector<half, 8> and vector<half, 9> may be interesting because they cover the 128-bit boundary that has been in place most of the time before and could be interesting for load/stores as well as how they work in groupshared.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,243 @@
+# Test Plan
+
+All tests are to be included in the HLK test binary which ships with the OS.
+This test binary is only built in the OS repo and based off of the 'ExecTests'
+'ExecTests' source code in the DXC repo. There is a script in the WinTools repo
+which generates and annotates the HLK tests.
+
+There are three test categories we are concerned with:
```
damyanp, line 8: Do we need special tests for groupshared?

Are there any other types of storage that might be interesting?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,243 @@
...
+     - Add a SM 6.9 HLK requirement. Includes updating the HLK requirements doc.
+     - Update mm_annotate_shader_op_arith_table.py to annotate the new test cases
+        with HLK GUIDS and requirements
+     - Add new tests to HLK playlist
+
+# Shipping
+1. Tests will be shared privately with IHV's along with the latest DXC and
+latest Agility SDK for tesing and validation.
+
+2. The tests will ship with the HLK at a TBD date in a later OS release.
```
damyanp, line 47: > The tests will ship with the HLK at a TBD date in a later OS release.

Maybe mention that this is our normal way of operating for new shader model features?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,243 @@
...
+completed.
+ - All new test cases pass when run locally against a WARP device
+ - All new test cases are confirmed to be lit up and available in the HLK when the
+    target device reports support
+ - All new tests/test cases are added to the HLK playlist
+
+# Requirements
+     - Greg's long vector changes. Check-in ETA of 3/7/25.
+     - WARP long vector support (Jesse). Needs Greg's work, or a private WIP branch
+        of Greg's work. ETA of ~1 week to implement.
```
damyanp, line 67: This is out of date now.  You should be able to work against microsoft/DirectXShaderCompiler#7188.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,243 @@
...
+
+# Notes
+     - The 'ExecTests' (ExecutionTest.cpp) are used to generate the HLK tests (DxilConf
+        tests)
+     - Private test binaries/collateral will be shared with IHVs for validation
+        purposes. This will enable IHVs to verify long vector functionality without
+        waiting for an OS/HLK release.
```
damyanp, line 74: These sections don't format how I think you think they should.

```suggestion
- The 'ExecTests' (ExecutionTest.cpp) are used to generate the HLK tests (DxilConf 
   tests)
- Private test binaries/collateral will be shared with IHVs for validation purposes. 
   This will enable IHVs to verify long vector functionality without waiting for an 
   OS/HLK release.
```

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,243 @@
...
+
+# Open Questions
+     - We speculate that the combiniation of test cases for HLSL operators and
+     HLSL intrinsics which map to DXIL ops should get us all, or most of the
+     coverage required for the 'emulated' intrinsics. But we should audit that.
```
damyanp, line 79: Same formatting issue as mentioned above.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,243 @@
...
+
+# HLSL Operators
+These operatores should generate LLVM native ops which use vectors.
```
damyanp, line 82: typo

```suggestion
These operators should generate LLVM native ops which use vectors.
```

damyanp, line 82: "LLVM native ops" - we should check correct terminology here.  I think it might be "LLVM instructions" or "LLVM IR"?

Maybe @bogner can advise.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,243 @@
...
+
+| Operator Name | Operator | Notes |
+|-----------|--------------|----------|
+Additive and Multiplicative Operators | +, -, *, /, % |
+Array Operator | [i] |
+Assignment Operators | =, +=, -=, *=, /=, %= |
+Binary Casts | C rules for float and int, C rules or HLSL intrinsics for bool |
```
damyanp, line 89: "Binary casts" - does this mean reinterpret casts?  Like asint() or asfloat()?

If not, wondering how this differs from Cast Operator

damyanp: (this may be my ignorance showing, btw)

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,243 @@
...
+
+# Notes
+     - The 'ExecTests' (ExecutionTest.cpp) are used to generate the HLK tests (DxilConf
+        tests)
+     - Private test binaries/collateral will be shared with IHVs for validation
+        purposes. This will enable IHVs to verify long vector functionality without
+        waiting for an OS/HLK release.
+
+# Open Questions
+     - We speculate that the combiniation of test cases for HLSL operators and
```
damyanp, line 77: Typo 
```suggestion
     - We speculate that the combination of test cases for HLSL operators and
```

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,243 @@
...
+Bitwise Operators | ~, <<, >>, &, \|, ^, <<=, >>=, &=, \|=, ^= | Only valid on int and uint vectors
+Boolean Math Operators | & &, ||, ?: |
+Cast Operator | (type) |
+Comma Operator | , |
+Comparison Operators| <, >, ==, !=, <=, >= |
+Prefix or Postfix Operators| ++, -- |
+Structure Operator | . |
+Unary Operators | !, -, + |
+
+# Mappings of HLSL intrinsics to DXIL opcodes or LLVM native operations
```
damyanp, line 99: Wondering where you sourced this list of operators from?  Not all lists out there are exhaustive!

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: Actually, given that the DXC repo is public we should be able to just share build/run instructions for the tests once checked in?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: asfloat, asint, and asuint. I think it would be good for me to be explicit about that. Will update.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: These were sourced from [HLSL operators](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators 
)

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: Probably, I was thinking about this and I need to learn/explore more. I suspect I'm actually missing details on test case granularity. Does testing across data types matter? Do we care to verify the output of operations for correctness?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
damyanp: D'oh, I said "operators" and I meant "opcodes".  So specifically thinking about all the various intrinsics / DXIL operations that you want to exercise.  This is where some of Farzon's stuff, like https://github.com/farzonl/HLSLDocGenerator, might come in handy.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
damyanp: We should probably be testing for correctness, yes.  Sizes, especially around interesting alignment cases, are probably interesting.  It might be interesting to build cases that could conceivably generate overruns and check for them.  For example, most GPUs will operate on at least 32-bits at once, so if you have 16-bit values then what happens if you have an odd number of elements - could it accidentally overwrite the next variable if it is assuming alignment?

damyanp: And so the size of a type, and the element count, gets interesting.  I expect you could come up with some interesting boundary cases and reuse that across all the types without having to think hard about each combination individually.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: Thanks for pointing that out. I asked co-pilot to format them for me. Set up properly now (I think).

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: The HLSL intrinsics were sourced from [0026-hlsl-long-vector-type.md](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators#binary-casts )

damyanp: :) I think that doc's worth treating as authoritative - but not that link!  Try clicking it :)

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,254 @@
...
+   purposes. This will enable IHVs to verify long vector functionality without
+   waiting for an OS/HLK release.
+
+# Open Questions
+- We speculate that the combination of test cases for HLSL operators and HLSL
+intrinsics which map to DXIL ops should get us all, or most of the coverage
+required for the 'emulated' intrinsics. But we should audit that.
+
+# HLSL Operators
+These operators should generate LLVM native ops which use vectors.
```
damyanp, line 92: It might be worth digging a bit deeper here into which actual LLVM operations you want to be testing.

For example, I don't think that the `,` operator actually corresponds to anything the driver would see?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,254 @@
...
+
+Operator table from [Microsoft HLSL Operators](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators#binary-casts)
```
damyanp, line 94:
```suggestion
Operator table from [Microsoft HLSL Operators](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators)
```

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,254 @@
...
+latest Agility SDK for tesing and validation. IHVs will also be able to build
+and run the tests from the public DXC repo themselves.
+
+2. The tests will ship with the HLK at a TBD date in a later OS release.
+
+# New HLK Tests
+mm_annotate_shader_op_arith_table.py (WinTools repo) will need to be updated to
+recognize any new tests (not test cases) added. And additional GUIDs added for
+new test cases. mm_annotate_shader_op_arith_table.py is called by
+mm-hlk-update.py when converting from 'ExecTests' to the HLK 'DxilConf' tests.
```
pow2clk, line 62: Now I'm wondering if I'm mistaken. Maybe that's what this script refers to them as?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,254 @@
...
+| tanh      | DXIL::Opcode::Htan |  |
+
+## Math
+
+| Intrinsic | DXIL OPCode | Notes |
+|-----------|--------------|----------|
+| mod       | Emulated |              |
+| abs       | DXIL::OpCode::Imax |    |
+| ceil      | DXIL::OpCode::Round_pi ||
+| clamp     | DXIL::OpCode::UMax, UMin \ DXIL::OpCode::FMax/Fmin \ DXIL::OpCode::IMax/Imin ||
```
pow2clk, line 136: This is what has me wondering if "emulated" means what I think it means. This is "emulated" as much as atan2 is. It may be useful to list the DXIL Ops it uses though.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,254 @@
...
+| sinh      | DXIL::OpCode::Hsin |  |
+| tan       | DXIL::OpCode::Tan |   |
+| tanh      | DXIL::Opcode::Htan |  |
+
+## Math
+
+| Intrinsic | DXIL OPCode | Notes |
+|-----------|--------------|----------|
+| mod       | Emulated |              |
+| abs       | DXIL::OpCode::Imax |    |
```
pow2clk, line 134: This can use OpCode::Fabs too for floating types.

pow2clk (commented):
Looks good! Just trying to add detail and answer questions where I can.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: Thanks, I've updated to reflect these test cases.

Can either of you help point me in a direction to better understand how this could be interesting when testing load/stores and how they work in groupshared?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: I had naively checked what worked on existing vectors in godbolt. Updating the table based on renewed understanding and these comments.

alsepkow: @pow2clk I think it would be helpful for me to discuss casts a little more offline. And I can update notes here based on that.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,254 @@
...
+    their mapped DXIL OpCodes. We just need coverage for each DXIL OpCode. If
+    the DXIL OpCode column lists 'emulated' then this means that there are
+    multiple LLVM/DXIL Ops used to compute the intrinsic. We will need to do an
+    audit of them to confirm coverage of all vector ops.
+
+2. Implement LLVM Native operation tests:
+     - These are the test cases for the 'basic' HLSL operators listed in the
+     HLSL operators table below.
+     - We have added some test cases for these with the long vectors work, but
+     we will want to make sure we add HLK coverage for all of these.
```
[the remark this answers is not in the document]
alsepkow: I'm probably confused here. I was assuming most, if not all, operators would map to some LLVM or DXIL operator? If that's true then wouldn't it make sense to have test cases in the HLK that use those operators with vector arguments?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: I just checked and I see them referred to as 'ExecTests' in the pipelines and an mm build script. Probably better to use 'ExecutionTests' as you suggested.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,254 @@
...
+|-----------|--------------|-----------|
+| WaveActiveAllEqual | DXIL::OpCode::WaveActiveAllEqual | |
+| WaveMatch          | DXIL::OpCode::WaveMatch | |
+| asdouble           | DXIL::OpCode::MakeDouble | |
+| asfloat            | Emulated | |
+| asfloat16          | Emulated | |
+| asint              | Emulated | |
+| asint16            | Emulated | |
+| asuint             | DXIL::OpCode::SplitDouble | |
+| asuint16           | Emulated | |
```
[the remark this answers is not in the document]
alsepkow: 'emulated' was essentially a placeholder. When I first started going through the list of HLSL intrinsics I was taking some notes on the LLVM/DXIL ops that I could see it distilled down to. I got stuck on several and gave up and went with 'emulated' for now. 

I'll go back through everything that is marked as 'emulated' so we have an inventory of the operators used.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,254 @@
...
+| sin       | DXIL::OpCode::Sin |   |
+| sinh      | DXIL::OpCode::Hsin |  |
+| tan       | DXIL::OpCode::Tan |   |
+| tanh      | DXIL::Opcode::Htan |  |
+
+## Math
+
+| Intrinsic | DXIL OPCode | Notes |
+|-----------|--------------|----------|
+| mod       | Emulated |              |
```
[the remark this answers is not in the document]
alsepkow: Not sure where this one came from. Although, for modf I saw that it mapped to multiple instructions and ended with a CreateFMul call. Hence the 'emulated'. Removing mod from the table and will resolve fmod/modf ops on next pass through 'emulated' ops.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,273 @@
...
+ExecutionTests source code in the DXC repo. There is a script in the WinTools
+repo which generates and annotates the HLK tests.
+
+There are three test categories we are concerned with:
+
+1. Implement DXIL Intrinsic tests:
+    - At the bottom of this document there is a table of all HLSL intrinsics and
+    their mapped DXIL OpCodes. We just need coverage for each DXIL OpCode. If
+    the DXIL OpCode column lists 'emulated' then this means that there are
+    multiple LLVM/DXIL Ops used to compute the intrinsic. We will need to do an
```
damyanp, line 14: > If the DXIL OpCode column lists 'emulated' then this means that there are multiple LLVM/DXIL Ops used to compute the intrinsic.

I think you've changed how this appears in the document now?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,273 @@
...
+2. Implement LLVM Native operation tests:
+     - These are the test cases for the 'basic' HLSL operators listed in the
+     HLSL operators table below.
+     - We have added some test cases for these with the long vectors work, but
+     we will want to make sure we add HLK coverage for all of these.
+
+3. Standard loading and storing of long vectors
+     - Ensure we have some basic tests doing standard loading/storing of long
+     vectors across buffer types. __TODO: Add details about which types of
+     buffers to test and why.__
```
damyanp, line 26: > TODO: Add details about which types of buffers to test and why

This comment here to remind us that there's a TODO.  If we want to complete the PR without doing the TODO then we should get a follow-up issue filed.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,273 @@
...
+Cast Operator | (type) | No direct operator, difference in GetElementPointer 
+||| or load type
+Comparison Operators| <, >, ==, !=, <=, >= |
+Prefix or Postfix Operators| ++, -- |
+Unary Operators | !, -, + |
+
+# Mappings of HLSL intrinsics to DXIL OPcodes or LLVM Operators
+
+Note: [ ] - Brackets are used in the below tables to signify that the operator
+is only used in specific paths. Operators sharing the same brackers are in the
```
damyanp, line 123:
```suggestion
is only used in specific paths. Operators sharing the same brackets are in the
```

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,273 @@
...
+
+## Trigonometry
+
+| Intrinsic | DXIL OPCode | LLVM Operator | Notes |
+|-----------|--------------|----------|-----------|
+| acos      | Acos | | |
+| asin      | Asin | | |
+| atan      | Atan | | |
+| atan2     | Atan | CreateFDiv, CreateFAdd, CreateFSub, CreateFCmpOLT,
+||| CreateFCpmOEQ, CreateFCmpOGE, CreateFCmpOLT, CreateAnd, CreateSelect | |
```
damyanp, line 134: I'm guessing that these "CreateXXX" function names came from the C++ code.  The LLVM instructions are things like `fdiv`, `fadd`, `fcmp`, `and`, `select` etc.

I also note that if I try `atan2` in godbolt it doesn't actually generate `fsub`. https://godbolt.org/z/7xqz6448P

damyanp (commented):
Looking good - there's still some open issues.

I'm also keen to see a prioritization of which tests will be implemented first.

These operations are good candidates for high-priority tests I think:

* Initializing a vector with another.
* Multiply all components of a vector with a scalar value
* Add all components of a vector with a scalar value
* Clamp all components of a vector to the range [c, t]
* Component wise minimum between 2 vectors
* Component wise maximum between 2 vectors
* Component wise multiply between 2 vectors
* Component wise add between 2 vectors
* Subscript access, vec[i] = c

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,273 @@
...
+     we will want to make sure we add HLK coverage for all of these.
+
+3. Standard loading and storing of long vectors
+     - Ensure we have some basic tests doing standard loading/storing of long
+     vectors across buffer types. __TODO: Add details about which types of
+     buffers to test and why.__
+
+# Vector Sizes to test
+
+General sizes to test are in the range [5, 1024]. Sizes < 5 are assumed to
```
anupamachandra, line 30: The new vector load/store intrinisics produce vectorized versions for small vectors (<5) right? So those intrinsic tests will cover the <5 cases as well?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
```diff
@@ -0,0 +1,273 @@
...
+- 128 bit boundaries : Memory access for Shader Model 5.0 and earlier operate on
+128-bit slots aligned on 128-bit boundaries. An example is vector<half, 7>,
+vector<half, 8> and vector<half, 9>. 112 bits, 128 bits, and 144 bits
+respectively. This boundary will tested for with 32-bit and 64-bit sized values
+as well.
+- vector<TYPE, 5> : Testing one above previous vector limit.
+- vector<TYPE, 16> : This size of 'vector' previously only appeared as matrices.
+- vector<TYPE, 17> : Larger than any vector previously possible.
+- vector<TYPE, 1024> : The new max size of a vector. Test for float, half,
+  double, and int64.
```
anupamachandra, line 45: " Test for float, half, double, and int64.'" this is for all the above sizes right? What about packed types?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: Yes, they're the C++ helper functions. I was intending to distill down further on a subsequent iteration. Wanted to sanity check that they do what I would expect (always return the llvm instruction in the name).

I'll have to peel a little more in TranslateAtan2, for this 'inventory' I went through the various Translate* functions in HLSLOperatorLower and noted the possible operations and instructions. For [TranslateAtan2 it looks like FSub](https://github.com/microsoft/DirectXShaderCompiler/blob/9e8a698deed37116e5e55cebd0d725c5c2be5e4c/lib/HLSL/HLOperationLower.cpp#L1849) is always used, so it must be optimized or simplified out along the way?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: I'd prefer to fill in the details before completing this PR.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: Asked for some help. The FSub is optimized out to normalize to FAdd. If you pass '-O0' instead you'll see that the FSub is there.

But this makes me wonder about the right testing approach to ensure coverage. Given that the goal is to test that DXIL Ops and LLVM Instructions are functional with vectors, maybe we want to disable optimizations for the test cases?

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: Yes, sizes less than 5 are covered by existing test cases. I'll update to clarify that point.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: Educated myself. Its LLVM instructions.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: Yes, for all above sizes. I'll clarify that. And have now included bool and packed types.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: Got some pointers offline for implementation details.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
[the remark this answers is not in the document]
alsepkow: We do want HLK coverage for anything that can end up at the graphics driver with vectorized arguments. This will include some LLVM instructions. But we expect to get coverage for the majority of those for 'free' just by having to write test code.

proposals/infra/INF-0005-Long-Vector-TestPlan.md
alsepkow: Let's discuss this further offline. I'll make a note to ask you.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,339 @@
...
+|-----------|--------------|----------|-----------|
+| WaveActiveAllEqual | WaveActiveAllEqual | |
+| WaveMatch          | WaveMatch | |
+
+## Type Casting Operations
+
+| Intrinsic | DXIL OPCode | LLVM Instruction | Notes |
+|-----------|--------------|----------|-----------|
+| WaveActiveAllEqual | WaveActiveAllEqual | |
+| WaveMatch          | WaveMatch | |
```
llvm-beanz, line 332: I think these aren't supposed to be here.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,339 @@
...
+| WaveMultiPrefixBitOr  | WaveMultiPrefixOp | |
+| WaveMultiPrefixBitXor | WaveMultiPrefixOp | |
+| WaveMultiPrefixProduct| WaveMultiPrefixOp | |
+| WaveMultiPrefixSum    | WaveMultiPrefixOp | |
+| WavePrefixSum         | WavePrefixOp | |
+| WavePrefixProduct     | WavePrefixOp | |
+| WaveReadLaneAt        | WaveReadLaneAt | |
+| WaveReadLaneFirst     | WaveReadLaneFirst | |
+
+## Wave Reductions
```
llvm-beanz, line 320: It's weird to say that the two operations below are reductions but the ones above aren't, when many of the operations above are reductions.

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -274,7 +274,8 @@ Verify that long vectors produce validation errors in:
 Correct behavior for all of the intrinsics listed in [allowed elementwise vector intrinsics](#allowed-elementwise-vector-intrinsics)
  will be verified with execution tests that perform the operations on long vectors and confirm correct results
  for the given test values.
-Where possible, these tests will be variations on existing tests for these intrinsics.
+Where possible, these tests will be variations on existing tests for these
+intrinsics. [Long-Vector-ExecutionTest-Plan](.\infra\INF-0005-Long-Vector-ExecutionTest-Plan.md)
```
pow2clk: This link doesn't work. I don't think you need the dot and the slashes should be forward.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+
+1. Implement DXIL OpCode tests:
+    * At the bottom of this document there is a table of all HLSL intrinsics and
+    their mapped DXIL OpCodes as well as LLVM Instructions. For this first
+    category we are concerned with getting coverage of each DXIL OpCode. In some
+    cases there are multiple DXIL OpCodes listed. '[]' brackets are used to
+    signify that the operator is only used in specific paths. Operators sharing
+    the same brackets are in the same logic path. If an operator is not in
+    brackets then it is used in all cases. Additonal opcodes are logic based,
+    for example there may be an int and float specific opcode.
```
pow2clk: I'm having a bit of trouble following this.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+
+2. Implement LLVM Instruction tests:
+    * These are the test cases for the LLVM Instructions listed in the in table
+     at the bottom of this document.
+    * Because we will use HLSL intrinsics to get coverage for the DXIL OpCode
+       tests we speculate that we will get most of the coverage needed for the
+       LLVM Instruction tests. After implementing the DXIL OpCode tests we
+       should be able to do a coverage audit and ammend test cases, or write
+       simple additional ones, as needed.
+    * Just as in '1. Implement DXIL Intrinsic Tests' some cases have multiple
```
pow2clk:
```suggestion
    * Just as in '1. Implement DXIL OpCode Tests' some cases have multiple
```

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+    * These could be covered in test categories 1 and 2. But I propose we break
+      out individual cases to ensure that we have more granular coverage.
+    * Ensure we have some basic tests doing standard loading/storing of long
+     vectors across [Buffer types to test](#buffer-types-to-test) and [Vector
+     element data types to test](#vector-element-data-types-to-test).
+
+5. 'Creative' test cases:
+  * Sizes around alignments and boundaries (details in [Vector Sizes to
+    Test])[#vector-sizes-to-test].
+  * Odd number of elements in vector
```
pow2clk: Does this mean not even or peculiar? I think both are interesting. Peculiar dimensions could be some that cut 4 element blocks at different points. Given the pervasiveness of the 4 element chunk, that might exhibit some issues.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+  * Anything else that comes up? TBD here.
+
+# Buffer types to test
+* Raw Buffers (ByteAddressBuffer)
```
pow2clk: It's a pet peeve, but wherever possible, to avoid ambiguity I would prefer being more specific and referring to ByteAddressBuffers as "Byte Address Buffers" even though it's a mouthful.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+* Structured Buffers (StructuredBuffer)
+* Typed Buffers (Buffer\<T>)
```
pow2clk: Structured buffers require templated types just as typed buffers do. Typed Buffers don't support long/native vectors.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+* vector<TYPE, 17> : Larger than any vector previously possible.
+* vector<TYPE, 1024> : The new max size of a vector. 
+* These sizes will be tested across [Vector element data types to test](#vector-element-data-types-to-test)
+
+Some noteable alignment cases:
+* Most GPUs operate on at least 32-bits at once, so waht happens if you use
+  16-bit values and an odd number of elements. Could accesssing the last element
+  expose issues where we could overwrite the next variable is it is assuming
+  alignment?
+* Additional interesting cases TBD.
```
pow2clk: These are some good "odd" sizes! Might even link to this for the mention of "odd" or whatever might replace it above.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+# High level test design
+1. The test will leverage the existing XML infrastructure currently used by the
+   existing execution tests. There are two XML files. This general design
+   pattern exists today in the execution tests.
+   First XML: Used to define shader source code and metadata about that shader
+   code. This XML file is parsed using a private class. This private class helps
+   faciliate creation of D3D resources and execution of the shader.
+   Second XML: Describes metadata about the specific test cases. Used by the
+   TAEF infrastructe for [TAEF Data Driven
+   Testing](https://learn.microsoft.com/en-us/windows-hardware/drivers/taef/data-driven-testing)
```
pow2clk, line 136: Might help to have some bullet points or even numbers for first and second XML

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+
+# Shipping
+Note that because DXC and the Agility SDK are both undocked from Windows it is
+our normal operating behavior for the HLK tests to become available with a later
+TBD OS release. The good news is that this doesn't prevent the tests from being
+available much earlier in the DXC repo. It just means that they are simply TAEF
+tests in the DXC repo. An HLK test includes an extra level of infrastructure for
+test gating, selection, and result submission for WHQL signing of drivers.
+
+1. Tests will be shared privately with IHV's along with the latest DXC and
```
pow2clk:
```suggestion
1. Tests will be shared privately with IHVs along with the latest DXC and
```

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+
+# Test Validation Requirements
+The following statements must be true and validated for this work to be
+considered completed.
+* All new test cases pass when run locally against a WARP device
+* All new test cases must verify applicable outputs for correctness.
+* All new test cases are confirmed to be present in HLK Studio and selectable to
+* be run when a target device satisfies the HLK ShaderModel 6.9 requirement.
+* All new tests/test cases are added to the official WHQL HLK playlist for the
+* OS release that the HLK tests will ship with.
```
pow2clk: I don't think these were meant to have their own bullet point. I like to indent line continuations like this. The MD parser will put them on the same line anyway.
```suggestion
  be run when a target device satisfies the HLK ShaderModel 6.9 requirement.
* All new tests/test cases are added to the official WHQL HLK playlist for the
  OS release that the HLK tests will ship with.
```

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
+# Test Plan
```
pow2clk: It's pretty nitpicky, but there is a md-lint addon for vscode that insists on one `#` title and that all the rest be `##` or more. In that mode, this should be the title of the entire file. That at least is still probably a good idea even if you don't change the other `#`s.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+| WaveMultiPrefixSum    | WaveMultiPrefixOp | |
+| WavePrefixSum         | WavePrefixOp | |
+| WavePrefixProduct     | WavePrefixOp | |
+| WaveReadLaneAt        | WaveReadLaneAt | |
+| WaveReadLaneFirst     | WaveReadLaneFirst | |
+| WaveActiveAllEqual | WaveActiveAllEqual | |
+| WaveMatch          | WaveMatch | |
+
+## Type Casting Operations
+Note:
```
pow2clk: Missing note or should this be removed?

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+
+### HLSL-Operators
+# HLSL Operators
+These operators generate LLVM instructions which use vectors.
+
+Operator table from [Microsoft HLSL Operators](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators)
+| Operator Name | Operator | Notes |
+|-----------|--------------|----------|
+| Addition | + | blah |
+| Subtraction | - | blah |
```
pow2clk:
```suggestion
| Addition | + | |
| Subtraction | - | |
```

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+| Multiplication | * |
+Additive and Multiplicative Operators | +, -, *, /, % |
+Array Operator | [i] | llvm:ExtractElementInst
+Assignment Operators | =, +=, -=, *=, /=, %= |
+Bitwise Operators | ~, <<, >>, &, \|, ^, <<=, >>=, &=, \|=, ^= | Only valid on
+||| int and uint vectors
+Boolean Math Operators | & &, ||, ?: |
+Cast Operator | (type) | No direct operator, difference in GetElementPointer 
+||| or load type
```
pow2clk: Not sure what happened here. If the entry gets too long, most renderers will use an additional line on their own. As-is, I don't think these look right.
```suggestion
Bitwise Operators | ~, <<, >>, &, \|, ^, <<=, >>=, &=, \|=, ^= | Only valid on
 int and uint vectors
Boolean Math Operators | & &, ||, ?: |
Cast Operator | (type) | No direct operator, difference in GetElementPointer 
or load type
```

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+These operators generate LLVM instructions which use vectors.
+
+Operator table from [Microsoft HLSL Operators](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators)
+| Operator Name | Operator | Notes |
+|-----------|--------------|----------|
+| Addition | + | blah |
+| Subtraction | - | blah |
+| Multiplication | * |
+Additive and Multiplicative Operators | +, -, *, /, % |
+Array Operator | [i] | llvm:ExtractElementInst
```
pow2clk: It can also be InsertElementInst when used to assign a value.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+| asin      | Asin | | range: -pi/2 to pi/2. Floating point types only. |
+| atan      | Atan | | range: -pi/2 to pi/2. |
+| cos       | Cos | | no range requirements. |
+| cosh      | Hcos | | no range requirements. |
+| sin       | Sin | | no range requirements. |
+| sinh      | Hsin | | no range requirements. |
+| tan       | Tan | | no range requirements. |
+| tanh      | Htan | | no range requirements. |
+| atan2     | Atan | FDiv, FAdd, FSub, FCmpOLT,
+||| FCpmOEQ, FCmpOGE, FCmpOLT, And, Select | Not required. Covered by other ops. |
```
pow2clk: Is this all meant to refer to atan2? Like this maybe?
```suggestion
 FCpmOEQ, FCmpOGE, FCmpOLT, And, Select | Not required. Covered by other ops. |
```

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+| degrees   | | FMul | Not needed. Covered by FMul. |
+| radians   | | FMul | Not needed. Covered by FMul. |
+
+## Math
+
+| Intrinsic | DXIL OPCode | LLVM Instruction | Notes |
+|-----------|--------------|----------|-----------|
+| abs       | [Imax], [Fabs] |
```
pow2clk: I think I understand what's going on here now in that the `[]` mean that whether they get generated or not depends. I think it would be helpful to mention what it depends on in the Notes. 
```suggestion
| abs       | [Imax], [Fabs] | Imax for ints. Fabs for floats. |
```

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+| ceil      | Round_pi ||
+| exp       | Exp | |
+| floor     | Round_ni ||
+| fma       | Fma | |
+| frac      | rc | |
+| frexp     | | FCmpUNE, SExt, BitCast, And, Add,
+||| AShr, SIToFP, Store, And, Or | |
```
pow2clk:
```suggestion
AShr, SIToFP, Store, And, Or | |
```

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+| ldexp     | Exp | FMul |  |
+| lerp      | | FSub, FMul, FAdd | |
+| log       | Log | FMul | |
+| mad       | IMad | |
+| max       | IMax | |
+| min       | IMin | |
+| pow       | [Log, Exp] | [FMul] , [FDiv] ||
```
pow2clk: Hoo boy. Now I regret suggesting listing the conditions in the notes because this one is pretty wild. Maybe just link to this wall of text and call it a day:
```suggestion
| pow       | [Log, Exp] | [FMul] , [FDiv] | See [This comment](https://github.com/microsoft/DirectXShaderCompiler/blob/50f53c6c200fd6b53f65268912e6f9e444ce9242/lib/HLSL/HLOperationLower.cpp#L761-L785)|
```

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+| or        | | Or, [ExtractElement, InsertElement] | Not required. Covered by select. |
+
+## Reductions
+
+| Intrinsic | DXIL OPCode | LLVM Instruction | Notes |
+|-----------|--------------|----------|-----------|
+| all       | | [FCmpUNE], [ICmpNE] ,
+||| [ExtractElement, And] |
+| any       | | [FCmpUNE], [ICmpNE] ,
+||| [ExtractElement, Or] | |
```
pow2clk:
```suggestion
 [ExtractElement, And] |
| any       | | [FCmpUNE], [ICmpNE] ,
 [ExtractElement, Or] | |
```

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,332 @@
...
+| Intrinsic | DXIL OPCode | LLVM Instruction | Notes |
+|-----------|--------------|----------|-----------|
+| ddx       | DerivCoarseX | |
+| ddx_fine  | DerivFineX | |
+| ddy       | DerivCoarseY | |
+| ddy_fine  | DerivFineY | |
+| fwidth    | QuadReadLaneAt | |
+| QuadReadLaneAcrossX | QuadOp | |
+| QuadReadLaneAcrossY | QuadOp | | Not requied. Covered by QuadReadLaneAcrossX |
+| QuadReadLaneAcrossDiagonal | QuadOp | Not required. Covered by QuadReadLaneAcrossX |
```
pow2clk: Disagree. While these all use QuadOp, the different parameters make them different operations. I think we need to test them all. 
```suggestion
| QuadReadLaneAcrossY | QuadOp | | Uses different QuadOp parameters leading to different behavior|
| QuadReadLaneAcrossDiagonal | QuadOp | Uses different QuadOp parameters leading to different behavior |
```

pow2clk (commented):
It's a lot of cleanups and nitpicks, but I think overall it looks really solid!

proposals/0026-hlsl-long-vector-type.md
alsepkow: Only works for me locally if I include the dot. Does it work for you once switched to forward slashes?

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: I think I meant non-even, but fair point.

Non-even sizes are trivial to me. I'm not clear what peculiar dimensions would actually be. What does it mean to cut element blocks at different points? Conceptually I get it, but I'm not clear how I would test that? Are you referring loading/storing on a ByteAddressBuffer?

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: In this instance I was meaning to reference the same type name as you would when writing HLSL source. Can you help me understand where the ambiguity is?

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: The preview in vs code kept stretching to fill my screen. I did this as a way to force the wrap. I'll remove it and just put all of the text in one cell.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: Yes, I was attempting to limit the width of the table.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: If I do that then the second line is left justified and looks wrong. I'll just move it to the same cell like ive been doing for the other instances.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: Haha, yah. I initially started providing some further context but realized quickly that wasn't going to be a good time. I can go through and link sections. But I'm also thinking that maybe just a link do HLOperationLower.cpp with a note that if you're interested then you can search for the applicable function. I'm note sure many people (if any) would really follow these links anyways? Main thought is I'm not sure its worth my effort :)

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: removed

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: Changed to bullet points, do you think that is clearer?

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: I've rewritten it to be more explicit. Hopefully that helps.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: Discussed offline. We are comfortable with keeping this at covering odd sized vectors right now.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
pow2clk: I think we resolve this in discussion.

pow2clk (approved):
Thanks for the changes!

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,363 @@
...
+This test binary is only built in the OS repo and based off of the
+ExecutionTests source code in the DXC repo. There is a script in the WinTools
+repo which generates and annotates the HLK tests.
+
+We break coverage down into five test categories.
+
+1. Implement DXIL OpCode tests:
+    * At the bottom of this document there are [tables](#hlsl-operators)
+      containing all HLSL operators (more on those in '3. HLSL Operator Tests')
+      and HLSL intrinsics that can be used with long vectors. The HLSL
```
bob80905, line 30: Just nits on typos:
"...have a DXIL OpCode column and an LLVM instruction column. These columns contain the intrinsic's mapped DXIL OpCodes as well as their LLVM instructions."

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,363 @@
...
+      intrinsics tables have a DXIL OpCode and LLVM instruction columns. These
+      columns contain the the intrinsics mapped DXIL OpCodes as well as there
+      LLVM instructions. All intrinsics have at least one DXIL OpCode or one
+      LLVM instruction.
+
+      Many instrinsics have trivial mappings. [Atan](#trigonometry) is an
+      example of an intrinsic with a trivial mapping. Other intrinsics have
+      multiple DXIL OpCodes. Some intrinsics will use all listed DXIL OpCodes
+      and/or LLVM instruvtions, while others will have additonal logic which
```
bob80905: `instructions`

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,363 @@
...
+      determines which OpCodes/Instructions are used. If an intrinsic relies on
+      additional logic to determine which OpCodes/Instructions are used then the
+      OpCode/Instructions will be enclosed in '[]' brackets. The [sign](#math)
+      intrinsic is an example of an intrinsic with addtional logic. If an
```
bob80905: `additional`

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,363 @@
...
+      OpCode/Instruction is not enclosed in '[]' then it is used in all paths
+      for that intrinsic.
+
+2. Implement LLVM Instruction tests:
+    * These are the test cases for the LLVM Instructions listed in the in table
```
bob80905: `in the table`

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,363 @@
...
+     at the bottom of this document.
+    * Because we will use HLSL intrinsics to get coverage for the DXIL OpCode
+       tests we speculate that we will get most of the coverage needed for the
+       LLVM Instruction tests. After implementing the DXIL OpCode tests we
+       should be able to do a coverage audit and ammend test cases, or write
+       simple additional ones, as needed.
+    * Just as in '1. Implement DXIL OpCode Tests' some cases have multiple
+       instructions listed. '[]' brackets are used in the same manner. And there
+       may also be multiple instructions.
+    additional OpCodes/Instructions are logic based (i.e float or int specific).
```
bob80905: `Additional`, also it looks like this needs a tab

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,363 @@
...
+    * These could be covered in test categories 1 and 2. But I propose we break
+      out individual cases to ensure that we have more granular coverage.
+    * Ensure we have some basic tests doing standard loading/storing of long
+     vectors across [Buffer types to test](#buffer-types-to-test) and [Vector
+     element data types to test](#vector-element-data-types-to-test).
+    * Additionally, the above buffer types and data types should be tested by
+      loading from a [ResourceDescriptorHeap](https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_DynamicResources.html)
+
+5. 'Creative' test cases:
+    * Sizes around alignments and boundaries (details in [Vector Sizes to
```
bob80905: For some reason this isn't linking right.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,363 @@
...
+* vector<TYPE, 256> : Arbitrarily picked.
+* vector<TYPE, 1024> : The new max size of a vector. 
+* These sizes will be tested across [Vector element data types to test](#vector-element-data-types-to-test)
+
+## Some noteable alignment cases
+
+* 128 bit boundaries : Memory access for Shader Model 5.0 and earlier operate on
+128-bit slots aligned on 128-bit boundaries. An example is vector<half, 7>,
+vector<half, 8> and vector<half, 9>. 112 bits, 128 bits, and 144 bits
+respectively. This boundary will tested for with 32-bit and 64-bit sized values
```
bob80905, line 117: `will be tested with`?

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,363 @@
...
+as well.
+
+* Most GPUs operate on at least 32-bits at once, so what happens if you use
+  16-bit values and an odd number of elements. Could accesssing the last element
+  expose issues where we could overwrite the next variable is it is assuming
```
bob80905: `the next variable assuming alignment?`

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,363 @@
...
+
+1. The test will leverage the existing XML infrastructure currently used by the
+   existing execution tests. There are two XML files. This general design
+   pattern exists today in the execution tests.
+
+   * 1st XML: Used to define shader source code and metadata about that shader
+   code. This XML file is parsed using a private class. This private class helps
+   faciliate creation of D3D resources and execution of the shader.
+   * 2nd XML: Describes metadata about the specific test cases. Used by the
+   TAEF infrastructe for [TAEF Data Driven
```
bob80905: `infrastructure`

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,363 @@
...
+   Testing](https://learn.microsoft.com/en-us/windows-hardware/drivers/taef/data-driven-testing)
+2. Test inputs will be hard coded in a c++ header file. This was chosen over
+   definining inputs in the second XML as this is cleaner and easier to parse
+   for different data types. This c++ header method also avoids needing to
+   repeat the data set in the XML for each individual test cast. Inputs will use
+   'value sets' which will typically be much smaller than the desired vector
+   test size. Values will be repeated until the vector is full.
```
bob80905: So with a value set {1, 2} and a vector<int, 1024>, are we guaranteed the sequence <1, 2, 1, 2, 1, 2, ..., 2, 1, 2> ?

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,363 @@
...
+
+Note that because DXC and the Agility SDK are both undocked from Windows it is
+our normal operating behavior for the HLK tests to become available with a later
+TBD OS release. The good news is that this doesn't prevent the tests from being
+available much earlier in the DXC repo. It just means that they are simply TAEF
+tests in the DXC repo. An HLK test includes an extra level of infrastructure for
+test gating, selection, and result submission for WHQL signing of drivers.
+
+1. Tests will be shared privately with IHVs along with the latest DXC and
+latest Agility SDK for tesing and validation. IHVs will also be able to build
```
bob80905: `testing`

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -0,0 +1,363 @@
...
+and run the tests from the public DXC repo themselves. If needed Microsoft can
+share further instructions when the tests are available.
+
+2. The tests will ship with the HLK at a TBD date in a later OS release.
+
+## New HLK Tests
+
+All of the scripts and yaml files mentioned are part of the Microsoft WinTools
```
bob80905: Unsure if we should / need to mention the internal WinTools repo, or if we can leave it out of this spec?

bob80905 (commented):
Excellent job on the table!

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: Good question! Yes, with a value set {1, 2} and a vector<int, 1024>, the pattern will be <1, 2, 1, 2, 1, 2, ..., 1, 2> repeating cyclically until all 1024 elements are filled. This ensures we get consistent, predictable test data while keeping the input definitions manageable.

I'll update the wording here to make that crystal clear.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: I think I'll remove it. It feels weird having it mentioned in the public space even though the tooling repo is private.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -131,16 +131,20 @@ as well.
...
    Testing](https://learn.microsoft.com/en-us/windows-hardware/drivers/taef/data-driven-testing)
 2. Test inputs will be hard coded in a c++ header file. This was chosen over
    definining inputs in the second XML as this is cleaner and easier to parse
    for different data types. This c++ header method also avoids needing to
    repeat the data set in the XML for each individual test cast. Inputs will use
    'value sets' which will typically be much smaller than the desired vector
-   test size. Values will be repeated until the vector is full.
+   test size. Values will be repeated cyclically until the vector is full. For
+   example, a value set `{1, 2, 3}` used to populate a `vector<int, 1024>` will
+   produce the pattern `<1, 2, 3, 1, 2, 3, 1, 2, 3, ..., 1, 2, 3>`, repeating
```
bob80905: Just thought I'd mention, 1024 isn't divisible by 3, so the final element would be `1`, since 1024 mod 3 = 1.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -216,7 +211,7 @@ Operator table from [Microsoft HLSL Operators](https://learn.microsoft.com/en-us
 | Multiplication | * | |
 | Additive and Multiplicative Operators | +, -, *, /, % | |
 | Array Operator | [i] | llvm:ExtractElementInst OR llvm:InsertElemtInst |
-| Assignment Operators | =, +=, -=, *=, /=, %= |
+| Assignment Operators | =, +=, -=, *=, /=, %= | |
```
bob80905, line 214: This seems like an accident?

alsepkow:
Blame copilot
________________________________

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: Which link? They're working for me, but the document links only work if im viewing the md in preview mode.

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
alsepkow: Do you mean to explicitly state which 32 bit and 64 bit types we will use?

proposals/infra/INF-0005-Long-Vector-ExecutionTest-Plan.md
bob80905: vector-sizes-to-test
image

bob80905: Perhaps you meant to use `Vector sizes and alignments to test`

alsepkow: Ah, Doh. Fixing.

PR #422, opened 2025-03-11
Add stride alignment, FP8 details, fix VectorMatrix to MatrixVector...

anupamachandra:
Covers Feedback:
FP8 Emulation Path
Agreement to make explicit in the spec that FP8 matrices must be in optimal layout, which gives the opportunity to store data in FP16. `DONE`
Agreement that the driver and compiler just need to have the same logic for which stride to use, which will be implicit for optimal layouts.
Agreement: spec will only require FP16 internal precision `Internal precision is an implementation detail not covered in spec`
Agreement that the conversion to FP8 optimal layout will be done precisely (exact expected value) `Implementation detail not covered in spec`

Stride
Agreement to add an alignment requirement for stride (multiple of 16) to the spec. `DONE`

Split from : https://github.com/microsoft/hlsl-specs/pull/420.
ReduceSumAccumulate renaming will be handled in a separate PR.

### DIFF
--- a/proposals/0029-cooperative-vector.md
+++ b/proposals/0029-cooperative-vector.md
@@ -546,7 +546,8 @@ typedef enum D3D12_COOPERATIVE_VECTOR_DATATYPE {
 typedef enum D3D12_COOPERATIVE_VECTOR_TIER
 {
     D3D12_COOPERATIVE_VECTOR_TIER_NOT_SUPPORTED,    
-    D3D12_COOPERATIVE_VECTOR_TIER_1_0
+    D3D12_COOPERATIVE_VECTOR_TIER_1_0,
+    D3D12_COOPERATIVE_VECTOR_TIER_1_1
 }
 
 // This struct may be augmented with more capability bits
@@ -556,7 +557,7 @@ typedef struct D3D12_FEATURE_DATA_D3D12_OPTIONSNN // NN tbd when implemented
     Out D3D12_COOPERATIVE_VECTOR_TIER CooperativeVectorTier;
 } D3D12_FEATURE_DATA_D3D12_OPTIONSNN;
 
-// Used for VectorMatrixMulAdd intrinsic
+// Used for MatrixVectorMulAdd intrinsic
 typedef struct D3D12_COOPERATIVE_VECTOR_PROPERTIES_INFERENCE
 {
     D3D12_COOPERATIVE_VECTOR_DATATYPE InputType;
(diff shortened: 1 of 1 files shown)


proposals/0029-cooperative-vector.md
```diff
@@ -699,6 +713,8 @@ the inputs required to calculate the necessary size. The same descriptor,
 updated with the calculated output size, is then passed to the conversion
 API. 
 
+The `DestStride` should be a multiple of 16 bytes.
```
damyanp, line 716:
```suggestion
The `DestStride` must be a multiple of 16 bytes.
```

damyanp (approved):

proposals/0029-cooperative-vector.md
```diff
@@ -546,7 +546,8 @@ typedef enum D3D12_COOPERATIVE_VECTOR_DATATYPE {
 typedef enum D3D12_COOPERATIVE_VECTOR_TIER
 {
     D3D12_COOPERATIVE_VECTOR_TIER_NOT_SUPPORTED,    
-    D3D12_COOPERATIVE_VECTOR_TIER_1_0
+    D3D12_COOPERATIVE_VECTOR_TIER_1_0,
```
amarpMSFT, line 549: If we rename COOPERATIVE_VECTOR to something like LINEAR_ALGEBRA, would this tiering system still make sense, in that mat*mat would layer on top?  Or would we need indepdendent tier progressions for vecMat vs matMat?

If we need separate tier progresssions, could the names be D3D12_LINEAR_ALGEBRA_VEC_MAT_TIER and (future) D3D12_LINEAR_ALGEBRA_MAT_MAT_TIER?  Or is there a more accurate naming?

amarpMSFT: We could do such a rename in a different PR, just trying to understand the implications for this tiering.

anupamachandra: I think D3D12_LINEAR_ALGEBRA_**2**_* for Mat-Mat operations, leaving D3D12_LINEAR_ALGEBRA_1_* should be sufficient?

tex3d (approved):

proposals/0029-cooperative-vector.md
damyanp: I think more discussion here is required, but it outside the scope of this PR so I'm going to go ahead and merge this one.

PR #428, opened 2025-03-14
Cooperative Vectors Test Plan

VladM1076:
I have created a HLK/ExecTest test plan for the Cooperative Vectors feature and I am starting to work on its implementation.

Since the spec is still in flux, I have based it the latest I am currently aware of and will keep it up to date as spec changes. 

TODO: Update naming once spec is finalized.

### DIFF
--- a/proposals/infra/INF-0006-CoopVec-TestPlan.md
+++ b/proposals/infra/INF-0006-CoopVec-TestPlan.md
@@ -1,0 +1,257 @@
+# Cooperative Vector DirectX Feature - Test Plan
+
+<a name="top"></a>
+
+## Executive Summary
+
+**DISCLAIMER: This is based on the WIP cooperative vector spec. Some details may change.**
+
+**TODO: Update naming once spec is finalized.**
+
+**Current status is: UNDER EXTERNAL REVIEW**
+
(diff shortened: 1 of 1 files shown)


VladM1076:
@microsoft-github-policy-service agree company="NVIDIA"

VladM1076:
I added it to the cooperative vector proposal document as requested.
However since this is the HLSL test plan, wouldn't it be more appropriate in the vector-matrix operations proposal? Or just both?

damyanp:
> I added it to the cooperative vector proposal document as requested. However since this is the HLSL test plan, wouldn't it be more appropriate in the vector-matrix operations proposal? Or just both?

The execution tests are really validating how the driver responds to the DXIL it receives.  It's just that we write the tests in HLSL so we can use the compiler to generate the DXIL.  That's much more convenient than hand-crafting DXIL!

VladM1076:
Added more details on testing different shader stages in section 2.5.
Enumerated shader stages that need testing and explicitly called out helper pixel tests.

VladM1076:
Added OOB tests to the plan and also updated InterlockedAdd to VectorAccumulate to align with new naming.

damyanp:
@vladm1076 - I'm going to close this as I don't believe we intend to merge the PR now?

PR #430, opened 2025-03-14
[0029] Clarify 16-byte underlying allocation requirement

damyanp:


proposals/0029-cooperative-vector.md
```diff
@@ -318,7 +318,9 @@ interpretation** and **matrix layout** behaving as described [above]
 The base address of **matrix resource** and **matrix offset** must be 128 byte
 aligned. Also note that the size of the underlying allocation is guaranteed to
 be a multiple of 16 bytes ensuring that the 16 bytes access of the last
-row/column of the matrix is valid memory
+row/column of the matrix is valid memory. Implementations may write to the
```
anupamachandra, line 321: One idea is to rewrite this as : Implemetations may need to use the padding space (e.g while using smaller types like FP16 that certain hardware might not natively support and are therefore emulated with bigger native types), so as a general rule using the padding space is undefined behavior.

Just to give the reader more information of why we have this restriction.

llvm-beanz (approved):

proposals/0029-cooperative-vector.md
damyanp: I think this is more about supporting cases where there is native support for the type but it has additional alignment requirements.  I think I might leave this as it is for now, can add clarification later, but that really belongs in the user documentation.

proposals/0029-cooperative-vector.md
```diff
@@ -920,7 +927,7 @@ void ID3D12CommandList::ConvertLinearAlgebraMatrix(D3D12_LINEAR_ALGEBRA_MATRIX_C
 * If SrcLayout is row-major or column-major, then SrcStride should be greater than the length of a row/column, and a
   multiple of the element size.
 * If DestLayout is row-major or column-major, then DestStride should be greater than the length of a row/column, and a
-  multiple of the element size.
+  multiple of 16.
```
jenatali, line 930: Is it correct that both of these bullets are talking about row-major/column-major, but only dests have this stride requirement? Perhaps the 16 byte stride requirement should be unconditionally applied to the dest wording here, i.e. a new bullet that's not coupled to the row-major/column-major bit.

damyanp: The stride value is ignored for non row-major/column-major layouts, so it isn't really unconditionally applied to the dest.  Or am I misunderstanding your suggestion?

jenatali: Oh right, forgot that spec detail.

jenatali (approved):

joaosaffran (approved):
The text LGTM

PR #442, opened 2025-03-21
[0029] Alternative way to query D3D device for supported CoopVec type combinations (no HLSL/DXIL impact)

amarpMSFT:
This proposes changing the way cooperative vector type combinations are reported from the driver from a list of all combinations to instead the app individually requesting whether the driver supports a given configuration defined by the app.

No HLSL/DXIL impact here, just a tweak to the D3D API.   Question for reviewers:  Does this make sense?  Note the commentary included in the changed spec text discussing how an app would use this new paradigm.

### DIFF
--- a/proposals/0029-cooperative-vector.md
+++ b/proposals/0029-cooperative-vector.md
@@ -687,28 +687,52 @@ typedef struct D3D12_COOPERATIVE_VECTOR_PROPERTIES_TRAINING
     D3D12_LINEAR_ALGEBRA_DATATYPE AccumulationType;
 };
 
-// CheckFeatureSupport data struct used with type D3D12_FEATURE_COOPERATIVE_VECTOR:
+typedef enum D3D12_COOPERATIVE_VECTOR_PROPERTIES_TYPE
+{
+    D3D12_COOPERATIVE_VECTOR_PROPERTIES_TYPE_MATRIX_VECTOR_MUL_ADD,
+    D3D12_COOPERATIVE_VECTOR_PROPERTIES_TYPE_OUTER_PRODUCT_ACCUMULATE,
+    D3D12_COOPERATIVE_VECTOR_PROPERTIES_TYPE_VECTOR_ACCUMULATE,
+} D3D12_COOPERATIVE_VECTOR_PROPERTIES_TYPE;
+
 typedef struct D3D12_FEATURE_DATA_COOPERATIVE_VECTOR
-{    
-    InOut UINT                                         MatrixVectorMulAddPropCount;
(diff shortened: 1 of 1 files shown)


proposals/0029-cooperative-vector.md
```diff
@@ -687,28 +687,52 @@ typedef struct D3D12_COOPERATIVE_VECTOR_PROPERTIES_TRAINING
...
+
+`D3D12_COOPERATIVE_VECTOR_PROPERTIES_INFERENCE` has a `BOOL TransposeSupported`
+field - if the app sets this to `TRUE` they are specifically asking for 
+the specified inference configuration with transpose.  The driver support is
+still reported out via the `ConfigurationSupported` field mentioned above 
+(as opposed to treating `TransposeSupported` as an output field).
+
+> To avoid apps having to guess any feature combination that might work,
+> a minimum required set is listed below (though `TransposeSupported` would
+> still need to be checked if needed).  For combinations other than the
```
jenatali, line 729: Consider a comment that, based on the cooperative vectors tier, if `TransposeSupported` is false and the configuration is in the minimum supported set, that D3D can/will answer the support question without asking the driver, so the driver only needs to report configurations above and beyond the baseline?

jenatali (approved):
I think a chattier interface that doesn't have to condense a 6D matrix into a 1D array makes sense.

proposals/0029-cooperative-vector.md
amarpMSFT: Added a note about this, though didn't incorporate tiering into it based on currently defined tiers.  Did I miss something?

jenatali: Yeah that's fine, the tier only impacts the training queries. For tier 0, we can just flat return false for all training queries, and tier 1 adds a minimum baseline there. I wouldn't be surprised to see future tiers that expand out the minimum support set over time.

anupamachandra: Couple of points about this:
1. If apps have to rely on vendor documentation for this guess and check approach, that is an indirection that is not desirable, also this documentation needs to be updated for different skus and potentially driver versions making it hard to maintain.
2. How will testing work in this case. The current plan is to test beyond the minimum support set, by iterating over the table returned by the driver. Same with caps viewer functionality.

amarpMSFT: Is there a risk the number of supported permutations could explode?

anupamachandra: Practically I don't think this will happen and the number of combinations will stay in the 10s. We do expect the set to increase as support for newer (smaller) types are more readily available, but unlikely that implementations would advertise support and maintain 100s of combinations. Adding a helper query seems like a good idea.. perhaps down the line?

PR #450, opened 2025-03-22
[0031] Add builtins

damyanp:
Although these "builtins" are not intended to be part of the HLSL language, and
no promises are made that these will continue to be available over time, this
proposal describes an implementation in terms of builtins such as this. For
this reason it is useful to have them described here as a reference point.

Each builtin corresponds to one of the operations described in [0029].

[0029]: 0029-cooperative-vector.md

----

For this review, I'm keen on feedback for whether or not this is a good way to describe the builtins?  At the moment they're HLSL code that I know can compile, but that might mean they've picked up HLSL-isms that don't necessarily apply to a specification of a builtin.

### DIFF
--- a/proposals/0031-hlsl-vector-matrix-operations.md
+++ b/proposals/0031-hlsl-vector-matrix-operations.md
@@ -174,23 +174,26 @@ namespace details {
 
 // dx.op.matvecmul
 template <typename TYo, int NUMo, typename TYi, int NUMi, typename RES>
-void __builtin_MatVecMul(vector<TYi, NUMi> InputVector,
-                         uint InputVectorInterpretation, RES MatrixResource,
-                         uint MatrixStartOffset, uint MatrixInterpretation,
-                         uint M, uint K, uint Layout, bool MatrixTranspose,
-                         uint MatrixStride, out vector<TYo, NUMo> OutputVector);
+void __builtin_MatVecMul(out vector<TYo, NUMo> OutputVector,
+                         bool IsOutputUnsigned, vector<TYi, NUMi> InputVector,
+                         bool IsInputUnsigned, uint InputVectorInterpretation,
+                         RES MatrixResource, uint MatrixStartOffset,
+                         uint MatrixInterpretation, uint M, uint K,
+                         uint MatrixLayout, bool MatrixTranspose,
+                         uint MatrixStride);
 
 // dx.op.matvecmuladd
 template <typename TYo, int NUMo, typename TYi, int NUMi, typename RESm,
           typename RESv>
-void __builtin_MatVecMulAdd(vector<TYi, NUMi> InputVector,
+void __builtin_MatVecMulAdd(out vector<TYo, NUMo> OutputVector,
+                            bool IsOutputUnsigned,
+                            vector<TYi, NUMi> InputVector, bool IsInputUnsigned,
                             uint InputVectorInterpretation, RESm MatrixResource,
                             uint MatrixStartOffset, uint MatrixInterpretation,
-                            uint M, uint K, uint Layout, bool MatrixTranspose,
-                            uint MatrixStride, RESv BiasVectorResource,
-                            uint BiasVectorOffset,
-                            uint BiasVectorInterpretation,
-                            out vector<TYo, NUMo> OutputVector);
+                            uint M, uint K, uint MatrixLayout,
+                            bool MatrixTranspose, uint MatrixStride,
+                            RESv BiasVectorResource, uint BiasVectorOffset,
+                            uint BiasVectorInterpretation);
 
 // dx.op.outerproductaccumulate
 template <typename TY, int M, int N, typename RES>
--- a/proposals/0031-hlsl-vector-matrix-operations.md
+++ b/proposals/0031-hlsl-vector-matrix-operations.md
@@ -173,11 +173,11 @@ namespace linalg {
 namespace details {
 
 // dx.op.matvecmul
-template <typename TYo, int NUMo, typename TYi, int NUMi, typename RES>
+template <typename TYo, int NUMo, typename TYi, int NUMi, typename RESm>
 void __builtin_MatVecMul(out vector<TYo, NUMo> OutputVector,
                          bool IsOutputUnsigned, vector<TYi, NUMi> InputVector,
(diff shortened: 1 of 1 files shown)


proposals/0031-hlsl-vector-matrix-operations.md
```diff
@@ -160,7 +160,58 @@ TBD
...
+Each builtin corresponds to one of the operations described in [0029].
+
+```c++
+namespace dx {
+namespace linalg {
+namespace details {
+
+// dx.op.matvecmul
+template <typename TYo, int NUMo, typename TYi, int NUMi, typename RES>
+void __builtin_MatVecMul(vector<TYi, NUMi> InputVector,
```
anupamachandra, line 177: Make the out vector arg in Mul and MulAdd as the first in sync with the implementation

proposals/0031-hlsl-vector-matrix-operations.md
```diff
@@ -160,7 +160,58 @@ TBD
...
+                            uint BiasVectorInterpretation,
+                            out vector<TYo, NUMo> OutputVector);
+
+// dx.op.outerproductaccumulate
+template <typename TY, int M, int N, typename RES>
+void __builtin_OuterProductAccumulate(vector<TY, M> InputVector1,
+                                      vector<TY, N> InputVector2,
+                                      RES MatrixResource,
+                                      uint MatrixStartOffset,
+                                      DataType MatrixInterpretation,
```
anupamachandra, line 201: should MatrixInterpretation and Layout be uints here or do we want enums?

joaosaffran (approved):
LGTM, seems reasonable to define HLSL built ins that way. I would suggest documenting the functions with known limitation, or requisites, such as for `__builtin_MatVecMul`, given that the Input Matrix is `MxN`, the input vector should be `N` elements or `M` if transpose is true (not sure if that is actual limitation, just some example of additional information we might want to add). Another example is document possible OpCodes we must use when lowering to DXIL/SPIRV.

proposals/0031-hlsl-vector-matrix-operations.md
damyanp: I think they need to be uints - I'm trying to define these builtins with the minimum number of dependencies, so I don't want to also have to define the enums for them.

proposals/0031-hlsl-vector-matrix-operations.md
```diff
@@ -160,7 +160,61 @@ TBD
 
 ### Builtins
 
-TBD
+Although these "builtins" are not intended to be part of the HLSL language, and
+no promises are made that these will continue to be available over time, this
+proposal describes an implementation in terms of builtins such as this. For
+this reason it is useful to have them described here as a reference point.
+
+Each builtin corresponds to one of the operations described in [0029].
```
alsepkow, line 168: The link to [0029] doesn't work.

damyanp: It works on my machine:

image

proposals/0031-hlsl-vector-matrix-operations.md
```diff
@@ -160,7 +160,61 @@ TBD
...
+namespace details {
+
+// dx.op.matvecmul
+template <typename TYo, int NUMo, typename TYi, int NUMi, typename RES>
+void __builtin_MatVecMul(out vector<TYo, NUMo> OutputVector,
+                         bool IsOutputUnsigned, vector<TYi, NUMi> InputVector,
+                         bool IsInputUnsigned, uint InputVectorInterpretation,
+                         RES MatrixResource, uint MatrixStartOffset,
+                         uint MatrixInterpretation, uint M, uint K,
+                         uint MatrixLayout, bool MatrixTranspose,
```
alsepkow, line 182: nit: 'MatrixTranpose' -> 'IsMatrixTransposed' would make the naming consistent with other bool parameters.

proposals/0031-hlsl-vector-matrix-operations.md
```diff
@@ -160,7 +160,61 @@ TBD
...
+                         uint MatrixStride);
+
+// dx.op.matvecmuladd
+template <typename TYo, int NUMo, typename TYi, int NUMi, typename RESm,
+          typename RESv>
+void __builtin_MatVecMulAdd(out vector<TYo, NUMo> OutputVector,
+                            bool IsOutputUnsigned,
+                            vector<TYi, NUMi> InputVector, bool IsInputUnsigned,
+                            uint InputVectorInterpretation, RESm MatrixResource,
```
alsepkow, line 191: This function uses 'RESm' while the other functions use 'RES'. 'RESm' is probably better.

proposals/0031-hlsl-vector-matrix-operations.md
damyanp: Updated

alsepkow: 'uint MatrixLayout' -> 'MatrixLayout layout'

proposals/0031-hlsl-vector-matrix-operations.md
damyanp: Updated `__builtin_MatVecMul` to match.  I'm happy with how the *Accumulate ones are named.

proposals/0031-hlsl-vector-matrix-operations.md
damyanp: Nice catch - but actually it's the other way around.  The builtins can't use the enums.

proposals/0031-hlsl-vector-matrix-operations.md
```diff
@@ -160,7 +160,61 @@ TBD
...
+Each builtin corresponds to one of the operations described in [0029].
+
+```c++
+namespace dx {
+namespace linalg {
+namespace details {
+
+// dx.op.matvecmul
+template <typename TYo, int NUMo, typename TYi, int NUMi, typename RESm>
+void __builtin_MatVecMul(out vector<TYo, NUMo> OutputVector,
```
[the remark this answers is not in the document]
damyanp: We can't use the enums in the builtins.

alsepkow (approved):
lgtm

PR #453, opened 2025-03-24
[SER] Specify how reordercoherent is encoded

rasmusnv:
In the DXIL section, specify how reordercoherent is encoded.

### DIFF
--- a/proposals/0027-shader-execution-reordering.md
+++ b/proposals/0027-shader-execution-reordering.md
@@ -1544,15 +1544,23 @@ This new resource attribute is encoded in DXIL in the following way:
 
 A new tag is added to the resource extended property tags:
 ```cpp
-  static const unsigned kDxilAtomic64UseTag = 3;
-+ static const unsigned kDxilReorderCoherentTag = 4;
+   static const unsigned kDxilAtomic64UseTag = 3;
++  static const unsigned kDxilReorderCoherentTag = 4;
 ```
 The tag carries an `i1` value that indicates whether the resource is reordercoherent. The resource is not reordercoherent when the tag is absent.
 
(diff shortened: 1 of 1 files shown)


proposals/0027-shader-execution-reordering.md
```diff
@@ -1537,6 +1537,24 @@ Validation errors:
...
+  static const unsigned kDxilAtomic64UseTag = 3;
++ static const unsigned kDxilReorderCoherentTag = 4;
+```
+The tag carries an `i1` value that indicates whether the resource is reordercoherent. The resource is not reordercoherent when the tag is absent.
+
+A flag is added to the Dxil Library Runtime Data (RDAT):
+```cpp
+  RDAT_ENUM_VALUE(Atomics64Use,             1 << 4)
++ RDAT_ENUM_VALUE(UAVReorderCoherent,       1 << 5)
+```
```
tex3d, line 1556: I think this is still missing potentially the most important location.  In `DxilResourceProperties.h`, we need to add a new bit after `SamplerCmpOrHasCounter`:
https://github.com/microsoft/DirectXShaderCompiler/blob/94596e1c97e10ef2f97cf21d33cbabdc0e7df2e8/include/dxc/DXIL/DxilResourceProperties.h#L47

starting at the next byte:
```diff
    // BYTE 2
-    uint8_t Reserved2;
+    uint8_t ReorderCoherent : 1;
+    uint8_t Reserved2 : 7;
```

This bit needs to be set so AnnotateHandle will indicate the property on the resource handle being used.

rasmusnv: Thanks. This has been added.

tex3d (approved):

PR #461, opened 2025-03-26
New Proposal: Draft proposal to modify resource typing in HLSL

Tobski:
In discussion with @llvm-beanz about how resources should be handled in future, and noting that the Vulkan WG has publicly noted they are working on changing how descriptors work, myself and others (including @llvm-beanz) have been working on a proposed overhaul to HLSL's resource model, with the following properties:

 - Resource type handling overhauled for consistency
 - Stronger typing added to resource heap access
 - "Raw" buffer types added for handling GPU VAs
 - Recursive typing (e.g. buffers can contain buffers or images)
 - "Bindless" access for root and local root data
 - Enables deprecation of descriptor tables

The big wins here for shader authors are the ability to reason clearly about how resource types map to real data, the ability to define recursive types, and having the ability to more strongly type their accesses.

For example, the following shader code would be valid and meaningful with this proposal:

```hlsl
struct SomeResources {
	Texture2D texture;
	RWBuffer<Texture2D> bufferOfTextures;
};

ConstantBuffer<SomeResources> resourceBuffer;
```

In the above, each resource is actually a handle into the heap, set as a 32-bit integer value in memory, noting that these can also be _written_ to memory as well.

This does not, as currently proposed, require any changes to the DX12 or Vulkan resource models, though some features may need some driver changes, and there are some open issues and suggested directions which *would* require changes.

### DIFF
--- a/proposals/0TBD-improved-resource-typing.md
+++ b/proposals/0TBD-improved-resource-typing.md
@@ -1,0 +1,524 @@
+<!-- {% raw %} -->
+
+# Typed resource heaps
+
+* Proposal: [0TBD](0TBD-improved-resource-typing.md)
+* Author(s): [Tobias Hector](https://github.com/tobski)
+* Sponsor: Chris Bieneman
+* Status: **Under Consideration**
+
+
+## Introduction
+
+This proposal adds a method to access the resource heap in HLSL with stronger
+typing, and adds "bindless" methods of accessing root constants and shader
+record data, fully removing the need to use descriptor tables.
+
+
+## Motivation
+
+The `ResourceHeap` added in Shader Model 6.6 exposed a typeless, unsized heap
+to shader authors which could be used to access any descriptor placed in the
+heap through the client API.
+
+While useful, without type information, it is incredibly difficult to
+validate whether an application is doing what it's supposed to be doing, both
+for debugging tools and for a shader author to reason about it in the first
+place.
+
+In addition to this, some APIs now provide methods to treat descriptors as
+plain old data (POD), even interleaving descriptors and other POD types in
+the heap. (E.g.
+[PSSL](https://gdcvault.com/play/1024241/Higher-Res-Without-Sacrificing-Quality)
+[see 11 minutes in].)
+
+While regular POD types are currently off the table for DirectX 12 in its
+current iteration of resource heaps, enabling stronger typing on the heap
+would lay groundwork for exposing interleaved POD in future.
+
+Notably, PSSL as linked above also enables resources to be accessed freely
+from any source; allowing nested structures and arbitrary handling of data.
+Until or unless hardware is modified to allow descriptors in arbitrary
+memory, handling actual descriptors in this way is not possible portably.
+However, it should be possible to better enable passing handles to the heap
+around by modifying the type system to handle resources more consistently.
+
+
+## Proposed solution
+
+### Consistent Resource Type Handling
+
+This change proposes that resource type declarations are _always_ considered
+as 32-bit offsets into the heap, allowing them to be declared in arbitrary
+external memory.
+This includes storing them in buffers of indeterminate size, allowing fairly
+arbitrary nesting of resource declarations.
+
+Example declarations:
+
+```hlsl
+// A constant buffer containing other resource indices, including a buffer that points to further resources.
+struct SomeResources {
+	Texture2D	texture;
+	RWBuffer<Texture2D> buffer;
+};
+
+ConstantBuffer<SomeResources> someResources;
+
+// RW Buffer; resource offsets can be written out too!
+RWBuffer<Texture2D> buffer;
+```
+
+Declaring these is equivalent to declaring integer values and passing those
+values to `ResourceDescriptorHeap` with the same resource type.
+This serves primarily to make accessing the heap and reasoning about
+descriptors in the heap easier, rather than passing loose integers around.
+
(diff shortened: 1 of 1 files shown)


proposals/0TBD-improved-resource-typing.md
```diff
@@ -0,0 +1,529 @@
...
+descriptor VA:
+
+ * `RawConstantBuffer<typename T>`
+ * `RawStructuredBuffer<typename T = float4, uint64_t Size = UINT64_MAX>`
+ * `RawRWStructuredBuffer<typename T = float4, uint64_t Size = UINT64_MAX>`
+ * `RawByteAddressBuffer<uint64_t Size = UINT64_MAX>`
+ * `RawRWByteAddressBuffer<uint64_t Size = UINT64_MAX>`
+ * `RawRasterizerOrderedBuffer<typename T = float4, uint64_t Size = UINT64_MAX>`
+ * `RawRasterizerOrderedByteAddressBuffer<uint64_t Size = UINT64_MAX>`
+ * `RawRaytracingAccelerationStructure`
```
jenatali, line 336: Capturing feedback from myself and others: The usage of "raw" is a quite overloaded term, with the D3D *API* taking a "raw buffer" to be the equivalent of an HLSL `ByteAddressBuffer`. We need a different term for that.

Tobski: Yea, this was always a placeholder anyway. There's also some overlap with what LLVM calls "raw" which might confuse anyone working on clang.

I am wildly open to suggestions here.

Tobski: The only idea that comes to my mind, which I don't love, is "VAConstantBuffer", since you know, it's come from a VA. Maybe we call that one the backup plan. It's probably better than "Raw" though.

Tobski: I've made the change to call them "VA\<type>" instead, we can still haggle on the name, but at least it's not "Raw" anymore.

Tobski:
Some external discussion has been thrown around about type handling, and I've retooled `ResourceEntry` to be a type now, with a 32-bit in-memory representation just like resource types are proposed as. The option for the function is still there, but under an open issue, with rationale for why it's not part of the proposal anymore.

Tobski:
Also just added an issue about bounds checking and VA buffer sizing as this has also come up in conversation repeatedly.

Jasper-Bekkers:
> Some external discussion has been thrown around about type handling, and I've retooled `ResourceEntry` to be a type now, with a 32-bit in-memory representation just like resource types are proposed as. The option for the function is still there, but under an open issue, with rationale for why it's not part of the proposal anymore.

Since this is a 32-bit valued type, can I construct this type from a uint myself? For example, we have this in our API and we bitpack the equivalent of this into our `GpuMaterial` struct together with a flat color. Since this proposal very closely matches our bindless rendering blog posts, I'd like to make sure we can still at least get that behavior (https://blog.traverseresearch.nl/bindless-rendering-setup-afeb678d77fc and co).

For us, this type is modeled with the following core features:

 - `RenderResourceHandle::unsafeFromUint(uint handle)` this disables all validation, but makes it so we can construct this type from any uint.
 - `RenderResourceHandle::invalid()`
 - `RenderResourceHandle::isValid()`

And the following validation features

 - `RenderResourceHandle::resourceTag()` (texture, buffer, tlas)
 - `RenderResourceHandle::isWritable()`
 - `RenderResourceHandle::version()`  (bumped if the slot gets recycled)
 - `RenderResourceHandle::readIndex()` (points to SRV)
 - `RenderResourceHandle::writeIndex()` (points to UAV)

Tobski:
> > Some external discussion has been thrown around about type handling, and I've retooled `ResourceEntry` to be a type now, with a 32-bit in-memory representation just like resource types are proposed as. The option for the function is still there, but under an open issue, with rationale for why it's not part of the proposal anymore.
> 
> Since this is a 32-bit valued type, can I construct this type from a uint myself? For example, we have this in our API and we bitpack the equivalent of this into our `GpuMaterial` struct together with a flat color. Since this proposal very closely matches our bindless rendering blog posts, I'd like to make sure we can still at least get that behavior (https://blog.traverseresearch.nl/bindless-rendering-setup-afeb678d77fc and co).
> 
> For us, this type is modeled with the following core features:
> 
>     * `RenderResourceHandle::unsafeFromUint(uint handle)` this disables all validation, but makes it so we can construct this type from any uint.
> 
>     * `RenderResourceHandle::invalid()`
> 
>     * `RenderResourceHandle::isValid()`
> 
> 
> And the following validation features
> 
>     * `RenderResourceHandle::resourceTag()` (texture, buffer, tlas)
> 
>     * `RenderResourceHandle::isWritable()`
> 
>     * `RenderResourceHandle::version()`  (bumped if the slot gets recycled)
> 
>     * `RenderResourceHandle::readIndex()` (points to SRV)
> 
>     * `RenderResourceHandle::writeIndex()` (points to UAV)

Right now there's an open issue about this - the issue with providing a "let me construct it" function is just that it enables developers to not construct it until the last second; making it hard for a compiler (or more importantly debugger) to reason about - but this is a good use case that I think is justification enough to add the constructor. I'm mostly just wary of people misusing it.

Tobski:
I do like that you have "unsafe" on the direct construction though, maybe it's worth doing the same.

Tobski:
@Jasper-Bekkers I've re-added the construction of resource entries from indices, so your use case should be covered.

PR #466, opened 2025-03-29
Remove validation mask language, update spec given implementation knowledge

bob80905:
This PR updates the spec given a rear-view perspective after having implemented the OMM feature. It removes undesired language concerning validation masks, and solidifies the diagnostic checking approach to the availability attribute mechanism, while making some small nits here and there.

It should be noted that RDAT information pulled from HLSL Subobjects is tested, but not using the D3DReflect part of FileCheckerTest. This is because the subobject data is lost due to how filecheck processes the compiled output. We instead run tests by checking the subobject dump in the disassembly comments. So, language referring to testing the feature using D3Dreflect has been changed to more accurately reflect the alternative methods used in testing the feature.

### DIFF
--- a/proposals/0024-opacity-micromaps.md
+++ b/proposals/0024-opacity-micromaps.md
@@ -78,7 +78,7 @@ independent of raytracing pipelines. For `RayQuery`, the template for
 instantiating the object includes a new optional `RAYQUERY_FLAGS` parameter:
 
 ```
-template<uint constRayFlags, uint constRayQueryFlags = RAYQUERY_FLAG_NONE>
+template<uint StaticRayFlags, uint RayQueryFlags = RAYQUERY_FLAG_NONE>
 class RayQuery;
 ```
 
--- a/proposals/0024-opacity-micromaps.md
+++ b/proposals/0024-opacity-micromaps.md
@@ -323,6 +323,8 @@ See [Opacity Micromaps][dxr-omm] in the Raytracing spec for details.
   it through to appropriate DXIL operation arguments.
 - Test optional template argument to RayQuery generates appropriate DXIL
   opreation and that the optional flag makes it through to the argument list.
+- Test that the new flag value, `RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS`,
+  appears in the `RaytracingPipelineConfig1` subobject's `Flags` field.
 
 ### Diagnostics
 
@@ -379,7 +381,6 @@ reachable in the current compilation.
 ## Acknowledgments (Optional)
 
 * Amar Patel
-* Tex Riddell
 * Josh Batista
 
 <!-- External References -->
--- a/proposals/0024-opacity-micromaps.md
+++ b/proposals/0024-opacity-micromaps.md
@@ -60,7 +60,6 @@ A new `RAY_FLAG` is proposed that allows ray tracing in shaders to force OMM
 
 Add a built-in `RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS` to the
 `RAYTRACING_PIPELINE_FLAG` flags in HLSL to enable OMM.
-Add built-in `RAY_FLAG_FORCE_OMM_2_STATE` to `RAY_FLAG` flags in HLSL.
 
 `RAYTRACING_PIPELINE_FLAG` flags are used in the
 [Raytracing pipeline config1][pipeline-config]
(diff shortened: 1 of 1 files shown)


proposals/0024-opacity-micromaps.md
```diff
@@ -60,7 +60,6 @@ A new `RAY_FLAG` is proposed that allows ray tracing in shaders to force OMM
 
 Add a built-in `RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS` to the
 `RAYTRACING_PIPELINE_FLAG` flags in HLSL to enable OMM.
-Add built-in `RAY_FLAG_FORCE_OMM_2_STATE` to `RAY_FLAG` flags in HLSL.
```
tex3d, line 63: Why was this removed?

bob80905: Because it's repeated on line 97, and this section seems to focus on RAYTRACING_PIPELINE_FLAG

proposals/0024-opacity-micromaps.md
```diff
@@ -75,11 +74,11 @@ overall.  See the D3D12 flag definition for more details:
 `RayQuery` objects, used with [RayQuery::TraceRayInline()][rq-trace], 
 also need to be aware that OMMs may be in use.  The above `RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS` 
 doesn't apply for inline raytracing however. `RayQuery` objects are 
-independent of raytracing pipelines. For `RayQuery` the template for 
+independent of raytracing pipelines. For `RayQuery`, the template for 
 instantiating the object includes a new optional `RAYQUERY_FLAGS` parameter:
 
 ```
-template<uint StaticRayFlags, uint RayQueryFlags = RAYQUERY_FLAG_NONE>
```
tex3d, line 82: I think this change is unnecessary, and potentially worse.  The reason for "Static" in "StaticRayFlags" is to say that the portion of ray flags that are static can be specified here, since you normally also supply ray flags at the TraceRayInline call (it differentiates the subset of ray flags supplied here).  This is the only place for RayQuery object flags, so I don't think it needs the same differentiation.  Since they are template arguments, they must be constant, but I don't think it's necessary to start each name with "const" here.

bob80905: Ok, I changed the template param names back.

proposals/0024-opacity-micromaps.md
```diff
@@ -325,13 +323,6 @@ See [Opacity Micromaps][dxr-omm] in the Raytracing spec for details.
   it through to appropriate DXIL operation arguments.
 - Test optional template argument to RayQuery generates appropriate DXIL
   opreation and that the optional flag makes it through to the argument list.
-- Use D3DReflect test to verify new flag value in `RaytracingPipelineConfig1`
-  subobject `Flags` field.
```
tex3d, line 329: The issue isn't quite that we can't check the flag in the RDAT.  When you check for the flag in the disassembly, it's only able to print the subobjects by decoding the RDAT at that point.  So by checking the disassembly, you are verifying that it's in the RDAT.  Just not from a D3DReflect RDAT dump.

The problem is that for a D3DReflect test, the current pipeline of parts constructed from the RUN line under the FileCheckerTest pass data from the compilation part to the next part in the form of the disassembled IR, rather than the original binary container output.  The subsequent D3DReflect stage has to re-assemble a shader from the disassembly before use.  While the disassembly contains a comment dumping subobjects from the RDAT, these are just in a comment, not in metadata that could be picked up by the assembler.  Therefore, subobjects are lost in transition.

proposals/0024-opacity-micromaps.md
```diff
@@ -371,49 +362,25 @@ subobject when the shader model is less than 6.9.
...
-the valid mask in the compiler?  We have a custom check for arguments such as
-the dimensions passed to `vector` and `matrix` templates.  It was suggested
-that we might be able to implement this with some sort of static assert
-instead.  However, I'm concerned that this approach may unintentionally impact
-unreachable uses of `RayQuery`.
-
 ## Acknowledgments (Optional)
 
 * Amar Patel
+* Tex Riddell
```
tex3d, line 382: No need to put me here, as I'm already the original author at the top.

tex3d (commented):
Primarily, I'd adjust the testing wording for verifying that the new flag is in the `RaytracingPipelineConfig1` subobject when specified in HLSL.

While we can't check it with D3DReflect (without an infrastructure change), checking it in the disassembly is equivalent, since that is printing the RDAT subobjects.

proposals/0024-opacity-micromaps.md
bob80905: Ok, for the first bullet point, I've reworded the way we express that a test was used to verify the presence of the new flag in the flags field.

tex3d (approved):
Looks good.

amarpMSFT (approved):

damyanp (approved):

PR #474, opened 2025-04-03
[0027] SER validation cleanup

### LINKED ISSUE #444: [0027] Clean up validation/diagnostic requirements
https://github.com/microsoft/hlsl-specs/blob/224971df77fb051354e6db4d330a1e83ff2c2486/proposals/0027-shader-execution-reordering.md?plain=1#L1368-L1373

First, there is a section under each new DXIL op labelled `Validation errors:` that has a point like this:
```md
- Validate that `opcode` equals `HitObject_MakeMiss`.
```
I'm not sure what this is trying to say.  Common infrastructure is responsible for verifying the correct intrinsic definitions using the correct `OpCodeClass` that determines the common function name and signature, based on the DXIL `OpCode`.  This point looks unnecessary to me.

Second, there is a section under some new DXIL ops labelled `Validation warnings:`.  I don't think there is such a thing as a DXIL validation warning.  If these are intended to be applied as warning diagnostics on the HLSL intrinsic (even if they use `DefaultError`), they should be described on the HLSL intrinsic, using "diagnostics" instead.

An alternative location for the HLSL Diagnostics and DXIL Validation specifications would be under new sections `Diagnostic Changes` and `Validation Changes`, respectively.

**Additional context**
I commented on a PR where a "Validation warnings:" section was being added [here](https://github.com/microsoft/hlsl-specs/pull/382/files#r2008190073).

rasmusnv:
Fixes #444

### DIFF
--- a/proposals/0027-shader-execution-reordering.md
+++ b/proposals/0027-shader-execution-reordering.md
@@ -1306,6 +1306,7 @@ declare %dx.types.HitObject @dx.op.hitObject_TraceRay.PayloadT(
 Validation errors:
 - Validate the resource for `acceleration structure handle`.
 - Validate the compatibility of type `PayloadT`.
+- Validate that `payload` is a valid pointer.
 
 #### HitObject_FromRayQuery
 
@@ -1317,6 +1318,9 @@ declare %dx.types.HitObject @dx.op.hitObject_FromRayQuery(
 
 ```
 This is used for the HLSL overload of `HitObject::FromRayQuery` that only takes `RayQuery`.
+
+Validation errors:
+- Validate that `ray query` is a valid ray query handle.
 
 #### HitObject_FromRayQueryWithAttrs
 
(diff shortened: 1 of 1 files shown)


damyanp (approved):
LGTM, but would like @tex3d to review (and happy for him to merge it if he approves).

tex3d (approved):
My only concern is whether we wanted to move items from "validation warnings" to Sema diagnostics on the intrinsic calls.  That was the intent of the following statement in the issue:

> If these are intended to be applied as warning diagnostics on the HLSL intrinsic (even if they use DefaultError), they should be described on the HLSL intrinsic, using "diagnostics" instead.

For now, I think this change is fine, but we might want to resurrect some of these warnings in a Sema diagnostic warning form, if we think they will help shader authors.

tex3d (changes requested):
Apologies for retracting my last review, I had looked at this before, and could have sworn it didn't remove RayQuery, payload, and attribute pointer validation.  I'm not sure why that change was applied.

These certainly are not part of addressing #444, so I think we should revert that additional change.

tex3d (approved):
In order to expedite review turnaround, I've reverted the extra change to remove RayQuery and pointer validation, which wasn't discussed in the attached issue.

If we need changes there, we should put up a separate PR and discuss.

PR #486, opened 2025-04-11
[0033] Assign DXIL 1.9 proposal number 0033

damyanp:
I hit complete too soon and forgot to update the proposal number.


proposals/0033-dxil19.md
tex3d, line 1: There should be a change to content to update the NNNN as well.

damyanp: Thanks!  More haste, less speed....

tex3d (approved):

Issue #492, opened 2025-04-17
[0029][0031] Specify MatrixLayout for OuterProductAccumulate should be OuterProductOptimal


anupamachandra:
**Which proposal does this relate to?**
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0031-hlsl-vector-matrix-operations.md

**Describe the issue or outstanding question.**
Update the spec enforcing that only OuterProductOptimal matrix layout is supported for OuterProductAccumulate

**Additional context**
Optimal Layouts are opaque and implementation defined. Implementations could define layouts that would enable efficient access patterns for atomics for the accumulate operation

PR #494, opened 2025-04-18
[0029][0031]Add Matrix Layout constraint to OuterProductOptimal for OuterProductAccumulate function

anupamachandra:
https://github.com/microsoft/hlsl-specs/issues/492

### DIFF
--- a/proposals/0029-cooperative-vector.md
+++ b/proposals/0029-cooperative-vector.md
@@ -279,7 +279,7 @@ that the input vector is an unsigned integer.
   following `ComponentType`s: `I16`, `U16`, `I32`, `U32`, `F16`, `F32`, `U8`,
   `I8`, `F8_E4M3`, `F8_E5M2`, 
 
-### Vector Outer Product
+### Vector-Vector Outer Product and Accumulate
 
 #### Syntax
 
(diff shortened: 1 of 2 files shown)


proposals/0029-cooperative-vector.md
```diff
@@ -312,8 +312,9 @@ The two input vectors are specified via **input vector 1** and **input vector
 
 The matrix is accumulated to the writeable raw-buffer specified by **matrix
 resource**, with **matrix offset**, **matrix interpretation**, **matrix
-layout**, and **matrix stride** behaving as described
-[above](#matrix-vector-multiply-and-multiply-add-operations).
+layout**, and **matrix stride** behaving as described 
+[above](#matrix-vector-multiply-and-multiply-add-operations), with the constraint
+that **matrix layout** must be `DXILMatrixLayout::OuterProductOptimal`.
```
damyanp, line 317: Rather than trying to pack all of this into one sentence, I suggest adding details about the constraints on matrix layout below.  There's always paragraphs talking about constraints on the address of the resource.

I also notice that it refers to the matrix stride needing to be 16-byte aligned, which isn't compatible with requiring OutputProductOptimal layout.

damyanp (changes requested):

proposals/0031-hlsl-vector-matrix-operations.md
```diff
@@ -665,6 +667,11 @@ void OuterProductAccumulate(
 } // namespace dx
 ```
 
+Validation:
```
tex3d, line 670: If we are talking about HLSL diagnostics, we should say "Diagnostics" not "Validation".  We use "Validation" to refer to DXIL Validation.

tex3d (approved):

proposals/0029-cooperative-vector.md
anupamachandra: Right good point, the matrix stride field itself is currently present only for future proofing for a higher tier that might support Row/column layouts for OPA. Does an update with "matrix stride should be '0' and will be ignored" for TIER_1_1, as this tier only support training_optimal layout for this function" sound ok?

damyanp: Maybe something like "Since only optimal layouts are currently supported for this operation, matrix stride must be 0"?

damyanp (approved):

PR #499, opened 2025-04-30
First pass at validator backcompat testing proposal

damyanp:
We want to make the internal validator the default one, but this impacts how we test that we haven't broken DXIL.  This proposal suggests a way to move these tests into the public github and how to modify DXC and the build system to run the tests.

### DIFF
--- a/proposals/infra/INF-NNNN-validator-backcompat-testing.md
+++ b/proposals/infra/INF-NNNN-validator-backcompat-testing.md
@@ -1,0 +1,111 @@
+<!-- {% raw %} -->
+
+# Feature name
+
+## Instructions
+
+* Proposal: [INF-NNNN](INF-NNNN-validator-backcompat-testing.md)
+* Author(s): [Damyan Pepper](https://github.com/damyanp)
+* Sponsor: [Damyan Pepper](https://github.com/damyanp)
+* Status: **Under Consideration**
+* Impacted Project(s): DXC
+
+## Introduction
+
+We propose moving the DXC validator backwards compatability tests from a
+private, internal, repo to the public one, cleaning up the mechanisms that
+enable this as we go.
+
+## Motivation
+
+Currently, internal builds of DXC perform various tests that involve running DXC
+against older versions of the validator (dxil.dll). This is to ensure that
+modifications to DXC (including modifications to the validator) don't break
+backwards compatability with previously released versions. In the past, the
+validator was closed source and so these tests were run from internal builds.
+
+We've recently [open sourced the validator](./INF-0004-validator-hashing.md) and
+as part of this work we want to make DXC default to using the built-in,
+"internal", validator rather than one loaded from a dll, and this breaks these
+tests. 
+
+While we're fixing these tests, we're going to take this opportunity to move the
+back-compat testing into the public repo and simplify the infrastructure
+required to run them.
+
+
+## Proposed solution
+
+### How it works currently
+
+These tests only run for Windows x64 builds. The internal build pipeline runs
+`mm test --validator-backcompat-all`. This script has a list of old dxil.dll
+versions to test. The DLL's themselves are checked into the repo. Before running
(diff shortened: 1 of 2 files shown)


proposals/infra/INF-NNNN-validator-backcompat-testing.md
```diff
@@ -0,0 +1,147 @@
...
+A new option, `-dxil-dll-path`, can be used to specify the full path to dxil.dll
+to use. Example path: `c:\dxc\build\old-releases\1.7.123\bin\x64\dxil.dll`.
+
+Setting the path to an empty string forces the internal validator to be used.
+
+If a dll path is specified and isn't found then the compiler will fail with an
+error.
+
+When invoked via dxc.exe (ie not through the compiler API), DXC will check for a
+new environment variable, `DXC_DXIL_DLL_PATH`, that  specifies the full path to
```
bob80905, line 91: If the environment variable's value is also an empty string, will it default to internal validation behavior?

proposals/infra/INF-NNNN-validator-backcompat-testing.md
```diff
@@ -0,0 +1,147 @@
...
+
+A new option, `-dxil-dll-path`, can be used to specify the full path to dxil.dll
+to use. Example path: `c:\dxc\build\old-releases\1.7.123\bin\x64\dxil.dll`.
+
+Setting the path to an empty string forces the internal validator to be used.
+
+If a dll path is specified and isn't found then the compiler will fail with an
+error.
+
+When invoked via dxc.exe (ie not through the compiler API), DXC will check for a
```
bob80905, line 90: Even if it isn't necessary in the proposal itself, could you describe why the design decision is being made that the environment variable should only be checked when invoked via dxc.exe, not the compiler API?

tex3d (approved):
I think generally this is a good outline.  I have some concern about not being able to specify loading a DXIL.dll in the same path as the currently used DxCompiler.dll in a relative way.

tex3d:
Oh yeah, I forgot to comment on something else I had thought of.  You can override the environment variable with an explicit argument, but how do you choose the internal validator this way?

proposals/infra/INF-NNNN-validator-backcompat-testing.md
```diff
@@ -0,0 +1,111 @@
...
+
+Rather than check the binaries in to source control, we can download them from
+github.com as zip files. `ExternalProject_Add` can be used to do this.
+
+### Specifying which DXIL.dll to use
+
+The `-select-validator` option will be deprecated.  As this is not documented,
+it should be ok to just remove it entirely.
+
+A new option, `-dxil-dll-path`, can be used to specify the full path to dxil.dll
```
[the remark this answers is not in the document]
damyanp: My understanding here is that there are no scenarios where an end-user of dxc or the API would be using an external dxil.dll validator - they'd always use the one compiled in to dxcompiler.dll

Selecting dxil.dll is purely something we use internally for validating changes to dxc.

proposals/infra/INF-NNNN-validator-backcompat-testing.md
damyanp: Yes, unless the command-line parameter is specified.

damyanp:
> Oh yeah, I forgot to comment on something else I had thought of. You can override the environment variable with an explicit argument, but how do you choose the internal validator this way?

Something like `-dxil-dll-path ""` should do it.

proposals/infra/INF-NNNN-validator-backcompat-testing.md
damyanp: Text has been updated to include this.

bob80905 (changes requested):
I can approve this for now, though we will likely overhaul most of this with an internal spec with the most up-to-date details.
I'll approve after the proposal number is set in stone (line 7 and the file name)

bob80905 (approved):

PR #502, opened 2025-05-15
[0029] Fix packed types in min support table

damyanp:


anupamachandra:
LGTM!

proposals/0029-cooperative-vector.md
```diff
@@ -763,7 +763,8 @@ explicitly checked for the combinations below.
 | F16         | F16                 | F16                  | F16                | F16        |
 | F16         | F8_E4M3             | F8_E4M3              | F16                | F16        |
 | F16         | F8_E5M2             | F8_E5M2              | F16                | F16        |
-| PackedS8x32 | I8                  | I8                   | I32                | I32        |
+| PackedS8x32 | PackedS8x32         | I8                   | I32                | I32        |
+| U32         | PackedS8x32         | I8                   | I32                | I32        |
```
tex3d, line 767: Separating these lines makes it seem like they are different configurations with different behaviors, such as by converting U32s to PackedS8x32, but that's not the case.

Would it be possible to say something in that column like: `PackedS8x32 (U32)`?

tex3d: Or is the InputType exposed in the FeatureInfo API, so needs an entry for each?

tex3d (approved):

alsepkow (approved):
lgtm

PR #505, opened 2025-05-27
[New] Initial draft for memory and execution model

llvm-beanz:
This is the start of a proposal for the memory and execution model for HLSL.

### DIFF
--- a/proposals/0050-mem-and-exec-model.md
+++ b/proposals/0050-mem-and-exec-model.md
@@ -1,12 +1,14 @@
-<!-- {% raw %} -->
+---
+title: 0050 - Formalized Memory and Execution Model
+params:
+  authors:
+  - llvm-beanz: Chris Bieneman
+  sponsors:
+  - llvm-beanz: Chris Bieneman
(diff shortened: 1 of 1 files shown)


proposals/NNNN-mem-and-exec-model.md
```diff
@@ -0,0 +1,164 @@
...
+  GroupMemoryBarrierWithGroupSync(); // sync point!
+  if (X == 0)
+    X = 2; // How many threads are in the tangle here?
+}
+```
+
+One problem with this model is not having clearly defined memory ordering which
+can impact tangle formation. If a thread is allowed to execute the second `if`
+body before all threads have finished evaluating the condition, tangle formation
+becomes unintuitive.
```
spall: is this not just unintuitive but also undefined?

llvm-beanz:
> Could you use 51 please? It looks like this #557 is in the process of being completed with 50 and has a bunch of issues filed against it with that number already, which means that it would be more effort to update that one than this one.

Renumbering issues is easier. I'll fix them.

PR #512, opened 2025-06-09
[Shader Execution Reordering] Minor spec inconsistency: HitObject::MakeMiss missing Ray in argument listing

amarpMSFT:
The function signature is correct, but the definition list was missing the Ray argument.


amarpMSFT:
See this issue: https://github.com/microsoft/DirectX-Specs/issues/204
D3D DXR spec has been fixed.

damyanp (approved):

PR #514, opened 2025-06-13
[0029] Rename API call ConvertLinearAlgebraMatrix to ConvertLinearAlgebraMatrices

anupamachandra:
ConvertLinearAlgebraMatrix call allows conversion of several matrices in a single call, reflect that in the call name.


anupamachandra:
@amarpMSFT Should we change `ID3D12DevicePreview` to `ID3D12DeviceNN` as well?

amarpMSFT (approved):

Issue #50, opened 2023-04-26
[004] Are struct and array layouts specified by HLSL?


s-perron:
**Which proposal does this relate to?**
[Unions](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0004-unions.md)

**Describe the issue or outstanding question.**

To implement unions, SPIR-V will have to know the size of the union. I do not know if that is specified by HLSL or if the compiler can decide for itself. Depending on which features are going to be used, the entire layout must be specific.

**Additional context**
See #49 for the limitations on how to represent unions in spir-v.

To create a variable that is large enough to hold the entire union, the size of the union has to be known by the compiler.

Related to this, consider the example from the proposal:

```
struct vector {
  union {
    struct {
      float x;
      float y;
      float z;
      float w;
    };
    struct {
      float r;
      float g;
      float b;
      float a;
    };
    float f[4];
  };
};

vector v;
```

Can the user assume that `v.x`, `v.r`, and `v.f[0]` are the same memory location? If so, where in the HLSL spec does it say that the members of the struct and the elements of the array must be at the same offset? I cannot find it, but my search was not extensive.

s-perron:
FYI: @llvm-beanz

llvm-beanz:
The Unions spec is not fully specified, but it does say:
> Following C++ Unions are always sized large enough to contain their largest member type.

In practice, the `sizeof(T)` for a union is the size of the largest union member.

We do need to specify that and the memory layout in the spec. Each union member's layout starts at the start of the union, but that doesn't necessarily guarantee strict overlapping. In the example you provided in the issue the fields all overlap, but padding, alignment or other constraints can impact how they overlap.

s-perron:
Based on what you said above, I would expect the following to be undefined:

```
float func() {
  vector v;
  v.a = 1.0;
  return v.f[3]; // Undefined because it is not specified how the struct and the array overlap.
}
```

However, this statement in the proposal (emphasis mine) seems to suggest that you expect to be able to do this.

> One common case is when the **layout of data matches between two data types** and the user wants to be able to access the data interchangeably as the two types.

llvm-beanz:
I think you're misunderstanding (or I'm mis-speaking). First we need to answer #43 to define whether or not punning is allowed. By C++ rules, punning is always undefined behavior regardless of the layouts (which are fairly well defined).

[...]

Issue #318, opened 2024-09-12
`bit_cast` template


llvm-beanz:
**Is your feature request related to a problem? Please describe.**
HLSL has a bunch of different casting builtins that are all spelled differently and not templated. That makes it difficult for template code to cast to the appropriate type if the type is a template parameter.

**Describe the solution you'd like**
It would be awesome to add a templated `bit_cast` operation. Such a casting operation was added as an implementation detail in:

https://github.com/llvm/llvm-project/pull/107292

I suspect we could create a similar header-implemented interface for DXC as well using type traits.

devshgraphicsprogramming:
OUr implementations https://github.com/Devsh-Graphics-Programming/Nabla/blob/edd7e6242a3f07bbed957cb9d7b8778c547037be/include/nbl/builtin/hlsl/bit.hlsl#L34

damyanp:
(related to internal 30312655)

PR #520, opened 2025-06-24
[0004] Adding spec language for unions

### LINKED ISSUE #38: [0004] Update union spec to clarify casting behaviors
[0004-unions](proposals/0004-unions.md)

As per the discussion in #37, C++ union rules actually might not really be what we want for HLSL.

> The C++ standard states the following in §9.5.1:
> 
> In a union, at most one of the non-static data members can be active at any time, that is, the value of at most one of the non-static data members can be stored in a union at any time. [...] The size of a union is sufficient to contain the largest of its non-static data members. Each non-static data member is allocated as if it were the sole member of a struct. All non-static data members of a union object have the same address.

We should elaborate the unions spec to clarify some of these behaviors one way or the other.

**Additional context**
[`vk::BufferPointer<T,A>` review](https://github.com/microsoft/hlsl-specs/pull/37)
[Related `vk::BufferPointer<T,A>` issue](https://github.com/microsoft/hlsl-specs/issues/43)

### LINKED ISSUE #523: [0004] Finalize detailed design
[Unions](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0004-unions.md)

Detailed design needs finalization.

llvm-beanz:
This fleshes out the language design for union types.

Resolves #38, resolves #523

### DIFF
--- a/proposals/0004-unions.md
+++ b/proposals/0004-unions.md
@@ -99,6 +99,8 @@ non-static data member begins when the member is initialized, and ends either
 when the value is destroyed via an explicit destructor call, a different
 non-static data member is initialized, or the union's lifetime ends.
 
+A union may not contain any non-static data members of intangible type.
+
 If a union contains more than one standard-layout structures which share a
 common initial sequence of non-static members, the common initial sequence may
 be accessed through any non-static union member containing the sequence.
--- a/proposals/0004-unions.md
+++ b/proposals/0004-unions.md
@@ -65,14 +65,75 @@ specific clarifications.
 
 * Following C++ Unions are always sized large enough to contain their largest
   member type.
-* Union members cannot have HLSL semantics applied to them.
+* Union members and objects cannot have HLSL semantics applied to them.
+  * As such, unions may not be used as input arguments to shaders.
+* Union objects may not be stored in constant buffers.
 * Union objects in buffer layouts are aligned as elements of their
   most-restrictive member type.
 * Union objects in struct layouts behave as defined in C++.
-* Union objects cannot have semantics applied to them.
 * Union's cannot have user-defined constructors or destructors until the
   language supports them for other user defined data types (See:
-  [0007](0007-constructors.md)).
+  [0032](0032-constructors.md)).
+
+## Detailed Design
+
+This proposal depends on the adoption of [Strict Initializer
+Lists](0005-strict-initializer-lists.md) because initialization of a union type
+is incompatible with HLSL's object initialization rules.
+
+> Note: With current rules which member of the union should be initialized is
+> ambiguous in all cases and unions with differing numbers of scalar members
+> would result in the whole initializer list being ambiguous.
+
+HLSL unions will adopt the behavior defined in the [ISO C++ specification
+(ISO/IEC 14882:2011)](https://timsong-cpp.github.io/cppwp/n3337/class.union),
+except with the following additions and clarifications.
+
+### Proposed Spec Language
+
+A union is a class defined with the `union` keyword.
+
+Non-static members of a union share overlapping storage. As such, a union may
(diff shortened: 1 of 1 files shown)


llvm-beanz:
I think this PR also resolves https://github.com/microsoft/hlsl-specs/issues/50, but would like @s-perron's perspective.

The PR does assume that we're going to elsewhere in the spec define "standard layout", but I expect that to basically match C++'s definition with the additional constraint that a standard layout type cannot be intangible.

proposals/0004-unions.md
```diff
@@ -65,14 +65,77 @@ specific clarifications.
...
+
+### Proposed Spec Language
+
+A union is a class defined with the `union` keyword.
+
+Non-static members of a union share overlapping storage. As such, a union may
+only store the value of one non-static data member at a time. The lifetime of a
+non-static data member begins when the member is initialized, and ends either
+when the value is destroyed via an explicit destructor call, a different
+non-static data member is initialized, or the union's lifetime ends.
```
s-perron, line 100: This would seems to disallow type punning, right? I thought we decided in previous discussions to allow it.

Keenuts: Case to consider for type punning: if we explicitly disallow it, but "support it" the same way C++ tolerates it, we need to handle a bool/int union:
LLVM used `int1`, but in SPIR-V, a boolean value has no defined bit pattern, meaning we could have `((bool)union.int) != union.bool`.

llvm-beanz: This language aligns with the C++ specification, which is to say type-punning is defined if there is an overlap in initial "standard layout" members, but undefined otherwise.

This almost exactly matches the C++ behavior, which I think is what we need to do if we're going to allow unions to have constructors and destructors.

Keenuts: So in this case, this means AFAIU than the compiler must support type punning with:
```hlsl
union {
    bool a;
    int  b;
};
```

In SPIR-V, int-to-bool conversion requires specific instructions to convert from an int to a bool when it occurs (`OpSelect` and `OpINotEqual`) since `0b0` could mean `TRUE`.

This might require to always store booleans in unions as integers, and always emit the proper conversion at the interface. Might need some additional thought on how to handle unions in resources

llvm-beanz: No, that's not true. Given your union above, in C++ it is undefined behavior to set `a` and access `b` or vice versa. The relevant section of the C++ 20 specification is:

>At most one of the non-static data members of an object of union type can be active at any time, that is, the value of at most one of the non-static data members can be stored in a union at any time.

[Source](https://timsong-cpp.github.io/cppwp/n4868/class.union#general-2)

It may still make sense to store booleans as integers (it's actually kinda silly that SPIRV has a boolean type), but in C++ this is UB, and I'm proposing the same in HLSL.

Fun fact: it is also UB in C++ for a boolean to store any value other than `true` or `false`.

bogner: FWIW I think if we want an escape hatch so that people can type pun legally we should go the route of C++20's [std::bit_cast](https://en.cppreference.com/w/cpp/numeric/bit_cast.html), which was introduced almost entirely to work around the fact that the memcpy solution isn't useable in `constexpr` contexts, but really makes sense in any case where memcpy isn't appropriate.

TLDR, I agree with the direction to match C++ and make arbitrary type punning through unions undefined.

s-perron: Sounds good to me.

llvm-beanz: We do also have an issue tracking `bit_cast` (https://github.com/microsoft/hlsl-specs/issues/318). It's actually already implemented in Clang as an implementation detail for HLSL's bitcasting intrinsics (e.g. `asuint`, `asfloat`, etc).

PR #522, opened 2025-06-25
[0005] Adding spec language for strict initializer lists

### LINKED ISSUE #228: [0005] Strict Initializer Lists - Detailed Design
[0005 - Strict Initializer Lists](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0005-strict-initializer-lists.md)

This feature needs a completed detailed design. Since this feature implements a replacement for a non-conformant HLSL language feature that we may not want to implement in Clang, it would be nice to get this feature into HLSL 202x so that users can adopt the new feature in DXC before transitioning to Clang.

llvm-beanz:
This adds a detailed design including draft language spec wording for the strict initializer lists proposal.

Fixes #228

### DIFF
--- a/proposals/0005-strict-initializer-lists.md
+++ b/proposals/0005-strict-initializer-lists.md
@@ -80,7 +80,7 @@ of the HLSL specification.
 
 An _aggregate_ is a vector, matrix, array, or class which does not contain
 * user-declared or inherited constructors
-* non-public non-static data members, and
+* non-public non-static data members, or
 * non-public base classes
 
 For the purposes of aggregate initialization, anonymous bit-fields are not


proposals/0005-strict-initializer-lists.md
```diff
@@ -67,6 +67,86 @@ Fix-It will be backwards compatible to prior HLSL versions.
...
+HLSL 202y shall adopt specification language aligned with the [ISO C++
+specification (ISO/IEC
+14882:2011)](https://timsong-cpp.github.io/cppwp/n3337/dcl.init).
+
+The language in the next subsection will replace the **Decl.Init.Agg** section
+of the HLSL specification.
+
+### Specification Language
+
+An _aggregate_ is a vector, matrix, array, or class which contains
```
spall: maybe it would be better to move the 'no's to this line. 'contains no'

proposals/0005-strict-initializer-lists.md
```diff
@@ -67,6 +67,86 @@ Fix-It will be backwards compatible to prior HLSL versions.
...
+
+The subobjects of an aggregate have a defined _subobject order_. For vectors and
+arrays the order is increasing subscript order. For matrices it is increasing
+subscript order with the subscript nesting such that in the notation
+`Mat[M][N]`, the ordering is `Mat[0][0]...Mat[0][N]... Mat[M][0]...Mat[M][N]`.
+For classes the order is recursively the order of the base class, followed by
+member subobjects in declaration order. Anonymous bitfields are not included in
+the subobject ordering.
+
+When initialized by an initializer list the elements of the initializer list are
```
spall: When initialized by an initializer list,

proposals/0005-strict-initializer-lists.md
```diff
@@ -67,6 +67,86 @@ Fix-It will be backwards compatible to prior HLSL versions.
...
+                                 // [1][0] = 3,[1][1] = 4
+int y[3][2] = {{1,2}, {3}, {4}}; // [0][0] = 1, [0][1] = 2,
+                                 // [1][0] = 3, [1][1] = 0,
+                                 // [2][0] = 4, [2][1] = 0
+```
+
+Element initialization behaves as if an assignment-expression is evaluated
+from the initializer-clause to the initialized member. All implicit conversions
+are applied as appropriate to convert the initializer to the member type. If the
+assignment-expression is ill-formed and the element is a subobject, brace
```
spall, line 154: Could you give an example? I read this multiple times and don't fully understand.

proposals/0005-strict-initializer-lists.md
bogner: or even "... which does not contain"

proposals/0005-strict-initializer-lists.md
```diff
@@ -67,6 +67,86 @@ Fix-It will be backwards compatible to prior HLSL versions.
...
+
+The language in the next subsection will replace the **Decl.Init.Agg** section
+of the HLSL specification.
+
+### Specification Language
+
+An _aggregate_ is a vector, matrix, array, or class which contains
+* no user-declared or inherited constructors
+* no non-public non-static data members, and
+* no non-public base classes
```
bogner: The C++ [definition of an aggregate](https://timsong-cpp.github.io/cppwp/n3337/dcl.init#aggr-1) disallows base classes entirely. Why differ here?

proposals/0005-strict-initializer-lists.md
```diff
@@ -67,6 +67,86 @@ Fix-It will be backwards compatible to prior HLSL versions.
...
+
+The subobjects of an aggregate have a defined _subobject order_. For vectors and
+arrays the order is increasing subscript order. For matrices it is increasing
+subscript order with the subscript nesting such that in the notation
+`Mat[M][N]`, the ordering is `Mat[0][0]...Mat[0][N]... Mat[M][0]...Mat[M][N]`.
+For classes the order is recursively the order of the base class, followed by
+member subobjects in declaration order. Anonymous bitfields are not included in
+the subobject ordering.
```
bogner: I have a couple of concerns with these definitions. The wording that this is replacing from C++ is simply "subscript or member order", which by its lack of detail is precise while avoiding bringing up topics like standard layout classes. I see why we may want more wording here for matrices, but:
- The wording here for classes should probably be constrained to standard layout classes (I think this may currently be all classes in HLSL, but that could be a dangerous assumption going forward).
- If we're explicitly mentioning arrays and vectors we should probably explicitly mention classes and structs.
- I don't know what the comment about anonymous bitfields here is trying to say.

proposals/0005-strict-initializer-lists.md
```diff
@@ -67,6 +67,86 @@ Fix-It will be backwards compatible to prior HLSL versions.
...
+copy-initialized from the corresponding initializer-clause. If the
+initializer-clause is an expression which requires a narrowing conversion to
+convert to the subobject type, the program is ill-formed.
+
+An aggregate that is a class can also be initialized by a single expression not
+enclosed in braces.
+
+An array of unknown size may be initialized by a brace enclosed
+initializer-list. For arrays of unknown size the initializer list shall contain
+`n` initializers where `n > 0` and will produce an array containing `n`
```
bogner: Should we specifically call out that this is a one-dimensional array?
```suggestion
`n` initializers where `n > 0` and will produce a one-dimensional array containing `n`
```

proposals/0005-strict-initializer-lists.md
```diff
@@ -67,6 +67,86 @@ Fix-It will be backwards compatible to prior HLSL versions.
...
+elements.
+
+If an initializer-list contains more initializer-clauses than the number of
+subbjects being initialized, the program is ill-formed.
```
bogner:
```suggestion
subobjects being initialized, the program is ill-formed.
```

proposals/0005-strict-initializer-lists.md
```diff
@@ -67,6 +67,86 @@ Fix-It will be backwards compatible to prior HLSL versions.
...
+
+If an initializer-list contains fewer initializer-clauses than the number of
+subobjects being initialized, each member not explicitly initialized shall be
+initialized as if by an empty initializer list.
+
+If an aggregate class contains a subobject member that has no members in its
+subobject order, the initializer-clause for the empty subobject shall not be
+omitted from an initializer list for the containing class unless the
+initializer-caluses for all following members of the class are also omitted.
```
bogner: An example similar to the one in the C++ spec would be helpful here.

proposals/0005-strict-initializer-lists.md
llvm-beanz: Mostly that HLSL aggregates as they exist today can have a single base class (HLSL doesn't allow multiple inheritance).

proposals/0005-strict-initializer-lists.md
llvm-beanz: Standard layout is a really tricky notion in HLSL across the board, but this can't fit C++'s definition of standard layout, and it doesn't in C++ either.

Consider this code:

```c++
struct Base {
  int X;
};

struct Derived : Base {
  float F;
};
```

By C++'s definition of standard layout `Derived` is not standard layout, but it _is_ an aggregate.

> A standard-layout class is a class that:
> ...
> — either has no non-static data members in the most derived class and at most one base class with non-static data members, or has no base classes with non-static data members, and
> ...

To be a standard layout class in C++, _at most_ one class in the inheritance chain can have non-static data members. Despite not being standard layout, `Derived` does fit the definition of aggregate:

> An aggregate is an array or a class (Clause 9) with no user-provided constructors (12.1), no brace-or-equal-initializers for non-static data members (9.2), no private or protected non-static data members (Clause 11), no base classes (Clause 10), and no virtual functions (10.3).

The anonymous bitfield comment relates to something like this:

```c++
struct Doggo {
  uint32_t Legs : 2;
  uint32_t : 2; // reserve some bits
  uint32_t Tail : 1;
};

Doggo D = {4, 1}; // Note omitting the anonymous bitfield in the initializer.
```

C++ has similar wording:

> Static data members and anonymous bit-fields are not considered members of the class for purposes of aggregate initialization.

I'll simplify the ordering language, but I think the rest largely needs to stay.

proposals/0005-strict-initializer-lists.md
llvm-beanz: I'll add this to the spec too, but something like:

```hlsl
RWBuffer Buf;

struct ResourceRef {
  RWBuffer B;
  int Offset;
};

ResourceRef R = {0, Buf}; // Oh no! Parameters backwards!
```

Because an `RWBuffer` can't be initialized via assignment from an integer and an integer can't be initialized via assignment from an `RWBuffer` the initialization is ill-formed and fails.

llvm-beanz: Oh! Sorry I think I saw your comment on the wrong spot. This is an example for how brace elision works:

```c++
struct A {
    int X;
    float F;
};

struct B {
    A S;
    int Y; 
};

B T = {1,2,3}; // This is fine.

A U = 1; // This is ill-formed!
```

Basically when processing `B T = {1,2,3};` since the initializer `T.S = 1` is ill-formed, it assumes instead that you elided the braces, and initializes `T.S.X = 1` and `T.S.F = 2`. If the later expressions were ill-formed, the initialization would fail.

I suspect that emulating this odd carryover from C is part of how HLSL's behavior got so weird.

proposals/0005-strict-initializer-lists.md
```diff
@@ -67,6 +67,125 @@ Fix-It will be backwards compatible to prior HLSL versions.
...
+14882:2011)](https://timsong-cpp.github.io/cppwp/n3337/dcl.init).
+
+The language in the next subsection will replace the **Decl.Init.Agg** section
+of the HLSL specification.
+
+### Specification Language
+
+An _aggregate_ is a vector, matrix, array, or class which does not contain
+* user-declared or inherited constructors
+* non-public non-static data members, and
```
spall: should this be an or instead of an and?

proposals/0005-strict-initializer-lists.md
spall: thanks I think that explains my confusion.

PR #536, opened 2025-07-01
[0007] Adding spec language for const instance methods

### LINKED ISSUE #283: [0007] Finish proposal for const instance methods
[0007 Const Instance Methods](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0007-const-instance-methods.md)

This proposal is incomplete and should be completed for either 202x or 202y.

llvm-beanz:
This adds additional language for how the implicit object parameters are handled during overload resolution. The main key points in this language are:

1) No conversions may be applied to the implicit object parameter.
2) No temporary may be created to hold the implicit object argument.

Point 2 is what restricts binding a temporary to a const implicit object.

Closes #283


llvm-beanz:
Tagging @farzonl for review

bogner (approved):

PR #537, opened 2025-07-01
[0008] Adding spec language for overloading operators

### LINKED ISSUE #524: [0008] Finish detailed design - Non-member operator overloading
[Non-member operator overloading](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0008-non-member-operator-overloading.md)

Finish detailed design with draft specification language.

llvm-beanz:
This langauge clarifies that operators can be both non-static member functions and non-member functions.

Closes #524


proposals/0008-non-member-operator-overloading.md
```diff
@@ -44,6 +44,53 @@ under HLSL 2021 as an extension. Using the feature in HLSL 2021 mode as an
...
+    \terminal{\[\]}\br
+\end{grammar}
+```
+![Latex Rendering](0008-assets/OperatorGrammar.png)
+
+An _operator function_ is a function declared with an _operator-function-id_ as
+its name. An _operator function template_ is a function template
+operator-function-id as its name.
+
+Both the binary and unary forms of `+`, `-`, `*` and `&` may be overloaded.
```
farzonl: the picture `OperatorGrammar.png` declares more operator than you have listed here. Does that mean all the other operators are not overloadable in HLSL?

farzonl: may be and maybe confusion, figured I'd just change the word. The word Both is also excessive given you list the two types but not a big deal.
```suggestion
Both the binary and unary forms of `+`, `-`, `*` and `&` can be overloaded.
```

proposals/0008-non-member-operator-overloading.md
```diff
@@ -44,6 +44,53 @@ under HLSL 2021 as an extension. Using the feature in HLSL 2021 mode as an
...
+
+An operator function may be called explicitly by using the _operator-function_id
```
farzonl:
```suggestion
An operator function can be called explicitly by using the _operator-function_id
```

proposals/0008-non-member-operator-overloading.md
```diff
@@ -44,6 +44,53 @@ under HLSL 2021 as an extension. Using the feature in HLSL 2021 mode as an
...
+as the name of the function.
+
+An operator function shall either be a non-static member function or a
+non-member function. If an operator is a non-member function it must have at
+least one parameter of class or enumeration type.
+
+No argument to an operator function shall have a default value. Operators cannot
```
farzonl, line 87: The c++ docs seem to get into talking about lookup types for overloads? https://en.cppreference.com/w/cpp/language/operators.html
https://en.cppreference.com/w/cpp/language/adl.html

Should we be doing something similar with HLSL to clarify Argument Dependent lookup vs unqualified lookup?

proposals/0008-non-member-operator-overloading.md
llvm-beanz: This is specifically citing that both the binary and unary forms of those operators can be overloaded. The other operators don't have multiple forms.

proposals/0008-non-member-operator-overloading.md
llvm-beanz: I haven't written language yet for name lookup ([stub](https://microsoft.github.io/hlsl-specs/specs/hlsl.html#Basic.Lookup)). ADL and general lookup behavior is addressed separate from overload resolution in the C++ spec because it applies much wider than just choosing an overload from a set of candidates.

I have written a bunch about [overload resolution](https://microsoft.github.io/hlsl-specs/specs/hlsl.html#Overload) specifically, which does address how the best match is chosen from a set of candidates.

Since I don't think lookup will differ between this spec proposal and the base HLSL 2021 behavior that we need, I'll take an action to get up a PR for the language spec to detail name lookup (https://github.com/microsoft/hlsl-specs/issues/550).

bogner (approved):

proposals/0008-non-member-operator-overloading.md
```diff
@@ -44,6 +44,53 @@ under HLSL 2021 as an extension. Using the feature in HLSL 2021 mode as an
...
+    \terminal{\[\]}\br
+\end{grammar}
+```
+![Latex Rendering](0008-assets/OperatorGrammar.png)
+
+An _operator function_ is a function declared with an _operator-function-id_ as
+its name. An _operator function template_ is a function template
+operator-function-id as its name.
+
+Both the binary and unary forms of `+`, `-`, `*` and `&` can be overloaded.
```
V-FEXrt, line 75: It might be worth spending a sentence on how `operator-function-id`s are determined to be binary or unary. (I assume its implicit in the number of parameters?)

V-FEXrt (approved):

PR #542, opened 2025-07-02
Proposal for 16 bit IsSpecialFloat dxil op

spall:
Initial proposal for adding 16 bit IsSpecialFloat dxil op to SM 6.9

### DIFF
--- a/proposals/NNNN-16bit-isspecialfloat.md
+++ b/proposals/NNNN-16bit-isspecialfloat.md
@@ -1,0 +1,31 @@
+<!-- {% raw %} -->
+
+# 16 bit isSpecialFloat DXIL Op
+
+* Proposal: [NNNN](NNNN-16bit-isspecialfloat.md)
+* Author(s): [Sarah Spall](https://github.com/spall)
+* Sponsor: [Sarah Spall](https://github.com/spall)
+* Status: **Under Consideration**
+
+* Planned Version: Shader Model 6.9
+* PRs:
+* Issues: [#521](https://github.com/microsoft/hlsl-specs/issues/521)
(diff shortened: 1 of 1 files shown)


llvm-beanz:
Just as a side note: for the detailed design we'll want the specific opcodes and the scalar and vector overloads that we intend to require support for.

proposals/NNNN-16bit-isspecialfloat.md
```diff
@@ -0,0 +1,31 @@
...
+## Motivation
+
+The 'isinf', 'isnan', and 'isfinite' functions (there is no 'isnormal' function), support fp16 but instead of generating a 16 bit
+IsSpecialFloat DXIL Op, DXC extends to the 32 bit float op. See (https://github.com/microsoft/DirectXShaderCompiler/issues/7496).
+Some Vendor drivers support the 16 bit IsSpecialFloat op, but some do not. 
+
+## Proposed solution
+
+Beginning with SM 6.9 we would like to have DXC be able to generate this operation, and not extend to 32-bit.
+For shader models before 6.9 we want to emulate the functionality using LLVM IR rather than DXIL op
```
tex3d: nit: punctuation

We should also mention retroactively rolling back and disallowing (in the DXIL validator) the 16-bit DXIL overload for these operations for SM 6.8 and below.

It should be noted that current `f16` (`half`) HLK tests for these DXIL ops actually do a conversion and use the `f32` (`float`) DXIL op instead. Code generated for these tests will be impacted by a compiler change to emulate with an LLVM instructions instead.

spall: I will mention the thing about HLK tests in the detailed design.

PR #556, opened 2025-07-14
[0035] `linalg::Matrix` detailed design revision

llvm-beanz:
This is a big revision to the `linalg::Matrix` proposal. There are a few substantial changes in this proposal to the API formulation.

First it updates the `Matrix` template so that the component type is specified as an enumeration rather than an explicit type. This allows specifying matrix types that are not HLSL built-in types.

Second, it adds two new arguments to the `Matrix` template: `Use` and `Scope`. The `MatrixUse` enumeration specifies the type of matrix as an `A`, `B`, or `Accumulator`. This aligns with the Vulkan cooperative_matrix naming rather than the earlier DirectX `Left` and `Right` naming in an effort to be less confusing to users who are supporting both platforms. The additional `MatrixScope` enumeration denotes the scope of uniformity that the matrix value is operating under. Wave-uniformity is required for matrix-matrix operations, however vector-matrix operations can operate on thread-divergent matrices.

This proposal adds support for vector-matrix operations subsuming the features described in [cooperative vectors](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md).

Last, this proposal adds some rough initial DXIL operation specifications and updates the list of outstanding questions.

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -47,16 +47,34 @@ concept ArithmeticScalar = std::is_arithmetic<T>::value;
 
 namespace linalg {
 
-template <typename ComponentTy, uint M, uint N>
+enum class MatrixUse {
+  A = 0,
+  B = 1,
+  Accumulator = 2,
+};
+
+enum class MatrixScope {
+  Thread = 0,
+  Wave = 1,
+};
+
+enum class UnaryOperation {
+  negate,
+  abs,
+  sin,
+  cos,
+  tan,
+  // What elementwise unary operations make sense?
+};
+
+template <typename ComponentTy, uint M, uint N, MatrixUse Use,
+          MatrixScope Scope>
   requires ArithmeticScalar<ComponentTy>
 class Matrix {
-  template <typename NewCompTy> Matrix<NewCompTy, M, N> cast();
+  template <typename NewCompTy, MatrixUse NewUse = Use>
+  Matrix<NewCompTy, M, N, NewUse, Scope> cast();
 
-  Matrix operator+(Matrix);
-  Matrix operator-(Matrix);
-  Matrix operator*(Matrix);
-  Matrix operator/(Matrix);
-
+  // Element-wise arithmetic operations.
   template <typename T>
     requires ArithmeticScalar<T>
   Matrix operator+(T);
@@ -70,26 +88,87 @@ class Matrix {
     requires ArithmeticScalar<T>
   Matrix operator/(T);
 
-  static Matrix Splat(ElTy Val);
+  // Apply a unary operation to each element.
+  Matrix unaryOperation(UnaryOperation Op);
+
+  static Matrix Splat(ComponentTy Val);
   static Matrix Load(ByteAddressBuffer Res, uint StartOffset, uint Stride,
                      bool ColMajor, uint Align = sizeof(ComponentTy));
   static Matrix Load(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
                      bool ColMajor, uint Align = sizeof(ComponentTy));
 
-  static Matrix Load(groupshared ElTy Arr[], uint StartIdx, uint Stride,
-                     bool ColMajor);
+  static Matrix Load(/*groupshared*/ ComponentTy Arr[], uint StartIdx,
+                     uint Stride, bool ColMajor);
 
   void Store(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
              bool ColMajor, uint Align = sizeof(ComponentTy));
 
-  void Store(groupshared ElTy Arr[], uint StartIdx, uint Stride, bool ColMajor);
+  void Store(/*groupshared*/ ComponentTy Arr[], uint StartIdx, uint Stride,
+             bool ColMajor);
 
-  void MultiplyAccumulate(const ref Matrix<T, N, M>);
-  void SumAccumulate(const ref Matrix<T, N, M>);
+  template <typename T>
+    requires ArithmeticScalar<T>
+  std::enable_if_t<Use == MatrixUse::Accumulator, void>
+  MultiplyAccumulate(const Matrix<T, N, M, MatrixUse::A, Scope>,
+                     const Matrix<T, N, M, MatrixUse::B, Scope>);
+
+  template <typename T>
+    requires ArithmeticScalar<T>
+  std::enable_if_t<Use == MatrixUse::Accumulator, void>
+  SumAccumulate(const Matrix<T, N, M, MatrixUse::A, Scope>,
+                const Matrix<T, N, M, MatrixUse::B, Scope>);
+
+  // Cooperative Vector outer product accumulate.
+  template <typename T>
+  std::enable_if_t<Use == MatrixUse::Accumulator, void>
+  OuterProductAccumulate(const vector<T, M> &, const vector<T, N>);
 };
 
-template <typename T, uint M, uint N, uint K>
-Matrix<T, M, N> Multiply(const ref Matrix<T, M, K>, const ref Matrix<T, K, N>);
+template <typename T, uint M, uint N, uint K, MatrixScope Scope>
+Matrix<T, M, N, MatrixUse::A, Scope>
+Multiply(const Matrix<T, M, K, MatrixUse::A, Scope>,
+         const Matrix<T, K, N, MatrixUse::B, Scope>);
+
+// HLSL 202y+ with global operator overloading these become viable.
+template <typename T, uint M, uint N, uint K, MatrixScope Scope>
+Matrix<T, M, N, MatrixUse::Accumulator, Scope>
+operator+(const Matrix<T, M, K, MatrixUse::A, Scope>,
+          const Matrix<T, K, N, MatrixUse::B, Scope>);
+
+template <typename T, uint M, uint N, uint K, MatrixScope Scope>
+Matrix<T, M, N, MatrixUse::Accumulator, Scope>
+operator-(const Matrix<T, M, K, MatrixUse::A, Scope>,
+          const Matrix<T, K, N, MatrixUse::B, Scope>);
+
+template <typename T, uint M, uint N, uint K, MatrixScope Scope>
+Matrix<T, M, N, MatrixUse::Accumulator, Scope>
+operator*(const Matrix<T, M, K, MatrixUse::A, Scope>,
+          const Matrix<T, K, N, MatrixUse::B, Scope>);
+
+template <typename T, uint M, uint N, uint K, MatrixScope Scope>
+Matrix<T, M, N, MatrixUse::Accumulator, Scope>
+operator/(const Matrix<T, M, K, MatrixUse::A, Scope>,
+          const Matrix<T, K, N, MatrixUse::B, Scope>);
+
+// Cooperative Vector Replacement API
+// Cooperative Vector operates on per-thread vectors multiplying against B
+// matrices.
+
+template <typename OutputElTy, typename InputElTy, uint M, uint N, uint K,
+          typename MatrixBufferTy, typename InputDT, typename MatrixDT,
+          uint MatrixM, uint MatrixK, MatrixScope Scope, bool MatrixTranspose>
+vector<OutputElTy, M>
+Multiply(vector<InputElTy, N> InputVector,
+         Matrix<MatrixDT, M, K, MatrixUse::B, Scope> Matrix);
+
+template <typename OutputElTy, typename InputElTy, typename BiasElTy, uint M,
+          uint N, uint K, typename MatrixBufferTy, typename InputDT,
+          typename MatrixDT, uint MatrixM, uint MatrixK, MatrixScope Scope,
+          bool MatrixTranspose>
+vector<OutputElTy, M>
+MultiplyAdd(vector<InputElTy, N> InputVector,
+            Matrix<MatrixDT, M, K, MatrixUse::B, Scope> Matrix,
+            vector<BiasElTy, M> BiasVector);
 
 } // namespace linalg
 } // namespace hlsl
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -59,18 +59,19 @@ enum class MatrixScope {
 };
 
 enum class UnaryOperation {
-  nop = 0,
-  negate = 1,
-  abs = 2,
-  sin = 3,
-  cos = 4,
-  tan = 5,
-  // What elementwise unary operations make sense?
+  NOp = 0,
+  Negate = 1,
+  Abs = 2,
+  Sin = 3,
+  Cos = 4,
+  Tan = 5,
 };
 
 template <MatrixComponentType ComponentTy, uint M, uint N, MatrixUse Use,
           MatrixScope Scope>
 class Matrix {
+  using ElementType = __detail::ComponentTypeTraits<ComponentTy>::Type;
+
   template <MatrixComponentType NewCompTy, MatrixUse NewUse = Use>
   Matrix<NewCompTy, M, N, NewUse, Scope> cast();
 
@@ -90,51 +91,57 @@ class Matrix {
 
   // Apply a unary operation to each element.
   template<UnaryOperation Op>
-  Matrix unaryOperation();
+  Matrix ApplyUnaryOperation();
 
   template <typename T>
     requires ArithmeticScalar<T>
   static Matrix Splat(T Val);
   static Matrix
   Load(ByteAddressBuffer Res, uint StartOffset, uint Stride, bool ColMajor,
-       uint Align = sizeof(__detail::ComponentTyMapping<ComponentTy>::Type));
+       uint Align = sizeof(ElementType));
   static Matrix
   Load(RWByteAddressBuffer Res, uint StartOffset, uint Stride, bool ColMajor,
-       uint Align = sizeof(__detail::ComponentTyMapping<ComponentTy>::Type));
+       uint Align = sizeof(ElementType));
 
   template <typename T>
     requires ArithmeticScalar<T>
-  static Matrix Load(/*groupshared*/ T Arr[], uint StartIdx, uint Stride,
+  static Matrix Load(groupshared T Arr[], uint StartIdx, uint Stride,
                      bool ColMajor);
 
   void
   Store(RWByteAddressBuffer Res, uint StartOffset, uint Stride, bool ColMajor,
-        uint Align = sizeof(__detail::ComponentTyMapping<ComponentTy>::Type));
+        uint Align = sizeof(ElementType));
 
   template <typename T>
     requires ArithmeticScalar<T>
-  void Store(/*groupshared*/ T Arr[], uint StartIdx, uint Stride,
+  void Store(groupshared T Arr[], uint StartIdx, uint Stride,
              bool ColMajor);
 
   // Row accesses
-  vector<ComponentTy, M> GetRow(uint Index);
-  void SetRow(vector<ComponentTy, M> V, uint Index);
+  vector<ElementType, M> GetRow(uint Index);
+  void SetRow(vector<ElementType, M> V, uint Index);
 
   // Element access
-  ComponentTy Get(uint2 Index);
-  void Set(ComponentTy V, uint2 Index);
+  std::enable_if_t<__detail::ComponentTypeTraits<ComponentTy>::IsNativeScalar,
+                   ElementType>
+  Get(uint2 Index);
+  std::enable_if_t<__detail::ComponentTypeTraits<ComponentTy>::IsNativeScalar,
+                   void>
+  Set(ElementType V, uint2 Index);
 
-  template <typename T>
+  template <typename T, uint K>
     requires ArithmeticScalar<T>
-  std::enable_if_t<Use == MatrixUse::Accumulator, void>
-  MultiplyAccumulate(const Matrix<T, N, M, MatrixUse::A, Scope> &,
-                     const Matrix<T, N, M, MatrixUse::B, Scope> &);
+  std::enable_if_t<Use == MatrixUse::Accumulator && Scope == MatrixScope::Wave,
+                   void>
+  MultiplyAccumulate(const Matrix<T, M, K, MatrixUse::A, Scope> &,
+                     const Matrix<T, K, N, MatrixUse::B, Scope> &);
 
-  template <typename T>
+  template <typename T, unit K>
     requires ArithmeticScalar<T>
-  std::enable_if_t<Use == MatrixUse::Accumulator, void>
-  SumAccumulate(const Matrix<T, N, M, MatrixUse::A, Scope> &,
-                const Matrix<T, N, M, MatrixUse::B, Scope> &);
+  std::enable_if_t<Use == MatrixUse::Accumulator && Scope == MatrixScope::Wave,
+                   void>
+  SumAccumulate(const Matrix<T, M, K, MatrixUse::A, Scope>,
+                const Matrix<T, K, N, MatrixUse::B, Scope>);
 
   // Cooperative Vector outer product accumulate.
   template <typename T>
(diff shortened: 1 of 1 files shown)


proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +88,352 @@ class Matrix {
...
+  Load(ByteAddressBuffer Res, uint StartOffset, uint Stride, bool ColMajor,
+       uint Align = sizeof(__detail::ComponentTyMapping<ComponentTy>::Type));
+  static Matrix
+  Load(RWByteAddressBuffer Res, uint StartOffset, uint Stride, bool ColMajor,
+       uint Align = sizeof(__detail::ComponentTyMapping<ComponentTy>::Type));
 
-  static Matrix Load(groupshared ElTy Arr[], uint StartIdx, uint Stride,
+  template <typename T>
+    requires ArithmeticScalar<T>
+  static Matrix Load(/*groupshared*/ T Arr[], uint StartIdx, uint Stride,
```
damyanp: Could this also be loaded from an array in a cbuffer?  Wondering exactly how to interpret `/*groupsshared*/`

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +88,352 @@ class Matrix {
...
   Matrix operator/(T);
 
-  static Matrix Splat(ElTy Val);
-  static Matrix Load(ByteAddressBuffer Res, uint StartOffset, uint Stride,
-                     bool ColMajor, uint Align = sizeof(ComponentTy));
-  static Matrix Load(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
-                     bool ColMajor, uint Align = sizeof(ComponentTy));
+  // Apply a unary operation to each element.
+  template<UnaryOperation Op>
+  Matrix unaryOperation();
```
damyanp: Is this how unaryOperation would be invoked?

```c++
Matrix<...> M;
M.unaryOperation<UnaryOperation::abs>();
```

proposals/0035-linalg-matrix.md
llvm-beanz: Only `groupshared` memory. Loading from an array in any other address space would have potentially massive performance implications. CBuffer layouts are too inefficient for this, and local memory would have shuffling overhead for `Wave` scope matrices. We could potentially allow local memory for `Thread` scope matrices, but I'm not sure that would be useful.

proposals/0035-linalg-matrix.md
llvm-beanz: Yep, this forces the operation specifier to be a compile-time constant.

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+#### Matrix Use
+
+The `Use` parameter of an instance of a `linalg::Matrix` denotes which argument
+it can be in matrix-matrix operations or matrix-vector operations.
+
+There are three matrix usages: `A`, `B`, and `Accumulator`.
+* The `A` matrix usage denotes a matrix that can be the first argument to binary
+  or ternary algebraic operations.
+* The `B` matrix usage denotes a matrix that can the second argument to binary
+  or ternary algebraic operations.
```
V-FEXrt, line 199: `or ternary`

Is there a missing `C` usage?

V-FEXrt: I see, there is `Accumulator` for that. Is that going to be obvious to users or should we alias a `C` => `Accumulator`?

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+template<>
+struct ComponentTypeTraits<MatrixComponentType::F32> {
+    using Type = float;
+    static const bool IsNativeScalar = true;
+};
+
+template<>
+struct ComponentTypeTraits<MatrixComponentType::F64> {
+    using Type = double;
+    static const bool IsNativeScalar = true;
```
V-FEXrt, line 345: Do we have an  `IsNativeScalar = false` example? And how do we handle non-native scalars?

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+  requires ArithmeticScalar<T>
+Matrix Matrix::operator*(T);
+template <typename T>
+  requires ArithmeticScalar<T>
+Matrix Matrix::operator/(T);
+```
+
+For any arithmetic scalar type the `+`, `-`, `*` and `/` binary operators
+perform element-wise arithmetic on the matrix. The returned by-value `Matrix`
+contains the same handle and refers to the same (now modified) `Matrix`.
```
V-FEXrt, line 383: These are modifying one the of operand matrices? If so, can we use `=*` variants instead of `*`?

V-FEXrt: `Matrix Bar = Foo * 2;` reads hard as `Foo` is unmodified

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+Sets the specified matrix row to the value in the vector V. If the matrix scope
+is `Wave`, this behaves as if called as `SetRow(WaveReadLaneFirst(V), Index)`.
+If the `Index` is out of range of the matrix, this is a no-op.
+
+#### Matrix::Get(uint2)
+
+```c++
+std::enable_if_t<__detail::ComponentTypeTraits<ComponentTy>::IsNativeScalar,
+                 ElementType>
+Matrix::Get(uint2 Index);
```
V-FEXrt, line 486: nit: missing return type (`ElementType`?)

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+method is only available if the component type has has native scalar support in
+HLSL. If the `Index` parameter is out-of range for the matrix the result is `0`
+casted to `ElementType`.
+
+#### Matrix::Set(ElementType, uint2)
+
+```c++
+std::enable_if_t<__detail::ComponentTypeTraits<ComponentTy>::IsNativeScalar,
+                 void>
+Matrix::Set(ElementType V, uint2 Index);
```
V-FEXrt, line 499: nit: missing return type (`void`?)

V-FEXrt: seems like all the examples below here are also missing return types, please also add them there

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+std::enable_if_t<Use == MatrixUse::Accumulator && Scope == MatrixScope::Wave,
+                 void>
+Matrix::MultiplyAccumulate(const Matrix<T, M, K, MatrixUse::A, Scope>,
+                           const Matrix<T, K, N, MatrixUse::B, Scope>);
+```
+
+A matrix with the `Accumulator` use and `Wave` scope has a method
+`MultiplyAccumulate` which takes as parameters an M x K `A` matrix and a K x N
+`B` matrix. The matrix arguments are multiplied against each other and added
+back into the implicit object `Accumulator` matrix.
```
V-FEXrt: which one is the implict `Accumulator` matrix? It can't one of the parameters because that's a type change right?

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+
+#### Matrix::SumAccumulate(Matrix, Matrix)
+
+```c++
+template <typename T>
+  requires ArithmeticScalar<T>
+std::enable_if_t<Use == MatrixUse::Accumulator, void>
+Matrix::SumAccumulate(const Matrix<T, N, M, MatrixUse::A, Scope>,
```
```suggestion
Matrix::SumAccumulate(const Matrix<T, M, K, MatrixUse::A, Scope>,

```

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+              const Matrix<T, N, M, MatrixUse::B, Scope>);
```
```suggestion
              const Matrix<T, K, N, MatrixUse::B, Scope>);

```

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+`MultiplyAccumulate` which takes as parameters an M x K `A` matrix and a K x N
+`B` matrix. The matrix arguments are multiplied against each other and added
+back into the implicit object `Accumulator` matrix.
+
+#### Matrix::SumAccumulate(Matrix, Matrix)
+
+```c++
+template <typename T>
+  requires ArithmeticScalar<T>
+std::enable_if_t<Use == MatrixUse::Accumulator, void>
```
V-FEXrt: This is missing the Scope check right?

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+
+```llvm
+declare @dx.op.fillMatrix.[TY](
+  immarg i32,            ; opcode
+  %dx.types.MatrixRef *, ; matrix
+  [Ty]                   ; fill value
+  )
+```
+
+Fills a matrix with a scalar value. The scalar's type does not need to match the
```
V-FEXrt, line 675: Is the fill a bit fill or is the value casted to the correct type?

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+
+```llvm
+declare void @dx.op.castMatrix(
+  immarg i32,            ; opcode
+  %dx.types.MatrixRef *, ; matrix destination
+  %dx.types.MatrixRef *  ; matrix source
+  )
+```
+
+Converts the element and use type of the source matrix to the destination
```
V-FEXrt, line 687: Is the source handle invalid after the call to this op?

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+  )
+```
+
+Applies a unary math function to each element of the provided matrix.
+
+
+```llvm
+declare void @dx.op.matrixElementwiseBinaryOp.[TY](
+  immarg i32,            ; opcode
+  immarg i32,            ; unary operation (DXILMatrixElementwiseOperation)
```
```suggestion
  immarg i32,            ; binary operation (DXILMatrixElementwiseOperation)

```

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+  Store(RWByteAddressBuffer Res, uint StartOffset, uint Stride, bool ColMajor,
+        uint Align = sizeof(ElementType));
+
+  template <typename T>
+    requires ArithmeticScalar<T>
+  void Store(groupshared T Arr[], uint StartIdx, uint Stride,
+             bool ColMajor);
+
+  // Row accesses
+  vector<ElementType, M> GetRow(uint Index);
```
V-FEXrt, line 121: Should we have an (expensive?) `GetCol(uint Index)` function for a complete API

proposals/0035-linalg-matrix.md
```diff
@@ -70,47 +89,758 @@ class Matrix {
...
+
+#### Matrix Use
+
+The `Use` parameter of an instance of a `linalg::Matrix` denotes which argument
+it can be in matrix-matrix operations or matrix-vector operations.
+
+There are three matrix usages: `A`, `B`, and `Accumulator`.
+* The `A` matrix usage denotes a matrix that can be the first argument to binary
+  or ternary algebraic operations.
+* The `B` matrix usage denotes a matrix that can the second argument to binary
```
V-FEXrt, line 198: Its unclear to me (besides dimensions which are already in the type) why A is only allowed as the first parameter and B the second. Assuming the dimensions fit together why can't they be swapped?

proposals/0035-linalg-matrix.md
llvm-beanz: This would be for the packed data types:
https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_Pack_Unpack_Intrinsics.html#packed-datatypes

We don't have native HLSL scalar support for them.

proposals/0035-linalg-matrix.md
llvm-beanz: Yes, good catch. Will fix. I had mirrored the SPIRV feature, which isn't quite the right behavior for what I'm proposing here.

proposals/0035-linalg-matrix.md
llvm-beanz: The return type is `std::enable_if_t<...>`.

proposals/0035-linalg-matrix.md
llvm-beanz: The return type is `std::enable_if_t<...>`.

proposals/0035-linalg-matrix.md
llvm-beanz: The type for a template class includes the parameters. So an accumulator matrix is a matrix with `Use=Accumulator`.

proposals/0035-linalg-matrix.md
llvm-beanz: I'll add this into the outstanding questions section.

proposals/0035-linalg-matrix.md
llvm-beanz: The A, B, and Accumulator matrices are generally stored differently in the hardware, so we need to denote which is which because conversions between them can be expensive.

proposals/0035-linalg-matrix.md
llvm-beanz: Sum accumulate had a few problems in the signature which needed to be updated here and above. I'll have both updates in the next push.

proposals/0035-linalg-matrix.md
V-FEXrt: Ah that is not how I  thouth `enable_if` syntax worked

proposals/0035-linalg-matrix.md
V-FEXrt: Sorry, what I mean was where does the handle for the Accumulator matrix come into play in source?  The return is `void`, and `A`/`B` can't be the `Accumulator`

In trying to write this out in code I think I figured it out though. `this` is the Acc matrix

```c++
Matrix  Acc = MakeAnAccMat();
Matrix  A = MakeAnAMat();
Matrix  B = MakeABMat();

Acc.MultiplyAccumulate(A B);
```

And the result is stored in `Acc`, is this correct? 

Is there any possibility for a syntax like
```
Matrix  A = MakeAnAMat();
Matrix  B = MakeABMat();
Matrix Acc = Matrix::MultiplyAccumulate(A, B);
```

llvm-beanz: It is a member function, so the implicit object is the accumulator.

proposals/0035-linalg-matrix.md
llvm-beanz: The terminology I use here aligns with the existing terminology for the Vulkan cooperative matrix extension [GLSL_KHR_cooperative_matrix](https://github.com/KhronosGroup/GLSL/blob/main/extensions/khr/GLSL_KHR_cooperative_matrix.txt#L128). I don't know if it is the _best_ terminology, but without something that is clearly better I think it is better for us to align rather than to diverge.

proposals/0035-linalg-matrix.md
llvm-beanz: I filed an issue to fill out the conversions section to detail how conversions occur https://github.com/microsoft/hlsl-specs/issues/568.

proposals/0035-linalg-matrix.md
```diff
@@ -47,70 +47,814 @@ concept ArithmeticScalar = std::is_arithmetic<T>::value;
...
+
+  // Cooperative Vector outer product accumulate.
+  template <typename T>
+  std::enable_if_t<Use == MatrixUse::Accumulator, void>
+  OuterProductAccumulate(const vector<T, M> &, const vector<T, N> &);
 };
 
 template <typename T, uint M, uint N, uint K>
-Matrix<T, M, N> Multiply(const ref Matrix<T, M, K>, const ref Matrix<T, K, N>);
+Matrix<T, M, N, MatrixUse::A, MatrixScope::Wave>
```
```suggestion
Matrix<T, M, N, MatrixUse::Accumulator, MatrixScope::Wave>

```

proposals/0035-linalg-matrix.md
```diff
@@ -47,70 +47,815 @@ concept ArithmeticScalar = std::is_arithmetic<T>::value;
...
+`OuterProductAccumulate` which takes an M-element vector and an N-element
+vector. The operation performs an outer product of the two vectors to produce an
+MxN matrix which is then added back into the implicit object accumulator
+matrix.
+
+#### linalg::Multiply(Matrix, Matrix)
+
+```c++
+template <typename T, uint M, uint N, uint K>
+Matrix<T, M, N, MatrixUse::A, MatrixScope::Wave>
```
```suggestion
Matrix<T, M, N, MatrixUse::Accumulator, MatrixScope::Wave>

```

proposals/0035-linalg-matrix.md
```diff
@@ -47,70 +47,816 @@ concept ArithmeticScalar = std::is_arithmetic<T>::value;
...
+  Accumulator = 2,
+};
+
+enum class MatrixScope {
+  Thread = 0,
+  Wave = 1,
+};
+
+enum class UnaryOperation {
+  NOp = 0,
```
dneto0, line 62: Is this meant to be `Nop` for no-operation?

proposals/0035-linalg-matrix.md
```diff
@@ -47,70 +47,816 @@ concept ArithmeticScalar = std::is_arithmetic<T>::value;
...
+  Negate = 1,
+  Abs = 2,
```
dneto0, line 64: I don't know of particular use cases for Sin, Cos, Tan.  Maybe I need to learn something.  :-)


I would think these two have uses from ML:
* ReLU operator, i.e.  max-with-0
* tanh, for the classic sigmoid.  But it doesn't have to be high accuracy

proposals/0035-linalg-matrix.md
```diff
@@ -47,70 +47,816 @@ concept ArithmeticScalar = std::is_arithmetic<T>::value;
...
-  static Matrix Splat(ElTy Val);
-  static Matrix Load(ByteAddressBuffer Res, uint StartOffset, uint Stride,
-                     bool ColMajor, uint Align = sizeof(ComponentTy));
-  static Matrix Load(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
-                     bool ColMajor, uint Align = sizeof(ComponentTy));
+  template <typename T>
+    requires ArithmeticScalar<T>
+  static Matrix Splat(T Val);
+  static Matrix
+  Load(ByteAddressBuffer Res, uint StartOffset, uint Stride, bool ColMajor,
```
dneto0, line 100: The `Stride` parameter on the load and store functions need more explanation.

Are they byte stride? This is even more critical when operating on groupshared arrays, or when the component type is a packed type.

proposals/0035-linalg-matrix.md
```diff
@@ -47,70 +47,816 @@ concept ArithmeticScalar = std::is_arithmetic<T>::value;
...
 } // namespace linalg
 } // namespace hlsl
 ```
 
 ## Detailed design
 
+### HLSL API Concepts
+
+The new HLSL API introduces a new `linalg::Matrix` type which represents an
+opaque matrix object, and contains an intangible handle that refers to the
```
dneto0, line 183: What is the sense of intangible here?

The HLSL draft spec says

> standard intangible
types. Intangible types are types that have no defined object representation or value representation, as such the size is
unknown at compile time.

What I would need to know is: what declarations can use a matrix type?
* function param
* function return type
* declaration of a variable/const inside a function
* groupshared variable
* structured buffer, RWstructured buffer
* And can the above use arrays of these matrices, or use the matrix values inside a struct.

Issue [DXC/5935](https://github.com/microsoft/DirectXShaderCompiler/issues/5935) criticizedt he original  wave matrix proposal for not allowing arrays-of-matrices, as used in canonical code samples.

proposals/0035-linalg-matrix.md
```diff
@@ -47,70 +47,816 @@ concept ArithmeticScalar = std::is_arithmetic<T>::value;
...
+  I64 = 6,
+  U64 = 7,
+  F16 = 8,
+  F32 = 9,
+  F64 = 10,
+  SNormF16 = 11,
+  UNormF16 = 12,
+  SNormF32 = 13,
+  UNormF32 = 14,
+  SNormF64 = 15,
```
dneto0, line 256: I didn't know of a use case for the normalized F64 cases. Interesting!

proposals/0035-linalg-matrix.md
```diff
@@ -47,70 +47,816 @@ concept ArithmeticScalar = std::is_arithmetic<T>::value;
...
+  UNormF64 = 16,
+  PackedS8x32 = 17,
```
dneto0, line 258: Something should spell out this type: 4 8bit signed integers, packed into 32 bits.  Little endian when in memory.

proposals/0035-linalg-matrix.md
```diff
@@ -47,70 +47,816 @@ concept ArithmeticScalar = std::is_arithmetic<T>::value;
...
+Sets the specified matrix row to the value in the vector V. If the matrix scope
+is `Wave`, this behaves as if called as `SetRow(WaveReadLaneFirst(V), Index)`.
+If the `Index` is out of range of the matrix, this is a no-op.
+
+#### Matrix::Get(uint2)
+
+```c++
+std::enable_if_t<__detail::ComponentTypeTraits<ComponentTy>::IsNativeScalar,
+                 ElementType>
+Matrix::Get(uint2 Index);
```
dneto0, line 486: Is there a uniformity constraint here?
The interesting case is Index varies by thread, but the matrix has Wave MatrixScope. That's a good usecase (the obvious one), and it should work.  I'd like that called out.

proposals/0035-linalg-matrix.md
```diff
@@ -47,70 +47,816 @@ concept ArithmeticScalar = std::is_arithmetic<T>::value;
...
+            Matrix<MatrixDT, M, K, MatrixUse::B, Scope> Matrix,
+            vector<BiasElTy, K> BiasVector);
 
 } // namespace linalg
 } // namespace hlsl
 ```
 
 ## Detailed design
 
+### HLSL API Concepts
```
dneto0, line 180: Something should mention that the API will need a way to enumerate the valid combinations of the template parameters.  AFAIK there is no single set of parameters that works everywhere, and programmers should be cautioned about that.

proposals/0035-linalg-matrix.md
```diff
@@ -47,70 +47,816 @@ concept ArithmeticScalar = std::is_arithmetic<T>::value;
...
+
+Accesses a specific component of the matrix using two-dimensional indexing. This
+method is only available if the component type has has native scalar support in
+HLSL. If the `Index` parameter is out-of range for the matrix the result is `0`
+casted to `ElementType`.
+
+#### Matrix::Set(ElementType, uint2)
+
+```c++
+std::enable_if_t<__detail::ComponentTypeTraits<ComponentTy>::IsNativeScalar,
```
dneto0, line 497: Setting and getting are only available when the component has a native scalar type.
I imagine the use case for getting and setting is still interesting.

In future should there be getter and setter functionality in the additional cases, with implied conversions to a scalar HLSL type?

dneto0 (commented):
Thanks for this revision!

I've got feedback of various importance: minor, substantive, and some forward-looking.

At first look, this seems like it would support our prototyping cooperative matrix for WebGPU. (which is a subset of VK_KHR_cooperative_matrix).
I'd like @alan-baker to also take a look.

proposals/0035-linalg-matrix.md
llvm-beanz: I posted a PR here to clarify the language spec for the usage of intangible objects:
https://github.com/microsoft/hlsl-specs/pull/594

The main points are you cannot use them as parameters or return types for exported functions (the functions must be inlined) and you cannot store them in group shared or device memory since they don't have defined object representations.

We can support arrays of matrices, and we can store them in thread-local structs.

We have specific functions for loading matrices from device and group shared memory, and for storing matrices out to device and group shared memory.

Issue #458, opened 2025-03-24
[0029] clarify OOB handling for vector outer product and vector accumulate


dneto0:
**Which proposal does this relate to?**
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md#matrix-transpose

**Describe the issue or outstanding question.**
The matvec-mul and matvec-muladd operations say the matrix and bias operands are not bounds-checked.

The `outerproductaccumulate` and `vectoraccumulate` operations don't say how out of bounds is handled.
As far as I know D3D usually makes OOB reads 0 and makes OOB writes into no-ops.

llvm-beanz:
I feel like we should revisit this.

Since the matrices are just `ByteAddressBuffers` it seems inconsistent and inappropriate that HLSL have _no_ bounds checking on the matrix accesses.

It seems to me like a better approach would be to define the granularity of bounds checking that we expect for the matrix reads and writes. We could make it simple and say that if any part of the matrix is OOB, the result is 0 for a read and no-op for a write.

PR #558, opened 2025-07-15
[0029] Clarify OOB bahavior for all Coopvec linalg builtins

anupamachandra:
For Coopvec Linalg operations:
OOB Load: Entire load returns zero.
OOB Store: Entire Store is skipped.

spec changes per: https://github.com/microsoft/hlsl-specs/issues/458

### DIFF
--- a/proposals/0029-cooperative-vector.md
+++ b/proposals/0029-cooperative-vector.md
@@ -250,7 +250,7 @@ row/column of the matrix is valid memory.
 The **matrix stride** is 16-byte aligned.
 
 This operation doesn't perform bounds checking for matrix loads. If any part of
-the matrix load is out of bounds then the entire operation is undefined.
+the matrix load is out of bounds then the entire matrix load will return zero.
 
 
 ##### Bias Vector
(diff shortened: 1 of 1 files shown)


damyanp:
We should make sure that there are HLK tests planned to validate this.  (Maybe file issues for them?)

damyanp (approved):

proposals/0029-cooperative-vector.md
```diff
@@ -250,7 +250,7 @@ row/column of the matrix is valid memory.
 The **matrix stride** is 16-byte aligned.
 
 This operation doesn't perform bounds checking for matrix loads. If any part of
-the matrix load is out of bounds then the entire operation is undefined.
+the matrix load is out of bounds then the entire matrix load will return zero.
```
tex3d, line 253: This shouldn't say "doesn't perform bounds checking"
Instead, it should say something like:

```
When the **matrix resource** has bounds checking, bounds are only checked at
matrix granularity. If any part of the matrix load is out of bounds, then the
entire matrix load will return zero.

If the **matrix resource** does not have bounds checking, such as when it is
supplied through a root descriptor, and any part of the matrix load is out of
bounds, the operation is undefined.
```

And similar for bias load, and the corresponding store operations.

Issue #449, opened 2025-03-22
[0029] Clarify that rounding / denorm behavior is implementation defined for the accumulation operations


damyanp:

anupamachandra:
Yes, this makes sense as the rounding and denorm behavior for the new tensor h/w might not be standardized. So keeping this as implementation defined for now makes sense.

damyanp:
See https://devblogs.microsoft.com/directx/shader-model-6-9-and-the-future-of-cooperative-vector/

Issue #448, opened 2025-03-22
[0029] Clear up integer->float / float->integer conversion specification


damyanp:
Currently the spec says "Float to integer conversion is RTNE and saturating. Integer to float conversion is RTNE." with an associated TODO. We'd like the rounding for these conversions to be implementation dependent, similar to the specified float conversion rule.

A possible wording might be "implementation dependent rounded to **a** nearest integer" without the requirement that it's the **nearest even** integer.

anupamachandra:
RTNE is default in IEEE754, preferred for not introducing bias in the data. It is also what is used in SpirV. Do we want to change this?

PR #564, opened 2025-07-16
[0029] Clarify Rounding and Denorm behavior for Coopvec functions

anupamachandra:
Implements https://github.com/microsoft/hlsl-specs/issues/449 and https://github.com/microsoft/hlsl-specs/issues/448.

Updated the spec with:
Rounding: Round to Nearest EvenUhttps://github.com/microsoft/hlsl-specs/issues/448
Upconversion: Not intended behavior but allowed to support older hardware.

These rules apply to all coopvec functions.

### DIFF
--- a/proposals/0029-cooperative-vector.md
+++ b/proposals/0029-cooperative-vector.md
@@ -342,6 +342,8 @@ determine which combinations are supported. A list of combinations that are
 guaranteed to be supported on all implementations can be found in
 [Minimum Support Set].
 
+See [Conversion Rules] for rounding and denorm behavior.
+
 #### Validation
 
 * **matrix interpretation** must be a value corresponding to one of the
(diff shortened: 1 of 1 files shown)


proposals/0029-cooperative-vector.md
```diff
@@ -506,17 +508,12 @@ Non-"Packed" type interpretations are used to request arithmetic conversions.
...
 float. Integer to integer conversion saturates, float to float conversion is
 implementation dependent and preserves the value as accurately as possible.
-Float to integer conversion is RTNE and saturating. Integer to float conversion
-is RTNE.
-
-> TODO: These rules make sense for NN applications but diverge from HLSL
-> conversion rules
-> [here](https://microsoft.github.io/hlsl-specs/specs/hlsl.html#Conv).
+Float to integer conversion is RTNE (round to nearest even) and saturating.
+Integer to float conversion is RTNE. Given the diverse hardware used in
```
tex3d, line 512: For integer to float, what is the overflow behavior when the integer value is out-of-range of the selected float value?  Is it saturating, or does it generate +/-INF?  Or is it implementation defined?

proposals/0029-cooperative-vector.md
```diff
@@ -342,6 +342,8 @@ determine which combinations are supported. A list of combinations that are
 guaranteed to be supported on all implementations can be found in
 [Minimum Support Set].
 
+See [Conversion Rules] for rounding and denorm behavior.
```
tex3d, line 345: Missing `[Conversion Rules]` link elsewhere (see how `[Minimum Support Set]` above is linked).

proposals/0029-cooperative-vector.md
anupamachandra: It is implementation defined. Ideally that would be INF, but since we aren't specing out INF/NAN behavior due to varying hardware, this would be partial specifying INF generation which I think will make it more confusing.

PR #566, opened 2025-07-16
[0038] Talk about adding 'isnormal' function to HLSL in 16 bit IsSpecialFloat proposal

spall:
Currently there is no 'isnormal' function in HLSL which utilizes the IsNormal DXIL Op.  Add text to this proposal which describes the plan of adding an 'isnormal' function to HLSL which utilizes the IsNormal DXIL Op.

### DIFF
--- a/proposals/0038-16bit-isspecialfloat.md
+++ b/proposals/0038-16bit-isspecialfloat.md
@@ -20,7 +20,8 @@ operations were never generated for 16-bit types.
 
 ## Motivation
 
-The 'isinf', 'isnan', and 'isfinite' functions (there is no 'isnormal' function),
+The 'isinf', 'isnan', and 'isfinite' functions (currently there is no 'isnormal'
+function in HLSL),
 support fp16 but instead of generating a 16 bit IsSpecialFloat DXIL Op,
 DXC extends to the 32 bit float op. See
 (https://github.com/microsoft/DirectXShaderCompiler/issues/7496).
@@ -34,17 +35,18 @@ emulate the functionality using LLVM IR rather than a DXIL op.
 
 We should also update the DXIL validator to disallow the 16 bit DXIL overloads
 for the 'isinf', 'isnan', and 'isfinite', operations for SM 6.8 and below.
-We do not plan to update the 'isnormal' operation because it is not reachable
-via HLSL.
 
 ## Detailed design
 
 ### HLSL Additions
 
-There are no HLSL Additions.  The implementations for 16 bit float 'isinf',
+Currently there is no 'isnormal' HLSL function, and the plan is to add 'isnormal'
+to HLSL.  The 16 bit version of the new 'isnormal' function will be emulated
+using LLVM IR for SM6.8 and earlier and will generate the appropriate 16 bit DXIL
+Op in SM6.9 and later.  The implementations for 16 bit float 'isinf',
 'isnan', and 'isfinite' will be updated to use emulation via LLVM IR for
 SM6.8 and earlier, and will generate the appropriate 16 bit DXIL Op in SM6.9
-and later.  For the min16float type, the implementation will remain unchanged,
+and later. For the min16float type, the implementations will remain unchanged,
 DXC will continue to use the 32 bit DXIL ops.
 
 ### Interchange Format Additions
@@ -58,6 +60,7 @@ These are the existing Opcodes for the isSpecialFloat DXIL OpClass.
 8        | IsNaN       | Returns true if x is NAN or WNAN, false otherwise.
 9        | IsInf       | Returns true if x is +INF or -INF, false otherwise.
 10       | IsFinite    | Returns true if x is finite, false otherwise.
+11       | IsNormal    | Returns IsNormal
 
 The following overload will be added:
 ```DXIL
@@ -69,7 +72,7 @@ declare i1 @dx.op.isSpecialFloat.f16(
 #### SPIRV changes
 
 There are the following SPIRV Ops:
-OpIsInf, OpIsNan, OpIsFinite.
+OpIsInf, OpIsNan, OpIsFinite, OpIsNormal.
 
 Currently the 32 bit version of these Ops are being used, but they
 have 16 bit versions as well.  A possible plan is to always generate these 16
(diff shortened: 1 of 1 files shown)


proposals/0038-16bit-isspecialfloat.md
```diff
@@ -20,7 +20,8 @@ operations were never generated for 16-bit types.
 
 ## Motivation
 
-The 'isinf', 'isnan', and 'isfinite' functions (there is no 'isnormal' function),
+The 'isinf', 'isnan', and 'isfinite' functions (currently there is no 'isnormal'
+function in HLSL),
 support fp16 but instead of generating a 16 bit IsSpecialFloat DXIL Op,
 DXC extends to the 32 bit float op. See
```
tex3d: This is sort of a nit...

> support fp16 but instead of generating a 16 bit IsSpecialFloat DXIL Op

This is a bit confusing/misleading, since the HLSL functions do *not* support fp16, which is the problem.  Since `float16_t` inputs can be cast to `float`, and these functions are only defined as taking `float`, they get implicitly cast.  Ultimately this leads to the 32-bit op being used, but the reason is this implicit cast.

Side nit: It's also a bit odd to refer to IsSpecialFloat as a DXIL Op, since it's an OpCodeClass, which is a grouping of DXIL ops that share a function signature when the overload is the same. But that's a nit in earlier text you can ignore.  It makes a difference when talking about overloads because legal overload sets are defined per-DXIL op, whether those DXIL ops share an OpCodeClass or not.  In this case, all the ops in the class have the same set of overloads, but that's not always the case.

Since this is the motivation section, I think it could be clearer about "motivation".

The comment about driver support might be saved for later as justification for our approach of emulating the 16-bit operations on SM < 6.9.

What about this?

Though the `IsSpecialFloat` class of DXIL ops support 16-bit float overloads, when `float16_t` is used with `isinf`, `isnan`, or `isfinite` HLSL intrinsic functions, the argument gets implicitly cast to `float`, since there are no 16-bit overloads for these intrinsics defined in HLSL (bug: [#7496]).
This prevents the expected 16-bit DXIL overload from being used, even though this overload has existed since SM 6.2.
Even HLK tests for the 16-bit operations are impacted by this bug, so they do not test the intended overloads.
We would like to remove this implicit casting behavior which could impact performance and prevents 16-bit overload testing.

Additionally, the `IsSpecialFloat` DXIL OpCodeClass supports the `isnormal` operation, and SPIR-V supports this using `OpIsNormal`, but no HLSL intrinsic exists to target this.
The 32-bit `isnormal` op is tested in the HLK through DXIL IR replacement, but this testing method is awkward and isn't supported by the clang offload test suite, so we would prefer to expose the operation as an intrinsic in HLSL.

proposals/0038-16bit-isspecialfloat.md
```diff
@@ -34,17 +35,18 @@ emulate the functionality using LLVM IR rather than a DXIL op.
 
 We should also update the DXIL validator to disallow the 16 bit DXIL overloads
 for the 'isinf', 'isnan', and 'isfinite', operations for SM 6.8 and below.
```
tex3d: I think we need to include 'isnormal' in this disallowed set.

tex3d: We might also want to mention the driver support issue as justification here for disallowing the 16-bit operations for SM 6.8 and below.

proposals/0038-16bit-isspecialfloat.md
```diff
@@ -34,17 +35,18 @@ emulate the functionality using LLVM IR rather than a DXIL op.
...
 
 ### HLSL Additions
 
-There are no HLSL Additions.  The implementations for 16 bit float 'isinf',
+Currently there is no 'isnormal' HLSL function, and the plan is to add 'isnormal'
+to HLSL.  The 16 bit version of the new 'isnormal' function will be emulated
+using LLVM IR for SM6.8 and earlier and will generate the appropriate 16 bit DXIL
+Op in SM6.9 and later.  The implementations for 16 bit float 'isinf',
 'isnan', and 'isfinite' will be updated to use emulation via LLVM IR for
 SM6.8 and earlier, and will generate the appropriate 16 bit DXIL Op in SM6.9
```
tex3d: The way we do this in DXC is by down-conversion in `DxilFinalizeModule` in `DxilPreparePasses.cpp`.  See where we call `convertQuadVote` there.

This approach is necessary for linking target support - you can generate libraries with latest IR and link down to earlier targets, where certain operations can be down-converted to the correct IR for the target - but it also simplifies reasoning and creates more consistent optimization behavior between targets.

proposals/0038-16bit-isspecialfloat.md
```diff
@@ -58,6 +60,7 @@ These are the existing Opcodes for the isSpecialFloat DXIL OpClass.
 8        | IsNaN       | Returns true if x is NAN or WNAN, false otherwise.
 9        | IsInf       | Returns true if x is +INF or -INF, false otherwise.
 10       | IsFinite    | Returns true if x is finite, false otherwise.
+11       | IsNormal    | Returns IsNormal
```
tex3d: This description is pretty weak.

```suggestion
11       | IsNormal    | Returns false if x is zero, INF, NAN or subnormal (aka. denorm), true otherwise.
```

proposals/0038-16bit-isspecialfloat.md
```diff
@@ -108,6 +111,8 @@ not enabled.
 * How will the execution results be tested?
 There are existing SM6.2 HLK tests for 16 bit float 'isinf', 'isnan',
 and 'isfinite'. The plan is to copy these tests for the SM6.9 HLK.
+A new HLK test for 16 bit float 'isnormal' will need to be written and added
+to the SM6.9 HLK.
```
tex3d: The existing 32-bit `isnormal` test should also be modified to remove the IR modification and use the HLSL intrinsic instead.

tex3d (approved):
Looks basically fine, but I had some minor comments/clarifications.

proposals/0038-16bit-isspecialfloat.md
spall: Funny, I borrowed this description from DXIL.rst.

proposals/0038-16bit-isspecialfloat.md
spall: I decided to include a sentence about that in the motivation instead. I think that should be sufficient?

proposals/0038-16bit-isspecialfloat.md
spall: I added a note about this in the implementation issues.

PR #567, opened 2025-07-16
[0029] Updates to Bound's checking behavior per Tex's feedback

anupamachandra:
Updates per feedback in https://github.com/microsoft/hlsl-specs/pull/558

### DIFF
--- a/proposals/0029-cooperative-vector.md
+++ b/proposals/0029-cooperative-vector.md
@@ -249,8 +249,9 @@ row/column of the matrix is valid memory.
 
 The **matrix stride** is 16-byte aligned.
 
-This operation doesn't perform bounds checking for matrix loads. If any part of
-the matrix load is out of bounds then the entire matrix load will return zero.
+When the **matrix resource** has bounds checking, bounds are only checked at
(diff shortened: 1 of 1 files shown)


damyanp (approved):
LGTM - we should make sure we get at least one approval from a domain expert (ie not me) before merging this.

Before merging, can you update the PR title / description to be more about the substance of the change rather than it being in response to Tex's feedback please?

llvm-beanz (approved):

V-FEXrt (approved):

tex3d (approved):

jenatali:
> If any part of the matrix load is out of bounds then the entire matrix load will return zero.

That sounds like additional test content that needs to be added. Otherwise we should relax the wording to indicate that implementations *may* return zero from the entire matrix load.

V-FEXrt:
@jenatali IIRC `will`/`must` is the correct goal here. I believe there was a request for consistent behavior on OOB access

jenatali:
> @jenatali IIRC `will`/`must` is the correct goal here. I believe there was a request for consistent behavior on OOB access

Got it. Will be a pain to do in WARP but oh well.

PR #571, opened 2025-07-17
[NNNN] HLSL TriangleObjectPositions proposal

amarpMSFT:
HLSL portion of proposal to allow shaders access to raytracing hit triangle position data from BVH (so apps don't have to manually source a copy of this data).

### DIFF
--- a/proposals/NNNN-hlsl-triangle-object-positions.md
+++ b/proposals/NNNN-hlsl-triangle-object-positions.md
@@ -1,0 +1,299 @@
+<!-- {% raw %} -->
+
+# HLSL TriangleObjectPositions
+
+* Proposal: [NNNN](NNNN-triangle-object-positions.md)
+* Author(s): [Tex Riddell](https://github.com/tex3d)
+* Sponsor: [Tex Riddell](https://github.com/tex3d)
+* Status: **Under Consideration**
+
+---
+
+## Introduction
+
+This proposal adds intrinsics that can be called from an Any hit or Closest
+hit shader to obtain the positions of the vertices for the triangle that has
+been hit.
+
+---
+
+## Motivation
+
+Developers often need to know the positions of the vertices for the triangle
+that was hit in an Any/Closest hit shader. Currently, the only way to do this is
+for the developer to provide separate buffers containing this data. This
+information is also available in the acceleration structure, but there is
+currently no way to access it.
+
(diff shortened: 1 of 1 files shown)


proposals/NNNN-hlsl-triangle-object-positions.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Resolved Issues
+
+---
+
+### Return Type
+
+Other approaches have been proposed for the return type of this intrinsic:
+
+* Current approach: return built-in struct containing all three vertices as `float3` fields:
```
llvm-beanz, line 218: This is represented in the HLSL, but not correctly reflected in the DXIL proposal (https://github.com/microsoft/hlsl-specs/pull/572/files). I believe that the document that this was taken from was not correctly updated.

llvm-beanz: On re-reviewing the other doc I guess it isn't so much that it was out of date for the feedback as much as it didn't factor in the other changes in SM 6.9 (specifically [DXIL Vectorization](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0030-dxil-vectors.md)). I'll comment over in that PR about more specific thoughts.

proposals/NNNN-hlsl-triangle-object-positions.md
```diff
@@ -0,0 +1,299 @@
...
+BuiltInTrianglePositions TriangleObjectPositions();
+```
+
+Two new intrinsic methods on `RayQuery` to retrieve triangle object-space
+positions for the candidate or committed triangle hits:
+`CandidateTriangleObjectPositions`, and `CommittedTriangleObjectPositions`.
+
+```cpp
+template<uint ConstRayFlags>
+class RayQuery {
```
llvm-beanz, line 91: It will be simpler for us to comply with our namespacing policy (https://github.com/microsoft/hlsl-specs/blob/main/docs/Process.md#conforming-extensions) if we also provide this implementation for SPIRV.

I believe this is fully compatible with [SPV_KHR_ray_tracing_position_fetch](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_ray_tracing_position_fetch.html), but it would be good to get an opinion from @s-perron and/or @Keenuts.

PR #572, opened 2025-07-17
[NNNN] DXIL TriangleObjectPositions proposal

amarpMSFT:
DXIL portion of proposal to allow shaders access to raytracing hit triangle position data from BVH (so apps don't have to manually source a copy of this data).

### DIFF
--- a/proposals/NNNN-dxil-triangle-object-positions.md
+++ b/proposals/NNNN-dxil-triangle-object-positions.md
@@ -1,0 +1,173 @@
+<!-- {% raw %} -->
+
+# DXIL TriangleObjectPositions
+
+* Proposal: [NNNN](NNNN-triangle-object-positions.md)
+* Author(s): [Tex Riddell](https://github.com/tex3d)
+* Sponsor: [Tex Riddell](https://github.com/tex3d)
+* Status: **Under Consideration**
+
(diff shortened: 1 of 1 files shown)


proposals/NNNN-dxil-triangle-object-positions.md
```diff
@@ -0,0 +1,173 @@
...
+
+---
+
+### Interchange Format Additions
+
+```llvm
+; Availability: Shader Model (Version TBD)+
+; Availability: anyhit,closesthit
+; Function Attrs: nounwind readnone
+declare f32 @dx.op.triangleObjectPosition.f32(
```
llvm-beanz, line 24: This DXIL formulation requires generating 9 instructions to access the full structure. With SM 6.9's [DXIL Vectorization](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0030-dxil-vectors.md) change we can at least reduce this down to 3, but probably to 1.

We should look at having the DXIL operation just return the struct containing 3 vectors.

pacxx: Should we return a long vector instead? NV would be fine with such a change. (This could even be reflected at the HLSL level)

PR #587, opened 2025-07-23
[0035] Updates to the linalg::matrix proposal

### LINKED ISSUE #583: [0035] Fix namespace in docs

This proposal is targeting DirectX, so the feature should go in the DX namespace. We should also identify which parts of this can be easily bridged to Vulkan and see if that allows for a portable `hlsl::Matrix` implementation, but not all parts of this proposal will be portable.

llvm-beanz:
This is a slightly big update to clean up the HLSL pseudo implementaiton. I've added a few example usage bits and added a full HLSL implementation in an appendix.

Resolves #583

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -39,117 +39,97 @@ Matrix Operations](/proposals/0031-hlsl-vector-matrix-operations.md).
 Below is a proposed pseudo-HLSL API. The proposal uses C++20 concepts to
 represent template type constraints so as to avoid needing SFINAE complications.
 
+Some portion of this API surface is portable between DirectX and Vulkan using the [proposed
+DXIL](#dxil-operations) for DirectX and
+[SPV_KHR_cooperative_matrix](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc)
+for Vulkan. Not all features proposed here are supported in Vulkan, so the API
+as described is in the `dx` namespace.
+
+A subsequent revision to HLSL's [0021 - Vulkan Cooperative
+Matrix](0021-vk-coop-matrix.md) support could be considered separately to align
+on a base set of functionality for inclusion in the `hlsl` namespace.
+
 ```c++
-namespace hlsl {
-
-template <class T>
-concept ArithmeticScalar = std::is_arithmetic<T>::value;
-
+namespace dx {
 namespace linalg {
-
-enum class MatrixUse {
-  A = 0,
-  B = 1,
-  Accumulator = 2,
-};
-
-enum class MatrixScope {
-  Thread = 0,
-  Wave = 1,
-};
-
-enum class UnaryOperation {
-  NOp = 0,
-  Negate = 1,
-  Abs = 2,
-  Sin = 3,
-  Cos = 4,
-  Tan = 5,
-};
 
 template <MatrixComponentType ComponentTy, uint M, uint N, MatrixUse Use,
           MatrixScope Scope>
 class Matrix {
-  using ElementType = __detail::ComponentTypeTraits<ComponentTy>::Type;
+  using ElementType = typename __detail::ComponentTypeTraits<ComponentTy>::Type;
 
   template <MatrixComponentType NewCompTy, MatrixUse NewUse = Use>
   Matrix<NewCompTy, M, N, NewUse, Scope> cast();
 
   // Element-wise operations
   template <typename T>
-    requires ArithmeticScalar<T>
-  Matrix operator+=(T);
+  typename hlsl::enable_if<hlsl::is_arithmetic<T>::value, Matrix>::type
+  operator+=(T);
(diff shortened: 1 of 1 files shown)


proposals/0035-linalg-matrix.md
```diff
@@ -39,126 +39,106 @@ Matrix Operations](/proposals/0031-hlsl-vector-matrix-operations.md).
...
-        uint Align = sizeof(ElementType));
+  void Store(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
+             bool ColMajor, uint Align = sizeof(ElementType));
 
   template <typename T>
-    requires ArithmeticScalar<T>
-  void Store(groupshared T Arr[], uint StartIdx, uint Stride,
-             bool ColMajor);
+  typename hlsl::enable_if<hlsl::is_arithmetic<T>::value, void>::type
+  Store(/*groupshared*/ T Arr[], uint StartIdx, uint Stride, bool ColMajor);
```
bob80905, line 98: Why was this commented?

proposals/0035-linalg-matrix.md
```diff
@@ -39,126 +39,106 @@ Matrix Operations](/proposals/0031-hlsl-vector-matrix-operations.md).
...
+      Use == MatrixUse::Accumulator && Scope == MatrixScope::Wave, void>::type
   SumAccumulate(const Matrix<LHSTy, M, K, MatrixUse::A, Scope>,
                 const Matrix<RHSTy, K, N, MatrixUse::B, Scope>);
 
   // Cooperative Vector outer product accumulate.
   template <typename T>
-  std::enable_if_t<Use == MatrixUse::Accumulator, void>
-  OuterProductAccumulate(const vector<T, M> &, const vector<T, N> &);
+  typename hlsl::enable_if<Use == MatrixUse::Accumulator, void>::type
+  OuterProductAccumulate(const vector<T, M>, const vector<T, N>);
```
bob80905, line 128: would a comment like /* inout */ be useful here to act as a counterpart to &?

proposals/0035-linalg-matrix.md
```diff
@@ -859,4 +923,216 @@ Stores a row-vector to a matrix. Out of bounds writes no-op.
...
+  UNormF64 = 16,
+  PackedS8x32 = 17,
+  PackedU8x32 = 18,
+};
+
+namespace __detail {
+template <MatrixComponentType T> struct ComponentTypeTraits {
+  using Type = uint;
+  static const bool IsNativeScalar = false;
+};
```
bob80905, line 993: Why isn't there a struct for, say, `int`? I don't see a situation where `using Type = int` would exist. 
It also isn't represented in any of the examples below, so I'm wondering if this is an example list or if it's supposed to be complete?
Or maybe it's supposed to be `using Type = T`?

proposals/0035-linalg-matrix.md
```diff
@@ -859,4 +923,216 @@ Stores a row-vector to a matrix. Out of bounds writes no-op.
...
+vector<OutputElTy, K>
+Multiply(vector<InputElTy, M> InputVector,
+         Matrix<MatrixDT, M, K, MatrixUse::B, Scope> Matrix);
+
+template <typename OutputElTy, typename InputElTy, typename BiasElTy, uint M,
+          uint K, MatrixComponentType MatrixDT, MatrixScope Scope,
+          bool MatrixTranspose>
+vector<OutputElTy, K>
+MultiplyAdd(vector<InputElTy, M> InputVector,
+            Matrix<MatrixDT, M, K, MatrixUse::B, Scope> Matrix,
```
bob80905: nit: I'm assuming parameter names that are equal to types are alright according to the style rules?

V-FEXrt (approved):

proposals/0035-linalg-matrix.md
llvm-beanz: We don't actually support `groupshared` as a function parameter qualifier for user-defined functions. We have some special handling for the interlocked intrinsics, and we'll likely need something similar to support this.

proposals/0035-linalg-matrix.md
llvm-beanz: I mostly decided that we didn't actually need the reference since this will just be inlined it won't actually create a copy.

proposals/0035-linalg-matrix.md
llvm-beanz: The specialization at the top just allows treating any of the packed types as `uint`. There is a mapping of `I32` to `int32_t`. We don't need to have separate handling for `int` since they're the same type.

proposals/0035-linalg-matrix.md
llvm-beanz: I don't think we have an explicit style rule, but I'll clean that up.

bob80905 (approved):

proposals/0035-linalg-matrix.md
bob80905: Thanks, I recall that phenomenon causing some build problems on certain platforms.

PR #588, opened 2025-07-24
[0026] Clarify Buffer Memory layout

### LINKED ISSUE #543: [0026][LongVec] Clear long vector alignemnt in memory
[0026-hlsl-long-vector-type](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0026-hlsl-long-vector-type.md)

As for long vectors, there is no information saying "how long vectors are placed in memory". Specification describes that `N-element vectors are loaded and stored from ByteAddressBuffers`.

ByteAddressBuffers requires 4 bytes alignment. I'm not sure that means the offset should be always aligned to 4 bytes. And, does StructuredBuffers follows this limitation? (I mean if the offset of each vector is aligned to 4 bytes.)

Additionally, in NV's GLSL cooperative vector extension, vectors must be aligned to 16 bytes in memory.

Please consider we have 2 `vector<half, 3>`, how to place them in memory:
 + always packed: $\mathtt{offset}=i*\mathtt{sizeof}(\mathit{componentType})*\mathit{numComponent}$.
 + always align to 16 bytes: $\mathtt{offset}=i*\mathtt{roundUpTo16}(\mathtt{sizeof}(\mathit{componentType})*\mathit{numComponent})$
 + always align to 4 bytes: $\mathtt{offset}=i*\mathtt{roundUpTo4}(\mathtt{sizeof}(\mathit{componentType})*\mathit{numComponent})$
 + always align to power by 2 bytes: $\mathtt{offset}=i*\mathtt{roundUpToPow2}(\mathtt{sizeof}(\mathit{componentType})*\mathit{numComponent})$.
 + Align to 4 bytes for ByteAddressBuffers, and packed for StructuredBuffers.

**Additional context**
[GLSL_NV_cooperative_vector](https://github.com/KhronosGroup/GLSL/blob/main/extensions/nv/GLSL_NV_cooperative_vector.txt)
> For both coopVecLoadNV and coopVecStoreNV, 'v' can be a cooperative vector
    type with any supported type parameters. The 'buf' arrays must be in either
    buffer or shared storage, and the array that is passed in can be sized or
    unsized. The load or store is done using memory qualifiers taken from the
    declaration of 'buf'. offset must be a multiple of 16. For buffer storage,
    the start of 'buf' must be 16B aligned.

V-FEXrt:
Fixes #543 

Updates the spec to explicitly state that `Load`/`Store` of a vector into/out of a Raw Buffer requires alignment of the underlying buffer and that consuming APIs are allowed to further restrict alignment.

### DIFF
--- a/proposals/0026-hlsl-long-vector-type.md
+++ b/proposals/0026-hlsl-long-vector-type.md
@@ -118,12 +118,12 @@ myBuffer.Store< vector<T, N> >(StartoffsetInBytes + 100, val);
 
 ```
 
-Long Vectors initialzed from or dumped to a Raw Buffer must comply with the 
+Long Vectors initialized from or dumped to a Raw Buffer must comply with the
 alignment rules of the underlying buffer. Therefore a vector `Load`ed or `Store`d
 to a `ByteAddressBuffer` must either naturally align to or be padded to 4 bytes.
 Misaligned offsets results in undefined behavior. APIs built on top of Long Vectors
 such as Cooperative Vectors or Cooperative Matrix may further restrict memory
-alignment requirements but that is not specified or manidated by this spec.
+alignment requirements but that is not specified or mandated by this spec.
 
 StructuredBuffers with N-element vectors are declared using the template syntax
  with a long vector type as the template parameter.
--- a/proposals/0026-hlsl-long-vector-type.md
+++ b/proposals/0026-hlsl-long-vector-type.md
@@ -118,6 +118,13 @@ myBuffer.Store< vector<T, N> >(StartoffsetInBytes + 100, val);
 
 ```
 
+Long Vectors initialzed from or dumped to a Raw Buffer must comply with the 
+alignment rules of the underlying buffer. Therefore a vector `Load`ed or `Store`d
+to a `ByteAddressBuffer` must either naturally align to or be padded to 4 bytes.
+Misaligned offsets results in undefined behavior. APIs built on top of Long Vectors
+such as Cooperative Vectors or Cooperative Matrix may further restrict memory
+alignment requirements but that is not specified or manidated by this spec.
+
 StructuredBuffers with N-element vectors are declared using the template syntax
  with a long vector type as the template parameter.
 N-element vectors are loaded and stored from ByteAddressBuffers using the templated load and store methods
--- a/proposals/0026-hlsl-long-vector-type.md
+++ b/proposals/0026-hlsl-long-vector-type.md
@@ -119,11 +119,12 @@ myBuffer.Store< vector<T, N> >(StartoffsetInBytes + 100, val);
 ```
 
 Long Vectors initialized from or dumped to a Raw Buffer must comply with the
-alignment rules of the underlying buffer. Therefore a vector `Load`ed or `Store`d
-to a `ByteAddressBuffer` must either naturally align to or be padded to 4 bytes.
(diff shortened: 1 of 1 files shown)


proposals/0026-hlsl-long-vector-type.md
```diff
@@ -118,6 +118,13 @@ myBuffer.Store< vector<T, N> >(StartoffsetInBytes + 100, val);
 
 ```
 
+Long Vectors initialized from or dumped to a Raw Buffer must comply with the
+alignment rules of the underlying buffer. Therefore a vector `Load`ed or `Store`d
+to a `ByteAddressBuffer` must either naturally align to or be padded to 4 bytes.
```
llvm-beanz: 4 bytes seems possibly wrong. Should 64-bit types require 8-byte alignment? What about 16-bit types?

V-FEXrt: Maybe I'm reading the docs wrong?

> The byte value must be a multiple of four so that it is DWORD aligned

https://learn.microsoft.com/en-us/windows/win32/direct3d11/direct3d-11-advanced-stages-cs-resources#byte-address-buffer

It explicitly says multiples of four, but also looks like DWORD is always 32 bits/4 bytes on windows

V-FEXrt: Also, I'm not sure I understand how a 64 or 16 bit type comes into play for a byte address buffer.

It's just a buffer of bytes and the requirement is that the offset into the buffer be 4-byte aligned. A vector holding a 64-bit type will naturally fall into the 4-byte alignment and a single 16-bit value _will_ need to 4-byte align the offset into the buffer. (though a `vector<half,  2>` will naturally align)

anupamachandra: Looking at this document as the expected behavior in HLSL: https://microsoft.github.io/hlsl-specs/specs/hlsl.pdf (section 1.7.2) : "The alignment requirements of an offset into device memory space is the size in bytes of the largest scalar type
contained in the given aggregate type." . This suggests that for longvectors, the alignment for the load should be the size of the scalar element type (2B for 16, 8B for 64...). @V-FEXrt, to your question above, though it is a "Byte" Address Buffer, the offset should be adjusted so the load/store alignment is a multiple of the scalar element size. So, in both the specs, the HLSL spec linked above and the current PR, is my understanding correct that the alignment guarantees for load/store mean base+offset? @llvm-beanz to comment on the spec language in the HLSL spec above.

V-FEXrt: > This suggests that for longvectors, the alignment for the load should be the size of the scalar element type (2B for 16, 8B for 64...)

Ah yep, that makes a lot of sense. The `The byte value must be a multiple of four` is discussing the `uint` overloads of the function, not the `T` overload. I'll update the wording!

V-FEXrt: Updated! PTAL

llvm-beanz (approved):

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -118,6 +118,12 @@ myBuffer.Store< vector<T, N> >(StartoffsetInBytes + 100, val);
 
 ```
 
+Long Vectors initialized from or dumped to a Raw Buffer must comply with the
```
anupamachandra: Minor nit, can we just say `loaded from or stored to` instead of `initialized from or dumped to`?

PR #593, opened 2025-08-01
[0026] Specify slice vector operation

### LINKED ISSUE #578: [0026] Consider `slice` and `shuffle` intrinsics
[[0026] - long vectors](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0026-hlsl-long-vector-type.md)

Long vectors don't support swizzling operators, but it may be useful to slice a long vector to produce a smaller vector, or to shuffle a long vector.

We should consider what these intrinsics could look like.

V-FEXrt:
Fixes #578 

Species the API for the slice and shuffle vector operations

Example of how this could look in implementation: https://godbolt.org/z/aeM6brqv5

### DIFF
--- a/proposals/0026-hlsl-long-vector-type.md
+++ b/proposals/0026-hlsl-long-vector-type.md
@@ -139,8 +139,8 @@ They do not support any swizzle operations.
 #### Converting dimensions of long vectors
 
 HLSL vectors support extraction and recomposition of new vectors via swizzling, however
-long vectors don't support swizzling instead two new operations are made available on
-all vectors.
+long vectors don't support swizzling, instead the following operation is made available
+on all vectors.
 
 **`slice<Offset, Count, ComponentType, InSize>(vector<ComponentType, InSize> in)`**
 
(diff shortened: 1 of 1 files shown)


proposals/0026-hlsl-long-vector-type.md
```diff
@@ -136,6 +136,56 @@ myBuffer.Store(elementIndex, val);
...
+all vectors.
+
+**`slice<Offset, Count>()`**
+
+The `slice` member function returns a new `vector` containing elements starting
+at `Offset` containing `Count` number of components.
+
+```hlsl
+vector<int, 6> src = {0, 1, 2, 3, 4, 5};
+vector<int, 3> sub = src.slice<2, 3>() // sub = {2, 3, 4}
```
llvm-beanz: This would be a quite large change to the language since today vector isn't a type that contains member functions. I'm a little worried about that vs having a freestanding slice function that takes the source as an argument.

V-FEXrt: I see, I didn't realize vector didn't have member functions.

The main reason for adding it as a member function is to avoid repeating the source type in the template ala
```
vector<float, 6> In = ...
vector<float, 2> Out = slice<float, 6, 4, 2>(In);
```

I can't seem to write the template rule in such a way that the source vector type is deduced without it being a member function

llvm-beanz: Here you go:

```c++
template <int Off, int OSz, typename T, int ISz>
vector<T, OSz> slice(vector<T, ISz> In) {
    vector<T, OSz> Result = (vector<T, OSz>)0;
    [unroll]
    for (int I = 0; I < OSz; ++I)
      Result[I] = In[I+Off];
    return Result;
}

export void fn() {
    int4 i = {0,1,2,3};
    int2 s = slice<2, 2>(i);
}
```

https://godbolt.org/z/bn84n5rha

V-FEXrt: Thanks! Updated!

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -136,6 +136,58 @@ myBuffer.Store(elementIndex, val);
...
+```hlsl
+vector<int, 6> src = {0, 1, 2, 3, 4, 5};
+vector<int, 4> sub = slice<4>(src) // sub = {0, 1, 2, 3}
+```
+
+A compiler error will be raised if the slice operation would result in out of bounds access.
+
+**`shuffle<Idx1, Idx2, ..., ComponentType, InSize>(vector<ComponentType, InSize> in)`**
+
+*`shuffle` is only available in HLSL 202y*
```
llvm-beanz: We should remove `shuffle` from this proposal, and file an issue on it to re-evaluate with HLSL 202y. I don't want to lump this in with long vectors since we're supporting long vectors in DXC, and we are not supporting HLSL 202y in DXC.

V-FEXrt: Done! https://github.com/microsoft/hlsl-specs/issues/599

llvm-beanz (approved):

PR #594, opened 2025-08-04
Add clarifications on usage of intangible objects

llvm-beanz:
This adds clarifications to the language specification on the allowed usage of objects of intangible types.

This is in response to a comment by @dneto0 on the linalg::Matrix PR:

https://github.com/microsoft/hlsl-specs/pull/556#discussion_r2217019792

### DIFF
--- a/specs/language/basic.tex
+++ b/specs/language/basic.tex
@@ -420,8 +420,8 @@ resources in cbuffer declarations, but it does so by hoisting the resource
 declaration out to the global scope.}
 
 \item An object of intangible type may not be a parameter or return type of a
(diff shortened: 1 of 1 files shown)


bob80905 (approved):
LGTM. Assuming program linkage means external linkage.

llvm-beanz:
> LGTM. Assuming program linkage means external linkage.

Thank you for pointing this out, this prompted me to double check the spec language. I need to restrict both program and external linkage.

The language spec follows the C++ convention of having different levels of visibility. Program visibility is roughly analogous to "external" in DXC, but there is also a separate "external" visibility in the language spec which is more like the "external" notion in LLVM IR.

PR #598, opened 2025-08-06
[0035] Revise matrix element accesses

### LINKED ISSUE #575: [0035] Revisit Row/Col accessors

The per-element accessors may be unwieldy on some hardware, and whether or not it is efficient to access rows or columns varies by the underlying representation. For this reason I need to revisit the per-element accessors and understand the use cases and if any restrictions should be applied.

llvm-beanz:
This PR removes the row and element accessors in favor of a more general means of accessing per-thread vectors. This change allows initializing a matrix from per-thread vectors as well as splitting it out into constituent vectors.

The vectors represent rows in A matrices, and columns in B matrices. Since the layout of Accumulator matrices vary by vendor, this method may only be used to construct A or B matrices.

Resolves #575


proposals/0035-linalg-matrix.md
```diff
@@ -57,6 +57,9 @@ template <MatrixComponentType ComponentTy, uint M, uint N, MatrixUse Use,
           MatrixScope Scope>
 class Matrix {
   using ElementType = typename __detail::ComponentTypeTraits<ComponentTy>::Type;
+  static const uint ElementsPerScalar = __detail::ComponentTypeTraits<ComponentTy>::IsNativeScalar ? 1 : 4;
```
bfavela: Not sure if this is supposed to be obvious, but I think it would be helpful to explain why you have a literal '4' here.

proposals/0035-linalg-matrix.md
```diff
@@ -90,25 +93,37 @@ class Matrix {
...
+  typename hlsl::enable_if<Use == MatrixUse::A && Scope == MatrixScope::Wave &&
+                               UseLocal == Use,
+                           Matrix>::type
+      FromThreadVectors(vector<ElementType, MScalars>);
+
+  template <MatrixUse UseLocal = Use>
+  typename hlsl::enable_if<Use == MatrixUse::B && Scope == MatrixScope::Wave &&
+                               UseLocal == Use,
+                           Matrix>::type
+      FromThreadVectors(vector<ElementType, NScalars>);
```
bfavela, line 115: Did you purposely leave off `MatrixUse::Accumulator` here (and `GetThreadVector()` below)? If so, a comment as to why would be useful

proposals/0035-linalg-matrix.md
```diff
@@ -513,6 +547,32 @@ expected target data format. When read from `groupshared` memory, the data may
...
+Produces a matrix from per-thread vectors. An A matrix is produced from
+per-thread row vectors, while a B matrix is produced from per-thread column
+vectors. The `FromThreadVectors` construction method is not available for
+accumulator matrices which vary by hardware implementation.
+
+When creating an A matrix, the N dimension must be less than or equal to the
+wave size. When creating a B matrix, the M dimension must be less than or equal
+to the wave size. Threads outside the matrix size are discarded.
+
+Must be called from wave-uniform control flow.
```
bfavela, line 594: This seems like an unnecessary restriction. Wouldn't it be possible to partially pull a matrix from masked threads? Not 100% sure the use case (identity matrix loads, maybe?) but enforcing this might be power intensive if you only need certain portions of a matrix.

proposals/0035-linalg-matrix.md
bfavela: Ah, reading further I see the explanation later on. I'm ok to disregard this feedback, then.

proposals/0035-linalg-matrix.md
```diff
@@ -262,6 +276,26 @@ Throughout this document a matrix may be described as having a scope as
...
+In HLSL, matrix objects are intangible objects so they do not have defined size
+or memory layout. When in use, implementations are expected to spread the
+storage of matrices across the thread-local storage for all threads in a SIMD
+unit. At the DXIL level we represent a matrix through a handle object.
+
+An A matrix is a collection of per-thread vectors representing matrix rows,
+while a B matrix is a collection of per-thread vectors representing matrix
+columns.
+
+An Accumulator matrix may be either an A matrix, or a B matrix, and it varies by
```
bfavela, line 307: One additional intrinsic that may be useful related to this is a query whether an accumulator matrix is an A or B type.

proposals/0035-linalg-matrix.md
llvm-beanz: Ah yea... that's some silliness to deal with the 8-bit packed types where we put 4 of them into a uint32_t. Will add clarifications.

proposals/0035-linalg-matrix.md
llvm-beanz: I think we'd need to define what values get pulled from masked threads. Is it 0's or... something else?

bfavela: I can see two potentials: 0s would mean any matmul operation would get poisoned data and make it obvious that the threads were erroneously used. Or 1s would act like an identity for those masked threads. I don't know which is functionally desired in this case (guessing 0s to make debugging easier).
Anything but UB :)

nhaehnle: I think it's strongly preferable that wave-scope matrices can only be used from wave-uniform control flow. Otherwise you're making the divergence gods unhappy.

It'd be good to understand the use cases better, but I'd imagine that common use cases would have divergent control flow only while building the vectors. At the end of the divergent control flow is a natural point where you are back in wave-uniform control flow with one vector per thread, and at that point you can build the wave-scope matrix.

That said, on the size restriction: I could see us allowing also the case where `matrix size > wave size`. Semantically you'd probably want the out of bounds columns/rows of the matrix to be 0; in practice, that will often be something we only need to enforce *after* the matrices are used (we can leave the matrix contents as garbage before matrix multiply, and that only creates garbage in elements of the output that we know mathematically to be 0 -- assuming a 0 initial accumulator).

proposals/0035-linalg-matrix.md
```diff
@@ -534,51 +594,33 @@ matrix object. When storing to `groupshared` memory, the matrix component data
...
+GetThreadVector(uint Index = 0);
 ```
 
-Accesses a specific component of the matrix using two-dimensional indexing. This
-method is only available if the component type has has native scalar support in
-HLSL. If the `Index` parameter is out-of range for the matrix the result is `0`
-casted to `ElementType`.
+Returns the underlying vector for the associated thread in the matrix. The
+optional index is used when the matrix `K` dimension is larger than the wave
+size to compute the starting offset (i.e. `(Index * WaveSize) + ThreadID`).
```
nhaehnle: This should probably also be restricted to uniform control flow.

Similar considerations as for building a matrix from vectors. Unlike `FromThreadVectors`, this operation can at least be given clearly defined and naturally intuitive semantics if used in divergent control, but implementing those robustly requires driver compiler magic (similar to the kind of issues with derivatives in control flow).

In terms of how this function will be used, I again don't see any issues with requiring uniform control flow. Since the other operations on wave-scope matrix have to be in uniform control flow anyway, you can easily see a pattern of:
```
... divergent region ...

matrix = FromThreadVectors(vec);
... matrix ops ...
vec = matrix.GetThreadVector(0);

... divergent region ...
```
In general, I think it is a very clear and intuitive rule to say: If you want to do anything with a matrix of scope S, your control flow has to be uniform in S.

proposals/0035-linalg-matrix.md
```diff
@@ -534,51 +594,33 @@ matrix object. When storing to `groupshared` memory, the matrix component data
...
+on an Accumulator matrix because the matrix layout varies by hardware
+implementation.
 
-Sets a specified element of the matrix to the provided value. If the matrix
-scope is `Wave`, this behaves as if called as `Set(WaveReadLaneFirst(V), Index)`.
-If the `Index` is out of range, this is a no-op.
+Threads which correspond to threads outside the matrix size will return a vector
+with all elements zero initialized.
 
 #### Matrix::MultiplyAccumuate(Matrix, Matrix)
```
nhaehnle: typo:missing l

proposals/0035-linalg-matrix.md
```diff
@@ -862,6 +926,25 @@ Store a matrix to groupshared memory. Data conversions between opaque matrices
...
+  i32,                   ; Index
+  )
+```
+
+Extracts per-thread vectors from a matrix. For an A matrix the NUM corresponds
+to the M dimension while for a B matrix it corresponds to the N dimension. The
+NUM must match the matrix corresponding dimension, unless the element is a
+packed data type in which case it must be the number of 32-bit unsigned integers
+used to store M elements. This operation may not be used on Accumulator
+matrices.
```
nhaehnle, line 984: Also restrict to uniform control flow.

nhaehnle (commented):
I like the general direction, and it's much better than being able to directly query or set elements by (row, col).

In addition to my earlier comments, I do think it could still be valuable to expose a more general mechanism for per-element operations on matrices, a sort of `map` operation, perhaps one that provides row/col information to the lambda/functor/whatever.

If `map` is an issue due to general HLSL concerns, an alternative could be extending element-wise operations to matrices similar to what's done for vectors. Plus, a way to generate "iota" matrices where the value of each element is its row or its column number.

proposals/0035-linalg-matrix.md
llvm-beanz: > If you want to do anything with a matrix of scope S, your control flow has to be uniform in S.

The only concern I have with this statement is that we do explicitly want to support the vector x matrix operations with Wave-scope matrices, and those should be supported in divergent control flow. The use of a wave scope matrix in those contexts is really a good optimization hint.

proposals/0035-linalg-matrix.md
```diff
@@ -57,6 +57,14 @@ template <MatrixComponentType ComponentTy, uint M, uint N, MatrixUse Use,
           MatrixScope Scope>
 class Matrix {
   using ElementType = typename __detail::ComponentTypeTraits<ComponentTy>::Type;
+  // If this isn't a native scalar, we have an 8-bit type, so we have 4 elements
+  // packed in each scalar value.
+  static const uint ElementsPerScalar =
+      __detail::ComponentTypeTraits<ComponentTy>::IsNativeScalar ? 1 : 4;
+  static const uint MScalars =
```
V-FEXrt, line 66: Might be worth adding a small comment along the lines of: 

```
// Computes the number of M/N scalars actually stored in the matrix accounting for packing.
```

Took me a minute to work out what the value actually represented

proposals/0035-linalg-matrix.md
```diff
@@ -133,6 +153,8 @@ class Matrix {
   OuterProductAccumulate(const vector<T, M>, const vector<T, N>);
 };
 
+MatrixUse AccumulatorLayout();
```
V-FEXrt, line 160: Since we have this API, does it make sense for the API to have a "convert Acc to A/B Matrix" (and you have to use this to actually know which one you got?)

I'm assuming no since we'd lose the compile time checks since the compiler also don't know which matrix the converted one is

V-FEXrt: Mentioned later that the value is driver compile time known so maybe 

```
if (AccumulatorLayout() == A) {
  MakeAAMatrix();
} else {
  MakeABMatrix();
}
```
is sufficient with DCE

llvm-beanz: You kinda always need to know which type of matrix you have since we don't have dynamic typing and there isn't polymorphism here.

We have the `cast` method which can convert between matrix uses. Part of the benefit of this API is that if you know if a cast from accumulator is going to be free or costly, so if a user needed to use an accumulator as an a or b matrix, they could write code that is aware of which path is cheaper.

proposals/0035-linalg-matrix.md
```diff
@@ -833,6 +917,28 @@ Populates a matrix with data from a `groupshared` array. Data conversions
...
+Populates a matrix from per-thread vectors. For an A matrix the NUM corresponds
+to the M dimension while for a B matrix it corresponds to the N dimension. The
+NUM must match the matrix corresponding dimension, unless the element is a
+packed data type in which case it must be the number of 32-bit unsigned integers
+used to store M elements. This operation may not be used on Accumulator
+matrices.
+
+For an A matrix the N dimension must be less than or equal to the WaveSize. For
+a B matrix the M dimension must be less than or equal to the WaveSize. Values
+from additional threads are discarded.
```
tex3d, line 937: Don't you mean the `K` dimension for both A and B matrix case?
```suggestion
For an A matrix the K dimension must be less than or equal to the WaveSize. For
a B matrix the K dimension must be less than or equal to the WaveSize. Values
from additional threads are discarded.
```

proposals/0035-linalg-matrix.md
```diff
@@ -262,6 +287,26 @@ Throughout this document a matrix may be described as having a scope as
...
+A matrix with thread scope must behave as if each thread has a unique copy of
+the matrix, however the storage is not required to be thread local, and an
+implementation may coalesce identical matrices across threads.
+
+#### Matrix Storage
+
+In HLSL, matrix objects are intangible objects so they do not have defined size
+or memory layout. When in use, implementations are expected to spread the
+storage of matrices across the thread-local storage for all threads in a SIMD
+unit. At the DXIL level we represent a matrix through a handle object.
```
damyanp: Are these two statements compatible?

> the storage is not required to be thread local, and an implementation may coalesce identical matrices across threads.

> implementations are expected to spread the storage of matrices across the thread-local storage for all threads in a SIMD unit

damyanp (approved):

proposals/0035-linalg-matrix.md
llvm-beanz: This proposal doesn't consider `K` dimensions the way that the old WaveMatrix one did. All matrices are `M x N`, the `K` dimension only factors into the multiplication operations where the matrices are `M x K` and `K x N` to produce an `M x N` output.

proposals/0035-linalg-matrix.md
llvm-beanz: Yes, although probably confusingly worded...

The point is that the _full_ storage for a matrix does not need to be materialized into a single thread, instead an implementation is expected to spread storage of matrices across thread-local storage for all threads in a SIMD unit.

proposals/0035-linalg-matrix.md
```diff
@@ -513,6 +562,36 @@ expected target data format. When read from `groupshared` memory, the data may
...
+```
+
+Produces a matrix from per-thread vectors. An A matrix is produced from
+per-thread row vectors, while a B matrix is produced from per-thread column
+vectors. The `FromThreadVectors` construction method is not available for
+accumulator matrices which vary by hardware implementation.
+
+When creating an A matrix, the N dimension must be less than or equal to the
+wave size. When creating a B matrix, the M dimension must be less than or equal
+to the wave size. Threads outside the matrix size are discarded.
```
tex3d, line 592: Don't you mean the K dimension for both A and B matrix case?
```suggestion
When creating an A matrix, the K dimension must be less than or equal to the
wave size. When creating a B matrix, the K dimension must be less than or equal
to the wave size. Threads outside the matrix size are discarded.
```

proposals/0035-linalg-matrix.md
```diff
@@ -513,6 +562,36 @@ expected target data format. When read from `groupshared` memory, the data may
...
+template <MatrixUse UseLocal = Use>
+typename hlsl::enable_if<Use == MatrixUse::B && Scope == MatrixScope::Wave &&
+                              UseLocal == Use,
+                          Matrix>::type
+    FromThreadVectors(vector<ElementType, NScalars>);
+```
+
+Produces a matrix from per-thread vectors. An A matrix is produced from
+per-thread row vectors, while a B matrix is produced from per-thread column
+vectors. The `FromThreadVectors` construction method is not available for
```
tex3d: I think this is backwards:
A is M x N, where M is number of rows and matches the vector dimension, therefore each per-thread vector represents a column of the A matrix and a row of the B matrix.

PR #604, opened 2025-08-12
Some updates to Long Vector Execution Test Plan tables

alsepkow:
Adding some notes about the basic op type (unary, binary, or ternary). And some notes about output parameters where applicable.


inbelic (approved):
nit: The notes column is the last for some and not for others. Maybe best to make it consistent

alsepkow:
> nit: The notes column is the last for some and not for others. Maybe best to make it consistent

Thanks, I neglected to bring in the fix for that as I was also using a forked copy to help track implementation. I've decided that I think its helpful to just bring that over to this PR.

proposals/infra/INF-0006-Long-Vector-ExecutionTest-Plan.md
```diff
@@ -197,162 +196,164 @@ considered completed.
    waiting for an OS/HLK release.
 
 ## HLSL-Operators
+✅ - Means there was an explicit test case implemented for the intrinsic.
+☑️ - Means the intrinsic gets coverage via other intrinsics. For example 'exp2'
```
inbelic, line 200: nit: in the rich view this is all one line. Not sure that was intentional

PR #608, opened 2025-08-21
[0039] Flesh out the SPIRV and DXIL lowering

llvm-beanz:
This adds the remaining details for the HLSL, DXIL and SPIRV implementations of the debugbreak feature.

### DIFF
--- a/proposals/0039-debugbreak.md
+++ b/proposals/0039-debugbreak.md
@@ -56,4 +56,70 @@ void main(uint GI : SV_GroupIndex) {
 
 This aligns with C/C++ conventions that our users are already familiar with.
 
+## Detailed Design
+
+This proposal introduces a new HLSL `DebugBreak` intrinsic which has a
+runtime-defined behavior to facilitate shader debugging workflows. If the
+runtime does not support or is not configured to enable support for the
+corresponding DXIL instruction must be treated as a no-op by the driver.
+
+### HLSL Surface
+
+A new `DebugBreak` function is added with the signature:
+
+```
+void DebugBreak();
+```
+
+A new header `assert.h` is added and included with the compiler packaging which
+implements the `assert` macro:
+
+```c
+#if NDEBUG
+#define assert(cond) do { } while(false)
+#else
+#define assert(cond) do { if (!cond) DebugBreak();} while(false)
+#endif
+```
+
+### DXIL Lowering
+
+This change introduces a new DXIL operation:
+
+
+``` llvm
+declare void @dx.op.debugBreak(
+  immarg i32             ; opcode
+)
+```
+
+This DXIL operation must be treated as `convergent` even though it is not to
+prevent code motion.
+
+This instruction will only be valid in a new shader model.
+
+Because it is valid to treat this operation as a no-op, it is a required
+supported feature and does not require a capabilities bit.
+
+### SPIRV Lowering
+
+This change will utilize the existing `NonSemantic.DebugBreak` instruction.
+While this instruction is not widely supported by Vulkan debuggers, it is
+supported by NVIDIA's NSight and can be safely ignored by Vulkan runtimes.
+
+The SPIRV usage will utilize the following instructions:
+
+```
+%1 = OpExtInstImport "NonSemantic.DebugBreak"
+%2 = OpExtInst %void %1 DebugBreak
+```
+
+## Open Questions
+
+* Consider introducing the `convergent` attribute to DXIL.
+  * This should be "cheap" and would potentially address pre-existing bugs.
+  * This would preserve the requirement that this operation not be moved during
+    optimization in the final DXIL.
+
+
 <!-- {% endraw %} -->


proposals/0039-debugbreak.md
```diff
@@ -56,4 +56,70 @@ void main(uint GI : SV_GroupIndex) {
...
+``` llvm
+declare void @dx.op.debugBreak(
+  immarg i32             ; opcode
+)
+```
+
+This DXIL operation must be treated as `convergent` even though it is not to
+prevent code motion.
+
+This instruction will only be valid in a new shader model.
```
damyanp, line 104: > This instruction will only be valid in a new shader model.

I think that this will be quite limiting and we should consider ways to retrospectively add this as an optional feature to older shader models.

llvm-beanz: I don't believe we've ever retroactively added instructions or feature flags to a shader model.

This seems deeply concerning to me because I don't know how we would do it without a potential mess of binary compatibility issues. Drivers and runtimes for older shader models are shipped knowing the set of capabilities and DXIL operations they support. They often crash when the encounter bits of new shader models.

In theory you could tie this to an AgilitySDK _if_ the AgilitySDK gained the ability to rewrite DXIL which would pose a whole lot of problems (legal as well as technical).

damyanp: Alas I think this is necessary, but not easy.

llvm-beanz: I've filed an issue to go deeper on this: https://github.com/microsoft/hlsl-specs/issues/609

I think there are a lot of problems and nuances we'll need to explore here.

proposals/0039-debugbreak.md
```diff
@@ -56,4 +56,70 @@ void main(uint GI : SV_GroupIndex) {
...
+#endif
+```
+
+### DXIL Lowering
+
+This change introduces a new DXIL operation:
+
+
+``` llvm
+declare void @dx.op.debugBreak(
```
V-FEXrt, line 95: I still haven't internalized when we name new ops versus using the existing `unary/binary + opcode` but this seems like a perfect candidate for `nullary + opcode` https://en.wikipedia.org/wiki/Arity

```llvm
@dx.op.nullary(immarg i32 DEBUG_BREAK_OP)
```

proposals/0039-debugbreak.md
```diff
@@ -56,4 +56,70 @@ void main(uint GI : SV_GroupIndex) {
...
+### SPIRV Lowering
+
+This change will utilize the existing `NonSemantic.DebugBreak` instruction.
+While this instruction is not widely supported by Vulkan debuggers, it is
+supported by NVIDIA's NSight and can be safely ignored by Vulkan runtimes.
+
+The SPIRV usage will utilize the following instructions:
+
+```
+%1 = OpExtInstImport "NonSemantic.DebugBreak"
```
V-FEXrt, line 118: how does SPIRV manage the convergence of `NonSemantic.DebugBreak` is that also a consideration there or already handled?

proposals/0039-debugbreak.md
llvm-beanz: Yea... I'm not sure there's a good existing opcode class for this.

The nice benefit of giving it a named one is that it will make the DXIL less awful to read. We do have precedent with `IsHelperLane` for creating an opcode class for a single opcode that takes no arguments.

proposals/0039-debugbreak.md
llvm-beanz: SPIRV doesn't have an explicit convergent annotation, so it's really up to SPIRV-Tools to do something sane.

proposals/0039-debugbreak.md
```diff
@@ -56,4 +56,70 @@ void main(uint GI : SV_GroupIndex) {
 
 This aligns with C/C++ conventions that our users are already familiar with.
 
+## Detailed Design
+
+This proposal introduces a new HLSL `DebugBreak` intrinsic which has a
+runtime-defined behavior to facilitate shader debugging workflows. If the
+runtime does not support or is not configured to enable support for the
+corresponding DXIL instruction must be treated as a no-op by the driver.
```
spall: , it must

proposals/0039-debugbreak.md
```diff
@@ -56,4 +56,70 @@ void main(uint GI : SV_GroupIndex) {
...
+This change introduces a new DXIL operation:
+
+
+``` llvm
+declare void @dx.op.debugBreak(
+  immarg i32             ; opcode
+)
+```
+
+This DXIL operation must be treated as `convergent` even though it is not to
```
spall, line 100: this operation can't be moved but it doesn't prevent code being moved? (see line 121)

llvm-beanz: I tried to make the wording of this sentence more clear.

proposals/0039-debugbreak.md
tex3d: You do have to be careful when sharing OpCodeClass though. All opcodes in the same class need to have the same llvm attributes (like ReadNone etc) and should probably have the same properties/requirements for general reasoning about them, like convergent and such. Plus, as Chris said, there is some benefit to readability when using a new class.

We have not tried very much in the past to reuse OpCodeClass when a newly introduced operation has the same signature and properties as an existing operation. We certainly could have shared OpCodeClasses a lot more for existing ops than we have, but we have also made the mistake of accidentally sharing an OpCodeClass when the ops had different signatures and it shouldn't have been shared (see comment on `WavePrefixBitCount` where fortunately the overloads can't overlap so it doesn't cause any serious issue).

In this case, I think it makes sense to use a unique OpCodeClass. `nullary` implies nothing about the operation's llvm attributes/properties, which must match between all DXIL ops that share the OpCodeClass.

proposals/0039-debugbreak.md
tex3d: Perhaps the spec should be explicit that this operation should be treated as potentially reading/writing memory as well, to prevent DCE and reordering reads/writes across the `DebugBreak()`, so no `ReadNone` or `ReadOnly` attribute either.

And the code motion that `convergent` is meant to prevent would be movement of this instruction relative to surrounding control flow, right? I don't think there's any way to prevent movement of ordinary `ReadNone` instructions across this one, right?

proposals/0039-debugbreak.md
V-FEXrt: Yep makes sense! Thanks for the extra context. I definitely ran into some of these issue on the work I've been doing recently

damyanp (approved):

PR #610, opened 2025-08-25
Remove extra DXIL ops from LV spec

V-FEXrt:
Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/7712

DXIL Ops were added to the spec that don't make sense to accept vector parameters. This PR removes them.


llvm-beanz (approved):

alsepkow (commented):
lgtm

PR #615, opened 2025-08-28
[0027] Use new non-hit matrix description also for GetWorldToObject*

### LINKED ISSUE #496: [0027] References to 4x3 and 3x4 "identity" matrix misleading
[[0027] Shader Execution Reordering](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0027-shader-execution-reordering.md)

The spec references 4x3 and 3x4 identity matrices. It was pointed out to me that identity matrices must be square (see: https://en.wikipedia.org/wiki/Identity_matrix). I definitely didn't know this because I'm bad at math...

We should clarify exactly what those matrices are.

simoll:
Closes #496

### DIFF
--- a/proposals/0027-shader-execution-reordering.md
+++ b/proposals/0027-shader-execution-reordering.md
@@ -478,7 +478,7 @@ float3x4 dx::HitObject::GetWorldToObject3x4();
 
 Returns a matrix for transforming from world-space to object-space.
 
-Returns an identity matrix if the `HitObject` does not encode a hit.
+Returns a zero matrix with ones on the diagonal if the `HitObject` does not encode a hit.
 
 The only difference between this and `HitObject::GetWorldToObject4x3()` is
 the matrix is transposed – use whichever is convenient.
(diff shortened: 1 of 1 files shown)


damyanp (approved):

Icohedron (commented):
Could just be me, but I feel like "a zero matrix with ones on the diagonal" still isn't exactly clear what the resulting matrix looks like. Perhaps mention that the ones start from the top left and go down towards the bottom right of the matrix? Or more precisely: ones down the [main diagonal](https://en.wikipedia.org/wiki/Main_diagonal)

We could also define a name for these kinds of matrices. Like "partial identity matrix" or "rectangular identity matrix" and reuse the term.

V-FEXrt (approved):

damyanp:
I'm going to go ahead and merge this.

@Icohedron - thank you for your feedback.  If you feel strongly about this, please open a new issue (or even a PR with suggested changes).  This PR is updating a couple of places to use a new definition that was agreed upon, but were missed from the previous PR, and in that regard I think it's achieved its goal.

PR #622, opened 2025-09-02
[0030] Fix Op Class for vector dot,and,or

V-FEXrt:
Corrects the op class for the three new vector reduction operations previously specified.

These operations return their element type instead of a returning a vector matching the parameter type.

Also takes the opportunity to explicitly call out the allowed scalar types.

### DIFF
--- a/proposals/0030-dxil-vectors.md
+++ b/proposals/0030-dxil-vectors.md
@@ -138,7 +138,7 @@ This means that the same language-level vector (of any length) could be used
 
 #### `VectorReduce` OpCodeClass
 
-A new generic OpCodeClass `VectorReduce` is introduce for usage in new operations
+A new generic OpCodeClass `VectorReduce` is introduced for usage in new operations
 below. `VectorReduce` combines a vector of elements into a single element with
 the same type as the vector element type. The elements are combined using an
 operation specified by the opcode parameter.


damyanp (approved):
LGTM, but you should get a review from someone more familiar with this than me.

proposals/0030-dxil-vectors.md
```diff
@@ -142,41 +142,48 @@ This means that the same language-level vector (of any length) could be used
...
 
+The scalar type may be `i1`, `i16,` `i32`, or `i64`.
+
 ```C++
 DXIL::OpCode::VectorReduceAnd = 309
 ```
 
 ```asm
- [TYPE] @dx.op.unary.v[NUM][TY](309, <[NUM] x [TYPE]> operand)
+ [TYPE] @dx.op.unaryReduce.v[NUM][TY](309, <[NUM] x [TYPE]> operand)
```
tex3d: Maybe call out the new `OpCodeClass` explicitly?

proposals/0030-dxil-vectors.md
```diff
@@ -142,41 +142,48 @@ This means that the same language-level vector (of any length) could be used
...
+native vectors in DXIL `dot` can now be treated as a normal binary operation. Return
+type matches vector element type.
 
 Returns `op1[0] * op2[0] + op1[1] * op2[1] + ... + op1[NUM - 1] * op2[NUM - 1]`
 
+The scalar type may be `half` or `float`.
+
 ```C++
-DXIL::OpCode::VectorDotProduct = 311
+DXIL::OpCode::FDot = 311
```
tex3d, line 194: Does this need to have a `F` distinction? What if we decided to add int overloads? Couldn't we just use the same operation with int overload types? I don't think there can be any distinction between signed and unsigned int versions, without exposing a larger result.

proposals/0030-dxil-vectors.md
```diff
@@ -142,41 +142,48 @@ This means that the same language-level vector (of any length) could be used
...
 ```
 
 ```asm
- [TYPE] @dx.op.binary.v[NUM][TY](311, <[NUM] x [TYPE]> operand1, <[NUM] x [TYPE]> operand2)
+ [TYPE] @dx.op.binaryReduce.v[NUM][TY](311, <[NUM] x [TYPE]> operand1, <[NUM] x [TYPE]> operand2)
```
tex3d: Is there really need for a new general `OpCodeClass::BinaryReduce`, or could this be over-thinking it? What other binary reductions could there be?

V-FEXrt: This could very well be overthinking it but to me the cost is fairly minimal compared to the reverse. If we never need another binary reduce then we haven't lost anything but if we do need one in the future then we have to introduce a new class since the narrower naming wouldn't work.

proposals/0030-dxil-vectors.md
V-FEXrt: Right now, we already special case `int` lowering. I actually fully intended to support `int`s with the original FDot but after discussing with @llvm-beanz we decided to continue keeping `int`'s handled separately. 

Chris was saying we couldn't use the Dot op for everything because of the loss of signed/unsigned int knowledge but if that doesn't actually matter then I'm fine restoring it

proposals/0030-dxil-vectors.md
V-FEXrt: Just noting that a new op class was added and what it's shape is?

V-FEXrt: I'll just plan on adding a section about the new "vector reduction" classes once we make a decision on `binaryReduce`

tex3d: Sure, since it's a general class that we would likely use for other vector reductions in the future. I also think the OpCodeClass name VectorReduce might be more appropriate.  This matches llvm's intrinsic naming for such operations.  I think any non-unary reduction is going to be more specialized/complex, thus the unary case could be thought of as the general reduction, while other special cases would be named for their more specialized behavior.

proposals/0030-dxil-vectors.md
llvm-beanz: I don't believe it is in-scope for SM 6.9 to make signed integer wrapping well-defined. I think it is best to have this be `FDot`, and we can consider what we should do for integers in the future.

tex3d: I don't see this as a shader model change.

Signed integer wrapping was supposed to always be well-defined. It was well-defined for FXC and DXBC, and would have been for DXIL, if we hadn't forgotten to change default C behavior from Clang that emits `nsw` in DXIL for certain signed llvm instructions, which I consider to be a bug.

I see no advantage in committing ourselves to an unintended undefined behavior definition that could very easily and frequently be hit in the wild. We could easily modify DXC to retroactively stop emitting `nsw`, and I can't imagine it would create a regression.

In any case, if we used `Dot` instead of `FDot`, as I suggested, and somehow were forced to add `SDot` and `UDot` DXIL ops to differentiate signed/unsigned int operations, that would still work fine. There is no requirement to use `FDot` at this point to differentiate int types in the future.

We are only painting ourselves into a corner by using `FDot`, which forces us to use a different DXIL op for integers in the future, even if it's unnecessary.

tex3d: Ok, resolving this as SPIR-V already differentiates for signed/unsigned dot, so we are going to need to align anyway.

PR #628, opened 2025-09-08
[0040] Select type support proposal

kmpeng:
Language proposal for allowing Clang's `select` implementation to work with more general types (e.g. structs and arrays) instead of limiting it to the types allowed in DXC.


proposals/NNNN-select-type-support.md
```diff
@@ -0,0 +1,49 @@
+---
+title: "NNNN - Select type support"
+draft: true
```
damyanp: be sure to remove this before merging

```suggestion
```

PR #631, opened 2025-09-08
Add proposal for `groupshared` arguments

llvm-beanz:
This new proposal seeks to address a limitation in the HLSL language that prevents the HLSL standard library from being implementable in HLSL.

### DIFF
--- a/proposals/NNNN-groupshared-arguments.md
+++ b/proposals/NNNN-groupshared-arguments.md
@@ -1,6 +1,5 @@
 ---
 title: "NNNN - `groupshared` Arguments"
-draft: true
 params:
   authors:
     - llvm-beanz: Chris Bieneman
--- a/proposals/NNNN-groupshared-arguments.md
+++ b/proposals/NNNN-groupshared-arguments.md
@@ -1,0 +1,55 @@
+---
+title: "NNNN - `groupshared` Arguments"
+draft: true
+params:
+  authors:
+    - llvm-beanz: Chris Bieneman
+  sponsors:
+    - tbd: TBD
+  status: Under Consideration
+---
+
+* Planned Version: 202x
+
+## Introduction
+
+This proposal introduces a new use of the `groupshared` keyword for function
+arguments to allow passing `groupshared` arguments by address rather than by
+value or copy-in/copy-out.
+
+## Motivation
+
+DXC's implementation of HLSL includes a set of Interlocked functions
+implementing atomic operations on `groupshared` memory. These functions are not
+expressible in HLSL and rely on special case implementation in DXC.
+
+## Proposed solution
+
+HLSL 202x will allow the `groupshared` type annotation keyword on function
+parameter declarations. The keyword when applied to a parameter declaration of
+type `T`, alters the qualified type of the parameter to a `groupshared & T`
+(a reference to `groupshared` memory of type `T`).
+
+No implicit or explicit conversion can change the memory space of an object. To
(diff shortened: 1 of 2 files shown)


proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,55 @@
+---
+title: "NNNN - `groupshared` Arguments"
+draft: true
```
damyanp:
```suggestion
```

I need to figure out how to fix the template to not have this.  I don't want the templates to show up on the site, but I also don't want everyone copying the template to pick up the draft line!

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,55 @@
...
+DXC's implementation of HLSL includes a set of Interlocked functions
+implementing atomic operations on `groupshared` memory. These functions are not
+expressible in HLSL and rely on special case implementation in DXC.
+
+## Proposed solution
+
+HLSL 202x will allow the `groupshared` type annotation keyword on function
+parameter declarations. The keyword when applied to a parameter declaration of
+type `T`, alters the qualified type of the parameter to a `groupshared & T`
+(a reference to `groupshared` memory of type `T`).
```
damyanp: This only works / makes sense on function params?

What about struct members? Local vars? etc.?

llvm-beanz: Only function parameters. We currently only allow `groupshared` on global variable declarations, and this proposal shouldn't widen that any further.

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,54 @@
...
+
+DXC's implementation of HLSL includes a set of Interlocked functions
+implementing atomic operations on `groupshared` memory. These functions are not
+expressible in HLSL and rely on special case implementation in DXC.
+
+## Proposed solution
+
+HLSL 202x will allow the `groupshared` type annotation keyword on function
+parameter declarations. The keyword when applied to a parameter declaration of
+type `T`, alters the qualified type of the parameter to a `groupshared & T`
```
Flakebi: nit/question: Should the altered type rather be `groupshared T &`?
Thinking of C++, where it would look like `groupshared T &myvar` with identifier.

llvm-beanz: Yep. You're correct. I'll updated the language.

proposals/NNNN-groupshared-arguments.md
```diff
@@ -0,0 +1,54 @@
...
+HLSL 202x will allow the `groupshared` type annotation keyword on function
+parameter declarations. The keyword when applied to a parameter declaration of
+type `T`, alters the qualified type of the parameter to a `groupshared T &`
+(a reference to `groupshared` memory of type `T`).
+
+No implicit or explicit conversion can change the memory space of an object. To
+perform such a conversion, a user must declare a new object in the destination
+memory space and initialize it appropriately. For overload resolution, the
+parameter type must be an exact match in order for overload resolution to
+succeed since no conversions will be valid.
```
tex3d: Regarding the interaction of this feature with the existing LValue casting in HLSL, it should not be possible to perform an LValue cast changing the type/shape of a groupshared variable, resulting in a groupshared LValue reference.

llvm-beanz: I'll make sure to include rules about reference casting in the detailed design.

PR #634, opened 2025-09-09
Array accesses and bounds

### LINKED ISSUE #141: Remove unsized arrays
**Is your feature request related to a problem? Please describe.**
We allow unsized arrays in most contexts that C would, but they sometimes cause crashes in the compiler and other unexpected behaviors. Unsized arrays also break the basic rules of HLSL parameter passing (which isn't great).

**Describe the solution you'd like**
We need to design a way to support unsized arrays (or something functionally equivalent) for resource types, but we should remove support for unsized arrays from the language (at least until we support general pointers).

**Describe alternatives you've considered**
In Clang we're more clearly defining the semantics for unsized arrays which will allow us to avoid some of the bugginess in DXC. We could stick with that approach, but it still leaves the language handling of unsized array parameters to be a bit awkward.

**Additional context**
[See this documentation PR.](https://github.com/llvm/llvm-project/pull/75397)

llvm-beanz:
This adds language documenting that out of bounds accesses for arrays is undefined behavior, and that zero-sized arrays are disallowed.

This also restricts arrays of unknown size to strictly global resource declarations but allows omitting the array bound if it can be inferred from an initiailzer.

Resolves #141

### DIFF
--- a/specs/language/declarations.tex
+++ b/specs/language/declarations.tex
@@ -52,7 +52,7 @@
 
 \begin{grammar}
   \define{function-specifier}\br
-  export
+  \terminal{export}
 \end{grammar}
 
 \p The \texttt{export} specifier denotes that the function has program linkage (\ref{Basic.Linkage.Program}).


specs/language/declarations.tex
```diff
@@ -56,6 +56,64 @@
...
+\Sub{Arrays}{Decl.arrays}
+% This might need to nest deeper as we flesh out the sections.
+
+\p A declaration \texttt{T D} where \texttt{D} is of the form
+
+\begin{grammar}
+\textit{D} \terminal{[} \textit{N} \terminal{]} \opt{attribute-specifier-seq}\br
+\end{grammar}
+
+declares an array of \texttt{N} \texttt{T}, where \texttt{N} is the \textit{array
```
hekota: What about multi-dimensional arrays?

specs/language/declarations.tex
```diff
@@ -56,6 +56,74 @@
...
+\Sub{Arrays}{Decl.arrays}
+% This might need to nest deeper as we flesh out the sections.
+
+\p A declaration \textit{T D} where \textit{D} is of the form
+
+\begin{grammar}
+\textit{D`} \terminal{[} \textit{N} \terminal{]} \opt{attribute-specifier-seq}\br
+\end{grammar}
+
+declares an array of \textit{N} \textit{T`}, where \textit{N} is the
```
spall: what is the relationship between T and T'?

specs/language/declarations.tex
```diff
@@ -56,6 +56,74 @@
...
+\p The contained type of an array of either form shall be a complete type.
+
+\begin{note}
+  An array of arrays may be declared with infinite nesting of array declarators
+  as long as either all array bounds are known, or only the outermost array
+  bound is unknown. This is codified in the spec above with the requirement that
+  the contained type of an array be a complete type.
+\end{note}
+
+\p A declarator with an omitted array bound is valid if the when declaring an
```
spall: remove 'if the' before when? or remove 'the when' after if?

specs/language/declarations.tex
```diff
@@ -56,6 +56,74 @@
...
+object of array type which is not a non-static data member and is initialized
+with an initializer that follows the declarator. In this case the array bound is
```
spall: is this only allowed for global declarations of resource objects? or for all array declarations?

specs/language/declarations.tex
llvm-beanz: I'll see if I can reword this. The confusing thing is for cases like `int Arr[4][4]` where there is nesting occurring. `T` would be `int`, but you have two array declarators one where `D` is `Arr[4]` and T\` is `int[4]`, and one where `D` is `Arr` and T\` is `int`.

specs/language/declarations.tex
llvm-beanz: If we can imply an array bound from an initializer it can be used with any type. This would be for cases like `int Arr[] = {1,2,3}` to define an array of 3 `int`s.

spall (approved):

PR #635, opened 2025-09-09
[New] Introduce proposal to restrict unbound arrays

llvm-beanz:
This proposal seeks to restrict declarators for arrays without an array bound to only global resource declarations and array variables with an initializer where the bound is implied by the initializer.


V-FEXrt (approved):

V-FEXrt:
> LGTM. Thoughts on whether or not this should be backported to DXC?

202X implies DXC backport right?

llvm-beanz:
> LGTM. Thoughts on whether or not this should be backported to DXC?

I think this should be simple enough that we should consider it for DXC.

PR #649, opened 2025-09-18
[0035] Matrix dimensions HLSL and container handling

### LINKED ISSUE #586: [0035] Specify Matrix dimension constraints

There likely should be some constraints around the dimensions of matrices. The spec currently does not document any.

llvm-beanz:
This adds some size constriants for matrices in the HLSL types as well as a draft for how the matrix shapes and types will be captured in the PSV runtime information for the D3D runtime to inspect.

Fixes #586

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -7,9 +7,7 @@ params:
   status: Under Review
 ---
 
-
- 
-* Planned Version: SM 6.x
+* Planned Version: SM 6.10
 
 ## Introduction
 
@@ -411,6 +409,33 @@ columns.
 
 An Accumulator matrix may be either an A matrix, or a B matrix, and it varies by
 hardware implementation.
+
+#### Restrictions on Dimensions
+
+The HLSL API will enforce restrictions on the `K` dimension as found in the
+formula: `MxK * K*N = MxN`
+
+This restriction impacts the number of rows in an A matrix, and columns in a B
+matrix, but has no impact on an accumulator matrix.
+
+The minimum and maximum `K` dimension for Wave and Thread scope matrices is tied
+to the the minimum and maximum wave size, while the minimum and maximum `K`
+dimension for ThreadGroup matrices is tied to the thread group size.
+
+
+| Matrix Scope | Scalar element dimensions     |
+| ------------ | ----------------------------- |
+| Thread       | Powers of two between [4,128] |
+| Wave         | Powers of two between [4,128] |
+| ThreadGroup  | [1,1024]                      |
+
+Sizes for matrices of packed data types are 4 times the valid size for a scalar
+element.
+
+Not all hardware is required to support all possible dimensions for thread and
+wave scope matrices, or all possible element types. The shader compiler will
+encode the dimensions and input and output data types used by each shader in the
+[Pipeline State Validation metadata](#pipeline-state-validation-metadata).
 
 ### HLSL API Documentation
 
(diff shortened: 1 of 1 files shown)


proposals/0035-linalg-matrix.md
```diff
@@ -412,6 +410,33 @@ columns.
 An Accumulator matrix may be either an A matrix, or a B matrix, and it varies by
 hardware implementation.
 
+#### Restrictions on Dimensions
+
+The HLSL API will enforce restrictions on the `K` dimension as found in the
+formula: `MxK * K*N = MxN`
```
jenatali:
```suggestion
formula: `MxK * KxN = MxN`
```

proposals/0035-linalg-matrix.md
```diff
@@ -1312,6 +1337,45 @@ two input vectors. The matrix scope can be `Thread`, `Wave`, or `ThreadGroup`.
...
+
+The `MatrixUse` object is defined:
+
+```c
+struct MatrixUse {
+  uint32_t Dimensions[3]; // M, N, K
+  uint32_t Scope;
+  uint32_t Flags; // do we need this?
+  uint32_t OperandType;
+  uint32_t ResultType;
```
jenatali: I'd be tempted to bit-pack a bit more here. Maybe premature, but dimensions can't exceed `uint16_t`, and similarly the other types are enums with < 255 entries so could be `uint8_t`.

llvm-beanz: I'm not sure we've settled on a maximum bound for the M or N dimensions. `uint16_t` is probably reasonable for thread and wave scope, but there are definitely larger matrices we may want to support with thread group scope that could be well in excess of that. I'm hesitant to put us into any corners.

The scope and type enums could definitely likely be smaller safely. I'll take a stab at fleshing out the enum values for the type enumeration and condensing the structure a bit. We do need it to be an even number of dwords so that we don't mess up alignment when packing.

proposals/0035-linalg-matrix.md
```diff
@@ -412,6 +410,33 @@ columns.
...
+#### Restrictions on Dimensions
+
+The HLSL API will enforce restrictions on the `K` dimension as found in the
+formula: `MxK * KxN = MxN`
+
+This restriction impacts the number of rows in an A matrix, and columns in a B
+matrix, but has no impact on an accumulator matrix.
+
+The minimum and maximum `K` dimension for Wave and Thread scope matrices is tied
+to the the minimum and maximum wave size, while the minimum and maximum `K`
```
damyanp, line 422: "wave size" here I assume is referring to what is called "wave lane count" in [the docs](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_feature_data_d3d12_options1)?

damyanp (approved):

proposals/0035-linalg-matrix.md
jenatali: Oh right, there's only maximums for K, so maybe we keep those as `uint32_t[3]` then, and then just pack the enums together. That should all fit in one dword, making the total structure 4 dwords / 16 bytes.

proposals/0035-linalg-matrix.md
llvm-beanz: Yea, we've also called it wave size in other docs and language features (https://github.com/microsoft/directxshadercompiler/wiki/wave-intrinsics, https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_WaveSize.html)

proposals/0035-linalg-matrix.md
```diff
@@ -412,6 +410,33 @@ columns.
 An Accumulator matrix may be either an A matrix, or a B matrix, and it varies by
 hardware implementation.
 
+#### Restrictions on Dimensions
+
```
mbedy-amd, line 414: I realize this PR is already merged, but in reading this section, I wasn't sure how to interpret it. Could you expand on the details of the proposed restrictions?

llvm-beanz: An `A` matrix's dimensions are MxK, and a `B` matrix's dimensions are KxN. The dimension restriction only applies to the _K_ dimension, and the only requirement at the HLSL level is that they be powers of 2 between 4 and 128.

This is effectively to require that the matrix dimension align with the wave size. For thread group scope the size is allowed to go up to 1024 since that is the maximum thread count for DX today. I suppose we could allow that to be larger if we wanted.

mbedy-amd: Is the correct interpretation of this proposal that if wave size == 32, K must be 32?

PR #651, opened 2025-09-19
[0035] Rework elementwise operations

### LINKED ISSUE #623: [0035] Revisit Per-Element Matrix Operations
[0035 - linalg::Matrix](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

The current list of unary operations provided by [ApplyUnaryFunction](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md#matrixapplyunaryoperation) and binary operations provided by Matrix class operators are not sufficient for activation and reduction functions.  Furthermore, a programmable method would be preferred over a set of built-in functions.

The possible solutions could be:

1. **Preferred** A `map(func, Matrix)` function which takes a function reference as an argument.  The arguments of the mapped function would provide row and col information, or an element index that can be converted to row and col through an intrinsic function.
2. A `Matrix::Length` function, which returns the number of elements per-thread, for iterating over, along with a `Matrix::operator[]` for accessing the elements using a per-thread index.
3. Extend the current `enum class UnaryOperation` with additional unary operations (e.g., `sqrt`, `ceil`, `reciprocal`, `relu`, `sigmoid`, `tanh`, etc.) and add a `ApplyBinaryOperation` function with `enum class BinaryOperation` (e.g., `max`, `pow`, `atan2`, `hypot`, etc.).

Option 1 provides the most flexibility and also enables additional use cases.  While Option 3 is the most limiting for developers and requires constant updates to the specification.

**Additional context**
See also [PR 598](https://github.com/microsoft/hlsl-specs/pull/598#pullrequestreview-3107549616 ) comment from @nhaehnle .

llvm-beanz:
This revises elementwise operations to more closely align with SPIRV. The main changes here are:

* Removal of the APIs to create a matrix from per-thread vectors.
* Removal of the APIs to split a matrix into per-thread vectors.
* Removal of ApplyUnaryOperation API.
* Addition of Length API to query the number of elements stored locally in a thread.
* Addition of GetCoordinate API to convert a thread-local index to column and row.
* Addition of Get and Set APIs that use thread-local indices.
* Elementwise addition, subtraction, multiply and divide are implemented in HLSL, removing the binop DXIL operations.

Fixes #623

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -737,8 +737,8 @@ Matrix::Length();
 
 Returns the number of matrix components accessible to the current thread. The
 mapping and distribution of threads to matrix elements is opaque and
-implementation-specific. The value returned by `Length` may not be the same
-across all threads. The sum of the values returned by `Length` across all
+implementation-specific. The value returned by `Length` may be different
+for each thread. The sum of the values returned by `Length` across all
 threads must be greater than or equal to the total number of matrix elements.
 Some implementations may map multiple threads to the same matrix element.
 Therefore, developers should take this into consideration when programming


mapodaca-nv:
I have a draft that is 90% ready.  I will submit a PR early nest week for review.

damyanp (approved):

proposals/0035-linalg-matrix.md
```diff
@@ -734,38 +728,51 @@ This operation may be called in divergent control flow when loading a thread
...
+#### Matrix::Length
+
+```c++
+template <MatrixScope ScopeLocal = Scope>
+typename hlsl::enable_if<Scope != MatrixScope::Thread && ScopeLocal == Scope,
+                         uint>::type
+Matrix::Length();
+```
+
+Returns the number of components stored in the thread's thread local memory.
```
mapodaca-nv: Consider the following:

```md
Returns the number of matrix components accessible to the current thread. The
mapping and distribution of threads to matrix elements is opaque and
implementation-specific. The value returned by `Length` may not be the same
across all threads. The sum of the values returned by `Length` across all
threads must be greater than or equal to the total number of matrix elements.
Some implementations may map multiple threads to the same matrix element.
Therefore, developers should take this into consideration when programming
side-effects, such as atomic operations and/or UAV writes, within user-defined
matrix operations.

May be called from non-uniform control flow. However, given the above rules,
calling `Length` from divergent threads may result in unpredictable behavior.
For example, the number of matrix elements accessible to each thread will be
inconsistent across different implementations.
```

proposals/0035-linalg-matrix.md
```diff
@@ -734,38 +728,51 @@ This operation may be called in divergent control flow when loading a thread
...
+                         uint2>::type Matrix::GetCoordinate(uint);
 ```
 
-Produces a matrix from per-thread vectors. An A matrix is produced from
-per-thread column vectors, while a B matrix is produced from per-thread row
-vectors. The `FromThreadVectors` construction method is not available for
-accumulator matrices which vary by hardware implementation.
+Converts an index into the thread-local collection of matrix elements to the
+matrix column and row returned as a `uint2`. The mapping of thread local indices
+to matrix coordinates is implementation defined.
```
mapodaca-nv: Consider the following:

```md
Converts a specified index into row and column coordinates.  The valid
range of `Index` is `[0, Length()-1]`.  If the value of `Index` is out of
range, then the result value is undefined.
```

proposals/0035-linalg-matrix.md
```diff
@@ -734,38 +728,51 @@ This operation may be called in divergent control flow when loading a thread
...
-dimension must be less than or equal to the thread group size.
+```c++
+template <MatrixScope ScopeLocal = Scope>
+typename hlsl::enable_if<Scope != MatrixScope::Thread && ScopeLocal == Scope,
+                           ElementType>::type Matrix::Get(uint);
+```
 
-Threads outside the matrix size are discarded.
+Gets the element corresponding to the thread-local index provided. The mapping
+of thread local indices to matrix coordinates is implementation defined.
```
mapodaca-nv: Consider the following:

```md
Retrieves the value of a matrix component at the specified index.  The valid
range of `Index` is `[0, Length()-1]`.  If the value of `Index` is out of
range, then the result value is undefined.
```

proposals/0035-linalg-matrix.md
```diff
@@ -734,38 +728,51 @@ This operation may be called in divergent control flow when loading a thread
...
+
+```c++
+template <MatrixScope ScopeLocal = Scope>
+typename hlsl::enable_if<Scope != MatrixScope::Thread && ScopeLocal == Scope,
+                         void>::type Matrix::Set(uint, ElementType);
+```
+
+Sets the element at the corresponding thread-local index to the provided value.
+The mapping of thread local indices to matrix coordinates is implementation
+defined.
```
mapodaca-nv: Consider the following:

```md
Sets the value of a matrix component at the specified index.  The valid
range of `Index` is `[0, Length()-1]`.  If the value of `Index` is out of
range, then the result value is undefined.
```

mapodaca-nv (commented):
Adding a few alternatives to the language to consider from our proposal.

proposals/0035-linalg-matrix.md
```diff
@@ -1534,35 +1464,48 @@ class Matrix {
...
-  // Apply a unary operation to each element.
-  template <UnaryOperation Op, MatrixScope ScopeLocal = Scope>
-  typename hlsl::enable_if<Scope != MatrixScope::Thread && ScopeLocal == Scope,
-                           Matrix>::type
-  ApplyUnaryOperation();
+  operator/=(T V) {
+    for (uint I = 0; I < Length(); ++I)
+      Set(I, Get(I) / V);
+    return this;
+  }
```
pow2clk, line 1521: I expect you're aware, but for the sake of those who may not be, these operators are not supported by HLSL2021.

proposals/0035-linalg-matrix.md
llvm-beanz: Does this need to be undefined or can we define it as a no-op?

proposals/0035-linalg-matrix.md
llvm-beanz: Yes, and with the current API shape we really have two viable options. We could just not include these APIs (since they're trivial to implement), or we could propose relaxing the overload restrictions in HLSL 202x.

mapodaca-nv:
Remnant `Matrix::FromThreadVectors()` and `Matrix::GetThreadVector()` in Matrix Scope table.

https://github.com/microsoft/hlsl-specs/pull/651/files#diff-2821c00225d37c2253fbe4c219e02a45a7b309a9e2d1b9913ea1a09902ae595cR405

proposals/0035-linalg-matrix.md
```diff
@@ -126,17 +139,22 @@ class Matrix {
...
        MatrixLayout Layout);
 
-  template <MatrixUse UseLocal = Use>
-  typename hlsl::enable_if<Use == MatrixUse::A &&
-                               Scope != MatrixScope::Thread && UseLocal == Use,
-                           Matrix>::type
-      FromThreadVectors(vector<ElementType, MScalars>);
+  template <MatrixScope ScopeLocal = Scope>
+  typename hlsl::enable_if<Scope != MatrixScope::Thread && ScopeLocal == Scope,
+                           uint>::type
```
spall, line 144: nit: should Length() be on this line to match the others? looks like there is enough room.

proposals/0035-linalg-matrix.md
```diff
@@ -734,38 +726,66 @@ This operation may be called in divergent control flow when loading a thread
...
+template <MatrixScope ScopeLocal = Scope>
+typename hlsl::enable_if<Scope != MatrixScope::Thread && ScopeLocal == Scope,
+                         uint>::type
+Matrix::Length();
+```
+
+Returns the number of matrix components accessible to the current thread. The
+mapping and distribution of threads to matrix elements is opaque and
+implementation-specific. The value returned by `Length` may not be the same
+across all threads. The sum of the values returned by `Length` across all
```
V-FEXrt: nit: I think the positive is easier to parse

```suggestion
implementation-specific. The value returned by `Length` may be different
for each thread. The sum of the values returned by `Length` across all
```

V-FEXrt (approved):

proposals/0035-linalg-matrix.md
llvm-beanz: I do not know why, but clang-format is doing this one different.

PR #678, opened 2025-10-13
[HLSL] define the matrix element accessor semantics

### LINKED ISSUE #677: [HLSL][Matrix] Document the `._<numeric><numeric>` and ._m<numeric><numeric> accessors
Coresponding with this change https://github.com/llvm/llvm-project/issues/159438

We need a new section to indicate which accessors of matrix types are legal in the spec.

farzonl:
fixes #677

Moved the Vector Swizzle stuff that repeats with Matrix and moved it to a generic swizzle section.
image

The new Matrix Swizzle looks like this:
image

image

### DIFF
--- a/specs/language/expressions.tex
+++ b/specs/language/expressions.tex
@@ -245,7 +245,7 @@ postfix expression before the dot.
   matrix-zero-indexed-swizzle-sequence matrix-zero-indexed-swizzle\br
 
   \define{matrix-zero-indexed-swizzle}\br
-  \terminal{\_m} zero-index-value\br
+  \terminal{\_m} zero-index-value zero-index-value\br
 
   \define{zero-index-value} \textnormal{one of}\br
   \terminal{0 1 2 3 }\br
@@ -255,7 +255,7 @@ postfix expression before the dot.
   matrix-one-indexed-swizzle-sequence matrix-one-indexed-swizzle\br
 
   \define{matrix-one-indexed-swizzle}\br
-  \terminal{\_m} one-index-value\br
+  \terminal{\_} one-index-value one-index-value\br
 
   \define{one-index-value} \textnormal{one of}\br
   \terminal{1 2 3 4 }\br


specs/language/expressions.tex
```diff
@@ -167,6 +167,22 @@
 (\ref{Overload}).\footnote{HLSL does not support the base address of a subscript
 operator being the expression inside the braces, which is valid in C and C++.}
 
+\Sec{Matrix Member Expression(s)}{Expr.Post.Member.Matrix}
+\begin{grammar}
+  \define{postfix-expression}\br
+  primary-expression\br
+  postfix-expression [ expression ][ expression ]\br %
+  postfix-expression \terminal{.\_<integer-literal><integer-literal>}\br
+  postfix-expression \terminal{.\_m<integer-literal><integer-literal>}\br
```
farzonl: What is the propper way in the grammar to specify that `._m` and `._` can happen multiple times like in vector swizzle?

llvm-beanz: You'll want something like:
```suggestion
  postfix-expression matrix-zero-indexed-accessor\br
  postfix-expression matrix-one-indexed-accessor\br
  
\define{matrix-zero-indexed-accessor}\br
  \opt{matrix-zero-indexed-accessor} \terminal{_m} integer-literal\br

\define{matrix-one-indexed-accessor}\br
  \opt{matrix-one-indexed-accessor} \terminal{_} integer-literal\br
```

llvm-beanz: Oh, we also should not use `integer-literal` here, we should define separate index grammar elements because the valid characters are `0 1 2 3` and `1 2 3 4` not the full space of `integer-literal` which is more like `[0-9]+`.

specs/language/expressions.tex
```diff
@@ -167,6 +167,22 @@
...
+\end{grammar}
+\begin{itemize}
+\item \textbf{.\_} (math-style): subsequent subscripts use \textbf{1-based}
+ indexing for both row and column.
+\item \textbf{.\_m} (memory-style): subsequent subscripts use \textbf{0-based}
+ indexing for both row and column. This is equivalent to the double square
+ brackets subscript expression of the form \texttt{E0[E1][E2]}.
+\end{itemize}
```
llvm-beanz, line 268: We should generally avoid describing something in terms of something that isn't described in the language spec unless it is inherited behavior from C/C++. We don't have a description of matrix subscripts in the spec and they don't behave the same as array subscripts.
 
That aside, these aren't subscripts and they don't behave like subscripts. Generally speaking subscripts always turn into address math producing an address which is de-referenced to produce an l-value. Swizzles don't always produce an l-value. For example:

```
export void fn() {
    float2x2 F = {1,2,3,4};
    F._m00_m01 = 1.xx; // swizzle is an l-value
    F._m00_m00 = 2.xx; // swizzle is an r-value
}
``` 
[Compiler Explorer](https://godbolt.org/z/79n8Wjs6b)

A swizzle that contains repeated elements will produce an r-value, otherwise the swizzle should produce an l-value.

specs/language/expressions.tex
farzonl: need to escape the `_` via `\_` otherwise my latex compiler breaks.

As an asside I'm also seeing this warning everywhere
- `Underfull \hbox (badness 10000) in paragraph at lines 172--181`
- `Underfull \hbox (badness 10000) in paragraph at lines 129--131`

Doesn't seem to impact spec generation so I'm going to ignore it for now.

specs/language/expressions.tex
```diff
@@ -167,6 +167,26 @@
...
+  \define{postfix-expression}\br
+  primary-expression\br
+  postfix-expression [ expression ][ expression ]\br %
+  postfix-expression matrix-zero-indexed-accessor\br
+  postfix-expression matrix-one-indexed-accessor\br
+  \define{matrix-zero-indexed-accessor}\br
+    \opt{matrix-zero-indexed-accessor} \terminal{\_m} integer-literal\br
+  \define{matrix-one-indexed-accessor}\br
+    \opt{matrix-one-indexed-accessor} \terminal{\_} integer-literal\br
+\end{grammar}
```
llvm-beanz, line 262: I think I set you on the wrong path for the grammar. The grammar should probably be something more like:

```
\begin{grammar}
  \define {matrix-accessor-sequence}\br
  matrix-zero-indexed-accessor-sequence\br
  matrix-one-indexed-accessor-sequence\br

  \define{matrix-zero-indexed-accessor-sequence}\br
  matrix-zero-indexed-accessor\br
  matrix-zero-indexed-accessor-sequence matrix-zero-indexed-accessor\br

  \define{matrix-zero-indexed-accessor}\br
  \terminal{\_m} zero-index-value\br

  \define{zero-index-value}\br
  \terminal{0}\br \terminal{1}\br \terminal{2}\br \terminal{3}\br

  \define{matrix-one-indexed-accessor-sequence}\br
  matrix-one-indexed-accessor\br
  matrix-one-indexed-accessor-sequence matrix-one-indexed-accessor\br

  \define{matrix-one-indexed-accessor}\br
  \terminal{\_m} one-index-value\br

  \define{one-index-value}\br
  \terminal{1}\br \terminal{2}\br \terminal{3}\br \terminal{4}\br

\end{grammar}
```

Two things to note:

(1) you're not defining `postfix-expression`, that's defined above, you should be adding a formulation something like `postix-expression \terminal{.} matrix-accessor-sequence`.

(2) We should describe subscripts on matrices separately because they behave quite differently, but you actually don't need to describe the grammar since it works recursively from the already defined `[]` grammars in Expr.Post.

farzonl: Should we define `zero-index-value` and `one-index-value`  in `Lex.Literal.Int ` and reuse concepts like nonzero-digit

I had a thought to define it as 
```latex
\define{one-index-value}\br
  nonzero-dim-constraint-digit}\br

\[
\begin{array}{rcl}
\textit{nonzero-digit} &::=& 1 \mid 2 \mid 3 \mid 4 \mid 5 \mid 6 \mid 7 \mid 8 \mid 9 \\[4pt]
\textit{nonzero-dim-constraint-digit} &::=& \textit{nonzero-digit} \quad \text{where value} \le 4
\end{array}
\]
```

llvm-beanz: I wouldn't put them in the Lex chapter. They're only valid grammar constructs in this narrow space, and I think it's a lot simpler and clearer to read to just list the four possible values for each context.

specs/language/expressions.tex
```diff
@@ -167,6 +167,39 @@
...
+
+  \define{one-index-value}\br
+  \terminal{1 2 3 4 }\br
+\end{grammar}
+\begin{itemize}
+\item \textbf{.\_} (math-style): subsequent subscripts use \textbf{1-based}
+ indexing for both row and column.
+\item \textbf{.\_m} (memory-style): subsequent subscripts use \textbf{0-based}
+ indexing for both row and column.
+\end{itemize}
```
llvm-beanz, line 268: I think we need to dig into more of the semantics here. There are some key points to highlight:
1) A _postfix-expression_ followed by a `.` followed by a _matrix-accessor-sequence_ is a postfix expression (i.e. it can be used recursively).
2) The expression before the `.` must be an expression of matrix type
3) If the expression before the `.` is an lvalue and the matrix accessor sequence contains no duplicate members, the result is an lvalue, otherwise it is a PR value.
4) Out-of-range matrix accessors are ill-formed.

llvm-beanz: Oh, and we should detail the type of the swizzle expression being defined as a vector of the same element type as the matrix with the number of components specified in the matrix accessor sequence.

specs/language/expressions.tex
```diff
@@ -230,6 +234,46 @@
 refers to an element index that is out of range of the vector type of the
 postfix expression before the dot.
 
+\Sec{Matrix Swizzle}{Expr.Post.Swizzle.Matrix}
+\begin{grammar}
+  \define{matrix-swizzle-component-sequence}\br
+  matrix-zero-indexed-swizzle-sequence\br
+  matrix-one-indexed-swizzle-sequence\br
```
farzonl, line 241: @llvm-beanz  One unique thing I was thinking about indicating is that a matrix  swizzle can not be more than 4 elements. That seems to be implied later when we indicate that a swizzle expression is a vector. However, with long vectors coming in SM 6.9 does this 4 element limitation on matrix swizzles go away? https://hlsl.godbolt.org/z/r9o5PKPWr because I am still seeing it in godbolt: https://hlsl.godbolt.org/z/fdnWqq7cK.

llvm-beanz: We also don't allow vector swizzles to be more than 4 elements. I think we should keep the 4 element restriction.

specs/language/expressions.tex
```diff
@@ -173,7 +174,19 @@
 elements in the vector, the program is ill-formed. If the value is outside the
 range at runtime, the behavior is undefined.
 
-\Sec{Vector Swizzle}{Expr.Post.VectorSwizzle}
+\Sec{Swizzle}{Expr.Post.Swizzle}
+\p A \textit{postfix-expression} followed by a dot (\texttt{.}) and a sequence
+of one or more \textit{swizzle-components} is a postfix expression. The
```
llvm-beanz: Probably should have caught this in my PR, but this shouldn't be hyphenated since it isn't a grammar construct.
```suggestion
of one or more \textit{swizzle components} is a postfix expression. The
```

specs/language/expressions.tex
```diff
@@ -173,7 +174,19 @@
 elements in the vector, the program is ill-formed. If the value is outside the
 range at runtime, the behavior is undefined.
 
-\Sec{Vector Swizzle}{Expr.Post.VectorSwizzle}
+\Sec{Swizzle}{Expr.Post.Swizzle}
```
llvm-beanz:
```suggestion
\Sec{Swizzle Expressions}{Expr.Post.Swizzle}
```

specs/language/expressions.tex
```diff
@@ -173,7 +174,19 @@
...
+postfix expression before the dot is evaluated and must be of vector or matrix type.
+If the postfix expression before the dot is an lvalue, the swizzle expression may
+produce either an lvalue or a prvalue; otherwise it produces a prvalue.
+
+
+\p If the postfix expression before the dot is
+an lvalue and the \textit{swizzle-component-sequence} contains no repeated
+components, the swizzle expression is an lvalue; otherwise it is a prvalue.
+
+\Sec{Vector Swizzle}{Expr.Post.Swizzle.Vector}
```
llvm-beanz:
```suggestion
\Sub{Vector Swizzle}{Expr.Post.Swizzle.Vector}
```

specs/language/expressions.tex
```diff
@@ -230,6 +234,75 @@
 refers to an element index that is out of range of the vector type of the
 postfix expression before the dot.
 
+\Sec{Matrix Swizzle}{Expr.Post.Swizzle.Matrix}
```
llvm-beanz:
```suggestion
\Sub{Matrix Swizzle}{Expr.Post.Swizzle.Matrix}
```

specs/language/expressions.tex
```diff
@@ -230,6 +234,75 @@
...
+
+  \define{matrix-zero-indexed-swizzle}\br
+  \terminal{\_m} zero-index-value\br
+
+  \define{zero-index-value} \textnormal{one of}\br
+  \terminal{0 1 2 3 }\br
+
+  \define{matrix-one-indexed-swizzle-sequence}\br
+  matrix-one-indexed-swizzle\br
+  matrix-one-indexed-swizzle-sequence matrix-one-indexed-swizzle\br
```
V-FEXrt, line 255: I'm not exactly sure the best way to rewrite this, but as written this would require  `_m1_m2` instead of `_m12` right?

specs/language/expressions.tex
```diff
@@ -230,6 +234,75 @@
...
+
+  \define{matrix-one-indexed-swizzle}\br
+  \terminal{\_m} one-index-value\br
```
V-FEXrt:
```suggestion
  \terminal{\_} one-index-value\br
```

Should be just `_` not `_m` for 1-indexing right?

specs/language/expressions.tex
```diff
@@ -230,6 +234,75 @@
...
 
+\Sub{Matrix Swizzle}{Expr.Post.Swizzle.Matrix}
+\begin{grammar}
+  \define{matrix-swizzle-component-sequence}\br
+  matrix-zero-indexed-swizzle-sequence\br
+  matrix-one-indexed-swizzle-sequence\br
+
+  \define{matrix-zero-indexed-swizzle-sequence}\br
+  matrix-zero-indexed-swizzle\br
+  matrix-zero-indexed-swizzle-sequence matrix-zero-indexed-swizzle\br
```
V-FEXrt, line 245: I'm not exactly sure the best way to rewrite this, but as written this would require `_m1_m2` instead of `_m12` right?

specs/language/expressions.tex
```diff
@@ -230,6 +234,75 @@
...
+\end{grammar}
+\begin{itemize}
+\item \textbf{.\_} (math-style): subsequent subscripts use \textbf{1-based}
+ indexing for both row and column.
+\item \textbf{.\_m} (memory-style): subsequent subscripts use \textbf{0-based}
+ indexing for both row and column.
+\end{itemize}
+
+\p A \textit{matrix-swizzle-component-sequence} is a sequence of one but less
+than or equal to four swizzle components of either the
```
V-FEXrt, line 271: imo it would be better to encode this into the grammar

its not too crazy to write

```
matrix-one-indexed-swizzle-sequence:
  "_m", zero-index-value
  | "_m", zero-index-value,  zero-index-value
  | "_m", zero-index-value,  zero-index-value,  zero-index-value
  | "_m", zero-index-value,  zero-index-value,  zero-index-value,  zero-index-value
```

which also solves the `_m1_m2` issue mentioned above

specs/language/expressions.tex
farzonl: I think this can be fixed by just doing 
```latex
  \define{matrix-zero-indexed-swizzle}\br
  \terminal{\_m} zero-index-value zero-index-value\br
```
then it should be symantically similar to 
```hlsl
._m00_m01_m02_m03 
```

specs/language/expressions.tex
farzonl: same solution for zero index will just do 
```latex
  \define{matrix-one-indexed-swizzle}\br
  \terminal{\_} one-index-value one-index-value\br
```

then it should be symantically similar to 
```hlsl
._11_12_13_14 
```

specs/language/expressions.tex
farzonl: Most these comments are variations of the same concern. I beleive i've addressed a solution.

PR #680, opened 2025-10-14
[0041] Update triangle hit pos return values for HLSL and DXIL

jenatali:
Based on offline discussion, I think I've been convinced that this is the right way to go. Update the spec language and the open issues.

### DIFF
--- a/proposals/0041-triangle-object-positions.md
+++ b/proposals/0041-triangle-object-positions.md
@@ -326,6 +326,7 @@ Other approaches have been proposed for the return type of this intrinsic:
   * Benefit: user can use same struct elsewhere and easily assign to result of call
   * Benefit: no appearance of native dynamic indexing support
   * Benefit: should work with DXC intrinsic system relatively easily
+  * Benefit: will easily explicitly convert to long vector or matrix
   * Drawback: adds a new built-in type which effectively matches the memory layout of matrices or arrays
 * Add a vertex index parameter to the intrinsic and return a single position: `float3`:
   * Drawback: three calls required to store all vertices in a single variable in HLSL
@@ -344,7 +345,7 @@ Other approaches have been proposed for the return type of this intrinsic:
   * Benefit: avoids adding a new built-in type
   * Drawback: likely difficult to introduce this into intrinsic system
   * Drawback: language suggests native indexing supported, which would result in ugly codegen
-  * Drawback: might want to remove return array-by-value support in future HLSL version, as this is not allowed in C++.
+  * Drawback: current plans are to [disallow array return types in HLSL 202x](https://github.com/microsoft/hlsl-specs/issues/652)
 * Use a constant reference-to-array return type: `const float3[3] &`:
   * Benefit: checking static index is automatic
   * Drawback: likely difficult to introduce this into intrinsic system


proposals/0041-triangle-object-positions.md
```diff
@@ -317,22 +308,25 @@ Raytracing HLK tests.
...
   * Benefit: user can use same struct elsewhere and easily assign to result of call
   * Benefit: no appearance of native dynamic indexing support
   * Benefit: should work with DXC intrinsic system relatively easily
+  * Drawback: adds a new built-in type which effectively matches the memory layout of matrices or arrays
 * Add a vertex index parameter to the intrinsic and return a single position: `float3`:
   * Drawback: three calls required to store all vertices in a single variable in HLSL
   * Drawback: checking index requires an explicit diagnostic check
   * Benefit: A dynamic vertex index would easily lower directly to underlying DXIL op parameter without ugly codegen, if support for this was desired.
 * Use a matrix return type: `float3x3`:
-  * Drawback: It's not really a matrix
```
damyanp, line 330: It's still not really a matrix.

jenatali: That's fair.

proposals/0041-triangle-object-positions.md
```diff
@@ -317,22 +308,25 @@ Raytracing HLK tests.
 
 Other approaches have been proposed for the return type of this intrinsic:
 
-* Current approach: return built-in struct containing all three vertices as `float3` fields:
+* Return built-in struct containing all three vertices as `float3` fields:
```
damyanp, line 324: Does the struct actually have to be built-in?  Can't this all be done in a header, calling through to the builtin that returns the 9 x float?

llvm-beanz: The struct would need to be built-in because it will be referred to by other built-in constructs. That said, it's trivial to generate.

proposals/0041-triangle-object-positions.md
```diff
@@ -152,6 +152,22 @@ undefined. A shader can check for a triangle hit with
 
 Shader model 6.10 is required to use these intrinsics.
 
+The implementation for all of these can be implemented in terms of a built-in
+intrinsic which matches the DXIL definitions mentioned below:
```
llvm-beanz: Just a clarifying point. I believe we could implement the standalone function in terms of a builtin, but the RayQuery and HitObject implementations would themselves need to be built-in since the RayQuery and HitObject types themselves are built-in (we can't append to them in a header.

That likely means we're better off just implementing all the variants directly in the compiler because it would be weird to need to include a header for just one function.

jenatali: That makes sense, I didn't consider that, thanks.

proposals/0041-triangle-object-positions.md
```diff
@@ -317,22 +321,27 @@ Raytracing HLK tests.
...
   * Drawback: language suggests native indexing supported, which would result in ugly codegen
   * Drawback: layout isn't obvious
+  * Drawback: matrices can participate in implicit conversions that are likely to be programmer bugs, e.g. truncation to vector
 * Use a by-value array return type: `float3[3]`:
   * Benefit: checking static index is automatic
   * Benefit: convergence with SPIR-V
+  * Benefit: avoids adding a new built-in type
   * Drawback: likely difficult to introduce this into intrinsic system
   * Drawback: language suggests native indexing supported, which would result in ugly codegen
   * Drawback: might want to remove return array-by-value support in future HLSL version, as this is not allowed in C++.
```
llvm-beanz:
```suggestion
  * Drawback: current plans are to [disallow array return types in HLSL 202x](https://github.com/microsoft/hlsl-specs/issues/652)
```
This could probably be stronger.

proposals/0041-triangle-object-positions.md
```diff
@@ -317,22 +321,27 @@ Raytracing HLK tests.
 
 Other approaches have been proposed for the return type of this intrinsic:
 
-* Current approach: return built-in struct containing all three vertices as `float3` fields:
+* Return built-in struct containing all three vertices as `float3` fields:
   * Benefit: all values returned from one call
   * Benefit: user can use same struct elsewhere and easily assign to result of call
   * Benefit: no appearance of native dynamic indexing support
   * Benefit: should work with DXC intrinsic system relatively easily
```
llvm-beanz, line 328:
```suggestion
  * Benefit: should work with DXC intrinsic system relatively easily
  * Benefit: will easily explicitly convert to long vector or matrix
```

llvm-beanz (approved):
LGTM. I suggested a few updates to the benefits/drawbacks lists based on the conversation yesterday, but feel free take or leave them since I think we're aligned on direction.

PR #682, opened 2025-10-14
[Expr.Post.VectorSwizzle] Vector member swizzles

llvm-beanz:
This PR adds specification language for vector member swizzles to clarify how swizzles are parsed, how the fit in the grammar, what the component accesses mean, and whether the expression produces an lvalue or a prvalue.

Rendering of the grammar:
Screenshot 2025-10-14 at 3 14 39 PM

### DIFF
--- a/specs/language/expressions.tex
+++ b/specs/language/expressions.tex
@@ -188,17 +188,11 @@ range at runtime, the behavior is undefined.
   swizzle-component-xyzw\br
   swizzle-component-sequence-xyzw swizzle-component-xyzw\br
 
-  \define{swizzle-component-rgba}\br
-  \terminal{r}\br
-  \terminal{g}\br
-  \terminal{b}\br
-  \terminal{a}\br
+  \define{swizzle-component-rgba} \textnormal{one of}\br
+  \terminal{r g b a}\br
 
-  \define{swizzle-component-xyzw}\br
-  \terminal{x}\br
-  \terminal{y}\br
-  \terminal{z}\br
-  \terminal{w}
+  \define{swizzle-component-xyzw} \textnormal{one of}\br
+  \terminal{x y z w}
 \end{grammar}
 
 \p A \textit{postfix-expression} followed by a dot (\texttt{.}) and a sequence


specs/language/expressions.tex
```diff
@@ -167,6 +167,69 @@
...
+  \terminal{r}\br
+  \terminal{g}\br
+  \terminal{b}\br
+  \terminal{a}\br
+
+  \define{swizzle-component-xyzw}\br
+  \terminal{x}\br
+  \terminal{y}\br
+  \terminal{z}\br
+  \terminal{w}
```
farzonl: can we put these on one line
```suggestion
  \terminal{x y z w}\br
```

specs/language/expressions.tex
```diff
@@ -167,6 +167,69 @@
...
+
+  \define{swizzle-component-sequence-xyzw}\br
+  swizzle-component-xyzw\br
+  swizzle-component-sequence-xyzw swizzle-component-xyzw\br
+
+  \define{swizzle-component-rgba}\br
+  \terminal{r}\br
+  \terminal{g}\br
+  \terminal{b}\br
+  \terminal{a}\br
```
farzonl:
```suggestion
  \terminal{r g b a}\br
```

specs/language/expressions.tex
```diff
@@ -167,6 +167,69 @@
...
+of one or more \textit{swizzle-components} is a postfix expression. The
+postfix expression before the dot is evaluated and must be of vector type. If
+the postfix expression before the dot is an lvalue, the swizzle expression may
+produce either an lvalue or a prvalue; otherwise it produces a prvalue.
+
+\p A \textit{swizzle-component-sequence} is a sequence of one or more swizzle
+components of either the \textit{swizzle-component-rgba} or
+\textit{swizzle-component-xyzw} forms; the two forms may not be mixed in the
+same \textit{swizzle-component-sequence}. If the postfix expression before the
+dot is an lvalue and the \textit{swizzle-component-sequence} contains no
```
farzonl: can you change the langugage to if an lvalue it must contain no repeated components. I think what you are trying to capture in this line is this error case https://godbolt.org/z/v5G9Y7e1a but the way you wrote it makes it sound like if this condition does not apply it becomes a prvalue which doesn't seem correct.

specs/language/expressions.tex
llvm-beanz: That's not consistent with C++'s grammar style. That would mean this would match the token sequence `xyzw` not _one of_.

We could instead write it as something like:
```
\define{swizzle-component-rgba} \textnormal{one of}\br
\terminal{x y z w}
```
This would match the convention in Lex.

farzonl: That works for me.

specs/language/expressions.tex
llvm-beanz: There are two requirements for the resulting expression to be an lvalue:
1) The postfix expression to the left of the dot must be an lvalue
2) The component sequence must contain no repeated elements

These two requirements handle these cases:
```hlsl
2.r = 2.x;                                               // 2 as a constant produces an rvalue vector, so you cannot assign to it.
some_vec2.xx = some_other_vec2.xy; // repeated elements produce rvalues, so you cannot assign to it.
```

PR #684, opened 2025-10-15
[Lang] Update and restructure cbuffer spec

### LINKED ISSUE #671: Resolve inconsistencies between spec language and clang implementation for cbuffer declarations
[resources.tex](https://github.com/microsoft/hlsl-specs/blob/main/specs/language/resources.tex)

9.5.1 p6 reads:
> Other declarations in the cbuffer declaration group such as namespace-declaration, record-declaration or function declaration are not allowed

~~This is incorrect. The behaviour in dxc and fxc, and per the [implementation proposal for clang](https://github.com/llvm/wg-hlsl/blob/main/proposals/0016-constant-buffers.md#parsing-of-cbuffer-declaration), such constructs should transparently belong to the enclosing context.~~

However, clang allows some of these declarations. It should not.

llvm-beanz:
This change pulls the cbuffer section into the Declarations chapter instead of the Resources chapter since it defines a declaration type, and updates the grammar and specification language based on the updates to proposal 37.

The updated language avoids nesting of cbuffers and namepaces in cbuffers to simplify semantic validation. The updated langauge also adds additional language around storage durations which is referenced in the updated specification language.

Resolves #671

### DIFF
--- a/specs/language/declarations.tex
+++ b/specs/language/declarations.tex
@@ -179,9 +179,9 @@ referenced from within the translation unit, nor is it required to be unique.
 Each cbuffer declaration refers to a unique constant buffer resource
 (\ref{Resources.cnbuf}).
 
-\p Declarations within a cbuffer declaration that declare names, declare their
-names in the scope containing the cbuffer declaration. The cbuffer declaration
-itself does not declare a declaration scope. A cbuffer declaration may not
+\p The cbuffer declaration itself does not declare a declaration scope.
+Declarations within a cbuffer declaration that declare names, declare their names 
+in the scope containing the cbuffer declaration. A cbuffer declaration may not
 contain a \textit{namespace-declaration} or \textit{cbuffer-declaration}.
 \footnote{These declarations were previously allowed in HLSL reference compilers
 but are not supported in this specification.}


Icohedron (approved):
Reads good to me! LGTM

specs/language/declarations.tex
```diff
@@ -146,3 +154,41 @@
...
+
+\p A \textit{cbuffer declaration} is declared with the \texttt{cbuffer} keyword.
+The name of the cbuffer declaration does not declare a name, and cannot be
+referenced from within the translation unit, nor is it required to be unique.
+Each cbuffer declaration refers to a unique constant buffer resource
+(\ref{Resources.cnbuf}).
+
+\p Declarations within a cbuffer declaration that declare names, declare their
+names in the scope containing the cbuffer declaration. The cbuffer declaration
+itself does not declare a declaration scope. A cbuffer declaration may not
```
hekota:
```suggestion
\p The cbuffer declaration itself does not declare a declaration scope.
Declarations within a cbuffer declaration that declare names, declare their names 
in the scope containing the cbuffer declaration. A cbuffer declaration may not
```
I think it would read better if the first two sentences are flipped.

PR #685, opened 2025-10-15
[0037] Only restrict nesting cbuffer and namespace decls

llvm-beanz:
This slightly scales back the proposal to restricting namespaces in cbuffers and restricting nesting of cbuffers. This is likely a safe change for HLSL 202x, and we'll consider more significant changes in subsequent language updates.

This conclusion came out of the language design meeting discussion from 10/13/2025.

The language spec updates are posted in a draft as https://github.com/microsoft/hlsl-specs/pull/684.

### DIFF
--- a/proposals/0037-cbuffer-contexts.md
+++ b/proposals/0037-cbuffer-contexts.md
@@ -116,7 +116,7 @@ and cannot be modified in program code.
 ## Long-term Considerations
 
 During our discussion in the language design meeting additional restrictions to
-cbuffer declarations were considered. The general consesnsus was that additional
+cbuffer declarations were considered. The general consensus was that additional
 restrictions are desirable and will remove possible programmer errors, however
 there were concerns about HLSL 202x changing too much at once.
 


proposals/0037-cbuffer-contexts.md
```diff
@@ -69,39 +67,62 @@ To simplify HLSL's language semantics and the compiler implementation a new
...
+
+  \define{cbuffer-member-seq}\br
+  cbuffer-member-declaration\br
+  cbuffer-member-seq cbuffer-member-declaration\br
+
+  \define{cbuffer-member-declaration}\br
+  block-declaration\br
+  function-definition\br
+  template-declaration\br
+  empty-declaration
```
hekota, line 83: Where did the `variable-declaration` disappear? And what is `block-declaration`?

proposals/0037-cbuffer-contexts.md
```diff
@@ -69,39 +67,62 @@ To simplify HLSL's language semantics and the compiler implementation a new
...
-read-only storage in device memory that persists across the life of the
-dispatch.
+## Long-term Considerations
 
-Notably a `cbuffer` cannot contain function declarations, type declarations
-(classes, enumerations, typedefs), namespace declarations, constexpr or static
-variable declarations, `cbuffer` declarations, or any other declaration type not
-explicitly listed as allowed.
+During our discussion in the language design meeting additional restrictions to
+cbuffer declarations were considered. The general consesnsus was that additional
```
hekota:
```suggestion
cbuffer declarations were considered. The general consensus was that additional
```

proposals/0037-cbuffer-contexts.md
llvm-beanz: `block-declaration` is any declaration that can occur at block scope (i.e. any declaration you could have inside a function). That includes variables, but also aliases, static asserts, and other "simple" declarations like functions and classes.

The draft language spec PR includes some wider changes to the declarations chapter and grammars:
https://github.com/microsoft/hlsl-specs/pull/684/files#diff-11b83f8183acd1a33489925bf65fbeb9d420609588c72b0075263be859f6d186

hekota: Oh I see :) That's why I could not find it in the current version of the spec.

PR #688, opened 2025-10-18
[INF] Linear Algebra Implementation Plan

V-FEXrt:
This is far from a full implementation plan, but its a pretty good start. I mostly see this as "Part 1" which includes:
- Identify the types of work that need to be done
- Give a super basic, off-the-off time estimate of that work

I'm hoping we can use this PR to refine the time estimates a little bit and highlight any known work tasks that are missing from the list.

As further planning happens/more energy is focused at the implementation we can elaborate further on this document

### DIFF
--- a/proposals/infra/INF-XXXX-Linear-Algebra-Implementation-Plan.md
+++ b/proposals/infra/INF-XXXX-Linear-Algebra-Implementation-Plan.md
@@ -16,10 +16,9 @@ params:
 This implementation plan covers a high level description of the work necessary
 to implement the
 [linalg spec](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)
-for SM6.10 including back-of-the-napkin time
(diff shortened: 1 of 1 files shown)


proposals/infra/INF-XXXX-Linear-Algebra-Implementation-Plan.md
```diff
@@ -93,19 +184,29 @@ The work for each HLOp give then above context is roughly:
...
-## PSV0/RDAT Updates
-I don't have enough context to even posit a guess.
+### PSV0/RDAT Updates
+We need to update both, discussion was had on deprecating PSV0 and just doing
+RDAT however this doesn't align with near term goals for modernization.
 
-## Shader Reflection
-I don't have enough context to even posit a guess.
+### Shader Reflection
+Would require a refelction api version bump, reflection would be based on the
```
damyanp:
```suggestion
Would require a reflection api version bump, reflection would be based on the
```

PR #690, opened 2025-10-22
[0035] Add missing SumAccumulate OpCode

V-FEXrt:
`dx.op.matrixOp` was missing an opcode for Matrix Matrix Addition with Accumulation needed for `Matrix::SumAccumulate`.

This PR adds it to the spec

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -1149,9 +1149,10 @@ declare void @dx.op.matrixOp(
   )
 ```
 
-Two opcodes are available for this operation class, one for multiplying matrices
-and storing the result as `C = A * B`. The second for multiply accumulation `C
-+= A * B`.
+Three opcodes are available for this operation class:
+* Matrix Matrix Multiply: `C = A * B`
+* Matrix Matrix Multiply with Accumulation: `C += A * B`
+* Matrix Matrix Addition with Accumulation: `C += A + B`
 
 Validation rules will enforce that:
 * argument A is an `A` matrix


proposals/0035-linalg-matrix.md
```diff
@@ -1149,9 +1149,10 @@ declare void @dx.op.matrixOp(
   )
 ```
 
-Two opcodes are available for this operation class, one for multiplying matrices
-and storing the result as `C = A * B`. The second for multiply accumulation `C
-+= A * B`.
+Three opcodes are available for this operation class:
+* Matrix Matrix Multiply: `C = A * B`
+* Matrix Matrix Multiply with Accumulation: `C += A * B`
+* Matrix Matrix Addition with Accumulation: `C += A + B`
```
tex3d, line 1155: When I look up for the operation class, I see it would be `OpCodeClass::MatrixOp`.

Isn't "MatrixOp" a little too generic? The operation is limited to accepting three matrix operands, where it overwrites or accumulates into the first one. On that note, I think the operation class should be all you need to know to reason about the type of memory accesses for each matrix, but they are different for multiply vs. multiply or add with accumulate.

I think I'd split this into `MatrixBinaryOp` and `MatrixBinaryAccumulateOp`.

What do you think?

V-FEXrt: I don't have any issues with that, but @llvm-beanz should confirm. Chris do you have any thoughts here?

tex3d (approved):
I should clarify that my approval of this change isn't dependent on resolution of my comment about `OpCodeClass::MatrixOp`.

proposals/0035-linalg-matrix.md
V-FEXrt: Since you mentioned that you aren't blocking this PR I'll go ahead and merge as is but would love to keep the conversation going about the details here!

Further thoughts after chatting a bit with Chris:

Regardless of if we split these up or not they'll actually both end up with the same memory access annotations. `MatrixBinaryOp` would be readwrite* (read from A,B, write to C) and `MatrixBinaryAccumulateOp` would be readwrite* (read from A,B,C, write to C). 

Also if we do split them, `MatrixBinaryAccumulateOp` should be called something like `MatrixTernaryOp` because its actually more of shape `C' = A * B + C`

*: I use `readwrite` but iirc the actual annotation is none/the empty string which implies readwrite

V-FEXrt: Need to resolve to merge, going to resolve, merge, then file a follow up issue

PR #711, opened 2025-10-30
Increase Group Shared Memory Limit

JoeCitizen:
Proposes a mechanism to replace the static limit of 32kb of group shared memory for Compute, Mesh and Amplification shaders with a variable limit to allow implementations to expose larger sizes of local memory.


JoeCitizen:
@microsoft-github-policy-service rerun

Issue #693, opened 2025-10-23
[0035] Reflect 4x dimensions for packed types


pow2clk:
**Which proposal does this relate to?**
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md

**Describe the issue or outstanding question.**
The spec currently gives the matrix types the  static `ElementsPerScalar` value being assigned either to 1 for unpacked or 4 for 8bit packed integer types as well as `MScalars` and `NScalars` values indicating the practical dimensions of the vectors or matrices involved in operations ignoring the potential packed value 4x multiplier. However, these values aren't actually used in the template-enforced type restrictions so specifying the correct value for a packed type doesn't match the methods as written.

The assignment of `MScalars` versus `NScalars` to account for the packing multiplier should also be done for one or the other not both and depend on the `MatrixUse`.

**Additional context**
The way `ElementsPerScalar` is initialized is also among the characteristics of the header defined in the spec that isn't supported by DXC. Simply replacing the `IsNativeScalar` boolean used to initialize it with the  actual `ElementsPerScalar` value in the `ComponentTypeTraits` struct and assigning the value for the given element type to the static type in the templated matrix class type is a simple fix for that.

llvm-beanz:
I believe this has all been addressed in the current spec. Please let me know if there are any remaining issues.

Issue #568, opened 2025-07-16
[0035] Fill out spec section on conversions


llvm-beanz:
**Which proposal does this relate to?**
[0035 `linalg::Matrix`](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

**Describe the issue or outstanding question.**
We need to detail which matrix element conversions are legal and what the conversion behavior is for each conversion.

PR #712, opened 2025-11-03
[0035] Add Cooperative Vector interpretations

### LINKED ISSUE #570: [0035] Figure out how to handle `VectorRef`

The [cooperative vectors](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md) feature has a `linalg::VectorRef` type which allows bias vectors to be directly accessed from memory. We should consider how to apply this to the `linalg::Matrix` API as well.

### LINKED ISSUE #657: [0035] Interpretation parameters missing for linalg vector operations
[0035](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

[linalg](linalg::Multiply(vector, Matrix)) and linalg::MultiplyAdd(vector, Matrix, vector) do not provide a mechanism to declare interpretation parameters for input and bias vectors which would allow these values to be converted from native HLSL type to a different type (typically a non native type) for the operation. Conversion to a lower precision type that is not natively supported in HLSL is useful in ML applications.

**Additional context**
[0031](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0031-hlsl-vector-matrix-operations.md) and [0029](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md) have more details on interpretation parameters.

llvm-beanz:
This adds in the bias and input vector interpretation arguments to the linalg vector-matrix operations. To support these new arguments coming in from HLSL the API gains a new dx::linalg::InterpretedVector object, and a related dx::linalg::VectorRef object.

The dx::linalg::VectorRef object can be used to represent an InterpretedVector stored in a ByteAddressBuffer and can be used to materialize the vector into local memory immediately before the matmul call.

This update also has two general improvements to the HLSL header implementation:
1) This version now compiles with DXC so that it could include the packed data types which Clang doesn't yet support.
2) This version simplifies the SFINAE to instead use a partial specialization to separate out the thread-scope matrix type.

Resolves #657. Resolves #570.

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -54,25 +54,25 @@ on a base set of functionality for inclusion in the `hlsl` namespace.
 namespace dx {
 namespace linalg {
 
-template <MatrixComponentEnum ElementType, uint DimA> struct VectorRef {
+template <ComponentEnum ElementType, uint DimA> struct VectorRef {
   ByteAddressBuffer Buf;
   uint Offset;
 };
 
-template <typename T, int N, MatrixComponentEnum DT> struct InterpretedVector {
+template <typename T, int N, ComponentEnum DT> struct InterpretedVector {
   vector<T, N> Data;
-  static const MatrixComponentEnum Interpretation = DT;
+  static const ComponentEnum Interpretation = DT;
   static const SIZE_TYPE Size =
       __detail::ComponentTypeTraits<DT>::ElementsPerScalar * N;
 };
 
-template <MatrixComponentEnum DT, typename T, int N>
+template <ComponentEnum DT, typename T, int N>
 InterpretedVector<T, N, DT> MakeInterpretedVector(vector<T, N> Vec) {
   InterpretedVector<T, N, DT> IV = {Vec};
   return IV;
 }
 
-template <MatrixComponentEnum ComponentTy, SIZE_TYPE M, SIZE_TYPE N,
+template <ComponentEnum ComponentTy, SIZE_TYPE M, SIZE_TYPE N,
           MatrixUseEnum Use, MatrixScopeEnum Scope>
 class Matrix {
   using ElementType = typename __detail::ComponentTypeTraits<ComponentTy>::Type;
@@ -81,7 +81,7 @@ class Matrix {
   static const uint ElementsPerScalar =
       __detail::ComponentTypeTraits<ComponentTy>::ElementsPerScalar;
 
-  template <MatrixComponentEnum NewCompTy, MatrixUseEnum NewUse = Use>
+  template <ComponentEnum NewCompTy, MatrixUseEnum NewUse = Use>
   Matrix<NewCompTy, M, N, NewUse, Scope> Cast();
 
   template <typename T>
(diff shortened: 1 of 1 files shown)


V-FEXrt (approved):

damyanp (approved):

proposals/0035-linalg-matrix.md
```diff
@@ -54,131 +54,132 @@ on a base set of functionality for inclusion in the `hlsl` namespace.
...
-  template <MatrixComponentType LHSTy, MatrixComponentType RHSTy, uint K,
-            MatrixUse UseLocal = Use>
-  typename hlsl::enable_if<Use == MatrixUse::Accumulator &&
-                               Scope != MatrixScope::Thread && UseLocal == Use,
+  template <MatrixComponentEnum LHSTy, MatrixComponentEnum RHSTy, uint K,
+            MatrixUseEnum UseLocal = Use>
+  typename hlsl::enable_if<Use == MatrixUse::Accumulator && UseLocal == Use,
                            void>::type
   MultiplyAccumulate(const Matrix<LHSTy, M, K, MatrixUse::A, Scope>,
                      const Matrix<RHSTy, K, N, MatrixUse::B, Scope>);
```
pow2clk, line 139: It's good to see some usage of elementsperscalar and the DXC-compatible initialization as I suggested in #693. This doesn't quite address all the concerns that caused me to create that issue as these multiplies are still checking matching dimensions for number of scalars and not number of elements.

proposals/0035-linalg-matrix.md
```diff
@@ -54,131 +54,132 @@ on a base set of functionality for inclusion in the `hlsl` namespace.
 namespace dx {
 namespace linalg {
 
-template <MatrixComponentType ComponentTy, uint M, uint N, MatrixUse Use,
-          MatrixScope Scope>
+template <MatrixComponentEnum ElementType, uint DimA> struct VectorRef {
```
anupamachandra: Should this be just `ComponentTypeEnum` since we are using this for both Matrix and VectorRef?

proposals/0035-linalg-matrix.md
```diff
@@ -425,46 +474,61 @@ encode the dimensions and input and output data types used by each shader in the
...
-  F64 = 10,
-  SNormF16 = 11,
-  UNormF16 = 12,
-  SNormF32 = 13,
-  UNormF32 = 14,
-  SNormF64 = 15,
-  UNormF64 = 16,
-  PackedS8x32 = 17,
-  PackedU8x32 = 18,
+struct MatrixComponentType {
```
anupamachandra: SINT8 and UINT8 are missing? For vectors, this is not the packed case but when you want an arithmetic conversion from a higher precision to lower precision type. Also needed for Matrices when the value is store as 8bit integers.

proposals/0035-linalg-matrix.md
```diff
@@ -1632,14 +1721,33 @@ ByteAddressBuffer MBuf : register(t0);
...
+  VectorRef<MatrixComponentType::F8_E4M3, 16> MemBias = {MBuf,
+                                                         /*start offset*/ 4096};
+  vector<float16_t, 16> Layer3 = MultiplyAdd<float16_t>(Layer2, MatB, MemBias);
+
+  // Clang doesn't yet support packed types.
+#ifdef __hlsl_dx_compiler
+  vector<uint8_t4_packed, 4> SomeData = (vector<uint8_t4_packed, 4>)0;
+
+  vector<float16_t, 16> Layer4 = MultiplyAdd<float16_t>(
+      MakeInterpretedVector<MatrixComponentType::F8_E4M3>(SomeData), MatB,
```
anupamachandra: Do we want to support Packed->Float conversion of input vectors? We don't support this currently.

llvm-beanz: I need to address some wider questions around conversions so I'll track this as part of https://github.com/microsoft/hlsl-specs/issues/568.

proposals/0035-linalg-matrix.md
llvm-beanz: I'll address the open issues separately.

PR #727, opened 2025-11-11
[0045] Clustered Geometry:  Add ALLOW_CLUSTERED_GEOMETRY flag to raytracing pipeline state and RAYQUERY_FLAGS

amarpMSFT:
Pasting relevant text from motivation section in the spec:

> There also needs to be a way for apps to tell driver shader compilation to expect that 
acceleration structures being raytraced may contain clusters.  In the API this is a 
raytracing pipeline flag, `D3D12_RAYTRACING_PIPELINE_FLAG_ALLOW_CLUSTERED_GEOMETRY`. 
This flag needs to be available when raytracing pipeline state subobjects are 
authored in HLSL, including `RayQuery` objects for inline raytracing.
>
> The reason for the flag is if possibility of clusters in an acceleration structure requires 
a different path in the driver/hardware ray traversal algorithm.  This path might add 
overhead that can be avoided if it is known clusters will not be used.  In other words, 
a driver that supports clustered geometry should not regress the performance of applications 
that choose not to use it, or shipped before clustered geometry existed.

This addition builds on similar HLSL/DXIL functionality that was needed for Opacity Micromaps.  See https://github.com/microsoft/hlsl-specs/blob/main/proposals/0024-opacity-micromaps.md.  
So this time around it is easier to make the addition.

### DIFF
--- a/proposals/0045-clustered-geometry.md
+++ b/proposals/0045-clustered-geometry.md
@@ -4,6 +4,7 @@ params:
   authors:
   - jschmid-nvidia: Jan Schmid
   - simoll: Simon Moll
+  - amarpMSFT: Amar Patel
   sponsors:
   - jenatali: Jesse Natalie
   status: Under Review


proposals/0045-clustered-geometry.md
```diff
@@ -234,7 +356,13 @@ This proposal does not introduce or remove diagnostics or warnings.
...
 TODO: link the DXR spec proposal when available
 
-Note: No unique raytracing tier is required to use this intrinsic. If the device does not support clustered geometry, it must still support returning `CLUSTER_ID_INVALID` from this intrinsic.
+Note: No unique raytracing tier is required to use the `ClusterID` intrinsic. 
+If the device does not support clustered geometry, it must still support 
+returning `CLUSTER_ID_INVALID` from this intrinsic.
+
+If a raytracing pipeline is created with the 
+`RAYTRACING_PIPELINE_FLAG_ALLOW_CLUSTERED_GEOMETRY` flag on a device 
+without clustered geometry support, state object creation fails.
```
tex3d, line 365: I'm confused by this change. Aren't devices that lack support allowed to ignore the allow flag in other cases such as with RayQuery? Why would we fail state object creation in this case?

Additionally, if there is no optional feature flag requirement, it's not easy to tell if a compiled DXIL library contains a state object with this requirement that might cause a failure at runtime.

It seems the runtime should fail if a device doesn't support SM 6.10 when this flag is used, but should not require the device to support clustered geometry.

amarpMSFT: Older drivers might barf with the presence of the unrecognized new flag.  So might as well just lump in new drivers without cluster support as well.

> Additionally, if there is no optional feature flag requirement, it's not easy to tell if a compiled DXIL library contains a state object with this requirement that might cause a failure at runtime.

When we create an RTPSO in the runtime, we do look at the dxil subobjects and convert them to API ones.  So for subobjects there is effectively only one path, the API subobject that we validate.

> It seems the runtime should fail if a device doesn't support SM 6.10 when this flag is used, but should not require the device to support clustered geometry.

That could be fine too.   But given that ALLOW_CLUSTERED_GEOMETRY might incur perf overhead even if you aren't using the feature, apps should already be prepared to not set that flag unless they know they actually want to use clusters.  Even on a device that supports 6.10 and clusters.

tex3d: I suppose I'll resolve this, as a state object creation failure can be diagnosed using SDKLayers, and an application could use a different pipeline config for a device that lacks support for the feature.

tex3d (approved):

PR #729, opened 2025-11-12
[0052] Finalize Experimental DXIL Op proposal

V-FEXrt:
Presents the final proposed solution for experimental DXIL Ops which is to accept the 16bit FeatureID solution but to restrict the feature IDs to `0x0000` and `0x8000`

### DIFF
--- a/proposals/0052-experimental-dxil-ops.md
+++ b/proposals/0052-experimental-dxil-ops.md
@@ -76,10 +76,26 @@ for at least one shader model version it may be reused but this is discouraged.
 
 FeatureID `0x0000` must be used for stable opcodes released in the retail
 compiler. Any other value would break back compatibility.
-FeatureID `0x8000` is used for experimental opcodes released for preview.
+FeatureID `0x8000` is used for all new opcodes not yet finalized for a shader
+model release.
+These opcodes should be added (assigned) and merged into main as early as
+possible, even before they are used or tested. This reserves the opcode so
+subsequent code and tests may avoid opcode collisions with other features.
 
-When an opcode introduced as experimental is stablized it must be transitioned
-from FeatureID `0x0000` to `0x8000` which will completely renumber the opcode.
+During development, if an opcode needs to be changed in any breaking way from
+a version supported by an experimental driver, the opcode should be left behind
+and renamed as necessary to avoid a collision with the new version.
+Then a new opcode should be used to allow an easier transition for experimental
+development, allowing an experimental driver to support both versions for a
+period of time.
+
+When an opcode with FeatureID `0x8000` is finalized for the next DXIL release,
+it should be copied to FeatureID `0x0000` and assigned the next available
+opcode number.
+After the opcode is copied to a final DXIL opcode, the original opcode with
+FeatureID `0x8000` may be kept and renamed for transition compatibility for
+drivers, or it may be replaced with a reserved op to avoid reusing the opcode
+for at least one DXIL release.
 The opcodes spaces under each FeatureID are independent and no correlation in
 the underlying value may be assumed.
 
@@ -110,12 +126,12 @@ lacks process complexity it also lacks flexibility or resolution.
 
 The solution proposed is to implement the Top 16 Bit with the following restriction:
  * Feature IDs must be either `0x0000` or `0x8000` until an undetermined time in
-   the future where the restriction will either be lifted or permentantly codified.
+   the future where the restriction may be lifted if desired.
 
 Feature ID `0x0000` must be used for stable opcodes to ensure that existing
 opcodes are not renumbered however the choice of ID `0x80000` may seem
 arbitrary. `0x8000` is selected for one key feature. The set bit is the top
-most bit fo the opcode space. Therefore all opcodes defined with either ID will
+most bit of the opcode space. Therefore all opcodes defined with either ID will
 match the underlying values they would have in the Top 1 bit proposal.
 
 With this restriction, the compiled DXIL is proposal agnostic. Either original


proposals/0052-experimental-dxil-ops.md
```diff
@@ -7,7 +7,7 @@ params:
     - tex3d: Tex Riddell
   sponsors:
     - V-FEXrt: Ashley Coleman
-  status: Under Consideration
+  status: Accepted
```
V-FEXrt, line 10: Highlighting that this PR moves the proposal to accepted.

proposals/0052-experimental-dxil-ops.md
```diff
@@ -57,10 +57,97 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+utility given the remaining DXIL lifecycle. There is also process costs to
+maintain the list of exprimental feature IDs. Conversely while the 1bit proposal
+lacks process complexity it also lacks flexibility or resolution.
+
+The solution proposed is to implement the Top 16 Bit with the following restriction:
+ * Feature IDs must be either `0x0000` or `0x8000` until an undetermined time in
+   the future where the restriction will either be lifted or permentantly codified.
+
+Feature ID `0x0000` is to be used for stable opcodes that are published as part
+of the retail compiler. Feature ID `0x8000` is to be used for all expirmental
```
damyanp:
```suggestion
of the retail compiler. Feature ID `0x8000` is to be used for all experimental
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -57,10 +57,97 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+most bit fo the opcode space. Therefore all opcodes defined with either ID will
+match the underlying values they would have in the Top 1 bit proposal.
+
+With this restriction, the compiled DXIL is proposal agnostic. Either original
+proposal could have reasonably generated it. This is an imperfect compromise
+that has two key features:
+ * it unblocks high priority dependencies of the feature
+ * it cleanly collapses into either original proposal
+
+It is the intent of this proposal to serve as an expirement in its own right.
```
damyanp:
```suggestion
It is the intent of this proposal to serve as an experiment in its own right.
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -57,10 +57,97 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+At some point in the future this proposal should collapse into one of the above
+proposal in a non-breaking manner. After iterating through a number of
```
damyanp:
```suggestion
proposals in a non-breaking manner. After iterating through a number of
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -57,10 +57,97 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+development cycles, the level of complexity required to serve this feature will
+reveal itself. If the development challenges presented at the top of this
+document are resolved with the restrictions in place then the proposal will
+collapse into the 1bit solution and no further work is needed. However if the
```
damyanp:
```suggestion
collapse into the 1bit solution and no further work is needed. However, if the
```

proposals/0052-experimental-dxil-ops.md
V-FEXrt:
```suggestion
proposals in a non-breaking manner. After iterating through a number of
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -57,10 +57,97 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+into the 16bit proposal.
+
+### Implementation Considerations
+
+#### DXC
+DXC has considerable constraints to align implementation with existing
+infrastructure. All of the bit width proposals are reasonable to implement
+but the other proposals are beyond reasonable scope. For the bit width proposals
+the infrastructure needs to be updated to support multiple op code tables. The
+detailed explination of that infrastructure is listed below.
```
V-FEXrt:
```suggestion
detailed explanation of that infrastructure is listed below.
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -57,10 +57,97 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+setting the correct value for the opcode in `llvm/lib/Target/DirectX/DXIL.td`.
+
+For readabilty, the opcodes can be written in hex instead of the traditional
+decimal value. As an example the 12th experimental op code could be written as
+`0x8000000C` instead of `2147483660`.
+
+#### DXV
+The DXIL Validator should be updated in response to the proposed changes
+presented above. The smallest change with the notable impact would be to detect
+the usage of an experimental opcode (`opcode & 0x80000000`) and automatically
```
V-FEXrt:
```suggestion
the usage of an experimental opcodes (`opcode & 0x80000000`) and automatically
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -57,10 +57,97 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+compiler. That flag would set the DXIL flag and then the validator will error
+if the flag isn't set but an experimental opcode is used.
+
+There may be changes to IR lowering in experimental compilers that don't result
+in the emission of experimental opcodes. This means that the automatic check may
+miss some experimental uses.
 
 ## Existing DXIL Op and HLSL Intrinsic Infrastructure
 
+The follow details are specific to DXC, clang has a significantly different
```
V-FEXrt:
```suggestion
The details below are specific to DXC, clang has a significantly different
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -176,34 +263,57 @@ first step.
...
+
+No structural or shape changes to the DXIL opcode occur, the fact that the opcode
 has the high bit set informs that it is experimental. This makes it very easy
-for the compiler and drivers to detect experimental opcodes. When an opcode is
-transistioned to stable the opcode needs to be assigned a stable number.
+for the validator and drivers to detect experimental opcodes.
+
 This splits the 4 billion opcode space into two 2 billion partions. One for
-stable one for experimental. The proposal results in two separatlye contiguous
+stable one for experimental. The proposal results in two separatly contiguous
```
V-FEXrt:
```suggestion
stable one for experimental. The proposal results in two separately contiguous
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -176,34 +263,57 @@ first step.
...
 op code tables.
 
+When an opcode is transitioned to stable it must be moved to the stable opcode
+partition. These two tables are completely independent from each other so
+opcode transition will result in a complete renumbering. No assumption may be
+made about how the opcodenumber will change when moving to stable.
```
V-FEXrt:
```suggestion
made about how the opcode number will change when moving to stable.
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -221,16 +331,59 @@ partition for their own use without collision with other opcodes.
...
+exception is FeatureID `0x0000` which is reserved for past stable opcodes to 
+maintain their current value. 
 
-### Top 16 bits as "opcode partition" value
-Identical concept as above but with 64k partitions, each with 64k opcodes.
+The validator or some other mechanism must maintain a list of experimental
+FeatureIDs. When an experimental FeatureID is used the entire shader must
+be marked as preview.
+
+FeatureIDs may be resued once a FeatureID has been marked reserved for at least
```
V-FEXrt:
```suggestion
FeatureIDs may be reused once a FeatureID has been marked reserved for at least
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -57,10 +57,97 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+Both proposals use some number of bits from the top portion of the opcode to
+partition the remaining bits into seperate opcode sets. In DXC these sets map
+directly to tables of operations, while in clang individual opcode values are
+arbirarily set very late in lowering.
+
+The team saw notable merits and issues with both solutions such that a trivial
+decision was not possible. Those details are discussed in their respective
+"Alternatives Considered" section.
+
+Implementation complexity is within the same magnetude for both proposals.
```
llvm-beanz:
```suggestion
Implementation complexity is within the same magnitude for both proposals.
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -57,10 +57,97 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+However, the key debate between the proposals is that of
+process complexity. The 16bit proposal introduces a reasonable amount of
+process managment for reserving and retiring feature IDs which may have limited
```
llvm-beanz:
```suggestion
process management for reserving and retiring feature IDs which may have limited
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -57,10 +57,97 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+
+#### Clang
+Clang implementation for bit width proposals are trivial. Clang has a late
+lowering for convering high level notions of the DXIL ops into the specific
+opcode values. That mapping is arbitrary and can be implemented by simply
+setting the correct value for the opcode in `llvm/lib/Target/DirectX/DXIL.td`.
+
+For readabilty, the opcodes can be written in hex instead of the traditional
+decimal value. As an example the 12th experimental op code could be written as
+`0x8000000C` instead of `2147483660`.
```
llvm-beanz: I’m pretty sure we could even do something like `0x80000000 | 12` in TableGen since it has support for arithmetic and bitwise operations with some limitations.

proposals/0052-experimental-dxil-ops.md
```diff
@@ -221,16 +331,59 @@ partition for their own use without collision with other opcodes.
...
 
-### Top 16 bits as "opcode partition" value
-Identical concept as above but with 64k partitions, each with 64k opcodes.
+The validator or some other mechanism must maintain a list of experimental
+FeatureIDs. When an experimental FeatureID is used the entire shader must
+be marked as preview.
+
+FeatureIDs may be resued once a FeatureID has been marked reserved for at least
+one shader model. Ex: Feature ID `0xDEADBEEF` is introduced in 6.8 as
+experimental, marked as reserved for 6.9, so it may be recycled in 6.10
```
llvm-beanz, line 408: Technically we could reuse opcodes the same way. There is literally nothing to stop us from saying in a new shader model a previously used opcode has a new meaning, except that it would likely confuse people and have more harms than benefits.

proposals/0052-experimental-dxil-ops.md
```diff
@@ -57,10 +57,97 @@ solution. Thus this proposal explicitly avoids addressing these issues:
  * Process development to enable asynchronous non-colliding development
  * Metadata/RDAT/PSV0/Custom lowering are out of scope for this document
 
-
+## Proposed Solution
+
+The leading proposals and time of writing were:
+ * [Top 1 bit as experimental flag](###Top-1-bit-as-experimental-flag)
+ * [Top 16 bits as opcode partition](###Top-16-bits-as-opcode-partition)
```
Flakebi: The links don’t work. The correct one is `#top-16-bits-as-opcode-partition`.

Flakebi (commented):
Doing a fresh read, I agree with @llvm-beanz, that a description of the actual accepted state is missing.
A final summary under “Proposed Solution” would be good to read and also to remove any ambiguities that come from the current description (which reads like “merge these two proposals in your head and add this third restricition”).
The descriptions are not long (the “Pros” would not belong into “Proposed Solutions”), the problem is that it’s scattered and contains two “diffs” that you need to apply in your head.

proposals/0052-experimental-dxil-ops.md
```diff
@@ -58,9 +58,122 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+
+```
+0b 0000 0000 0000 0000 0000 0000 0000 0000
+                       ^^^^ ^^^^ ^^^^ ^^^^ opcode space
+   ^^^^ ^^^^ ^^^^ ^^^^-------------------- opcode partition / FeatureID
+```
+
+FeatureID `0x0000` must be used for stable opcodes released in the retail
+compiler. Any other value would break back compatibility.
+FeatureID `0x8000` is used for experimental opcodes released for preview.
```
tex3d:
```suggestion
FeatureID `0x8000` is used for all new opcodes not yet finalized for a shader
model release.
These opcodes should be added (assigned) and merged into main as early as
possible, even before they are used or tested. This reserves the opcode so
subsequent code and tests may avoid opcode collisions with other features.

During development, if an opcode needs to be changed in any breaking way from
a version supported by an experimental driver, the opcode should be left behind
and renamed as necessary to avoid a collision with the new version.
Then a new opcode should be used to allow an easier transition for experimental
development, allowing an experimental driver to support both versions for a
period of time.
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -58,9 +58,122 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+
+When an opcode introduced as experimental is stablized it must be transitioned
+from FeatureID `0x0000` to `0x8000` which will completely renumber the opcode.
+The opcodes spaces under each FeatureID are independent and no correlation in
+the underlying value may be assumed.
```
tex3d: Reversed FeatureIDs, and some more suggestions:
```suggestion
When an opcode with FeatureID `0x8000` is finalized for the next DXIL release,
it should be copied to FeatureID `0x0000` and assigned the next available
opcode number.
After the opcode is copied to a final DXIL opcode, the original opcode with
FeatureID `0x8000` may be kept and renamed for transition compatibility for
drivers, or it may be replaced with a reserved op to avoid reusing the opcode
for at least one DXIL release.
The opcodes spaces under each FeatureID are independent and no correlation in
the underlying value may be assumed.
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -58,9 +58,122 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+However, the key debate between the proposals is that of
+process complexity. The 16bit proposal introduces a reasonable amount of
+process management for reserving and retiring feature IDs which may have limited
+utility given the remaining DXIL lifecycle. There is also process costs to
+maintain the list of exprimental feature IDs. Conversely while the 1bit proposal
+lacks process complexity it also lacks flexibility or resolution.
+
+The solution proposed is to implement the Top 16 Bit with the following restriction:
+ * Feature IDs must be either `0x0000` or `0x8000` until an undetermined time in
+   the future where the restriction will either be lifted or permentantly codified.
```
tex3d: It feels like it will already be enforcing the top-bit-only, so I'm not sure what "permentantly codified" would really mean other than as a spec detail.
```suggestion
   the future where the restriction may be lifted if desired.
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -58,9 +58,122 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+
+Feature ID `0x0000` must be used for stable opcodes to ensure that existing
+opcodes are not renumbered however the choice of ID `0x80000` may seem
+arbitrary. `0x8000` is selected for one key feature. The set bit is the top
+most bit fo the opcode space. Therefore all opcodes defined with either ID will
```
tex3d:
```suggestion
most bit of the opcode space. Therefore all opcodes defined with either ID will
```

proposals/0052-experimental-dxil-ops.md
```diff
@@ -58,9 +58,122 @@ solution. Thus this proposal explicitly avoids addressing these issues:
...
+presented above. The smallest change with the notable impact would be to detect
+the usage of an experimental opcodes (`opcode & 0x80000000`) and automatically
+set the preview hash. A more robust change would be to add a DXIL flag for
+expirmental allowed which is set by a `--hlsl-experimental` flag on the
+compiler. That flag would set the DXIL flag and then the validator will error
+if the flag isn't set but an experimental opcode is used.
+
+There may be changes to IR lowering in experimental compilers that don't result
+in the emission of experimental opcodes. This means that the automatic check may
+miss some experimental uses.
```
tex3d, line 185: There's some more to consider for having the option flag:
We could add an experimental flag to HLSL intrinsics and availability diagnostics when experimental features are used without it. This would be removed (with shader model requirement updated) for each intrinsic when finalized for the next shader model.

We can differentiate an opcode from the next shader model being used from experimental opcode usage in the validator. This could be very useful in transition to finalized opcodes, where you would expect a test to only be using finalized opcodes, and would then remove the `--hlsl-experimental` option. You would see a validation failure if you accidentally produced an experimental opcode, rather than a final opcode in the next shader model. Both of these would be given the `PREVIEW` hash, but one is ready for the next release, and the other is not.

This also makes things more predictable when targeting released shader models when longer-running experimental opcodes may still exist. If we don't have the flag, and you were to accidentally use an experimental opcode on a released shader model, it would silently just give you a `PREVIEW` hash on the final result, with no validation failure or visible indication that this shader is not going to conform to the released shader model you targeted.

proposals/0052-experimental-dxil-ops.md
```diff
@@ -221,16 +357,59 @@ partition for their own use without collision with other opcodes.
...
+FeatureIDs may be reused once a FeatureID has been marked reserved for at least
+one shader model. Ex: Feature ID `0xDEADBEEF` is introduced in 6.8 as
+experimental, marked as reserved for 6.9, so it may be recycled in 6.10
+
+Pros:
+ * Could be implemented "by hand" today by hard coding opcodes in clang,
+   DXC requires some updates to opcode generation code.
+ * Enables basic opcodes extension system
+ * Very flexible for opcode asignments
+ * Decreases the occurance of opcode collison merge conflicts
```
tex3d, line 415: This also allows for external parties to privately experiment with new DXIL operations before proposing features, without having to constantly update those opcodes with breaking changes every time a new experimental op is added to main. Then transitioning to a public proposal requires far fewer changes than transitioning from other hacky methods of extending DXIL for such private experiments today.

However, another solution for that using a single experimental table would be to reserve a bunch of opcodes for such purposes and just never reuse those. Still, we would need to anticipate and reserve sufficient opcodes for this use case.

Other advantages are:
- The ability to identify specific experimental (or extended) features being used, which could be checked in tests and even reported in disassembly comments.
- We could emit a metadata table listing experimental feature usage, and even keep any additional metadata needed for each feature isolated from the rest of the finalized metadata, including details of feature usage that will ultimately be integrated into runtime validation data once finalized.
- could allow for specific feature checking on shaders and in the runtime, since some preview drivers may only support certain features, which makes it easier to test the right things on experimental drivers.
- could allow individual feature enablement flags and availability attributes.

V-FEXrt: I'll file an issue to add this as a follow up. It doesn't effect the accepted proposal so I don't want to block them merge before I leave for vacation

proposals/0052-experimental-dxil-ops.md
V-FEXrt: I'll file an issue to add this as a follow up. It doesn't effect the accepted proposal so I don't want to block them merge before I leave for vacation

proposals/0052-experimental-dxil-ops.md
```diff
@@ -58,9 +58,138 @@ solution. Thus this proposal explicitly avoids addressing these issues:
  * Metadata/RDAT/PSV0/Custom lowering are out of scope for this document
 
 
+## Accepted Solution
+
+The top 16 bits of the opcode shall be used to partition the opcode into ~64k
+partitions each with ~64k opcodes. The top 16 bits of the opcode are called
+the `FeatureID` and the only valid FeatureIDs are `0x0000` and `0x8000`.
+Within a given partition, opcodes must be contiguous. When opcodes are retired
```
llvm-beanz, line 66: > opcodes must be contiguous

I don't think this is technically true. The fact that DXC requires reserving opcodes is an oddity of DXC, but we could leave gaps, the only reason not to do that is that it has an undue cost to DXC.

PR #735, opened 2025-11-13
[0049] Declare DXIL Metadata for Variable Group Shared Memory

JoeCitizen:
To allow for the runtime to validate against device limits:

- Add kDxilGroupSharedLimitTag
- Add GroupSharedLimit to PSV0

### DIFF
--- a/proposals/0049-variable-groupshared-memory.md
+++ b/proposals/0049-variable-groupshared-memory.md
@@ -11,31 +11,6 @@ params:
 * Planned Version: Shader Model 6.10
 * Issues: (TBD)
 
-## Contents
-
-- [Introduction](#introduction)
-- [Motivation](#motivation)
-- [Proposed Solution](#proposed-solution)
-  - [Examples](#examples)
-    - [Example 1: `GroupSharedLimit` declared but not exceeded](#example-1-groupsharedlimit-declared-but-not-exceeded)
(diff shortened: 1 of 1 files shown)


proposals/0049-variable-groupshared-memory.md
```diff
@@ -11,6 +11,31 @@ params:
...
+  - [Interchange Format Additions](#interchange-format-additions)
+    - [DXIL Metadata](#dxil-metadata)
+    - [PSV0 (Pipeline State Validation) Metadata](#psv0-pipeline-state-validation-metadata)
+  - [Validation Changes](#validation-changes)
+  - [Runtime Additions](#runtime-additions)
+    - [Capability Bit / Query](#capability-bit--query)
+    - [Pipeline Compilation / Load](#pipeline-compilation--load)
+- [Testing](#testing)
+- [Alternatives Considered](#alternatives-considered)
+
```
llvm-beanz: I don't think we should hard code a table of contents here. If we want one we should have Hugo generate it. @damyanp thoughts?

llvm-beanz (commented):
Content LGTM. Would like @damyanp's thoughts on formatting.

proposals/0049-variable-groupshared-memory.md
damyanp: Agreed.

JoeCitizen: updated to remove the table of contents

PR #736, opened 2025-11-14
Update 0026-hlsl-long-vector-type.md

ohmaya:
The 2nd part is for StructuredBuffers instead of for ByteAddressBuffers.

### DIFF
--- a/proposals/0026-hlsl-long-vector-type.md
+++ b/proposals/0026-hlsl-long-vector-type.md
@@ -131,8 +131,8 @@ in undefined behavior.
 
 StructuredBuffers with N-element vectors are declared using the template syntax
  with a long vector type as the template parameter.
-N-element vectors are loaded and stored from StructuredBuffers using the non-templated load and store methods
-with the element index parameters.
+N-element vectors are loaded and stored from StructuredBuffers using the load and store methods
+or subscript operators with the element index parameters.
 
 ```hlsl
 RWStructuredBuffer< vector<T, N> > myBuffer;


damyanp (approved):
This change looks correct to me.

Can you confirm please, @pow2clk or @V-FEXrt?

llvm-beanz (changes requested):
`StructuredBuffer`s don't have templated `Load` methods. This is for the syntax:

```hlsl
ByteAddressBuffer Buf;

[numthreads(1,1,1)]
void main() {
    vector<float,8> V = Buf.Load<vector<float,8> >(0);
}
```
[Compiler Explorer](https://godbolt.org/z/x9d74MWrT)

Edit:

For context, the usage for `StructuredBuffer`s is:

```hlsl
StructuredBuffer<vector<float,8> > Buf;

[numthreads(1,1,1)]
void main() {
    vector<float,8> V = Buf.Load(0);
}
```
[Compiler Explorer](https://godbolt.org/z/G8qaab8jb)

Note that the `Load` method is not templated.

ohmaya:
Then we can **add a "non-" before "templated"**.

It seems the whole sentence, "N-element vectors are loaded and stored from ByteAddressBuffers using the templated load and store methods with" (twice at the moment), was just copied from the section for ByteAddressBuffer and **pasted without modifying every places for StructuredBuffer**.

proposals/0026-hlsl-long-vector-type.md
```diff
@@ -131,7 +131,7 @@ in undefined behavior.
 
 StructuredBuffers with N-element vectors are declared using the template syntax
  with a long vector type as the template parameter.
-N-element vectors are loaded and stored from ByteAddressBuffers using the templated load and store methods
+N-element vectors are loaded and stored from StructuredBuffers using the non-templated load and store methods
 with the element index parameters.
```
llvm-beanz: How about:
```suggestion
N-element vectors are loaded and stored from StructuredBuffers using the load and store methods
or subscript operators with the element index parameters.
```

ohmaya: Looks good.

I also got the information about subscript operators from: https://devblogs.microsoft.com/directx/hlsl-native-and-long-vectors/. I was curious why the spec did not mention it.

V-FEXrt (approved):

PR #737, opened 2025-11-14
[0035] Clarify `Length` for packed types

llvm-beanz:
This PR clarifies that `Length` returns the count of HLSL native elements rather than the number of matrix elements such that `Length` for packed types will be 1/4th the actual thread-accessible members.

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -696,9 +696,9 @@ uint Matrix::Length();
 Requires `Wave` or `ThreadGroup` scope matrix.
 
 Returns the number of matrix components accessible to the current thread. If the
-matrix's elements are stored in a packed type `Length` will return the number of
-packed elements (e.g. if a thread has 8 accessible elements of `int8` type which
-are packed into 2 `int8_t4_packed`, `Length` will return 2). The mapping and
+matrix's elements are stored in a packed type, `Length` will return the number of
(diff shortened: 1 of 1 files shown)


proposals/0035-linalg-matrix.md
```diff
@@ -695,15 +695,18 @@ uint Matrix::Length();
...
-for each thread. The sum of the values returned by `Length` across all
-threads must be greater than or equal to the total number of matrix elements.
-Some implementations may map multiple threads to the same matrix element.
-Therefore, developers should take this into consideration when programming
-side-effects, such as atomic operations and/or UAV writes, within user-defined
-matrix operations.
+Returns the number of matrix components accessible to the current thread. If the
+matrix's elements are stored in a packed type `Length` will return the number of
+packed elements (e.g. if a thread has 8 accessible elements of `int8` type which
+are packed into 2 `int8_t4_packed`, `Length` will return 2). The mapping and
```
hekota:
```suggestion
Returns the number of matrix components accessible to the current thread. If the
matrix's elements are stored in a packed type, `Length` will return the number of
packed elements (e.g. if a thread has 8 accessible elements of `int8` type 
packed into 2 `int8_t4_packed`, `Length` will return 2). The mapping and
```

hekota (approved):
LGTM, just one suggestion.

PR #739, opened 2025-11-18
[0048] - GroupWaveIndex - Change to using an intrinsic vs system value and also add WaveGetCount

JoeCitizen:
- Change from using a system value for the Group Wave Index to an intrinsic function
- Introduce WaveGetCount as a second intrinsic to determine how many hardware waves are executing in the thread group

### DIFF
--- a/proposals/0048-group-wave-index.md
+++ b/proposals/0048-group-wave-index.md
@@ -200,6 +200,26 @@ No new metadata is required. The use of `GetGroupWaveIndex` and
 `GetGroupWaveCount` is indicated by the presence of the respective `dx.op.*`
 operations in the shader.
 
+#### SPIR-V Support
+
+For SPIR-V targets (Vulkan), these intrinsics map to existing GLSL built-in
+variables and their corresponding SPIR-V built-ins:
+
+**`GetGroupWaveIndex()`** maps to:
+- GLSL: `gl_SubgroupID` 
+- SPIR-V: `BuiltIn SubgroupId`
+- Decoration: `OpDecorate %var BuiltIn SubgroupId`
+
+**`GetGroupWaveCount()`** maps to:
+- GLSL: `gl_NumSubgroups`
+- SPIR-V: `BuiltIn NumSubgroups`
+- Decoration: `OpDecorate %var BuiltIn NumSubgroups`
+
+Both built-ins are part of the Vulkan 1.1 core specification. They provide the
+same semantics as the HLSL intrinsics: `SubgroupId` returns the index of the
+subgroup (wave) within the workgroup (thread group), and `NumSubgroups` returns
+the total number of subgroups executing the workgroup.
+
 ### Diagnostic Changes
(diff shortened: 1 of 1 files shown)


proposals/0048-group-wave-index.md
```diff
@@ -15,9 +15,10 @@ params:
 
 ## Introduction
 
-The proposal is for a new shader construct:
+The proposal is for two new shader intrinsics:
 
-* `SV_GroupWaveIndex`: the index of the wave in the thread group
+* `WaveGetGroupIndex()`: returns the index of the wave in the thread group
+* `WaveGetCount()`: returns the number of waves executing the thread group
```
tex3d: Since there's already a `SV_GroupIndex`, this might be confusing as worded. Also, `WaveGetCount()` might seem ambiguous to someone who doesn't see that lane count is retrieved via `WaveGetLaneCount()`. Plus, these don't pair up to make a nice pattern, and can we assume a wave count/index would be for the thread group? What other scope could it be for?

What about one of these?

If we really need to keep the `Wave` prefix. This seems most consistent with the existing `WaveGetLaneCount` while extending the pattern to `WaveCount`, even if it sounds a little repetitive.
```suggestion
* `WaveGetWaveIndex()`: returns the index of the wave in the thread group
* `WaveGetWaveCount()`: returns the number of waves executing the thread group
```

Shorter and still seems clear enough:
```suggestion
* `GetWaveIndex()`: returns the index of the wave in the thread group
* `GetWaveCount()`: returns the number of waves executing the thread group
```

Include `Group` (without the `Wave` prefix):
```suggestion
* `GetGroupWaveIndex()`: returns the index of the wave in the thread group
* `GetGroupWaveCount()`: returns the number of waves executing the thread group
```

Each of these options more intuitively pair together.

JoeCitizen: I think GetGroupWave* makes the most sense

proposals/0048-group-wave-index.md
```diff
@@ -75,101 +86,119 @@ perform different tasks and coordinate their work efficiently.
...
+- Returns the total number of waves N executing the thread group
+- Guaranteed to be at least:
+  ```hlsl
+  N >= ceil((numthreads.x * numthreads.y * numthreads.z) / WaveGetLaneCount())
+  ```
+- Where `numthreads` is specified by the `[numthreads(x, y, z)]` attribute
+- Implementation may launch additional waves beyond the minimum required, so N
+  may be larger than the computed minimum
+- All threads in the thread group (across all waves) return the same count value
+- The value is consistent throughout the execution of the thread group
```
llvm-beanz, line 137: Do these both have SPIRV equivalents? If so can you document them? If not the intrinsics need to be added to the `dx` namespace so we should call that out too.

JoeCitizen: Documented the SPIR-V equivalent

PR #743, opened 2025-11-24
Add language about bit-fields

spall:
Add language about conversions to and from bit-fields.

### DIFF
--- a/specs/language/classes.tex
+++ b/specs/language/classes.tex
@@ -1,0 +1,27 @@
+\Ch{Classes}{Classes}
+\Sec{Bit-field}{Classes.Bitfield}
+\p A bit-field is a class member with an explicit size specified in bits. Bit-fields must be of an underflying integer type.
+\begin{HLSL}
+struct Foo {
+  uint A : 10;
+  bool B : 1;
+};
+\end{HLSL}
+\p A bit-field can be unnamed.  Unnamed bit-fields cannot be accessed or initialized.
+\begin{HLSL}
+struct Foo {
+  uint : 10;
+};
+\end{HLSL}
+\p The explicit size of a named bit-field can be any number of bits greater than zero and less than or equal to the size of the specified type.
+\p The explicit size of an unnamed bit-field can also be zero.
+\p A bit-field cannot be a static member.
+\Sec{Static Members}{Classes.Static}
+\Sec{Conversions}{Classes.Conversions}
+\p A glvalue of a named bit-field with underlying type \texttt{T} can be converted
+to a cxvalue of type \texttt{T$^\prime$} if there exists an implicit conversion
+from \texttt{T} to \texttt{T$^\prime$}.
+\p A glvalue or a prvalue of a type \texttt{T} can be converted to a glvalue
+of a named bit-field of type \texttt{T$^\prime$} if there exists an implicit
+conversion from \texttt{T} to \texttt{T$^\prime$}.  After the conversion from
+\texttt{T} to \texttt{T$^\prime$} the value is truncated to the size of the bit-field.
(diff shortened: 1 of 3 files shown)


hekota (approved):
LGTM!

inbelic (approved):

llvm-beanz (commented):
I'm not sure this is the right way to spec bit-fields. Bit-fields aren't really a type, and there aren't really conversions in and out of bitfields the way other conversions occur.

The only parts of the conversion chapter that probably need to talk about bit-fields is the integral conversions and promotions because they may be restricted based on whether or not the value fits into the bit-field's size.

I think the main language we need around bitfields is how they're represented in aggregate initialization and inside classes (although none of the classes chapter has been written yet).

llvm-beanz (approved):

specs/language/classes.tex
```diff
@@ -0,0 +1,27 @@
...
+  bool B : 1;
+};
+\end{HLSL}
+\p A bit-field can be unnamed.  Unnamed bit-fields cannot be accessed or initialized.
+\begin{HLSL}
+struct Foo {
+  uint : 10;
+};
+\end{HLSL}
+\p The explicit size of a named bit-field can be any number of bits greater than zero and less than or equal to the size of the specified type.
```
llvm-beanz:
```suggestion
\p The explicit size of a named bit-field must be a constant expression of integer type, and must resolve to a number greater than zero and less than or equal to the size of the specified type in bits.
```

specs/language/classes.tex
```diff
@@ -0,0 +1,27 @@
...
+\p The explicit size of an unnamed bit-field can also be zero.
+\p A bit-field cannot be a static member.
+\Sec{Static Members}{Classes.Static}
+\Sec{Conversions}{Classes.Conversions}
```
llvm-beanz, line 20: This conversions section is only true for integral conversions (since bit-fields are only integral types), and should probably be folded in the conversions chapter.

specs/language/classes.tex
```diff
@@ -0,0 +1,27 @@
...
+\p A glvalue of a named bit-field with underlying type \texttt{T} can be converted
```
llvm-beanz: This seems correct to me. This is an area where HLSL `inout` gets to be a bit different from C++ because in C++ you cannot bind a bit-field to a reference (https://godbolt.org/z/Gn7xva6dE). In HLSL this should be "legal" because of the copy-in/copy-out.

I think you could put this paragraph in the "Integer Promotions" section largely unchanged.

specs/language/classes.tex
```diff
@@ -0,0 +1,27 @@
...
+to a cxvalue of type \texttt{T$^\prime$} if there exists an implicit conversion
+from \texttt{T} to \texttt{T$^\prime$}.
+\p A glvalue or a prvalue of a type \texttt{T} can be converted to a glvalue
```
llvm-beanz: I think we can drop this paragraph. There isn't really a conversion happening to the bit-field. Older C++ specifications had notes about assigning signed values into bitfields being implementation defined, but a lot of that language is removed in newer specs because it isn't implementation-defined if we accept that signed integers are twos complement.

C++ addresses bit-field truncation (and general storing) not through conversions, but through definitions of the operators and initializers. I think that probably makes the most sense for us too since we don't represent bit-fields as types in our AST.

specs/language/conversions.tex
```diff
@@ -31,6 +31,9 @@
 \p If the glvalue refers to an array of type \texttt{T}, the prvalue will refer
 to a copy of the array, not memory referred to by the glvalue.
 
+\p If the glvalue refers to a bit-field of type \texttt{T}, the conversion
+produces an object of the underlying type of the bit-field and not a prvalue.
```
llvm-beanz:
```suggestion
produces an object of the underlying type of the bit-field.
```

It does still produce a prvalue.

llvm-beanz (approved):

PR #748, opened 2025-12-03
[0043] Update groupshared args proposal with information about overloads

### LINKED ISSUE #745: [0043] update groupshared arguments proposal
Update groupshared argument proposal with details about older compilers errors when programs use groupshared parameters. 
Update proposal with information about how we are going to handling overloads and mangling.

spall:
Update groupshared args proposal with info about overloads and error produced by old compilers.

Closes #745

### DIFF
--- a/proposals/0043-groupshared-arguments.md
+++ b/proposals/0043-groupshared-arguments.md
@@ -137,8 +137,8 @@ void caller() {
 
 The `groupshared` type annotation keyword will be allowed on function parameter
 declarations.  In language modes before HLSL 202x, a warning will be produced,
-but it should still be supported if the compiler supports the feature.  If the
-compiler does not support the feature an error should be produced.
+but it should still be supported if the compiler supports HLSL 202x.  If the
+compiler does not support HLSL 202x an error should be produced.
 
 ```
 error: 'groupshared' is not a valid modifier for a parameter


proposals/0043-groupshared-arguments.md
```diff
@@ -85,11 +85,64 @@ this proposal is preferred to waiting until references can be finalized.
...
+}
+```
+
 ### Errors and Warnings
 
 The `groupshared` type annotation keyword will be allowed on function parameter
 declarations.  In language modes before HLSL 202x, a warning will be produced,
-but it should still be supported if the compiler supports the feature.
+but it should still be supported if the compiler supports the feature.  If the
+compiler does not support the feature an error should be produced.
```
V-FEXrt: Is it optional for a compiler to support groupshared parameters? Or is this more of a "clang doesn't do this yet, but will eventually"?

llvm-beanz: Any compiler that supports HLSL 202x should support this feature.

tex3d: Since all compilers under development should support the feature, I think this is just meant to capture the fact that prior versions of DXC compilers will produce an error in this case.

spall: Yes it essentially means any compiler which supports 202x supports this feature even if you don't use the 202x language mode. And yes to what @tex3d said.

V-FEXrt: I see. I don't feel strongly about it but maybe a better wording is 

```
In language modes before HLSL 202x, a warning will be produced,
but it should still be supported if the compiler supports HLSL 202x.  If the
compiler does not support HLSL 202x an error should be produced.
```

"this feature" feels a bit vague while calling out that "support HLSL 202x implies support groupshared pre-HLSL 202x" feels a bit more precise

Not blocking on it though

Issue #696, opened 2025-10-23
[0035] Inconsistency in form of matrix producing dxil ops


pow2clk:
**Which proposal does this relate to?**
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md

**Describe the issue or outstanding question.**
`dx.op.matrixOuterProduct` returns a MatrixRef which makes sense since it produces a matrix and doesn't consume any aspect of any previous matrix to produce it. Other DXIL operations that produce matrices without input from any aspect of the output matrix return void and take the output matrix as a parameter. This complicates some optimizations as already the output nature of the parameter is not clear to most passes while a return type is always understood. It also seems inconsistent enough to surprise DXIL consumers. It's possible these forms were inherited from the an earlier related spec which was fond of void operators and pointer parameters, though this made merging duplicate annotate and other calls impossible.

I propose changing the Load, Cast, and Splat/Fill operators to return matrixref and lose the output parameter. (Note that, as written, the fillMatrix operator had no return type, but could be inferred to be void)

```llvm
declare %dx.types.MatrixRef @dx.op.fillMatrix.[TY](
  immarg i32,            ; opcode
  [Ty]                         ; fill value
  )

declare  %dx.types.MatrixRef @dx.op.castMatrix(
  immarg i32,                   ; opcode
  %dx.types.MatrixRef    ; matrix source
  )

declare %dx.types.MatrixRef @dx.op.matrixLoadFromDescriptor(
  immarg i32,               ; opcode
  %dx.types.Handle,   ; ByteAddressBuffer
  i32,                            ; Offset
  i32,                            ; Stride
  i32,                            ; matrix layout
  )

declare %dx.types.MatrixRef @dx.op.matrixLoadFromMemory.p[Ty](
  immarg i32,                 ; opcode
  [Ty] * addrspace(4),   ; groupshared T[M * N]
  i32,                               ; Offset
  i32,                               ; Stride
  i32,                               ; matrix layout
  )
```

**Additional context**
This will have some effect on the placement of the annotate calls that can be discussed at length in a more efficient forum.

PR #750, opened 2025-12-03
[0035] Revise createMatrix to create a unique reference

pow2clk:
After losing all its type information to annotateMatrix, createMatrix had no purpose and would have resulted in all calls being merged as redundant, however, some means of uniquely identifying a matrix is needed that annotateMatrix can't do because it needs to be generated for parameters as well as locals and globals (assuming we allow globals).

Previous handles for resources have had global variables that are able to uniquely identify them and workgraph handles have had parameters to the entry functions. Lacking either of these, it was impossible to distinguish allocas created for the initial declaration of a matrix and one that was created for a parameter that should get merged when inlining that function. By generating a unique createMatrix call where the local variable is declared or in the entry block for global variables.

Addresses some issues in #696 though not in the way I initially proposed there

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -1009,10 +1009,14 @@ represents other attributes of the created matrix.
 ```llvm
 declare %dx.types.MatrixRef @dx.op.createMatrix(
   immarg i32  ; opcode
+  %dx.types.MatrixRef *,       ; Alloca representing the unique matrix
   )
 ```
 
-Creates a new uninitialized matrix handle.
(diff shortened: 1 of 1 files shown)


damyanp:
@pow2clk - I think that things have moved on enough that we should probably close this PR?

damyanp:
Ok - please resolve the conflicts with this PR so we can get it reviewed and merged.

llvm-beanz:
The alternate approach to this is in #769, which removes the `createMatrix` op entirely.

pow2clk:
I hadn't looked it over much. It definitely does. Closing.

PR #752, opened 2025-12-04
Add 'IsDebuggerPresent' and detail DebugBreak more

JoeCitizen:
**HLSL Debugging Intrinsics**

- **Broadened Proposal Scope**: The proposal now covers two new debugging intrinsics:  
  - `DebugBreak()`: Triggers a breakpoint when a debugger is attached.
  - `dx::IsDebuggerPresent()`: Returns whether a graphics debugger is currently attached (DirectX only).

- **Motivation Expanded**: Highlights the need for conditional debug checks and selective breakpoints, allowing shader authors to perform expensive debugging operations only when needed.

- **Detailed API & Usage Examples**:  
  - Provides updated HLSL signatures and usage snippets for both intrinsics.
  - Demonstrates conditional execution of validation code based on debugger presence.

- **Implementation Details Added**:
  - Specifies DXIL lowering for both intrinsics (`dx.op.debugBreak`, `dx.op.isDebuggerPresent`).
  - Clarifies convergence requirements to prevent optimizations that might move breakpoints.
  - Notes that both intrinsics are only valid in Shader Model 6.10+.
  - Explains runtime and driver behavior.

- **SPIR-V Targeting**:
  - Outlines use of `NonSemantic.DebugBreak` for `DebugBreak()` in SPIR-V.
  - Specifies that `dx::IsDebuggerPresent()` has no SPIR-V equivalent.

- **Testing Section**: Introduces compiler, validation, and execution testing requirements to ensure correct integration and behavior.

### DIFF
--- a/proposals/0039-debugbreak.md
+++ b/proposals/0039-debugbreak.md
@@ -18,11 +18,12 @@ This proposal specifies two new HLSL debugging intrinsics:
 
 1. **`DebugBreak()`**: Triggers a breakpoint when a debugger is attached,
    allowing developers to pause execution and inspect shader state.
-2. **`IsDebuggerPresent()`**: Returns whether a graphics debugger is currently
+2. **`dx::IsDebuggerPresent()`**: Returns whether a graphics debugger is currently
    attached to the process, enabling conditional debug-only code paths.
 
-These intrinsics will lower to new DXIL operations for DirectX and to
-appropriate SPIR-V instructions for Vulkan targets.
+`DebugBreak()` will lower to new DXIL operations for DirectX and to appropriate
+SPIR-V instructions for Vulkan targets. `dx::IsDebuggerPresent()` is a DirectX
+extension and has no Vulkan/SPIR-V equivalent.
 
 ## Motivation
 
@@ -49,7 +50,7 @@ This proposal introduces two new HLSL intrinsics for debugging shader code.
 
 ```hlsl
 void DebugBreak();        // Trigger a breakpoint if debugger attached
-bool IsDebuggerPresent(); // Query if a debugger is attached
+bool dx::IsDebuggerPresent(); // Query if a debugger is attached (DirectX only)
 ```
 
 ### Example Usage
@@ -58,7 +59,7 @@ bool IsDebuggerPresent(); // Query if a debugger is attached
 [numthreads(8,1,1)]
 void main(uint GI : SV_GroupIndex) {
     // Conditional expensive debug checks
-    if (IsDebuggerPresent()) {
+  if (dx::IsDebuggerPresent()) {
         // Expensive validation only when debugging
         ValidateComplexInvariants();
     }
@@ -88,10 +89,10 @@ Triggers a breakpoint if a graphics debugger is attached. If no debugger is
 attached or the runtime does not support this operation, it is treated as a
 no-op. Execution continues after the breakpoint.
 
-#### `IsDebuggerPresent()`
+#### `dx::IsDebuggerPresent()`
 
 ```hlsl
-bool IsDebuggerPresent();
+bool dx::IsDebuggerPresent();
 ```
 
 Returns `true` if a graphics debugger is currently attached to the process,
@@ -99,7 +100,7 @@ Returns `true` if a graphics debugger is currently attached to the process,
 debug validation code only when a debugger is present:
 
 ```hlsl
-if (IsDebuggerPresent()) {
+if (dx::IsDebuggerPresent()) {
     // Expensive bounds checking, validation, etc.
     for (uint i = 0; i < arraySize; ++i) {
         if (data[i] < 0.0f || data[i] > 1.0f) {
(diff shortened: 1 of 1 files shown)


proposals/0039-debugbreak.md
```diff
@@ -29,99 +36,406 @@ millions of times without issue, but in one instance produces a bad result.
...
 
+Terminates shader execution abnormally. This function does not return. The
+implementation signals abnormal termination to the runtime, which may:
+- Terminate the dispatch/draw
+- Signal a device removed/lost condition
+- Log diagnostic information
+- Trigger a debugger break before termination (implementation-defined)
+
+The exact behavior is implementation-defined, but the shader must not continue
+execution past this point. This matches the semantics of C's `abort()` function.
```
llvm-beanz: I think that a proper `abort` should force termination of the thread immediately, and should be specified to terminate all threads in a dispatch "eventually". The tricky thing about "eventually" being that it would be legal to terminate the thread that calls `abort` and allow all other threads to terminate naturally by completing execution.

proposals/0039-debugbreak.md
```diff
@@ -29,99 +36,406 @@ millions of times without issue, but in one instance produces a bad result.
...
-This proposal introduces a new HLSL intrinsic `DebugBreak()`, a new header
-`assert.h` which will define the `assert()` macro in a C-compatible interface.
+This proposal introduces three new HLSL intrinsics and a new header `assert.h`
+which will define the `assert()` macro in a C-compatible interface.
+
+### Intrinsics
+
+```hlsl
+void DebugBreak();        // Trigger a breakpoint if debugger attached
+void Abort();             // Terminate execution abnormally
```
llvm-beanz: While most HLSL functions are `CamelCase` the ones that have direct C equivalents (like the math functions) we tend to match the C spellings. So I think `abort()` would be preferable.

@tex3d thoughts?

proposals/0039-debugbreak.md
```diff
@@ -29,99 +36,406 @@ millions of times without issue, but in one instance produces a bad result.
...
+2. **A/B Testing**: Enable assertions for a subset of users to detect issues
+   in production without impacting all users.
+
+3. **Development Builds**: Always enable abort for internal testing while
+   keeping it disabled for public releases.
+
+4. **Per-Pipeline Control**: Enable assertions only for specific shaders that
+   are suspected of containing bugs, minimizing performance impact.
+
+#### Backend Compiler Optimization
```
llvm-beanz: I don't think this should ever be allowed.

proposals/0039-debugbreak.md
```diff
@@ -29,99 +36,406 @@ millions of times without issue, but in one instance produces a bad result.
...
+capability bits.
+
+`Abort()` requires runtime support for abnormal termination signaling, but as
+all conforming implementations must handle this (even if by treating it as a
+no-op in release drivers), it also does not require a capability bit.
+
+### D3D12 Runtime Behavior for Abort
 
-The SPIRV usage will utilize the following instructions:
+#### Default Behavior: Abort as No-Op
```
llvm-beanz: Per my other comment, I'm pretty strongly opposed to this being allowed.

llvm-beanz (commented):
I think we need to work on the `abort` semantics (comment inline).

proposals/0039-debugbreak.md
damyanp: FastFail might be alternative as well.  I don't think that the actual semantics we'll end up with will really exactly match those of C's `abort()`.

proposals/0039-debugbreak.md
gracezhang72: For some more context, we've gotten feedback from game developers that it'll be nice to have a way to enable and disable asserts/aborts at PSO creation time so that they can enable validation without needing to recompile all their shaders. That way, they can ship with aborts in their shader disabled by default, and if they see a spike in GPU crashes they can temporarily enable the aborts for a small percentage of users and trigger a dump or additional logging for offline analysis

llvm-beanz: I'm not sure that complexity is warranted at the DXIL level. A root constant could be used to enable or disable custom assert macros if a specific software vendor needs such a case. Ideally this is a use case for specialization constants, but since we don't have those in DXIL it isn't really viable, but increasing the complexity and testing matrix for a DXIL operation by allowing it to behave differently under different runtime state is sub-optimal and should be avoided.

There are additional reasons why I'm opposed to this which cannot be discussed publicly at this time, but I'm happy to discuss privately.

proposals/0039-debugbreak.md
JoeCitizen: It would be nice if we can repurpose the existing FXC abort since it matches the C function in name at least

proposals/0039-debugbreak.md
JoeCitizen: At the language level terminating the thread (wave?) is about all we can do right? What happens after that is a runtime issue? i.e. if it should always result in a TDR for example.

proposals/0039-debugbreak.md
```diff
@@ -29,99 +34,293 @@ millions of times without issue, but in one instance produces a bad result.
...
-```c
-#if NDEBUG
-#define assert(cond) do { } while(false)
-#else
-#define assert(cond) do { if (!cond) DebugBreak();} while(false)
-#endif
+The value returned is uniform across all threads in a dispatch/draw and remains
+constant for the duration of shader execution.
+
+#### Enabling Semi-Optimized Development Builds
```
llvm-beanz: I don't know that this section really needs to be here. HLSL specs aren't user documentation, they're for compiler authors to implement language features.

proposals/0039-debugbreak.md
```diff
@@ -29,99 +34,293 @@ millions of times without issue, but in one instance produces a bad result.
...
-A new header `assert.h` is added and included with the compiler packaging which
-implements the `assert` macro:
+Triggers a breakpoint if a graphics debugger is attached. If no debugger is
+attached or the runtime does not support this operation, it is treated as a
+no-op. Execution continues after the breakpoint.
+
+#### `IsDebuggerPresent()`
+
+```hlsl
+bool IsDebuggerPresent();
```
llvm-beanz: This will need to be in the `dx` namespace as there is no Vulkan equivalent that we can lower to for SPIRV generation.

proposals/0039-debugbreak.md
```diff
@@ -29,99 +34,293 @@ millions of times without issue, but in one instance produces a bad result.
...
+
+### Shader Model Requirements
+
+These instructions will only be valid in Shader Model 6.10 or later.
+
+Because `DebugBreak()` and `IsDebuggerPresent()` can be treated as no-ops when
+no debugger is present, they are required supported features and do not require
+capability bits.
+
+### D3D12 Runtime Behavior for DebugBreak
```
llvm-beanz: I would prefer to remove D3D-focused language from the HLSL specifications. This isn't where we want users looking for documentation.

proposals/0039-debugbreak.md
```diff
@@ -29,99 +34,293 @@ millions of times without issue, but in one instance produces a bad result.
...
+While this instruction is not widely supported by Vulkan debuggers, it is
+supported by NVIDIA's NSight and can be safely ignored by Vulkan runtimes.
+
+#### IsDebuggerPresent
+
+There is no direct SPIR-V equivalent. Implementations may:
+- Lower to a specialization constant that can be set by the debugger
+- Use a vendor-specific extension
+- Return a conservative value (e.g., always `false` in release, always `true`
+  in debug builds)
```
llvm-beanz: I commented this above, but it seems more appropriate that we put this function in the `dx` namespace and make it unavailable to Vulkan targets if it has no lowering. There is no vendor extension equivalent that I'm aware of, and using a specialization constant isn't really something the compiler can just do it would require user involvement.

proposals/0039-debugbreak.md
```diff
@@ -29,99 +34,293 @@ millions of times without issue, but in one instance produces a bad result.
...
+capability bits.
+
+### D3D12 Runtime Behavior for DebugBreak
+
+#### Default Behavior: DebugBreak as No-Op
+
+By default, when no debugger is attached, `DebugBreak()` is treated as a no-op
+i.e the `dx.op.debugBreak` operation has no effect at runtime.
+
+The allows for driver back end compilers to optimize away these instructions
```
gracezhang72: The D3D spec currently has the default as no-op such that the backend compiler can optimize away instructions that lead up to the debug break and even if a debugger was attached after PSO creation the debugbreak would still be a no-op, we still want the three debugbreak configuration options (no-op, debugbreak if debugger else no-op, and abort/trigger TDR), is that right?

gracezhang72 (approved):

proposals/0039-debugbreak.md
```diff
@@ -29,99 +35,177 @@ millions of times without issue, but in one instance produces a bad result.
...
-```c++
-#include <assert.h>
+### Example Usage
 
+```hlsl
 [numthreads(8,1,1)]
 void main(uint GI : SV_GroupIndex) {
-    assert(GI < 8);
+    // Conditional expensive debug checks
+  if (dx::IsDebuggerPresent()) {
```
llvm-beanz, line 62: nit:
```suggestion
    if (dx::IsDebuggerPresent()) {
```

Issue #581, opened 2025-07-22
[0035] Address feedback from David Neto


llvm-beanz:
**Which proposal does this relate to?**
[[0035]- `linalg::Matrix`](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

**Describe the issue or outstanding question.**

@dneto0 left some feedback on the `linalg::Matrix` proposal after it merged see the comments from this review:
https://github.com/microsoft/hlsl-specs/pull/556#pullrequestreview-3034953629

llvm-beanz:
Closing as I believe the latest spec addresses all this feedback, or the relevant bits have changed making the feedback not applicable.

PR #590, opened 2025-07-25
[0035] Add Multiply overload for explicit output type

llvm-beanz:
This adds a second overload of linalg::Multiply to support differing input matrix types if an explicit output type is specified. To make sure this works I also fixed the SFINAE magic in the HLSL implementation which caused some cascading source updates.

This additional overlod was mentioned in the feedback in #257

PR #759, opened 2025-12-17
[0035] Address a swath of feedback

### LINKED ISSUE #574: [0035] Define which stages each bit of functionality is available in

The matrix-matrix operations are available in compute-like shader stages (Amplification, Mesh and Compute), but the matrix-vector operations should be available in any stage.

### LINKED ISSUE #618: [0035] SumAccumulate matrix operand dimensions and types
[0035 - linalg::Matrix](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

The DirectX WaveMatrix feature that was previously previewed had a `SumAccumulate` operation which I translated into the new spec, but I didn't properly reason about the matrix dimensions. Matrix sum operations should require the same dimensions, which either means we should have both operands be the same matrix use, or we require the matrices be square.

We should follow up on this to figure out the right solution.

### LINKED ISSUE #653: [0035] Add transpose to cast

We should have an optional transpose flag on the `Cast` operation.

**Additional context**
Converting matrices between types may or may not require transposing, and we should make that optional for the user to control.

### LINKED ISSUE #681: [0035] For thread scope matrices, make layout of matrix in Matrix::load a compile-time constant

In the example below:
```cpp
[NumThreads(1,1,1)]
void main(){
 using namespace dx::linalg;
  using MatrixBTy =
      Matrix<MatrixComponentType::U32, 32, 16, MatrixUse::B, MatrixScope::Thread>;

  vector<uint32_t, 32> Vec = (vector<uint32_t, 32>)0;
  uint32_t check = Buf.Load<uint32_t>(0);
  MatrixBTy MatB ;
  if (check == 0) {
  	MatB = MatrixBTy::Load(Buf, 0, 32 * 4, MatrixLayout::RowMajor);
  } else {
  	MatB = MatrixBTy::Load(Buf, 200, 32 * 4, MatrixLayout::RowMajor);
  }
  
  vector<uint32_t, 16> Accum = Multiply<uint32_t>(Vec, MatB);
  RWBuf.Store< vector<uint32_t, 16> >(0, Accum);
```
Not knowing the layout of the loaded matrix at compile time can result in poor codegen as matrix loads maybe optimized based on the layout. In the above example, a requirement to have the layouts as a compile-time constant and the same value in all branches(?) would be beneficial.

### LINKED ISSUE #696: [0035] Inconsistency in form of matrix producing dxil ops

### LINKED ISSUE #744: [0035] Need to specify `Matrix::Accumulate` is an atomic operation
[0035-linalg-matrix.md](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

The specification does not explicitly state that `Matrix::Accumulate(RWByteAddressBuffer)` performs atomics operations, and the granularity of the atomics (e.g., component-wise).

We may want to consider renaming this function to `Matrix::InterlockedAccumulate` for consistency.

### LINKED ISSUE #755: [0035] Consider removing SumAccumulate

I don't think `Matrix::SumAccumulate` is an appropriate operation to include. As far as I can tell, there is no type of hardware acceleration for this operation. I don't see a reason for it to be defined as `C += A + B`, when it could just as easily be `C += A` and `C += B` if those were interesting things to do. The current header also defines `A` and `B` as MxK and KxN, but for elementwise addition, they'd both need to be MxN instead (matching dimensions with C).

llvm-beanz:
This PR folds in a bunch of changes to address feedback that has been backed up. One big change is the removal of `annotateMatrix` in favor of type mangling. Using type mangling reduces the need for the compiler to emit and de-duplicate `annotateMatrix` calls, and aligns more closely to how we'll support this in modern LLVM using target extension types. This is a first step to addressing matrix lifetime issues.

* Mangle matrix attributes into `AttributedMatrixRef`
* Added transpose compile-time argument to `cast` (Resolves #653).
* Renamed `Accumulate` operations that operate on memory to `InterlockedAccumulate`, and clarify atomicity (Resolves #744).
* Replaced `SumAccumulate` with new `Accumulate` on A and B matrices (Resolves #755, Resolves #618).
* Layout for `Matrix::Load` on thread-scope matrices is now compile-time constant (Resolves #681).
* `InterlockedAccumulate` for thead-scope matrices no longer takes a layout parameter since it must be OuterProductOptimal.
* Added clarification of matrix stage availability (Resolves #574).
* Fixed the form of the outerProduct DXIL op to not create a matrix (Resolves #696).
* Clarify `Stride` parameters (partial feedback from #581).

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -1203,7 +1203,7 @@ The validator will ensure that the group shared target memory is large enough
 for the write.
 
 ```llvm
-declare i32 @dx.op.matrixQueryAccumulatorLayout.v[NUM][TY](
+declare i32 @dx.op.matrixQueryAccumulatorLayout(
   immarg i32,            ; opcode
   )
 ```
@@ -1215,7 +1215,7 @@ layout while a return value of `1` will denote that accumulator matrices are `B`
 layout.
 
 ```llvm
-declare void @dx.op.matrixOp(
+declare void @dx.op.matrixMulOp(
   immarg i32,            ; opcode
   %dx.types.MatrixRef,   ; matrix A
   %dx.types.MatrixRef,   ; matrix B
@@ -1241,7 +1241,7 @@ Must be called from wave-uniform control flow.
 
 
 ```llvm
-declare void @dx.op.matrixBinOp(
+declare void @dx.op.matrixAccumulate(
   immarg i32,            ; opcode
   %dx.types.MatrixRef,   ; matrix RHS
   %dx.types.MatrixRef,   ; matrix LHS
@@ -1409,7 +1409,7 @@ struct MatrixUse {
 ```
 
 This object will encode each matrix shape and element type as used by the DXIL
-operations in the `matrixOp` and `matvecmuladd` opcode classes.
+operations in the `matrixMulOp` and `matvecmuladd` opcode classes.
 
 The Scope field will encode one of the values defined in the [`DXILMatrixScope`
 enumeration](#dxil-enumerations).


damyanp (approved):

proposals/0035-linalg-matrix.md
```diff
@@ -1725,9 +1783,8 @@ void CoopVec() {
       Matrix<ComponentType::F16, 16, 16, MatrixUse::A, MatrixScope::Thread>;
 
   vector<float16_t, 16> Vec = (vector<float16_t, 16>)0;
-  MatrixATy MatA = MatrixATy::Load(
-      MBuf, 0, /* Row stride = number of columns * element size */ 16 * 4,
-      MatrixLayout::RowMajor);
+  MatrixATy MatA = MatrixATy::Load<MatrixLayout::RowMajor>(
```
V-FEXrt, line 1786: Is there a particular reason the Layout is a template parameter for the CoopVec use case but a formal parameter for the WaveMatrix use case? Having them different feels like an API wrinkle

proposals/0035-linalg-matrix.md
```diff
@@ -1196,7 +1226,6 @@ declare void @dx.op.matrixOp(
 Three opcodes are available for this operation class:
```
```suggestion
Two opcodes are available for this operation class:
```

proposals/0035-linalg-matrix.md
```diff
@@ -1210,6 +1239,27 @@ Validation rules will enforce that:
 
 Must be called from wave-uniform control flow.
 
+
+```llvm
+declare void @dx.op.matrixBinOp(
```
V-FEXrt: Not a strong opinion but I think we talked about this before but this isn't really a BinOp right? Since its not taking two values and producing a third instead it's taking two values and then assigning one of them.

If we only expect to have the one operation maybe it could be the more direct `matrixAccumulate`?

proposals/0035-linalg-matrix.md
```diff
@@ -1196,7 +1226,6 @@ declare void @dx.op.matrixOp(
 Three opcodes are available for this operation class:
 * Matrix Matrix Multiply: `C = A * B`
```
V-FEXrt, line 1227: Can't leave a comment above since its behind the "expand code" bit but if `matrixOp` is just multiply variants now then `dx.op.matrixMulOp` is worth considering as a name

proposals/0035-linalg-matrix.md
```diff
@@ -775,44 +794,54 @@ matrix object. When storing to `groupshared` memory, the matrix component data
...
-                   MatrixLayout Layout);
-```
+Matrix::InterlockedAccumulate(/*groupshared*/ T Arr[], uint StartIdx,
+                              uint Stride, MatrixLayoutEnum Layout);
 
-The following table specifies the valid values for the `Layout` parameter
-given the `Accumulate` method type and matrix scope.  All other combinations
-are unsupported:
+// When Scope == Thread, the following overload is available:
+void Matrix::InterlockedAccumulate(RWByteAddressBuffer Res, uint StartOffset,
```
V-FEXrt, line 822: is `Stride` missing from the param list here?

proposals/0035-linalg-matrix.md
```diff
@@ -1603,11 +1660,12 @@ template <ComponentEnum ComponentTy, SIZE_TYPE M, SIZE_TYPE N,
...
   using ElementType = typename __detail::ComponentTypeTraits<ComponentTy>::Type;
 
+  template <MatrixLayoutEnum Layout>
   static Matrix Load(ByteAddressBuffer Res, uint StartOffset, uint Stride,
-                     MatrixLayoutEnum Layout, uint Align = sizeof(ElementType));
+                     uint Align = sizeof(ElementType));
 
-  void Accumulate(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
-                  MatrixLayoutEnum Layout, uint Align = sizeof(ElementType));
+  void InterlockedAccumulate(RWByteAddressBuffer Res, uint StartOffset,
```
V-FEXrt, line 1667: repeat comment: Is `Stride` missing here?

proposals/0035-linalg-matrix.md
```diff
@@ -121,29 +122,35 @@ class Matrix {
...
+                        MatrixLayoutEnum Layout,
+                        uint Align = sizeof(ElementType));
 
   template <typename T, MatrixUseEnum UseLocal = Use>
   typename hlsl::enable_if<hlsl::is_arithmetic<T>::value &&
                                Use == MatrixUse::Accumulator && UseLocal == Use,
                            void>::type
-  Accumulate(/*groupshared*/ T Arr[], uint StartIdx, uint Stride,
-             MatrixLayoutEnum Layout);
+  InterlockedAccumulate(/*groupshared*/ T Arr[], uint StartIdx, uint Stride,
```
V-FEXrt, line 133: is `Align` missing here or implied in some other way

proposals/0035-linalg-matrix.md
```diff
@@ -991,44 +1023,40 @@ enum class DXILComponentType {
...
-  %dx.types.MatrixRef,       ; Matrix being annotated
-  %dx.types.MatrixProperties ; Struct containing annotation info
-  )
-```
-
-Defines a matrix as having the specified component type, dimensions, use, and
-scope.
-
 ```llvm
 declare @dx.op.fillMatrix.[TY](
```
```suggestion
declare void @dx.op.fillMatrix.[TY](
```

V-FEXrt: comment stolen from https://github.com/microsoft/hlsl-specs/issues/696 :)

proposals/0035-linalg-matrix.md
```diff
@@ -775,44 +794,54 @@ matrix object. When storing to `groupshared` memory, the matrix component data
...
+`groupshared` array.
+
+#### Matrix::InterlockedAccumulate
 
 ```c++
-template <MatrixScope ScopeLocal = Scope>
-typename hlsl::enable_if<Use == MatrixUse::Accumulator && ScopeLocal == Scope,
+
+// When Scope != Thread, the following overloads are available:
+template <MatrixUseEnum UseLocal = Use>
```
tex3d, line 807: I don't understand this template parameter `UseLocal`. It's not used except to make sure it's equal to `Use` (the default) which is required to be equal to `MatrixUse::Accumulator`. I see it on a number of methods where it's unused. Is this to satisfy some requirement elsewhere?

proposals/0035-linalg-matrix.md
llvm-beanz: For thread-scope matrices this needs to be a compile-time constant (https://github.com/microsoft/hlsl-specs/issues/681), so I moved it to a template parameter to make it easy to enforce in Sema rather than only verifiable in DXIL validation.

proposals/0035-linalg-matrix.md
llvm-beanz: Stride must be 0 for thread scope because the layout must be outer product optimal, so I removed the parameter.

proposals/0035-linalg-matrix.md
llvm-beanz: Same as above, stride must be 0, so I removed it.

proposals/0035-linalg-matrix.md
llvm-beanz: We don't really have a way to specify alignment against group shared allocations.

proposals/0035-linalg-matrix.md
llvm-beanz: It's how SFINAE works (see: https://github.com/microsoft/hlsl-specs/pull/590#discussion_r2254351451).

proposals/0035-linalg-matrix.md
V-FEXrt: Can we have a unified API and make all of them compile-time constant or is that too restrictive?

llvm-beanz: We don't strictly need them to be compile-time constant for other scopes and my guess is that it could be useful to parameterize it. We could consider APIs both ways. I'm not entirely sure what the best path is.

V-FEXrt: Sounds good, not going to block on this comment just a gut feeling on the api.

Lets leave it more flexible for now and we can always add the extra API later if its useful

PR #764, opened 2026-01-06
Store Group Shared Bytes used and not limit into metadata

JoeCitizen:
The runtime is interested in the actual bytes used not the limit requested.

### DIFF
--- a/proposals/0049-variable-groupshared-memory.md
+++ b/proposals/0049-variable-groupshared-memory.md
@@ -192,8 +192,8 @@ Validator must:
 * Sum byte sizes of all groupshared globals (respect alignment / padding like
 today).
 * Check attribute presence & argument correctness.
-* Ensure attribute appears only in compute/node/mesh/amplification and 
-SM >= 6.10.
+* Ensure that `kDxilGroupSharedLimitTag` metadata appears only in
+  compute/node/mesh/amplification and SM >= 6.10.
 * Check that the sum of all groupshared usage is less than or equal to the 
 specified limit (if present) OR the legacy 32k/28k limit (whichever is less).
 


proposals/0049-variable-groupshared-memory.md
```diff
@@ -153,38 +153,37 @@ argument`.
...
+A new entry-point metadata field is added to communicate the group shared
+memory usage to the runtime:
 
-* **`kDxilGroupSharedLimitTag`** (constant id TBD): An optional i32 metadata
-node storing the declared limit in bytes.
-  - If the `GroupSharedLimit` attribute is present, this metadata contains the
-  declared byte limit.
-  - If the attribute is absent, this metadata is omitted (or contains a value
-  of 0 to indicate no explicit limit was declared).
+* **`kDxilGroupSharedUsageTag`** (constant id TBD): An i32 metadata
```
tex3d: Actually, the module metadata needs to keep the limit, since this is used when verifying that usage is within the specified limit, late in compilation. If desired, we could add a usage tag to preserve the calculated usage in the module metadata, but we shouldn't remove the other tag.

proposals/0049-variable-groupshared-memory.md
```diff
@@ -196,8 +195,7 @@ today).
 * Ensure attribute appears only in compute/mesh/amplification and SM >= 6.10.
 * Emit / retain static usage metadata (existing) for runtime comparison against
 device capability.
-* Populate the new PSV0 `GroupSharedLimit` field with the attribute value (or 0
-if absent).
+* Populate the new PSV0 `GroupSharedUsage` field with the computed static usage.
```
tex3d: Actually, the validator doesn't populate this value, it should have been calculated and placed into PSV0 before validation. The validator will re-calculate the value, but just for comparison to make sure the value in PSV0 matches.

The validator also does not:
* Emit / retain static usage metadata (existing) for runtime comparison against
device capability.

This is placed in PSV0 during container assembly.

proposals/0049-variable-groupshared-memory.md
tex3d: I suppose, technically, we could remove the tag in final DXIL, since it's only needed up to container assembly, but that presents a problem if loading / reassembling the module, unless this value is reloaded from the PSV0 or RDAT (which would be usage instead of max).

Avoiding any new tag in DXIL metadata would have the advantage that this feature could be enabled for any shader model, as long as the runtime supports the new validation. But if we were to do that, we probably would need a new feature flag when using more than default max to prevent running on old runtimes that don't check.

proposals/0049-variable-groupshared-memory.md
```diff
@@ -193,11 +192,10 @@ Validator must:
 * Sum byte sizes of all groupshared globals (respect alignment / padding like
 today).
 * Check attribute presence & argument correctness.
-* Ensure attribute appears only in compute/mesh/amplification and SM >= 6.10.
-* Emit / retain static usage metadata (existing) for runtime comparison against
-device capability.
-* Populate the new PSV0 `GroupSharedLimit` field with the attribute value (or 0
-if absent).
+* Ensure attribute appears only in compute/node/mesh/amplification and 
+SM >= 6.10.
```
tex3d: I think you mean the `kDxilGroupSharedLimitTag` metadata, right?
```suggestion
* Ensure that `kDxilGroupSharedLimitTag` metadata appears only in
  compute/node/mesh/amplification and SM >= 6.10.
```

tex3d: This might be implied, but this metadata shouldn't be required when usage is <= default limits, right? Also, we could decide whether to enforce that (ie. limit must not be present on shader that has usage <= default limits).

tex3d: Or maybe we want to allow it when smaller, just so a shader can specify a smaller limit and force a compiler failure if they exceed that?

PR #769, opened 2026-01-12
[0035] Rework specification to treat matrices as values

### LINKED ISSUE #756: [0035] Ill defined memory/lifetime semantics for linalg::Matrix

I don't think the memory semantics for the `linalg::Matrix` type are well enough defined.

The crux of the problem is that in HLSL the object looks like it holds or automatically manages the memory for the contents of the matrix. This object holds a handle type that translates to a `dx.types.MatrixRef` in DXIL, which just references a matrix (like a pointer). The MatrixRef value in DXIL does not represent the actual matrix memory allocated, and the `dx.op.createMatrix` operation marks a point of allocation for the matrix referred to by the MatrixRef, though you don't know what's allocated or where until the handle is used in an annotation call.

For instance, depending on the `linalg::MatrixScope`, the `linalg::Matrix` could refer to per-thread matrix loaded from global mem, which may be meant to be folded into an operation through optimization without actually allocating any local memory. Or it could refer to local per-thread fragments making up a wave-scope matrix, or to groupshared memory, which isn't counted in our groupshared memory accounting for a shader.

The current plan is to use the constructor to call `dx.op.createMatrix`, but what happens when one matrix variable overwrites another? I assume the handle of the second would replace the first, leaving a dangling matrix allocation, if we don't fully optimize that away somehow.

So, how do we track matrix content allocation lifetime? What happens when you assign one `linalg::Matrix` to another, or pass it through a function parameter (which copies the handle)? What rules guarantee the driver can track lifetimes and eliminate copies or re-use memory? How can you tell when a particular matrix allocation is no longer in use and may be freed?

I think the vagueness in this area of the spec is the root problem with coming up with a correct, reliable implementation plan for the compiler.

llvm-beanz:
This updates the linalg spec to treat the AttributedMatrixRef objects as value objects in the SSA graph. This should address concerns about object lifetimes.

Fixes #756

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -1159,7 +1159,7 @@ If the index is out of range for the values stored in this thread the result is
 0.
 
 ```llvm
-declare %dx.types.LinAlgMatrix<mangling> @dx.op.matrixSetElement.[MatTy].[Ty](
+declare %dx.types.LinAlgMatrix<mangling> @dx.op.matrixSetElement.[MatTy].[MatTy].[Ty](
   immarg i32,                         ; opcode
   %dx.types.LinAlgMatrix<mangling>,   ; input matrix
   i32,                                ; thread-local index


proposals/0035-linalg-matrix.md
```diff
@@ -1022,16 +1022,8 @@ enum class DXILComponentType {
...
-
-
-```llvm
-  %dx.types.MatrixRef     = type { i8 * }
-```
-
-The compiler will also generate a permutation of typed matrix handles with names
-of the format `%dx.types.AttributedMatrixRef<mangling>`. The mangling scheme for
+The compiler will generate a permutation of typed matrix handles with names of
+the format `%dx.types.AttributedMatrixRef<mangling>`. The mangling scheme for
```
tex3d: If this type represents an abstract/proxy value of a matrix, why keep the "Ref" language?

proposals/0035-linalg-matrix.md
```diff
@@ -1044,23 +1036,32 @@ each type name will capture the type parameterization with the tokens `C`,
   %dx.types.AttributedMatrixRefC11M16N16U2S1    = type { i8 * }
 ```
 
-DXIL validation will enforce that an `AttributedMatrixRef` of any type may be
-bitcast to a `%dx.types.MatrixRef`, but the inverse cast will be disallowed.
+DXIL validation will enforce that an `AttributedMatrixRef` of any type may not
+be bitcast to any other type.
```
tex3d, line 1043: Why did you drop the bitcast to generic on input? Was it because we already will be overloading based on one matrix type for returned matrix, and bitcast on input would always insert a bitcast before the next use, so you figured it would be easier to just overload on each matrix input as well?

proposals/0035-linalg-matrix.md
```diff
@@ -1044,23 +1036,32 @@ each type name will capture the type parameterization with the tokens `C`,
...
+; Matrix<ComponentType::F32, 16, 16, MatrixUse::Accumulator, MatrixScope::Wave>
+!3 = !{%dx.types.AttributedMatrixRefC11M16N16U2S1* undef, i32 11, i32 16, i32 16, i32 2, i32 1 }
 ```
 
-Creates a new uninitialized matrix handle.
+### DXIL Operations
 
 ```llvm
-declare void @dx.op.fillMatrix.[TY](
+declare %dx.types.AttributedMatrixRef<mangling> @dx.op.fillMatrix.[TY](
```
```suggestion
declare %dx.types.AttributedMatrixRef<mangling> @dx.op.fillMatrix.[MatTy].[TY](
```

nit: this is missing the `MatTy` right?

tex3d: Missing return matrix overload:
```suggestion
declare %dx.types.AttributedMatrixRef<mangling> @dx.op.fillMatrix.[MatTy].[TY](
```

V-FEXrt: if yes, then it looks like `dx.op.matrixLoadFromDescriptor` is missing it too

proposals/0035-linalg-matrix.md
```diff
@@ -1070,11 +1071,10 @@ matrix component's type, a type conversion is applied following the rules
 documented in the [Conversions](#conversions) section.
 
 ```llvm
-declare void @dx.op.copyConvertMatrix(
-  immarg i32,            ; opcode
-  %dx.types.MatrixRef,   ; matrix destination
-  %dx.types.MatrixRef,   ; matrix source
-  immarg i1,             ; transpose
+declare %dx.types.AttributedMatrixRef<mangling> @dx.op.copyConvertMatrix.[MatTy][MatTy](
```
tex3d: Overload types should be separated by `.`
```suggestion
declare %dx.types.AttributedMatrixRef<mangling> @dx.op.copyConvertMatrix.[MatTy].[MatTy](
```

proposals/0035-linalg-matrix.md
```diff
@@ -1084,9 +1084,8 @@ unmodified after this operation is applied. Validation shall enforce that both
 matrices have the same scope and dimensions.
 
 ```llvm
-declare void @dx.op.matrixLoadFromDescriptor(
+declare %dx.types.AttributedMatrixRef<mangling> @dx.op.matrixLoadFromDescriptor(
```
tex3d: Missing matrix overload
```suggestion
declare %dx.types.AttributedMatrixRef<mangling> @dx.op.matrixLoadFromDescriptor.[MatTy](
```

proposals/0035-linalg-matrix.md
```diff
@@ -1185,13 +1183,13 @@ Validation rules will enforce that:
...
 
 ```llvm
-declare void @dx.op.matrixStoreToMemory.p[Ty](
-  immarg i32,            ; opcode
-  %dx.types.MatrixRef,   ; matrix
-  [Ty] *,                ; groupshared T[M * N]
-  i32,                   ; Offset
-  i32,                   ; Stride
-  i32,                   ; matrix layout
+declare void @dx.op.matrixStoreToMemory.[MatTy].p[Ty](
```
tex3d: For DXIL's overload naming, we don't add `p` to the overload type if a parameter is a pointer to the type. Instead, we handle that in the way we translate to/from the `llvm::Type` to overload string.
```suggestion
declare void @dx.op.matrixStoreToMemory.[MatTy].[Ty](
```

proposals/0035-linalg-matrix.md
```diff
@@ -1241,31 +1238,32 @@ Must be called from wave-uniform control flow.
 
 
 ```llvm
-declare void @dx.op.matrixAccumulate(
+declare %dx.types.AttributedMatrixRef<mangling> @dx.op.matrixAccumulate.[MatTyC].[MatTyA].[MatTyB](
```
tex3d: Shouldn't overload naming match comments below?
```suggestion
declare %dx.types.AttributedMatrixRef<mangling> @dx.op.matrixAccumulate.[MatTyC].[MatTyLHS].[MatTyRHS](
```

Also, here's a case where we could, in theory, use one overload type for both the return and LHS matrix, since they must be the same type, right? There is currently a restriction enforced in hctdb.py on multi-overload types that you must use each overload once, and in order, but we could probably relax that, as this doesn't really have to match llvm's built-in intrinsic convention.

If we keep the separate overload types, we should call out the needed validation constraint to ensure the types are the same.

proposals/0035-linalg-matrix.md
```diff
@@ -1154,11 +1152,11 @@ If the index is out of range for the values stored in this thread the result is
...
 
 ```llvm
-declare void @dx.op.matrixSetElement.[Ty](
-  immarg i32,            ; opcode
-  %dx.types.MatrixRef,   ; matrix
-  i32,                   ; thread-local index
-  [Ty]                   ; value
+declare void @dx.op.matrixSetElement.[MatTy].[Ty](
+  immarg i32,                                ; opcode
+  %dx.types.AttributedMatrixRef<mangling>,   ; matrix
```
tex3d: This operation is the only one (I think) that still treats the matrix value as if it is a reference to mutable storage and would write to that storage. This breaks the model of treating the matrix value type as a value of the matrix.

Two approaches fix this:
- Make this take a pointer, which forces some mutable storage location to be declared (like alloca), instead of still considering the value type to be a reference to some storage that the compiler can't reason about.
- Return the modified matrix value, just like how insertelement works for vectors.

The `@dx.op.matrixAccumulate` operation is similar, but was changed to return the resulting value. If we decide to take a pointer here instead, we could change `@dx.op.matrixAccumulate` to follow the same pattern if desired.

proposals/0035-linalg-matrix.md
```diff
@@ -1241,31 +1238,32 @@ Must be called from wave-uniform control flow.
...
-  %dx.types.MatrixRef,   ; matrix RHS
-  %dx.types.MatrixRef,   ; matrix LHS
+  %dx.types.AttributedMatrixRef<mangling>,   ; matrix LHS
+  %dx.types.AttributedMatrixRef<mangling>,   ; matrix RHS
   )
 ```
 
 This operation accumulates an `A` or `B` matrix into an accumulator following
-the form `LHS += RHS`.
+the form `LHS = LHS + RHS`.
```
tex3d, line 1256: Here, there's an alternative approach possible, where we take a pointer to matrix storage. See the related comment on `@dx.op.matrixSetElement`.

proposals/0035-linalg-matrix.md
```diff
@@ -1241,31 +1238,32 @@ Must be called from wave-uniform control flow.
...
 
 Must be called from wave-uniform control flow.
 
 ``` llvm
-declare <[NUMo] x [TYo]> @dx.op.matvecmul.v[NUMo][TYo].v[NUMi][TYi](
-  immarg i32,           ; opcode
-  %dx.types.MatrixRef,  ; matrix A
-  <[NUMi] x [TYi]>,     ; input vector
-  immarg i32            ; input interpretation type (DXILComponentType)
+declare <[NUMo] x [TYo]> @dx.op.matvecmul.[MatTy].v[NUMo][TYo].v[NUMi][TYi](
```
tex3d: I believe the overload ordering needs to be changed to match: return type, then parameter types in order:
```suggestion
declare <[NUMo] x [TYo]> @dx.op.matvecmul.v[NUMo][TYo].[MatTy].v[NUMi][TYi](
```

proposals/0035-linalg-matrix.md
```diff
@@ -1277,13 +1275,13 @@ Validation will enforce that:
...
 
 ``` llvm
-declare <[NUMo] x [TYo]> @dx.op.matvecmuladd.v[NUMo][TYo].v[NUMi][TYi].v[NUMo][TYb](
-  immarg i32,            ; opcode
-  %dx.types.MatrixRef,   ; matrix A
-  <[NUMi] x [TYi]>,      ; input vector
-  immarg i32,            ; input interpretation type (DXILComponentType)
-  <[NUMo] x [TYb]>,      ; bias vector
-  immarg i32             ; bias interpretation type (DXILComponentType)
+declare <[NUMo] x [TYo]> @dx.op.matvecmuladd.[MatTy].v[NUMo][TYo].v[NUMi][TYi].v[NUMo][TYb](
```
tex3d: Return before argument ordering:
```suggestion
declare <[NUMo] x [TYo]> @dx.op.matvecmuladd.v[NUMo][TYo].[MatTy].v[NUMi][TYi].v[NUMo][TYb](
```

proposals/0035-linalg-matrix.md
```diff
@@ -1339,9 +1337,8 @@ The validator will ensure that the group shared target memory is large enough
 for the write.
 
 ```llvm
-declare void @dx.op.matrixOuterProduct.v[M][TY].v[N][TY](
+declare %dx.types.AttributedMatrixRef<mangling> @dx.op.matrixOuterProduct.[MatTy].v[M][TY].v[N][TY](
```
tex3d: I think we will need validation to enforce use of the same `[TY]` for both vectors, as well as compatibility with the matrix type in the overload.

Requiring some custom validation code to enforce certain constraints across overloaded types is a general requirement that applies across a variety of matrix operations, I suppose.

proposals/0035-linalg-matrix.md
```diff
@@ -1520,7 +1517,7 @@ struct MatrixLayout {
     RowMajor = 0,
     ColMajor = 1,
     MulOptimal = 2,
-    OuterProductOptimal = 3,
+    OuterProductOptimal = 3
```
tex3d: Strange, the trailing comma is used for other enums above, so it's a bit odd that it was removed here, particularly at this time, in this PR.

tex3d (commented):
Treating the matrix value as the value of a matrix will almost completely fix the problems I was having, however there are some remaining issues that need to be decided.

First, for operations that modify a matrix, we must do one of two things:
- return the modified value (as `@dx.op.matrixAccumulate` does), or
- accept a pointer to the matrix type instead of a value, forcing explicit storage to be declared - through an alloca or global (static).

Thread-scoped matrix will not fit with this scheme well, since the point is to track the connection between one operation and another. In the pure SSA case without phi/select, it's possible to track and skip creating some thread-local matrix, but the restrictions required may appear arbitrary to a user, or overly restrictive in HLSL if we attempt to define high-level language restrictions and apply them in Sema.

If a value is stored and loaded, or it's used in a phi or select, this tracking will be obfuscated, and fusing the final operation in the back end may be difficult, or even impossible.

Since such restrictions would only apply to thread-scoped matrix objects, they would appear very odd and disconnected from most basic language constructs you can reason about.

As I have thought about this for a little while, I don't think there is a good clean solution other than exposing fused operations directly to the user, rather than pretending you can have a local thread scoped matrix value of this type.

Separating the operations can introduce a variety of potential problems not only for later translation, but also for reasoning about when memory accesses occur, in what control flow branch a memory operation is performed, and so on.

proposals/0035-linalg-matrix.md
llvm-beanz: The generic type no longer exists in the latest PR since every operation uses the attributed form.

proposals/0035-linalg-matrix.md
llvm-beanz: I'm happy to change the naming to anything that makes sense. I was minimizing churn.

proposals/0035-linalg-matrix.md
llvm-beanz: See below where the validation requirement is called out:
> Type of LHS is the same as the return type

proposals/0035-linalg-matrix.md
llvm-beanz: This was just an oversight from me moving too quickly trying to get something posted before an arbitrary deadline.

proposals/0035-linalg-matrix.md
llvm-beanz: Validation of the matrix element type being compatible with the vectors is already captured in the list of validation requirements below. I've added that the element type of the A and B vectors must be the same.

proposals/0035-linalg-matrix.md
llvm-beanz: I'm not sure how that got dropped... I wasn't intentionally changing the code samples. Re-added.

proposals/0035-linalg-matrix.md
tex3d: To me, this PR changes the meaning of the type from a reference to some memory managed elsewhere to a proxy value of a matrix. It makes breaking changes to most operations, and eliminates the generic type and bitcasting.

Therefore, this seems like the right time to change the name, otherwise it will generate more confusion down the road, and it will become harder to change the longer we leave it.

proposals/0035-linalg-matrix.md
```diff
@@ -1070,11 +1071,10 @@ matrix component's type, a type conversion is applied following the rules
 documented in the [Conversions](#conversions) section.
 
 ```llvm
-declare void @dx.op.copyConvertMatrix(
-  immarg i32,            ; opcode
-  %dx.types.MatrixRef,   ; matrix destination
-  %dx.types.MatrixRef,   ; matrix source
-  immarg i1,             ; transpose
+declare %dx.types.AttributedMatrixRef<mangling> @dx.op.copyConvertMatrix.[MatTy].[MatTy](
```
anupamachandra: overload [MatTy1].[MatTy2]?

proposals/0035-linalg-matrix.md
anupamachandra: On the topic of names, do we need `Attribute`? Why not just dx.types.Matrix<mangling>?

pow2clk: It seems that no unattributed matrixrefs remain. Is that correct? There is still some language under "DXIL Validation" that refers both to matrixref, creatematrix ~~and bitcasts~~  (edit: sorry no bitcasts)

proposals/0035-linalg-matrix.md
```diff
@@ -1044,23 +1036,32 @@ each type name will capture the type parameterization with the tokens `C`,
...
-Creates a new uninitialized matrix handle.
+### DXIL Operations
 
 ```llvm
-declare void @dx.op.fillMatrix.[TY](
+declare %dx.types.AttributedMatrixRef<mangling> @dx.op.fillMatrix.[MatTy].[TY](
   immarg i32,            ; opcode
-  %dx.types.MatrixRef,   ; matrix
   [Ty]                   ; fill value
   )
```
pow2clk, line 1073: Potentially applicable to other calls, is this going to be readonly/readwrite/something else? If it's not readwrite, I suspect two fills that create distinct matrices will get merged. If it's not, then I don't see how we could remove unused fills.

proposals/0035-linalg-matrix.md
```diff
@@ -1154,11 +1152,10 @@ If the index is out of range for the values stored in this thread the result is
...
 ```llvm
-declare void @dx.op.matrixSetElement.[Ty](
-  immarg i32,            ; opcode
-  %dx.types.MatrixRef,   ; matrix
-  i32,                   ; thread-local index
-  [Ty]                   ; value
+declare %dx.types.AttributedMatrixRef<mangling> @dx.op.matrixSetElement.[MatTy].[Ty](
+  immarg i32,                                ; opcode
+  i32,                                       ; thread-local index
+  [Ty]                                       ; value
```
mjbedy: I'm not sure I understand how SetElement interacts with the AttributedMatrixRef - this is only partially modifying a matrix, so won't have a full matrix value to return?

tex3d: You're right, you caught the missing parameter!
```suggestion
declare %dx.types.AttributedMatrixRef<mangling> @dx.op.matrixSetElement.[MatTy].[Ty](
  immarg i32,                                ; opcode
  %dx.types.AttributedMatrixRef<mangling>    ; input matrix
  i32,                                       ; thread-local index
  [Ty]                                       ; value
```

proposals/0035-linalg-matrix.md
```diff
@@ -1044,23 +1036,32 @@ each type name will capture the type parameterization with the tokens `C`,
...
-### DXIL Operations
 
-```llvm
-declare %dx.types.AttributedMatrixRef<mangling> @dx.op.createMatrix<mangling>(
-  immarg i32  ; opcode
-  )
+```
+!dx.targetTypes = !{!1, !2, !3}
+; Matrix<ComponentType::F16, 16, 16, MatrixUse::A, MatrixScope::Wave>
+!1 = !{%dx.types.AttributedMatrixRefC10M16N16U0S1* undef, i32 10, i32 16, i32 0, i32 1 }
```
anupamachandra: having the undefs as pointers would complicate things when downstream compilers use IR versions with opaque pointers. Can this be a value instead of a pointer?

proposals/0035-linalg-matrix.md
```diff
@@ -1022,16 +1022,8 @@ enum class DXILComponentType {
...
-```llvm
-  %dx.types.MatrixRef     = type { i8 * }
-```
-
-The compiler will also generate a permutation of typed matrix handles with names
-of the format `%dx.types.AttributedMatrixRef<mangling>`. The mangling scheme for
+The compiler will generate a permutation of typed matrix handles with names of
+the format `%dx.types.AttributedMatrixRef<mangling>`. The mangling scheme for
 each type name will capture the type parameterization with the tokens `C`,
 `M`, `N`, `U` and `S` denoting each encoded property.
```
pow2clk, line 1031: How will this be represented in the header? A templated Matrix "handle" to pass into the builtins could be trivially converted into the attributed type, but it can't be declared there using the mangling syntax

pow2clk: In offline discussion, it was said that the plan would be to have a __builtin_linalg_matrix (or something like that) type that requires attributes to represent this handle in the header.

proposals/0035-linalg-matrix.md
pow2clk: offline discussion determined that readonly makes sense and I agree

proposals/0035-linalg-matrix.md
```diff
@@ -1022,45 +1022,50 @@ enum class DXILComponentType {
...
+A new named metadata `dx.targetTypes` will be added to contain mappings of
+attributed matrix types to their type parameters avoiding needing to parse the
+type mangling. For the given examples above metadata of the form below will be
+generated:
+
+
+```
+!dx.targetTypes = !{!1, !2, !3}
+; Matrix<ComponentType::F16, 16, 16, MatrixUse::A, MatrixScope::Wave>
+!1 = !{%dx.types.LinAlgMatrixC10M16N16U0S1 undef, i32 10, i32 16, i32 0, i32 1 }
```
hekota:
```suggestion
!1 = !{%dx.types.LinAlgMatrixC10M16N16U0S1 undef, i32 10, i32 16, i32 16, 32 0, i32 1 }
```
Is this missing one of the dimensions?

proposals/0035-linalg-matrix.md
```diff
@@ -1022,45 +1025,50 @@ enum class DXILComponentType {
...
-  %dx.types.AttributedMatrixRefC10M16N16U1S1    = type { i8 * }
+  %dx.types.LinAlgMatrixC10M16N16U1S1    = type { i8 * }
   ; Matrix<ComponentType::F32, 16, 16, MatrixUse::Accumulator, MatrixScope::Wave>
-  %dx.types.AttributedMatrixRefC11M16N16U2S1    = type { i8 * }
+  %dx.types.LinAlgMatrixC11M16N16U2S1    = type { i8 * }
 ```
 
-DXIL validation will enforce that an `AttributedMatrixRef` of any type may be
-bitcast to a `%dx.types.MatrixRef`, but the inverse cast will be disallowed.
+DXIL validation will enforce that an `LinAlgMatrix` of any type may not
```
```suggestion
DXIL validation will enforce that a `LinAlgMatrix` of any type may not
```

proposals/0035-linalg-matrix.md
```diff
@@ -1154,11 +1159,11 @@ If the index is out of range for the values stored in this thread the result is
...
 
 ```llvm
-declare void @dx.op.matrixSetElement.[Ty](
-  immarg i32,            ; opcode
-  %dx.types.MatrixRef,   ; matrix
-  i32,                   ; thread-local index
-  [Ty]                   ; value
+declare %dx.types.LinAlgMatrix<mangling> @dx.op.matrixSetElement.[MatTy].[Ty](
+  immarg i32,                         ; opcode
+  %dx.types.LinAlgMatrix<mangling>    ; input matrix
```
```suggestion
  %dx.types.LinAlgMatrix<mangling>,   ; input matrix
```

proposals/0035-linalg-matrix.md
```diff
@@ -1106,9 +1112,8 @@ Validation rules will enforce that:
 * `Stride` is `0` if the `Layout` is not `RowMajor` or `ColMajor`
 
 ```llvm
-declare void @dx.op.matrixLoadFromMemory.p[Ty](
+declare %dx.types.LinAlgMatrix<mangling> @dx.op.matrixLoadFromMemory.[MatTy].p[Ty](
```
V-FEXrt: Are we keeping or dropping the `.p`? I see it here but removed in other places

proposals/0035-linalg-matrix.md
```diff
@@ -1241,31 +1245,32 @@ Must be called from wave-uniform control flow.
 
 
 ```llvm
-declare void @dx.op.matrixAccumulate(
-  immarg i32,            ; opcode
-  %dx.types.MatrixRef,   ; matrix RHS
-  %dx.types.MatrixRef,   ; matrix LHS
+declare %dx.types.LinAlgMatrix<mangling> @dx.op.matrixAccumulate.[MatTyC].[MatTyLHS].[MatTyRHS](
```
V-FEXrt, line 1248: Should this only have 2 overloads instead of 3 since the return type and LHS must match?

proposals/0035-linalg-matrix.md
llvm-beanz: Doh. We can drop it.

```suggestion
declare %dx.types.LinAlgMatrix<mangling> @dx.op.matrixLoadFromMemory.[MatTy].[Ty](
```

proposals/0035-linalg-matrix.md
llvm-beanz: I think it is better to have each type represented in the signature. I know it gets a bit verbose, but it makes it consistent.

V-FEXrt: I think there is another opcode where you only list the overload once even though there are two matrix types. I assumed this was intentional since the return must match the parameter.

I think it is MatrixSetElement?

V-FEXrt: okay yep, MatrixSetElement. I'm fine either way, my vote is also just for consistency:)

proposals/0035-linalg-matrix.md
```diff
@@ -1154,11 +1159,11 @@ If the index is out of range for the values stored in this thread the result is
 0.
 
 ```llvm
-declare void @dx.op.matrixSetElement.[Ty](
-  immarg i32,            ; opcode
-  %dx.types.MatrixRef,   ; matrix
-  i32,                   ; thread-local index
-  [Ty]                   ; value
+declare %dx.types.LinAlgMatrix<mangling> @dx.op.matrixSetElement.[MatTy].[Ty](
```
V-FEXrt: The way the infra is set up we actually need all three here.
```suggestion
declare %dx.types.LinAlgMatrix<mangling> @dx.op.matrixSetElement.[MatTy].[MatTy].[Ty](
```

proposals/0035-linalg-matrix.md
V-FEXrt: closing this since 3 is actually enforced by the infra :)

proposals/0035-linalg-matrix.md
```diff
@@ -1070,11 +1078,10 @@ matrix component's type, a type conversion is applied following the rules
...
 ```llvm
-declare void @dx.op.copyConvertMatrix(
-  immarg i32,            ; opcode
-  %dx.types.MatrixRef,   ; matrix destination
-  %dx.types.MatrixRef,   ; matrix source
-  immarg i1,             ; transpose
+declare %dx.types.LinAlgMatrix<mangling> @dx.op.copyConvertMatrix.[MatTy1].[MatTy2](
+  immarg i32,                         ; opcode
+  %dx.types.LinAlgMatrix<mangling>,   ; matrix source
+  immarg i1,                          ; transpose
```
pow2clk, line 1084: This confused me at first since the use and type applied to the source matrix now come from the overload instead of the destination matrix as it is no longer a parameter. The description below still suggests drawing from a destination matrix and it's not clear which overload is the source and which the target.

proposals/0035-linalg-matrix.md
pow2clk, line 1232: For accumulation, how will the operation know which matrix to accumulate into?

proposals/0035-linalg-matrix.md
```diff
@@ -1339,9 +1344,8 @@ The validator will ensure that the group shared target memory is large enough
 for the write.
 
 ```llvm
-declare void @dx.op.matrixOuterProduct.v[M][TY].v[N][TY](
+declare %dx.types.LinAlgMatrix<mangling> @dx.op.matrixOuterProduct.[MatTy].v[M][TY].v[N][TY](
```
pow2clk, line 1347: It would be useful to have some description of how MatTy is expected to be represented here. I expect similar to the mangling of the types themselves, but probably not quite so long.

proposals/0035-linalg-matrix.md
V-FEXrt: We should update the spec to be more clear, but following the convention that the overloads are "in order" and return type is "param 0" then the first overload is dest while the second is source

proposals/0035-linalg-matrix.md
V-FEXrt: Hmm, before this update C was provided as a parameter. We probably need to update this to still take C for multiply with accumulate. Chris is out this week, but once he's back we can get an opinion here.

proposals/0035-linalg-matrix.md
V-FEXrt: It still needs to be put into the spec but to answer your question, it will be `mCxMxNxUxSx` where `x` is the specific parameter value as an integer

proposals/0035-linalg-matrix.md
V-FEXrt: This PR is getting pretty heavy so we are punting this feedback to the next round of updates! https://github.com/microsoft/hlsl-specs/issues/780

proposals/0035-linalg-matrix.md
V-FEXrt: This PR is getting pretty heavy so we are punting this feedback to the next round of updates! https://github.com/microsoft/hlsl-specs/issues/781

proposals/0035-linalg-matrix.md
V-FEXrt: This PR is getting pretty heavy so we are punting this feedback to the next round of updates! https://github.com/microsoft/hlsl-specs/issues/782

Issue #781, opened 2026-01-28
[0035] Mat Mat Mul with Accumulation doesn't have anything to accumulate into


V-FEXrt:
**Which proposal does this relate to?**
[[0035]- `linalg::Matrix`](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

**Describe the issue or outstanding question.**
Matrix Matrix Multiply with Accumulation is defined as  `C += A * B`  but it currently produces a new C which would be uninitialized. The spec either needs to accept C as a parameter or clarify that we are accumulating into A or B

**Additional context**
https://github.com/microsoft/hlsl-specs/pull/769#discussion_r2737153299

llvm-beanz:
I think there are two ways we can handle this:
1) we make a dxil op that takes 3 matrices (A, B & C) and returns the new accumulator as `Result = C + (A * B)`.
2) we allow `@dx.op.matrixAccumulate` to accept "Accumulator" matrices as RHS arguments, and implement the multiply accumulate as multiply then accumulate.

I don't have a strong preference. Curious for thoughts from hardware vendors as this may impact backends if fusion is profitable (cc @anupamachandra & @mbedy-amd).

llvm-beanz:
@V-FEXrt, since I'm going to be quite busy next week can you handle updating the spec for this?

PR #791, opened 2026-02-06
[0035] Fix LinAlg MatrixMulAccumulate

V-FEXrt:
Fixes https://github.com/microsoft/hlsl-specs/issues/781

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -1224,16 +1224,15 @@ layout while a return value of `1` will denote that accumulator matrices are `B`
 layout.
 
 ```llvm
-declare %dx.types.LinAlgMatrix<mangling> @dx.op.linAlgMatrixMulOp.[MatTyC].[MatTyA].[MatTyB](
+declare %dx.types.LinAlgMatrix<mangling> @dx.op.linAlgMatrixMultiply.[MatTyC].[MatTyA].[MatTyB](
   immarg i32,                        ; opcode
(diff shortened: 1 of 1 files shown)


proposals/0035-linalg-matrix.md
```diff
@@ -1269,6 +1268,32 @@ Validation rules will enforce that:
 
 Must be called from wave-uniform control flow.
 
+```llvm
+declare %dx.types.LinAlgMatrix<mangling> @dx.op.linAlgMatrixMultiplyAccumulate.[MatTyR].[MatTyA].[MatTyB].[MatTyC](
```
pow2clk, line 1272: I'm late, but I encountered this same problem and came up with my own solution. The only way it differed is that I placed C first so it could have just a 3 way overload since return and C need to be the same anyway with only minor change to the overload checking logic instead of having to allow out of order extended overloads. 

So:
```
declare %dx.types.LinAlgMatrix<mangling> @dx.op.linAlgMatrixMultiplyAccumulate.[MatTyC].[MatTyA].[MatTyB](
  immarg i32,                        ; opcode
  %dx.types.LinAlgMatrix<mangling>   ; matrix C
  %dx.types.LinAlgMatrix<mangling>,  ; matrix A
  %dx.types.LinAlgMatrix<mangling>   ; matrix B
)
```
This still requires some change to the script logic to allow reusing an extended overload, but I think that we should allow such reuse just as we do for singular overloads

Issue #787, opened 2026-02-04
[0035] MatVecMul Operation doesn't provide an option to Transpose input Matrix prior to the Multiplication, need for Thread Scope Matrices


anupamachandra:
**Which document does this relate to?**
[[0035]- linalg::Matrix](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

**Describe the issue you see with the spec**
Since ThreadScope matrices donot support the ConvertMatrix Functionality to transpose the input Matrix, we require an `isTranspose` parameter in the MatVecMul(Add) operation to pass this information. Corresponding capability bit is needed in the Runtime spec to check if the formats published support transpose.

**Additional context**
Similar to the earlier CoopVec spec:
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md#matrix-transpose

tex3d:
Should this be combined into the `MatrixLayoutEnum` for the load, such as with a `MulOptimalTransposed` option, or could it be based on whether you're loading an `A` matrix or a `B` matrix?

When looking through the API in the spec, I noticed some confusing discrepancies around this:
- The `Thread` scope row of the table for `Matrix::Load(ByteAddressBuffer)` lists `any` for the Matrix Layout. Should this really support `OuterProductOptimal` layout?
- `Matrix::Load(ByteAddressBuffer)` does not specify any `MatrixUse` constraint, nor does the `@dx.op.linAlgMatrixLoadFromDescriptor` DXIL op. I thought this was supposed to be the key constraint to separate Matrix specializations usable between groupings: `Load + Mul[Add]` vs. `OuterProduct + Accumulate`.
- The HLSL definitions for `linalg::Multiply` and `linalg::MultiplyAdd` are inconsistent. The first says it takes a `B` matrix, and the second says it takes an `A` matrix. The DXIL op only says it takes an `A` matrix. And these constraints are inconsistent with the unconstrained `Matrix::Load(ByteAddressBuffer)` HLSL and DXIL operations.

PR #803, opened 2026-02-25
[0035] Specify a way to transpose Thread Scope matrices on load and constraint loads to A-type Matrices.

anupamachandra:
This PR fixes [PR#787](https://github.com/microsoft/hlsl-specs/issues/787).
Changes:
1. Extends the `MatrixLayoutEnum` to contain Transpose types that specify that the matrices will be transposed on load. This applies only to Thread scope optimal layout matrices. InterlockedAccumulate will continue to support only `OuterProductOptimal` layout for thread scope matrix, and not the transpose variant.
2. Also, we were missing a restriction on the Thread scope matrix loads, that they are only useful for A-type matrices as the only user of these loads are the Multiply(Matrix,Vector)/MultiplyAdd(Matrix,Vector) operations which only accept A Matrices. That constraint is added as well.

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -1660,7 +1660,7 @@ in the [`DXIL::ComponentType` enumeration](#dxil-enumerations).
 
 ## Appendix 1: HLSL Header
 
-[Compiler Explorer](https://godbolt.org/z/d3dsczPMj)
+[Compiler Explorer](https://godbolt.org/z/dGrfY1T69)
 > Note: this mostly works with Clang, but has some issues to work out still.
 
 ```cpp


proposals/0035-linalg-matrix.md
```diff
@@ -699,6 +701,13 @@ This operation may be called in divergent control flow when loading a thread
 scope matrix, and must be called in uniform control flow when loading a wave
 scope matrix.
 
+This operation permits loads from `RowMajor`, `ColumMajor` and `Optimal` layouts
+for Thread scope matrices, with an option to traspose the matrix by setting the
```
damyanp: typo 
```suggestion
for Thread scope matrices, with an option to transpose the matrix by setting the
```

proposals/0035-linalg-matrix.md
```diff
@@ -699,6 +701,13 @@ This operation may be called in divergent control flow when loading a thread
 scope matrix, and must be called in uniform control flow when loading a wave
 scope matrix.
 
+This operation permits loads from `RowMajor`, `ColumMajor` and `Optimal` layouts
+for Thread scope matrices, with an option to transpose the matrix by setting the
+`Layout` parameter to `MulOptimalTranspose` or `OuterProductOptimalTranspose`.
+Not all component types support transposing, it is implementation specific.
+Applications need to query the driver to determine if a matrix transpose is
+supported.
+
```
llvm-beanz, line 808: This sounds like we need some changes to the metadata and PSV0 to ensure that we are surfacing whether or not transpose is used.

anupamachandra: Yep, I can add a `Layout` field, but before that a couple of questions. Currently we have,
```
struct MatrixUse {
  uint32_t Dimensions[3]; // M, N, K
  uint8_t Scope;
  uint8_t OperandType;
  uint8_t ResultType;
  uint8_t RESERVED; // unused but reserved for padding/alignment.
  uint32_t Flags; // do we need this?
};
``` 
It seems to capture all the required details to validate the Matrix x Matrix Multiply operation, but not for Matrix x Vector operations, OuterProduct and the recently ratified VectorAccumulate. Should we add new structs to capture the relevant validation information for those?
```
struct MatrixVectorOpsUse {
uint8_t InputVecType;
uint8_t MatrixType;
bool MatrixIsTranposed;
uint8_t BiasType;
}
```
and a similar structs for OuterProdAcc and VectorAcc. If this seems ok, I'll add these to the PR.

llvm-beanz: The right approach might be to make a bigger change. Something like:

```
struct MatrixUseBase {
  uint8_t Scope;
  uint8_t OperandType;
}

struct MatrixMatrixUse : public MatrixUseBase {
  uint8_t ResultType;
  uint8_t RESERVED; // unused but reserved for padding/alignment.
  uint32_t Flags; // do we need this?
  uint32_t Dimensions[3]; // M, N, K
};

struct MatrixVectorUse : public MatrixUseBase{
   uint8_t InputVecType;
   uint8_t BiasType;
   uint32_t Flags; // 0x0001 - IsTransposed
   uint32_t Dimensions[2]; // M, N
   uint32_t Unused; // Only exists to make use struc
};

static_assert(sizeof(MatrixMatrixUse) == sizeof(MatrixVectorUse), "These should match!");
```

This allows us to have a single list of uses that are slightly different.

@jenatali & @JoeCitizen, this will have impact on the runtime side of things. Do you have thoughts?

llvm-beanz: Tagging @tex3d too, since a related topic came up in a meeting a few minutes ago.

jenatali: @llvm-beanz your proposal sounds good to me, yeah.

proposals/0035-linalg-matrix.md
jenatali: Seems like the right PR to fix this bad comment.

```suggestion
// Cooperative Vector operates on per-thread vectors multiplying against A
```

proposals/0035-linalg-matrix.md
anupamachandra: @tex3d If you are ok with the above, I'll make the changes to the PR.

tex3d: More extensive changes than these are needed. Leave the PSV0 updates to me, and I should have a PR up for these later today.

tex3d: Forgot to leave follow-up comment. PR for PSV0/RDAT updates is here:
https://github.com/microsoft/hlsl-specs/pull/832

PR #807, opened 2026-02-27
[Spec] Move paragraph numbers into the margin

llvm-beanz:
This is just a minor formatting fix to move the paragraph numbers into the margin instead of at the beginning of the first line.


llvm-beanz:
For visualization, the old behavior was:
Screenshot 2026-02-27 at 2 49 08 PM
The new behavior is:
Screenshot 2026-02-27 at 2 48 44 PM

PR #812, opened 2026-03-11
 [0040] Add alternative approach

kmpeng:
This PR documents an alternative approach that came up during discussion where `select` only supports value types and sampler support in DXC is reverted.


proposals/0040-select-type-support.md
kmpeng, line 8: Should this still be marked "Accepted"?

bogner: I think it's fine - the important thing is that the alternative is documented so that when this comes up in the TC we can include it in any discussion.

PR #817, opened 2026-03-12
[0035] Rework enumerations and type handling

### LINKED ISSUE #569: [0035] `linalg::Matrix` component types
[[0035] `linalg::Matrix`](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

We need to fully enumerate the supported component types and if any types are optional how that is exposed and handled.

### LINKED ISSUE #730: [0035] LinAlg Matrix DXIL enum naming
https://github.com/microsoft/hlsl-specs/blob/eaa900cfaa8324ed4fe913ccab984755eb508254/proposals/0035-linalg-matrix.md?plain=1#L952

This spec defines DXIL enumerations: `DXILMatrixUse`, `DXILMatrixScope`, and `DXILComponentType`. There are a couple issues with this:

DXIL enumerations are already defined under a `DXIL` namespace, so adding `DXIL` to the name is redundant.

There already exists a `DXIL::ComponentType`, and it already contains the new types referenced in this enum, only with different enum values. This existing enum is already used in a number of places in DXIL and the runtime interface, and it was originally the intention to use the same enum for matrix component types, otherwise we will have a higher likelihood of bugs and confusion. The existing type to numeric mappings already shipped in prior versions of DXIL cannot be changed without breaking DXIL.

We should not introduce another almost-identical enum with conflicting values just for this feature.

See in DXC:
https://github.com/microsoft/DirectXShaderCompiler/blob/201a5b598caed7158c0ea2aadf1a6e1590afa218/include/dxc/DXIL/DxilConstants.h#L164-L193

```cpp
enum class ComponentType : uint32_t {
  ...
```

See in llvm-project:
https://github.com/llvm/llvm-project/blob/66da12ae69757690dff48e4c43816c753797babd/llvm/include/llvm/Support/DXILABI.h#L59-L80

See https://github.com/llvm/llvm-project/issues/167784 for tangent on the `ElementType` name used here.

### LINKED ISSUE #779: Update ComponentType enum in LinAlg Matrix spec

The spec defines `ComponentType` enum, and DXC already has an enum with the same name and slightly different values in [DxilConstants.h](https://github.com/microsoft/DirectXShaderCompiler/blob/e7346acd9d4000035d8383c713ee966a5abdd7cb/include/dxc/DXIL/DxilConstants.h#L164). The enum values in the spec should be updated to match existing values in DXC.

### LINKED ISSUE #798: [0035] F8 componentEnum values don't match the compiler internals

[0035 - Linear Algebra Matrix](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

The values of ComponentType defined in the linagl.h header match the corresponding enum values internal values for the most part. The exception are the F8 types at the end, which follow sequentially. Since the internal enum has 8-bit integer types before those, they are offset by 2. This was something we tried to do for Cooperative Vector. It still seems sensible for LinAlg Matrix.

```hlsl
struct ComponentType {
  enum ComponentEnum {
    Invalid = 0,
    I1 = 1,
    I8 = 2,
    U8 = 3,
    I16 = 4,
    U16 = 5,
    I32 = 6,
    U32 = 7,
    I64 = 8,
    U64 = 9,
    F16 = 10,
    F32 = 11,
    F64 = 12,
    SNormF16 = 13,
    UNormF16 = 14,
    SNormF32 = 15,
    UNormF32 = 16,
    SNormF64 = 17,
    UNormF64 = 18,
    F8_E4M3 = 19,
    F8_E5M2 = 20,
  };
};
```

llvm-beanz:
This change seeks to address a few issues relating to the enumerations defined in the spec. Some of these are cosmetic, but there is one functional change. A recent conversation we decided (at least for our initial preview) to disallow elementwise operations for types that are not natively representable in HLSL. There is a small header change and spec wording to adjust to that.

The remaining changes here separate the linalg::ComponentType enum from the dxil::ComponentType enum. The enumerations are named the same and their values match, but the linalg::ComponentType is a subset of the dxil::ComponentType enum to allow only the valid cases through the header. I've also updated the spec to detail validation requirements at the DXIL level to enforce only valid values.

Fixes #569, Fixes #730, Fixes #779, Fixes #798

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -80,6 +80,8 @@ class Matrix {
   // packed in each scalar value.
   static const uint ElementsPerScalar =
       __detail::ComponentTypeTraits<ComponentTy>::ElementsPerScalar;
+  static const bool IsNativeScalar =
+      __detail::ComponentTypeTraits<ComponentTy>::IsNativeScalar;
 
   template <ComponentEnum NewCompTy, MatrixUseEnum NewUse = Use,
             bool Transpose = false>
@@ -100,13 +102,21 @@ class Matrix {
   Load(/*groupshared*/ T Arr[], uint StartIdx, uint Stride,
        MatrixLayoutEnum Layout);
 
-  uint Length();
+  template<ComponentEnum LocalComp = ComponentTy>
+  typename hlsl::enable_if<LocalComp == ComponentTy && IsNativeScalar, uint>::type
+  Length();
 
-  uint2 GetCoordinate(uint);
+  template<ComponentEnum LocalComp = ComponentTy>
+  typename hlsl::enable_if<LocalComp == ComponentTy && IsNativeScalar, uint2>::type
+  GetCoordinate(uint);
 
-  ElementType Get(uint);
+  template<ComponentEnum LocalComp = ComponentTy>
+  typename hlsl::enable_if<LocalComp == ComponentTy && IsNativeScalar, ElementType>::type
+  Get(uint);
 
-  void Set(uint, ElementType);
+  template<ComponentEnum LocalComp = ComponentTy>
+  typename hlsl::enable_if<LocalComp == ComponentTy && IsNativeScalar, void>::type
+  Set(uint, ElementType);
 
   void Store(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
              MatrixLayoutEnum Layout, uint Align = sizeof(ElementType));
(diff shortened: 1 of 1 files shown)


proposals/0035-linalg-matrix.md
```diff
@@ -985,40 +1039,49 @@ This feature adds the following new DXIL enumerations, which used as immediate
 arguments to the new operations.
 
 ```c++
-enum class DXILMatrixUse {
+namespace DXIL {
```
hekota, line 1042: The namespace is lowercase on lines 499 and 1585 and uppercase here. Which is correct?

llvm-beanz: They are two different enumerations. This is referring to the DXC DXIL headers for C++ which is uppercase, in the HLSL header (lines 499 & 1585), I'm using lowercase to align with other namespace naming conventions that already exist. (`vk`, `dx`, `hlsl`).

proposals/0035-linalg-matrix.md
```diff
@@ -485,29 +495,70 @@ encode the dimensions and input and output data types used by each shader in the
...
+  U64 = 7,
+  F16 = 8,
+  F32 = 9,
+  F64 = 10,
+  SNormF16 = 11,
+  UNormF16 = 12,
+  SNormF32 = 13,
+  UNormF32 = 14,
+  SNormF64 = 15,
+  UNormF64 = 16,
```
pow2clk, line 519: SNorms and UNorms were not to be included.

MortimerSC (commented):
[Fixes #

PR #818, opened 2026-03-12
[0035] Clarify uniformity and synchronization

llvm-beanz:
This adds some clarifications about uniformity and implicit synchronization of execution. It also explicitly states that _explicit_ synchronization may be required around `Load` and `Store` operations.

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -393,9 +393,14 @@ There are three supported matrix scopes: `Thread`, `Wave`, and `ThreadGroup`.
 
 Operations are categorized by their scope requirements. Some operations require
 uniform scope matrices (`Wave` or`ThreadGroup`), while others can operate on
-non-uniform (`Thread`) scope matrices. Operations that support non-uniform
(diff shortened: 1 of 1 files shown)


proposals/0035-linalg-matrix.md
```diff
@@ -703,6 +708,9 @@ represents the row or column stride in bytes. For the `Load` operations on
 `groupshared` arrays, the `Stride` argument is the count of elements in the
 `groupshared` array.
 
+Reads from memory through `Load` functions are not atomic and may require
```
V-FEXrt, line 711: nit: typo?

```suggestion
Reads from memory though `Load` functions are not atomic and may require
```

V-FEXrt: I'm actually not 100% sure if this is or isn't a typo lol, but if it is the same one was made for Store

llvm-beanz: I think I wrote it correctly.

"though" is a synonym of "despite", whereas "through" would mean "by way of" or "using". Might be more clear to write it as "using".

PR #819, opened 2026-03-17
[0035] Adding conversion primitives

### LINKED ISSUE #568: [0035] Fill out spec section on conversions

### LINKED ISSUE #786: [0035] dx.op.linAlgMatVecMulAdd dx.op. doesn't capture the signedness of the input and output vectors
[[0035]- linalg::Matrix](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

ThreadScope: The MatVecMul(Add) functions allow signed and unsigned integer component types in the declaration but this information is absent in the dxil op. The interpretation values allow conversions and cannot be used to distinguish between unsigned and signed declared types.

**Additional context**
See is "input/output unsigned" operands in the [Coopvec spec](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md)

llvm-beanz:
This provides a different approach to addressing some of the issues around types and conversions. Instead of allowing the arithmetic operations to convert data during the operations it implements a conversion operation that converts data before the operation.

Fixes #786, Fixes #568

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -1546,13 +1546,13 @@ documented [conversion rules](#data-conversion-rules).
 Validation will ensure that:
 * If the input interpretation type enum refers to a type that has a native DXIL
   scalar representation the input vector type matches that scalar type,
-  ohterwise the input vector type should be `i32` as if storing 32-bit opaque
+  otherwise the input vector type should be `i32` as if storing 32-bit opaque
   values.
 * If the output interpretation type enum refers to a type that has a native DXIL
   scalar representation the output vector type matches that scalar type,
-  ohterwise the output vector type should be `i32` as if storing 32-bit opaque
+  otherwise the output vector type should be `i32` as if storing 32-bit opaque
   values.
-* The output vector lenght must be equal to `NUMi` multiplied by number of
+* The output vector length must be equal to `NUMi` multiplied by number of
   elements per scalar in the input interpretatation divided by the number of
   elements per scalar in the output interpretation (see the `__detail::DstN`
   template).


proposals/0035-linalg-matrix.md
```diff
@@ -678,6 +715,23 @@ implementation detail to enable mapping `ComponentType` values to their
...
+#### linalg::Convert
+
+```c++
+template <ComponentEnum DestTy, ComponentEnum OriginTy, typename T, int N>
+vector<typename __detail::ComponentTypeTraits<DT>::Type,
+       __detail::DstN<DT, OriginTy, N>::Value>
+linalg::Convert(vector<T, N> Vec);
+```
+
+Converts a vector of date interpreted as the `OriginTy` to a vector of data in
```
nhaehnle: typo
```suggestion
Converts a vector of data interpreted as the `OriginTy` to a vector of data in
```

proposals/0035-linalg-matrix.md
```diff
@@ -320,6 +339,18 @@ void CoopVec() {
...
+      MatA, MakeInterpretedVector<ComponentType::F8_E4M3>(SomeData), MemBias);
+
+  // This example creates an interpreted vector where the data needs to be
+  // converted from a source type to a destination type.
+  vector<uint, 16> SomeData2 = (vector<uint, 16>)0;
+  vector<float16_t, 16> Layer7 = MultiplyAdd<float16_t>(
+      MatA,
+      MakeInterpretedVector<ComponentType::F8_E4M3, ComponentType::U32>(
+          SomeData2),
+      MemBias);
```
nhaehnle, line 342: I stumbled over this slightly sneaky combination of conversion and making an "interpreted vector".

Does this even compile, given that there is no ComponentTypeTraits specialization for F8_E4M3?

proposals/0035-linalg-matrix.md
```diff
@@ -66,12 +66,31 @@ template <typename T, int N, ComponentEnum DT> struct InterpretedVector {
...
 };
 
+template <ComponentEnum DestTy, ComponentEnum OriginTy, typename T, int N>
+vector<typename __detail::ComponentTypeTraits<DestTy>::Type,
+       __detail::DstN<DestTy, OriginTy, N>::Value>
+Convert(vector<T, N> Vec) {
+  vector<typename __detail::ComponentTypeTraits<DestTy>::Type,
+         __detail::DstN<DestTy, OriginTy, N>::Value>
+      Result;
+  /*__builtin_convert(Result, Vec, OriginTy, DestTy);*/
```
pow2clk: This is the first inclusion of a builtin in the spec. I'm not saying it's a bad idea, but given its uniqueness, was it indended? Should others include this brief nod to implementation?

proposals/0035-linalg-matrix.md
pow2clk: F8_E4M3 will get the default traits of type = uint, IsNativeScalar = false, and ElementsPerScalar = 4. This indicates it is packed into a 32-bit integer, though it's not explicitly called out and might benefit from some explicit mention of what types are packed, how they are used, and how that influences the dimension checking.

proposals/0035-linalg-matrix.md
```diff
@@ -66,12 +66,31 @@ template <typename T, int N, ComponentEnum DT> struct InterpretedVector {
       __detail::ComponentTypeTraits<DT>::ElementsPerScalar * N;
 };
 
+template <ComponentEnum DestTy, ComponentEnum OriginTy, typename T, int N>
```
anupamachandra: I’m a bit concerned about the design of `linalg::Convert` as a free function that returns a plain `vector<..., N>` rather than an `InterpretedVector`. In the current form, a call like:
```
vector<uint, 16> src = ...;
auto packedF8 =
  linalg::Convert<ComponentType::F8_E4M3, ComponentType::U32>(src);
```
yields a `vector` whose type is just something like `vector<uint, M>`. From that type alone, there is no way to know that the data stored in those uints is actually F8_E4M3‑packed data; the `ComponentType::F8_E4M3` information is essentially lost once you store the result in packedF8. Given a standalone `Convert` that only returns `vector<...>`, it seems easy to misuse: you end up with a `vector<uint, M>` whose actual meaning (F8 vs U32) is not known by the type system. For LinAlg ops this cannot be fed back without the interpreted vector wrapper.
Would it make more sense to:

1. Make `Convert` an operation on `InterpretedVector` (e.g. InterpretedVector<..., OriginTy>::Convert<DestTy>() returning an InterpretedVector<..., DestTy>), or
2. Define linalg::Convert so that result type is an InterpretedVector<..., DestTy>, keeping the interpretation information attached to the data?

proposals/0035-linalg-matrix.md
```diff
@@ -66,12 +66,31 @@ template <typename T, int N, ComponentEnum DT> struct InterpretedVector {
...
+  return Result;
+}
+
 template <ComponentEnum DT, typename T, int N>
 InterpretedVector<T, N, DT> MakeInterpretedVector(vector<T, N> Vec) {
   InterpretedVector<T, N, DT> IV = {Vec};
   return IV;
 }
 
+template <ComponentEnum DT, ComponentEnum OriginTy, typename T, int N>
```
pow2clk: If this is expected to close #786, then I expect the intent here is that when you want to use unsigned 32-bit integers, you need to use this to create an interpreted vector that includes the conversion that can be used to identify them as such? When it comes to interpreting the output, it would need to pass through a similar operation to keep the signed information at the DXIL level. That seems counterintuitive.

proposals/0035-linalg-matrix.md
```diff
@@ -1722,6 +1810,17 @@ InterpretedVector<T, N, DT> MakeInterpretedVector(vector<T, N> Vec) {
...
 }
 
+template <ComponentEnum DestTy, ComponentEnum OriginTy, typename T, int N>
+InterpretedVector<typename __detail::ComponentTypeTraits<DestTy>::Type,
+                  __detail::DstN<DestTy, OriginTy, N>::Value, DestTy>
+Convert(vector<T, N> Vec) {
+  vector<typename __detail::ComponentTypeTraits<DestTy>::Type,
+         __detail::DstN<DestTy, OriginTy, N>::Value>
+      Result;
+  /*__builtin_convert(Result, Vec, OriginTy, DestTy);*/
```
```suggestion
```

proposals/0035-linalg-matrix.md
```diff
@@ -72,6 +72,17 @@ InterpretedVector<T, N, DT> MakeInterpretedVector(vector<T, N> Vec) {
...
```
V-FEXrt: nit: we have been careful to not mention a specific intrinsic backing for the header
```suggestion

```

pow2clk: Not sure why my comment isn't showing up anymore, but I said much the same https://github.com/microsoft/hlsl-specs/pull/819#discussion_r2955113565

proposals/0035-linalg-matrix.md
```diff
@@ -675,6 +704,23 @@ implementation detail to enable mapping `ComponentType` values to their
 native HLSL element types and differentiating between types that have native
 scalar support.
 
+#### linalg::Convert
+
+```c++
+template <ComponentEnum DestTy, ComponentEnum OriginTy, typename T, int N>
+vector<typename __detail::ComponentTypeTraits<DT>::Type,
```
pow2clk: This description still returns a native vector instead of an InterpretedVector as indicated above and below.

proposals/0035-linalg-matrix.md
```diff
@@ -1702,6 +1784,12 @@ __MATRIX_SCALAR_COMPONENT_MAPPING(ComponentType::I64, int64_t)
 __MATRIX_SCALAR_COMPONENT_MAPPING(ComponentType::U64, uint64_t)
 __MATRIX_SCALAR_COMPONENT_MAPPING(ComponentType::F64, double)
 
+template <ComponentEnum DstTy, ComponentEnum SrcTy, int SrcN> struct DstN {
+  static const int Value =
+      (SrcN * ComponentTypeTraits<SrcTy>::ElementsPerScalar) /
```
pow2clk, line 1809: This can produce an output vector of greater than the maximum vector length which produces a confusing error message: https://godbolt.org/z/W8T6MEbhj

proposals/0035-linalg-matrix.md
```diff
@@ -1479,6 +1524,43 @@ Validation will ensure that:
...
 
+```llvm
+declare <[NUMo] x [TYo]> @dx.op.linAlgConvert.v[NUMo][TYo].v[NUMi][TYi](
+  immarg i32,                         ; opcode
+  <[NUMi] x [TYi]>,                   ; input vector
+  immarg i32,                         ; input interpretation type (DXIL::ComponentType)
+  immarg i32                          ; output interpretation type (DXIL::ComponentType)
+)
+```
+
```
V-FEXrt, line 1541: We should add language on validation rules ala the other dxil ops.

One that stands out is that the 2 vectors have the same dimension

proposals/0035-linalg-matrix.md
```diff
@@ -1479,6 +1524,57 @@ Validation will ensure that:
...
+Converts an input vector containing data of the input interpretation type to a
+vector containing data of the output interpretation type following the
+documented [conversion rules](#data-conversion-rules).
+
+Validation will ensure that:
+* If the input interpretation type enum refers to a type that has a native DXIL
+  scalar representation the input vector type matches that scalar type,
+  ohterwise the input vector type should be `i32` as if storing 32-bit opaque
```
```suggestion
  otherwise the input vector type should be `i32` as if storing 32-bit opaque
```

proposals/0035-linalg-matrix.md
```diff
@@ -1479,6 +1524,57 @@ Validation will ensure that:
...
+  values.
+* If the output interpretation type enum refers to a type that has a native DXIL
+  scalar representation the output vector type matches that scalar type,
+  ohterwise the output vector type should be `i32` as if storing 32-bit opaque
```
```suggestion
  otherwise the output vector type should be `i32` as if storing 32-bit opaque
```

proposals/0035-linalg-matrix.md
```diff
@@ -1479,6 +1524,57 @@ Validation will ensure that:
...
+  values.
+* The output vector lenght must be equal to `NUMi` multiplied by number of
```
```suggestion
* The output vector length must be equal to `NUMi` multiplied by number of
```

proposals/0035-linalg-matrix.md
llvm-beanz: I think at the moment this is behaving correctly unless you have an alternative suggestion.

proposals/0035-linalg-matrix.md
```diff
@@ -1722,6 +1830,17 @@ InterpretedVector<T, N, DT> MakeInterpretedVector(vector<T, N> Vec) {
   return IV;
 }
 
+template <ComponentEnum DestTy, ComponentEnum OriginTy, typename T, int N>
+InterpretedVector<typename __detail::ComponentTypeTraits<DestTy>::Type,
+                  __detail::DstN<DestTy, OriginTy, N>::Value, DestTy>
+Convert(vector<T, N> Vec) {
```
tex3d, line 1836: Couldn't we have a version that takes an `InterpretedVector` as input, for convenience?

llvm-beanz: We could. It would be fully header implemented so it isn't a big change. I wrote it this way because I expect the common usage pattern is to have a vector of native typed elements and want to convert it to a vector of non-native typed elements. I think with a bit of SFINAE magic we could also deduce the `OriginTy` for the likely common case.

llvm-beanz: I've filed an issue to track this refinement as a follow-up. Since it doesn't change the DXIL we can handle it post preview:
#835

PR #832, opened 2026-03-27
[0035] LinAlg: Add runtime metadata details: SVI0, PSV0, and RDAT

tex3d:
This change adds:

* LinearAlgebra feature use flag to the FeatureInfo DXIL container part (SFI0)
* extended feature use gathering process/rules and details for PSV0 and RDAT DXIL container parts for feature use beyond LinAlg Tier 1 minimum requirements.


proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+common gathering process is described first.
+
+This extended usage information is only captured for usage outside the required
+Tier 1 feature set.
+
+Basic process:
+
+* Iterate over specific LinAlg DXIL op calls.
+* If operation is outside minimum requirements, gather and merge detailed usage
+  information.
```
damyanp: Wondering if we should just always gather the detailed usage, and then only the runtime will need to be taught what the minimum requirements are?

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+  dimension. For example, a 32x24x16 shape can be merged with an existing
+  32x8x16 shape. Only the smaller shape needs to be recorded.
+* Merging can only occur within the same operation, scope, and component type
+  combination.
+* If, according to Tier 1 requirements, the shape merges with the minimum
+  required shape for a required operation/scope/type combination, it can be
+  considered as fitting within the minimum requirements, and need not be
+  recorded into extended usage data.
+
+Open questions:
```
damyanp: Is it really important to implement this shape merging?  What are we looking to save here?

proposals/0035-linalg-matrix.md
tex3d: We could do that. It would end up with extra data that doesn't need to be checked, but we could consider this an optimization to reduce container size and perhaps a bit of extra runtime validation cost.

proposals/0035-linalg-matrix.md
tex3d: We could consider this an optional size/runtime efficiency optimization.

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+const uint64_t ShaderFeatureInfo_LinearAlgebra = 0x200000000;
+```
+
+In addition to the feature flag, detailed information about the usage of linalg
+matrix types and operations is gathered if the program uses linalg matrix types
+or operations outside of the required Tier 1 feature set.
+
+This information is required by the runtime to verify compatibility with a given
+device. Depending on whether the target is a shader or a library,  this
+information is encoded in either the Pipeline State Validation (PSV0) or the
```
llvm-beanz: Is "either" the right word here? If yes, in what situations is it encoded in PSV0 vs RDAT? My guess is that it goes to RDAT for library shaders and PSV0 for everything else, is that correct?

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+struct PSVLinAlgMatrixOperationShape0 {
+  uint32_t M; // Rows in matrix A
+  uint32_t N; // Columns in matrix B
+  uint32_t K; // Columns in matrix A / Rows in matrix B / Unused (0)
+};
+
+struct PSVLinAlgMatrixShapeArrayReference {
+  uint32_t ShapesIndex; // Index into SemanticIndexTable where array of indexes
+                        // into LinAlgMatrixOperationShape table is located
+  uint32_t Count;
```
llvm-beanz, line 1946: What does this `Count` signify?

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+struct PSVLinAlgRuntimeInfo0 {
+  // Presence of each table indicated by non-zero count.
+  // Tables are serialized in this order, with each starting with the record
+  // stride in bytes, followed by the records.
+  uint32_t MatrixOperationShapeCount;
+  uint32_t ThreadVectorMatrixMultiplyCount;
+  uint32_t WaveMatrixMultiplyCount;
+  uint32_t ThreadGroupMatrixMultiplyCount;
+  uint32_t OuterProductCount;
+  uint32_t AccumulateStoreCount;
```
llvm-beanz, line 1932: I wonder if we're better off encoding the use information as typed unions rather than a bunch of counts followed by separate structure definitions. I know that a typed union could be more data, but it is also trivially easy to generate and parse.

proposals/0035-linalg-matrix.md
tex3d: One thing these two optimizations help with is canonicalization of runtime data, which is important for validation. Serializing unneeded data is an opportunity for more variation in runtime data parts that wouldn't make a difference at runtime.

Currently, the validator does not independently check each element of the runtime data against the module to see whether it's valid. Instead, it expects the runtime data to exactly match what's generated by the validator (with some ordering independence). If we serialize unnecessary data, we have to either keep serializing that unnecessary data in the future or change the way we validate the runtime data to be much more sophisticated for both PSV0 and RDAT.

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+and at least the size of the record structure defined.
+
+The following are record definitions associated with each new table type. Each
+record definition corresponds to the record format used in the associated PSV
+structure, with some adjustments to fit RDAT patterns.
 
-The `OperandType` and `ResultType` fields will encode one of the values defined
-in the [`DXIL::ComponentType` enumeration](#dxil-enumerations).
+```cpp
+RDAT_ENUM_START(LinAlgThreadVectorMatrixMultiplyFlag, uint8_t)
```
llvm-beanz: I'm not really sure that a bunch of code using DXC's RDAT macros is something we should rely on for specification. This isn't meaningful outside the DXC codebase and really lacks a lot of context.

I'd prefer if we defined the new structures that we're encoding either C-structs (similar to how we've done with PSV0), or as a markdown table something more like:

**LinAlgMatrixOperationShape: Size 12 bytes**
| Offset | Type | Contents |
|-------|----|----|
| 0 | 32-bit unsigned integer | M dimension (rows in matrix A) |
| 4 | 32-bit unsigned integer | N dimension (columns in matrix B) |
| 8 | 32-bit unsigned integer | K dimension (columns in matrix A / rows in matrix B) |

If instead you prefer to define these as C structs, we also shouldn't just have a blob of C struct definitions, we should include explanations of the structure members and what each structure means.

proposals/0035-linalg-matrix.md
tex3d: This was supposed to indicate that:
> Depending on whether the target is a shader or a library,

But I can rewrite to clarify. How about:

> This information is stored in a separate DXIL container part from the DXIL IR. For library targets, it's in the DXIL Library Runtime Data (RDAT) part, otherwise, it's stored in the Pipeline State Validation (PSV0) part.

proposals/0035-linalg-matrix.md
tex3d: The size of the array. For each component type combination, this lists the set of unique dimensions used. The `ShapesIndex` is an index into the SemanticIndex buffer, which contains indices into the shared Shapes table, and Count is the number of shapes to index.

Perhaps the confusion is the name `ShapesIndex`, since this is an index into an index array buffer, not directly into the shape records.

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+* If, according to Tier 1 requirements, the shape merges with the minimum
+  required shape for a required operation/scope/type combination, it can be
+  considered as fitting within the minimum requirements, and need not be
+  recorded into extended usage data.
+
+Open questions:
+
+1. Do we need to gather more usage information for operations other than the
+   ones listed above when the matrix isn't captured by one of these operations?
+2. Do we need to capture component conversions with CopyConvertMatrix?
```
jenatali: Do we want to require all drivers to be able to load and store all possible SM6.10 component types? I'm waffling on whether that seems trivial or onerous.

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+
+In addition to the feature flag, detailed information about the usage of linalg
+matrix types and operations is gathered if the program uses linalg matrix types
+or operations outside of the required Tier 1 feature set.
+
+This information is required by the runtime to verify compatibility with a given
+device. Depending on whether the target is a shader or a library,  this
+information is encoded in either the Pipeline State Validation (PSV0) or the
+Dxil Library Runtime Data (RDAT) DXIL container part. The following sections
+describe the proposed encoding of this information for each of these container
```
pow2clk: I think saying "proposed" is implicit in the current status of the spec and might more consistently be described in definite terms, which might make it a little less confusing when it is read after being accepted
```suggestion
describe the encoding of this information for each of these container
```

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+* Iterate over specific LinAlg DXIL op calls.
+* If operation is outside minimum requirements, gather and merge detailed usage
+  information.
+
+Extended usage information is gathered for the following operation groups:
+
+* `ThreadVectorMatrixMultiply`
+  * Iterate `LinAlgMatVecMul`/`LinAlgMatVecMulAdd` calls and gather shapes and
+    types.
+  * Matrix operand must be `MatrixLoadFromDescriptor` for transposed flag.
```
pow2clk: I don't understand this. It sounds like the transposed flag takes a matrix operand and it has to be MatrixLoadFromDescriptor?

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+  * Iterate `LinAlgMatrixOuterProduct` calls and gather types.
+* `AccumulateStore`
+  * Iterate `LinAlgMatrixAccumulateToDescriptor` calls and gather type and
+    layout for transposed flag.
+
+Shape merging rules:
+
+* A shape can be merged with an existing shape if all but one dimension matches,
+  and the differing dimension is a factor of the existing shape's corresponding
+  dimension. For example, a 32x24x16 shape can be merged with an existing
```
pow2clk: Can we get a definition of the shape before its first usage? I see the structure below, but it's confusing when reading from start to finish

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+  uint32_t WaveMatrixMultiplyCount;
+  uint32_t ThreadGroupMatrixMultiplyCount;
+  uint32_t OuterProductCount;
+  uint32_t AccumulateStoreCount;
+};
+
+struct PSVLinAlgMatrixOperationShape0 {
+  uint32_t M; // Rows in matrix A
+  uint32_t N; // Columns in matrix B
+  uint32_t K; // Columns in matrix A / Rows in matrix B / Unused (0)
```
pow2clk: Unused only if we choose to provide shapes for mat/vec mul I guess?

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+
+* Iterate over specific LinAlg DXIL op calls.
+* If operation is outside minimum requirements, gather and merge detailed usage
+  information.
+
+Extended usage information is gathered for the following operation groups:
+
+* `ThreadVectorMatrixMultiply`
+  * Iterate `LinAlgMatVecMul`/`LinAlgMatVecMulAdd` calls and gather shapes and
+    types.
```
pow2clk: The below indicates no storage of shapes for mat/vec multiplies as yet. Is that right?

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+};
+
+enum class PSVLinAlgThreadVectorMatrixMultiplyFlag : uint8_t {
+  None = 0x00000000,
+  // Whether the matrix operand is loaded as transposed:
+  MatrixTransposed = 0x00000001,
+};
+
+struct PSVLinAlgThreadVectorMatrixMultiply0 {
+  // Do we need shapes? If so, K would be unused (0)
```
pow2clk: Since only K can be unused, I presume you'd be using M and N to represent a muliplication of an MxN matrix by an N-element vector? That's fine and I think preferred, but the rest of the spec refers to it as multiplication of an MxK matrix by a K-element vector. I've been hesitant to suggest changing it, but this might be sufficient motivation. Otherwise, we might keep consistent dimension terminology by making N the potentially unused dimension

pow2clk (commented):
A few potential clarifications and consistency requests

proposals/0035-linalg-matrix.md
anupamachandra: I had misunderstood this. The Matrix operand for ThreadScope is always MatrixLoadFromDescriptor or Phi right?

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+    types.
+  * Matrix operand must be `MatrixLoadFromDescriptor` for transposed flag.
+* `WaveMatrixMultiply`/`ThreadGroupMatrixMultiply`
+  * Iterate `LinAlgMatrixMultiply`/`LinAlgMatrixMultiplyAccumulate` calls and
+    gather shapes and types.
+* `OuterProduct`
+  * Iterate `LinAlgMatrixOuterProduct` calls and gather types.
+* `AccumulateStore`
+  * Iterate `LinAlgMatrixAccumulateToDescriptor` calls and gather type and
+    layout for transposed flag.
```
anupamachandra: Just wanted to note that when VectorAccumulate is added back to the spec, this needs to capture that as well : {InputType, AccType}

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+
+Shape merging rules:
+
+* A shape can be merged with an existing shape if all but one dimension matches,
```
anupamachandra: For thread scope, we don't have shape rules, just min and max. So does this apply only to Wave and TG scope?

proposals/0035-linalg-matrix.md
```diff
@@ -1499,42 +1499,367 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+  combination.
+* If, according to Tier 1 requirements, the shape merges with the minimum
+  required shape for a required operation/scope/type combination, it can be
+  considered as fitting within the minimum requirements, and need not be
+  recorded into extended usage data.
+
+Open questions:
+
+1. Do we need to gather more usage information for operations other than the
+   ones listed above when the matrix isn't captured by one of these operations?
```
anupamachandra: With Conversion of InputVectors separated out from MatVec, we'll probably have a runtime API for supported conversions and can capture the parameters of that operation here?

proposals/0035-linalg-matrix.md
tex3d: We need to trace the matrix operand back to a MatrixLoadFromDescriptor, otherwise we can't determine the matrix layout, which indicates whether this matrix is considered loaded as transposed for this operation (via MulOptimalTranspose layout). We might need to disallow any complex paths obfuscating this def-use chain, such as phi, at least for now.

proposals/0035-linalg-matrix.md
tex3d: I'm cutting shape merging for now. PR be updated soon.

proposals/0035-linalg-matrix.md
tex3d: That was the idea, but now, different components may be unused depending on the matrix type for MatrixConstruction. Update incoming soon.

proposals/0035-linalg-matrix.md
tex3d: As an intermediate step, I've provided macro definitions and comments to clarify the record definitions for now. One reason I don't yet want to switch to a completely different definition structure is that it will be much harder to keep it in sync with the real definitions being used to iterate and verify correctness.

proposals/0035-linalg-matrix.md
tex3d: I've updated the wording. Let me know if it's still confusing.

proposals/0035-linalg-matrix.md
tex3d: Thanks, I added it in anticipation.

proposals/0035-linalg-matrix.md
tex3d: I've deleted shape merging.

proposals/0035-linalg-matrix.md
tex3d: As this API is still TBD, I'll update this as necessary to supply that information. For now, there's an open question:

> Do we need to capture component conversions with CopyConvertMatrix?

Which I think should cover the need to sort this out for vector conversions as well. At least that's the intent.

proposals/0035-linalg-matrix.md
```diff
@@ -1810,42 +1810,542 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+* Current preference is to keep separate records and tables for each operation,
+  as it is simpler to implement and maintain, and allows for standard record
+  versioning through extension without issues with unions getting in the way.
 
-> Open questions:
-> 1) Do we need the M and N dimensions or just the K dimension?
-> 2) Do we need both operand types, or should we expect the operands to be the
->    same type?
-> 3) What flags do we need?
+* Another approach coulld be to follow the RDAT pattern where we have a list of
```
inbelic, line 2339: nit:
```suggestion
* Another approach could be to follow the RDAT pattern where we have a list of
```

proposals/0035-linalg-matrix.md
```diff
@@ -1810,42 +1810,542 @@ per-element bounds checking only converting the out of bounds stores to no-ops.
...
+Here, basic definitions are provided for serialization format with comments for
+semantic meanings.
+
+```cpp
+// Basic RDAT enum definitions stored as sTy, accessed as eTy.
+// eTy defined under hlsl::RDAT namespace (in DXC)
+
+#define RDAT_ENUM_START(eTy, sTy)           enum class eTy : sTy {
+#define RDAT_ENUM_VALUE(name, value)        name = value,
+#define RDAT_ENUM_VALUE_ALIAS(name, value)  name = value,
```
inbelic, line 2096:
```suggestion
```
seems unused

inbelic (commented):
Just some nits from a first pass through

damyanp (changes requested):
I think we should take the time now to actually write out the specification of the binary format. I don't think that the reader should need to run the C preprocessor in order to reverse engineer the binary format from the spec.  If we don't do this now, then we'll just need to repeat this exercise when we implement this in upstream llvm.

PR #833, opened 2026-03-30
LinAlg spec update - add SFINAE restrictions, argument names, clarify matrix use

hekota:
- Add SFINAE restrictions:
  - thread-scope `Matrix::Load` to require `A` matrix
  - thread-scope `Matrix::InterlockedAccumulate` to require `Accumulator` matrix
  - groupshared `Matrix::InterlockedAccumulate` to require `Wave` scope
  - `Multiply*` for matrix and vector to require arithmetic type
  - `Size` template parameter to groupshared `Matrix::Load` and `Matrix::InterlockedAccumulate`
- Add argument names to functions that were missing them. Some of the functions already had named arguments, so this is done for consistency. There were also cases where a function description was referring to an argument by name and the function declaration did not have it named.
- Fix array size condition on `Matrix::Store`
- Update Vector-Matrix operations to be Matrix `A` only

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -150,7 +150,7 @@ class Matrix {
                         uint Align = sizeof(ElementType));
 
   template <typename T, MatrixUseEnum UseLocal = Use,
-            MatrixScopeEnum ScopeLocal = Scope, SIZE_T Size>
+            MatrixScopeEnum ScopeLocal = Scope, SIZE_TYPE Size>
   typename hlsl::enable_if<
       hlsl::is_arithmetic<T>::value && Use == MatrixUse::Accumulator &&
           UseLocal == Use && (M * N / ElementsPerScalar <= Size) &&
@@ -943,7 +943,7 @@ Matrix::InterlockedAccumulate(RWByteAddressBuffer Res, uint StartOffset,
                               uint Align = sizeof(ElementType));
 
 template <typename T, MatrixUseEnum UseLocal = Use, 
-          MatrixScopeEnum ScopeLocal = Scope, SIZE_T Size>
+          MatrixScopeEnum ScopeLocal = Scope, SIZE_TYPE Size>
 typename hlsl::enable_if<
       hlsl::is_arithmetic<T>::value && Use == MatrixUse::Accumulator &&
           UseLocal == Use && (M * N / ElementsPerScalar <= Size) &&
@@ -1678,7 +1678,7 @@ in the [`DXIL::ComponentType` enumeration](#dxil-enumerations).
 
 ## Appendix 1: HLSL Header
 
-[Compiler Explorer](https://godbolt.org/z/dGrfY1T69)
+[Compiler Explorer](https://godbolt.org/z/Peh4jxrx3)
 > Note: this mostly works with Clang, but has some issues to work out still.
 
 ```cpp
@@ -1949,7 +1949,7 @@ class Matrix {
                         uint Align = sizeof(ElementType));
 
   template <typename T, MatrixUseEnum UseLocal = Use,
-            MatrixScopeEnum ScopeLocal = Scope, SIZE_T Size>
+            MatrixScopeEnum ScopeLocal = Scope, SIZE_TYPE Size>
   typename hlsl::enable_if<
       hlsl::is_arithmetic<T>::value && Use == MatrixUse::Accumulator &&
           UseLocal == Use && (M * N / ElementsPerScalar <= Size) &&


proposals/0035-linalg-matrix.md
```diff
@@ -723,10 +718,12 @@ static Matrix Matrix::Load(RWByteAddressBuffer Res, uint StartOffset,
...
-template <typename T>
-static typename hlsl::enable_if<hlsl::is_arithmetic<T>::value, Matrix>::type
-Matrix::Load(/*groupshared*/ T Arr[], uint StartIdx, uint Stride,
-             MatrixLayoutEnum Layout);
+template <typename T, SIZE_TYPE Size>
+static typename hlsl::enable_if<hlsl::is_arithmetic<T>::value &&
+                                    (M * N / ElementsPerScalar <= Size),
+                                Matrix>::type
+  Load(/*groupshared*/ T Arr[Size], uint StartIdx, uint Stride,
+       MatrixLayoutEnum Layout);
```
tex3d: I thought `Stride` should be number of elements per row in groupshared array (which could be col, depending on orientation), since the StartIdx should be the scalar array index into the groupshared array (not index into an array of matrices). If so, the size check is dependent on Layout, and should check `Stride * N` or Stride * M` instead of `M * N / ElementsPerScalar`. Of course, Stride is not a template parameter, so you can’t actually check this here.

proposals/0035-linalg-matrix.md
```diff
@@ -876,10 +867,11 @@ Matrix::InterlockedAccumulate(RWByteAddressBuffer Res, uint StartOffset,
                               uint Align = sizeof(ElementType));
 
 template <typename T, MatrixUseEnum UseLocal = Use>
-typename hlsl::enable_if<hlsl::is_arithmetic<T>::value &&
-                             Use == MatrixUse::Accumulator && UseLocal == Use,
+  typename hlsl::enable_if<
+      hlsl::is_arithmetic<T>::value && Use == MatrixUse::Accumulator &&
+          UseLocal == Use && (M * N / ElementsPerScalar <= Size),
```
tex3d: Same issue as above.

proposals/0035-linalg-matrix.md
```diff
@@ -1758,9 +1750,11 @@ class Matrix {
   static Matrix Load(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
                      MatrixLayoutEnum Layout, uint Align = sizeof(ElementType));
 
-  template <typename T>
-  static typename hlsl::enable_if<hlsl::is_arithmetic<T>::value, Matrix>::type
-  Load(/*groupshared*/ T Arr[], uint StartIdx, uint Stride,
+  template <typename T, SIZE_TYPE Size>
+  static typename hlsl::enable_if<hlsl::is_arithmetic<T>::value &&
+                                      (M * N / ElementsPerScalar <= Size),
```
tex3d: See other comment about Stride.

proposals/0035-linalg-matrix.md
```diff
@@ -1769,22 +1763,22 @@ class Matrix {
...
-  Set(uint, ElementType);
+  Set(uint Index, ElementType Value);
 
   void Store(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
              MatrixLayoutEnum Layout, uint Align = sizeof(ElementType));
 
   template <typename T, SIZE_TYPE Size>
   typename hlsl::enable_if<hlsl::is_arithmetic<T>::value &&
-                               (M * N / ElementsPerScalar >= Size),
+                               (M * N / ElementsPerScalar <= Size),
```
tex3d, line 1938: Same as above. But also thought: you can’t really do this validation here, since you can’t account for the StartIdx either. I’m not sure how valuable this check is in the first place. We can check at the builtin if things (StartIdx, Stride) are constant which would be superior to this weak partial check that probably would have a worse diagnostic.

proposals/0035-linalg-matrix.md
```diff
@@ -1797,29 +1791,30 @@ class Matrix {
                         MatrixLayoutEnum Layout,
                         uint Align = sizeof(ElementType));
 
-  template <typename T, MatrixUseEnum UseLocal = Use>
-  typename hlsl::enable_if<hlsl::is_arithmetic<T>::value &&
-                               Use == MatrixUse::Accumulator && UseLocal == Use,
+  template <typename T, MatrixUseEnum UseLocal = Use, SIZE_TYPE Size>
+  typename hlsl::enable_if<
+      hlsl::is_arithmetic<T>::value && Use == MatrixUse::Accumulator &&
+          UseLocal == Use && (M * N / ElementsPerScalar <= Size),
```
tex3d: Ditto.

proposals/0035-linalg-matrix.md
```diff
@@ -723,10 +718,12 @@ static Matrix Matrix::Load(RWByteAddressBuffer Res, uint StartOffset,
                            uint Align = sizeof(ElementType));
 
 // Not available on Thread scope matrices.
-template <typename T>
-static typename hlsl::enable_if<hlsl::is_arithmetic<T>::value, Matrix>::type
-Matrix::Load(/*groupshared*/ T Arr[], uint StartIdx, uint Stride,
-             MatrixLayoutEnum Layout);
+template <typename T, SIZE_TYPE Size>
+static typename hlsl::enable_if<hlsl::is_arithmetic<T>::value &&
```
tex3d, line 785: Oh yeah, isn’t the set of allowed component types a subset of arithmetic types? Could we capture that with something more specific?

proposals/0035-linalg-matrix.md
damyanp: Check that this has the correct enable_if'd on it.  The thread scope version should require an accumulator matrix.

proposals/0035-linalg-matrix.md
hekota: I filed an issue for this to follow-up post preview: #837

proposals/0035-linalg-matrix.md
hekota: #837

proposals/0035-linalg-matrix.md
hekota: #837

proposals/0035-linalg-matrix.md
hekota: #837

proposals/0035-linalg-matrix.md
hekota: Arithmetic types are a subset of allowed component types for matrices. For example `F8_E4M3` is not considered an arithmetic type.

PR #838, opened 2026-04-01
[0035] Clarify FP8 formats and emulation behavior

### LINKED ISSUE #827: [0035] Represent potentially emulated 8-bit floats with a separate enum
[0035 LinAlg Matrix](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

For platforms without native support for 8-bit float types, an enum that indicates 8-bit floats where supported and 16-bit floats (or whatever really) elsewhere is needed to give the user the ability to insist on them where necessary along with the careful conversion and math checks associated, but allow for single shader that runs everywhere as well. 

**Additional context**
An API capability bit will indicate when 8-bit types are natively supported to prevent submitting a shader requiring it to hardware that doesn't support it.

llvm-beanz:
This updates the specification to clarify the FP8 formats, their special values and the allowed emulation behaviors.

Closes #827 (this PR takes a different approach)

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -110,33 +110,37 @@ class Matrix {
 
   template <typename T, SIZE_TYPE Size>
   static typename hlsl::enable_if<hlsl::is_arithmetic<T>::value &&
-                                 (M * N / ElementsPerScalar <= Size),
-                                 Matrix>::type
+                                      (M * N / ElementsPerScalar <= Size),
+                                  Matrix>::type
   Load(/*groupshared*/ T Arr[Size], uint StartIdx, uint Stride,
        MatrixLayoutEnum Layout);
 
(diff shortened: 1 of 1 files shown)


proposals/0035-linalg-matrix.md
```diff
@@ -328,7 +328,7 @@ void CoopVec() {
   vector<float16_t, 16> NullBias = (vector<float16_t, 16>)0;
   vector<float16_t, 16> Layer2 = MultiplyAdd<float16_t>(MatA, Layer1, NullBias);
 
-  VectorRef<ComponentType::F8_E4M3, 16> MemBias = {MBuf,
+  VectorRef<ComponentType::F8_E4M3FN, 16> MemBias = {MBuf,
                                                    /*start offset*/ 4096};
```
tex3d: I don't know if you care, but this isn't going to match clang-format anymore.

PR #842, opened 2026-04-03
[0035] Fix dimensions of Cast with Transpose=true

llvm-beanz:
We need to swap the M and N dimensions on the return type of the Cast when Transpose is true.


jenatali (approved):

V-FEXrt:
Can you update the spec language for the copy convert op? Right now it says validation will ensure the dims are the same.

Should be updated to note they will swap on transpose

V-FEXrt (changes requested):
should have left the comment above here, sorry!

V-FEXrt (approved):

Issue #839, opened 2026-04-01
[0035] Alignment guarantee for Matrix Base, Offset, Stride and Size


anupamachandra:
**Which document does this relate to?**
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md

**Describe the issue you see with the spec**
Alignment guarantees for Matrix and Vector Accumulate buffers missing in the linAlg spec.

**Additional context**
In the earlier cooperative-vector [spec](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md) buffers had the following alignment guarantees:

*Matrix Buffer(for Mul/MulAdd/OuterProduct):*
The matrix base address and offset guaranteed to be aligned to at least **128 bytes**, and the matrix stride is guaranteed to be aligned to at least **16 bytes** and the matrix size was guaranteed to be a multiple of **16 bytes** ensuring that the **16 bytes* access of the last row/column of the matrix is valid memory.

*OutputArray (Vector Accumulate):*
The base address and output array offset must be **64 byte** aligned. Also note that the size of the underlying allocation is guaranteed to be a multiple of **16 bytes**, ensuring that there is valid memory between the end of the array and the **16 byte** boundary. Implementations may write to the contents of the padding between the end of the matrix and the **16 byte** boundary, so developers should not use this padding space for anything else.

This allows for efficient martrix loads/stores and cases where implementations might require padding the data due to h/w constraints. This guanratee is missing in the current linAlg spec. What we have in its place are align parameters for the relevant HLSL/DXIL ops, while this can be used by users to inform the driver/compiler of the alignment in conjuction with proper allocation of these resources.. just having the restrictions from the older spec without the flexibility seems adequate for most uses, provides portable performance and is sufficiently futureproof.

anupamachandra:
Clarifications under "Matrix Buffer":

*Shouldn't these apply to the [RW]ByteAddressBuffer operations: linAlgMatrixLoadFromDescriptor, linAlgMatrixStoreToDescriptor, and linAlgMatrixAccumulateToDescriptor operations? If not, can you specify the exact DXIL operations?*
[...]

PR #844, opened 2026-04-07
Add vector accumulate support

### LINKED ISSUE #751: [0035] Atomic vector accumulation missing from the spec.

The LinAlg spec is missing the functionality to accumulate Fp16 elements of a vector atomically-elementwise to corresponding elements in memory. 

**Additional context**
The earlier CoopVec spec had a `vectoraccumulate` function ( support set: Fp16), one important use case was accumulation of gradients in ML workloads. This piece is missing in the current spec. I am not sure if this functionality should be added to this spec, which is more matrix-centric, or as an independent feature.

pow2clk:
Add an atomic vector accumulate dxil operation with corresponding matrix method and language describing their behavior

Fixes #751

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -271,10 +271,14 @@ MultiplyAdd(Matrix<MatrixDT, M, K, MatrixUse::A, MatrixScope::Thread> MatrixA,
             InterpretedVector<InputElTy, VecK, InputInterp> InterpVec,
             VectorRef<BiasElTy, M> BiasRef);
 
-// Outer product functions
 template <ComponentEnum OutTy, typename InputElTy, SIZE_TYPE M, SIZE_TYPE N>
 Matrix<OutTy, M, N, MatrixUse::Accumulator, MatrixScope::Thread>
 OuterProduct(vector<InputElTy, M> VecA, vector<InputElTy, N> VecB);
+
+template <typename InputElTy, SIZE_TYPE M>
+typename hlsl::enable_if<hlsl::is_arithmetic<InputElTy>::value, void>::type
+InterlockedAccumulate(vector<InputElTy, M> Vec, RWByteAddressBuffer Res,
+                      uint StartOffset);
 
 } // namespace linalg
 } // namespace dx
@@ -479,6 +483,7 @@ The following table summarizes the operations supported for each matrix scope:
 | `linalg::Multiply(Matrix, vector)` | ✓ | ✗ | ✗ |
 | `linalg::MultiplyAdd(Matrix, vector, vector)` | ✓ | ✗ | ✗ |
 | `linalg::OuterProduct(vector, vector)` | ✓ | ✗ | ✗ |
+| `linalg::InterlockedAccumulate(RWByteAddressBuffer, vector)` | ✓ | ✗ | ✗ |
 
 Throughout this document a matrix may be described as having a scope as
 specified by the `Scope` parameter (e.g. a matrix with `Scope == Thread` is a
@@ -1044,7 +1049,20 @@ An accumulator matrix with wave or thread group scope has a method `Accumulate`
 which takes as a parameter an M x N `A` or `B` matrix. The method adds the
 provided matrix argument into the accumulator matrix.
 
-#### Matrix::AccumulatorLayout()
+#### linalg::InterlockedAccumulate
+
+```c++
+template <typename InputElTy, SIZE_TYPE M>
+typename hlsl::enable_if<hlsl::is_arithmetic<InputElTy>::value, void>::type
+InterlockedAccumulate(vector<InputElTy, M> Vec, RWByteAddressBuffer Res,
+                      uint StartOffset);
+```
+
+Atomically adds the vector data of `Vec` to the `RWByteAddressBuffer` target
+`Res`. When accumulating to the `RWByteAddressBuffer` object, the accumulation
+addition is performed on the component type of the vector.
+
+#### linalg::AccumulatorLayout()
 
 ```c++
 MatrixUse linalg::AccumulatorLayout();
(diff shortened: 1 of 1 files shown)


proposals/0035-linalg-matrix.md
```diff
@@ -271,11 +271,15 @@ MultiplyAdd(Matrix<MatrixDT, M, K, MatrixUse::A, MatrixScope::Thread> MatrixA,
             InterpretedVector<InputElTy, VecK, InputInterp> InterpVec,
             VectorRef<BiasElTy, M> BiasRef);
 
-// Outer product functions
```
pow2clk, line 274: Incidental, but clarifying. There's just one outer product function now, making this comment rather unnecessary and since I felt it appropriate to put the vector accumulate below, potentially misleading unless I added a similar comment for that single function.

proposals/0035-linalg-matrix.md
```diff
@@ -1044,7 +1049,20 @@ An accumulator matrix with wave or thread group scope has a method `Accumulate`
 which takes as a parameter an M x N `A` or `B` matrix. The method adds the
 provided matrix argument into the accumulator matrix.
 
-#### Matrix::AccumulatorLayout()
```
pow2clk, line 1047: Incidental, AccumulatorLayout is in the linalg namespace, not the Matrix class.

pow2clk (commented):
Just some comments to hopefully aid and expedite review.

proposals/0035-linalg-matrix.md
```diff
@@ -479,6 +483,7 @@ The following table summarizes the operations supported for each matrix scope:
 | `linalg::Multiply(Matrix, vector)` | ✓ | ✗ | ✗ |
 | `linalg::MultiplyAdd(Matrix, vector, vector)` | ✓ | ✗ | ✗ |
 | `linalg::OuterProduct(vector, vector)` | ✓ | ✗ | ✗ |
+| `linalg::InterlockedAccumulate(RWByteAddressBuffer, vector)` | ✓ | ✗ | ✗ |
```
jenatali, line 486: Does this really belong in the `linalg` namespace? It seems like this really should just be a loop over new overloads of `InterlockedAdd` for `float` and `half`.

proposals/0035-linalg-matrix.md
```diff
@@ -1681,6 +1699,20 @@ represent all values of the format used in the shader's DXIL.
 > an F8 type, but the source value can be represented accurately in the emulated
 > FP type, this may cause expected behavior differences.
 
+``` llvm
+declare void @dx.op.vectorAccumulateToDescriptor.v[NUM][TY](
```
jenatali, line 1703: It seems to me like this could be accomplished with new overloads of `dx.op.atomicBinOp`. If the vectorization is really important, then we can just enable that op to be vectorized.

proposals/0035-linalg-matrix.md
pow2clk: Are you asking if it belongs in the spec at all? In an earlier discussion we agreed to add it in. It was a part of the cooperative vector spec.

proposals/0035-linalg-matrix.md
pow2clk: I don't mind how it's designated particularly, that would require allowing vector overloads of `atomicBinOp` that would only be allowable for the addition variant, which is either warty or future proof depending on how you look at it or what you intend to do with it.

proposals/0035-linalg-matrix.md
jenatali: I agree we need support for fp32 and fp16 atomic addition that is impossible to do today but necessary for the linalg use cases. I can be convinced that it needs to be a vectorized DXIL op where the current `atomicBinOp` is not. I agree that it belongs in shader model 6.10. I'm not sure that the LinAlg spec and the `linalg` namespace are the right place for that.

tex3d: Considering some of the additional alignment requirements mentioned in #839, under _OutputArray (Vector Accumulate):_, I'm now tempted to lean towards keeping this as a more specialized op under LinAlg until we have a more solid plan for a more general operation.

proposals/0035-linalg-matrix.md
tex3d: As I mentioned [here](https://github.com/microsoft/hlsl-specs/pull/844/changes#r3053548379), I'm now leaning towards keeping this as a more specialized operation for now, considering the alignment requirements mentioned in #839. A general op that meets the same needs will take more time to negotiate and finalize, and we could still do that, but not in time for the preview. Perhaps we should keep an open issue in the spec about a desire for potential unification with a more general operation outside the linalg namespace/feature space.

jenatali: Fair enough! Agreed on keeping an open issue. Would be nice to unify for SM6.10 retail but not the end of the world either way.

jenatali (approved):

llvm-beanz (approved):

tex3d (approved):

proposals/0035-linalg-matrix.md
```diff
@@ -1681,6 +1699,20 @@ represent all values of the format used in the shader's DXIL.
...
+    immarg i32,       ; opcode
+    <[NUM] x [TY]>,   ; input vector
+    %dx.types.Handle, ; destination RWByteAddressBuffer
+    i32)              ; buffer offset
+```
+
+Accumulates a vector to a RWByteAddressBuffer at a specified offset. Each
```
V-FEXrt, line 1710: This can be in a follow up PR but we should add wording so clarify if `offset` is in bytes or "elements of the vector component type" ala https://github.com/microsoft/hlsl-specs/pull/862

I'm assuming bytes but would be good to explictly state it

V-FEXrt (approved):

PR #845, opened 2026-04-07
LinAlg packed type clarifications/protections

### LINKED ISSUE #823: [0035] Disallow native packed types in LinAlg header vector methods
[0035 LinAlg Matrix](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

Offline discussion made clear that using native packed types in vectors involved in multiplication and outerproduct operations was meant to be disallowed by the header's SFINAE enforcement. As currently written and updated by #817, it does not. See line 430 of [this](https://godbolt.org/z/PTc4TeeE5) slight modification of the example code in the spec.

**Additional context**
If these packed types are to be allowed, the dimension checking will have to become a lot more sophisticated to account for native (non-interpreted) vectors that contain a different number of scalars and elements.

pow2clk:
Packed types are only intended to be used as inputs to makeinterpreted vector after which they will be used only through that wrapper. Using them directly in the APIs introduces complicated dimension checking and isn't intended to be supported.

Mostly sprinkling heavy usage of is_arithmetic checks for fall native vector inputs. Some of these were added already for MultiplyAdd, but left out elsewhere.

Removes the restriction on using packed types for groupshared load/store/accumulate operations.

Makes a few incidental typo corrections here and there.

Removes mention of matrices of packed types as matrices cannot have packed types, though they may have types that can only be represented as packed when converted and wrapped in interpreted vectors

Fixes #823

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -111,8 +111,7 @@ class Matrix {
                      MatrixLayoutEnum Layout, uint Align = sizeof(ElementType));
 
   template <typename T, SIZE_TYPE Size>
-  static typename hlsl::enable_if<hlsl::is_arithmetic<T>::value &&
-                                      (M * N / ElementsPerScalar <= Size),
+  static typename hlsl::enable_if<M * N / ElementsPerScalar <= Size,
                                   Matrix>::type
   Load(/*groupshared*/ T Arr[Size], uint StartIdx, uint Stride,
        MatrixLayoutEnum Layout);
@@ -141,8 +140,7 @@ class Matrix {
              MatrixLayoutEnum Layout, uint Align = sizeof(ElementType));
 
   template <typename T, SIZE_TYPE Size>
-  typename hlsl::enable_if<hlsl::is_arithmetic<T>::value &&
-                               (M * N / ElementsPerScalar <= Size),
+  typename hlsl::enable_if<M * N / ElementsPerScalar <= Size,
                            void>::type
   Store(/*groupshared*/ T Arr[Size], uint StartIdx, uint Stride,
         MatrixLayoutEnum Layout);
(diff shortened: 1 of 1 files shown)


proposals/0035-linalg-matrix.md
```diff
@@ -526,9 +531,6 @@ DXIL validation.
 | Wave         | [4,128]                   |
 | ThreadGroup  | [1,1024]                  |
 
-Sizes for matrices of packed data types are 4 times the valid size for a scalar
-element.
```
pow2clk, line 530: Since this is talking about matrices and they don't support "packed" types anymore as such and only refer to them by the number of elements, I think this is more confusing than clarifying.

proposals/0035-linalg-matrix.md
```diff
@@ -1116,15 +1118,17 @@ parameter for the output matrix element type.
...
+                         vector<OutputElTy, M> >::type
 linalg::MultiplyAdd(Matrix<MatrixDT, M, K, MatrixUse::A, MatrixScope::Thread> MatrixA,
                     vector<InputElTy, K> Vec, vector<BiasElTy, M> Bias);
 ```
 
 Requires `Thread` scope matrix input, may be called from divergent control flow.
 
 The `linalg::MultiplyAdd` function has an overload that takes an MxK `A` matrix
-with `Thread` scope, a `K`-element vector `Vec`, and a `M`-element vector
+with `Thread` scope, a `K`-element vector `Vec`, and an `M`-element vector
```
pow2clk: Minor fixup. Elsewhere "an" is used to refer to `M`-element vectors.

proposals/0035-linalg-matrix.md
```diff
@@ -1544,9 +1548,9 @@ declare void @dx.op.linAlgMatrixAccumulateToDescriptor.[MatTy](
 Accumulates a matrix to a RWByteAddressBuffer at a specified offset. This
 operation is only available for matrices with `MatrixUse::Accumulator`. The
 matrix data is added to the existing data in the buffer. The matrix component
-data is converted to the target arithmetic or packed data type if the data types
-do not match, then added to the existing data in memory. This operation must
-observe [bounds checking behavior](#bounds-checking-behavior) described below.
+data is added to the existing data in memory using the component type of the
+matrix. This operation must observe
+[bounds checking behavior](#bounds-checking-behavior) described below.
```
pow2clk, line 1552: Language about type conversion is removed since byteaddressbuffers have no type, so the only type comes from the matrix and no conversion is necessary.

pow2clk (commented):
Nearly every change is just adding `is_arithmetic` checks to prevent usage of native packed types where they can't be handled.

Added comments for some of the other changes to try to pre-empt questions and expedite review.

There is another change that I couldn't disentangle from this one that has to wait until this goes in.

jenatali (approved):

tex3d (approved):
These changes all look good to me.

V-FEXrt (approved):
Let's do this in a follow up to not generate more churn but the godbolt link needs to be updated one this goes in

PR #849, opened 2026-04-09
[0035] LinAlg: Remove Stride and Layout params from Thread scoped InterlockedAccumulate

### LINKED ISSUE #850: [0035] LinAlg: extra parameters in Thread scope InterlockedAccumulate def in header appendix
https://github.com/microsoft/hlsl-specs/blob/40ede0ed4dbad3ec9875bdbde721daed282f0eec/proposals/0035-linalg-matrix.md?plain=1#L202-L204

An earlier definition of the thread-scope InterlockedAccumulate omits Stride and Layout parameters because it's required to accumulate in OuterProductOptimal layout (where Stride is also device-dependent). However, in the header version of the same method definition, these parameters are present. This looks like an oversight.

tex3d:
The thread-scope overload wasn't supposed to have Stride and Layout, since the Layout must be OuterProductOptimal for thread scope. This change removes these two parameters from the version of the function in Appendix 1: HLSL Header.

Fixes #850

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -2069,8 +2069,7 @@ class Matrix<ComponentTy, M, N, Use, MatrixScope::Thread> {
   template <MatrixUseEnum UseLocal = Use>
   typename hlsl::enable_if<Use == MatrixUse::Accumulator && UseLocal == Use,
                            void>::type
-  InterlockedAccumulate(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
-                        MatrixLayoutEnum Layout,
+  InterlockedAccumulate(RWByteAddressBuffer Res, uint StartOffset,
                         uint Align = sizeof(ElementType));
 };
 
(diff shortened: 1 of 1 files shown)


proposals/0035-linalg-matrix.md
jenatali: There's a second copy of the header at the top of this file. These probably should be de-duplicated...

jenatali (changes requested):

proposals/0035-linalg-matrix.md
tex3d: Thanks! Man, all these copies sure make it frustrating. I also found an example call to update. Let me know if it looks good now.

tex3d: Also, it turns out that the DXC implementation also includes these extra parameters. I filed an issue there: https://github.com/microsoft/DirectXShaderCompiler/issues/8360.

jenatali (approved):

alsepkow (approved):

PR #854, opened 2026-04-15
[0035] Always set align to 128

V-FEXrt:
The minimum common alignment is 128 and internal discussion landed on that also likely being the largest useful value.

At least for preview, always set that value in the header. We can reevaluate in the future but we are leaning on not changing the DXIL op so the value can be easily updated.

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -105,10 +105,10 @@ class Matrix {
   Splat(T Val);
 
   static Matrix Load(ByteAddressBuffer Res, uint StartOffset, uint Stride,
-                     MatrixLayoutEnum Layout, uint Align = sizeof(ElementType));
+                     MatrixLayoutEnum Layout, uint Align = 128);
 
   static Matrix Load(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
-                     MatrixLayoutEnum Layout, uint Align = sizeof(ElementType));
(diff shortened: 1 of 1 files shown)


tex3d (approved):
Looks good, though it's hard from these diffs to tell whether every instance was modified, oh well! Note that some of these changes could change formatting when clang-format is applied, but for now it's more readable, so I prefer the minimal diff.

Note related issue #839. This PR doesn't fix the spec according to this issue, but makes the default Align values in the HLSL API more in line with it. It will be important that buffer base and offsets/strides passed in by HLSL API users actually are aligned by the promised amount though!

V-FEXrt:
> This PR doesn't fix the spec according to this issue, but makes the default Align values in the HLSL API more in line with it. It will be important that buffer base and offsets/strides passed in by HLSL API users actually are aligned by the promised amount though!

Right now I'm going for correct-enough-for-preview :) Do you believe this change would be enough to reach that goal/punt the larger discussion in #839 or do you think we need to incorporate the details over there for preview?

The goal for this PR isn't so much as to "fix" the specs alignment requirement as it is the fix the header bug where we are just outright setting the wrong alignment values

tex3d:
> do you think we need to incorporate the details over there for preview?

I just wanted to link the related issue but also note that this PR didn't resolve the issue. I didn't expect anything different from this PR. It looks good to me!

tex3d (approved):

jenatali (approved):

PR #855, opened 2026-04-15
[0035] Add description of `__detail` structs and update `DstN` calculations

hekota:
Updates `DstN` calculation to make sure the calculated value is rounded. 
Added description of the `DimMN` and `DstN` helper structs from the `__detail` namespace.


tex3d (approved):
LGTM!

V-FEXrt (approved):

PR #871, opened 2026-05-14
Update DebugBreak proposal langauge

llvm-beanz:
This is mostly clarifications that the behavior isn't actually dependent on a debugger being attached, but rather on runtime state (which will be set by a debugger, but may also be set separately).

I've also renamed `dx::IsDebuggerAttached` to `dx::IsDebuggingEnabled` to more clearly reflect the change.

### DIFF
--- a/proposals/0039-debugbreak.md
+++ b/proposals/0039-debugbreak.md
@@ -9,20 +9,20 @@ params:
     status: Under Review
 ---
 
- 
+
 * Issue(s): https://github.com/microsoft/hlsl-specs/issues/33
 
 ## Introduction
 
 This proposal specifies two new HLSL debugging intrinsics:
 
-1. **`DebugBreak()`**: Triggers a breakpoint when a debugger is attached,
+1. **`DebugBreak()`**: Triggers a breakpoint when debugging is enabled
    allowing developers to pause execution and inspect shader state.
-2. **`dx::IsDebuggerPresent()`**: Returns whether a graphics debugger is currently
-   attached to the process, enabling conditional debug-only code paths.
+2. **`dx::IsDebuggingEnabled()`**: Returns whether debugging is enabled for the
+   currently executing thread, enabling conditional debug-only code paths.
 
 `DebugBreak()` will lower to new DXIL operations for DirectX and to appropriate
-SPIR-V instructions for Vulkan targets. `dx::IsDebuggerPresent()` is a DirectX
+SPIR-V instructions for Vulkan targets. `dx::IsDebuggingEnabled()` is a DirectX
 extension and has no Vulkan/SPIR-V equivalent.
 
 ## Motivation
@@ -36,8 +36,8 @@ Conditional breakpoints can be a powerful tool for shader authors to narrow down
 and identify these complex rare-occurring problems.
 
 Additionally, shader authors need a way to conditionally enable expensive debug
-checks only when a debugger is attached, avoiding runtime overhead in production
-scenarios.
+checks only when debugging is enabled in the runtime, avoiding runtime overhead
+in production scenarios.
 
 This proposal introduces two intrinsics that together provide a debugging
 toolkit for shader development.
@@ -49,8 +49,8 @@ This proposal introduces two new HLSL intrinsics for debugging shader code.
 ### Intrinsics
 
 ```hlsl
-void DebugBreak();        // Trigger a breakpoint if debugger attached
-bool dx::IsDebuggerPresent(); // Query if a debugger is attached (DirectX only)
+void DebugBreak();        // Trigger a breakpoint if debugging is enabled.
+bool dx::IsDebuggingEnabled(); // Query if debugging is enabled (DirectX only).
 ```
 
 ### Example Usage
@@ -59,11 +59,11 @@ bool dx::IsDebuggerPresent(); // Query if a debugger is attached (DirectX only)
 [numthreads(8,1,1)]
 void main(uint GI : SV_GroupIndex) {
     // Conditional expensive debug checks
-  if (dx::IsDebuggerPresent()) {
+  if (dx::IsDebuggingEnabled()) {
         // Expensive validation only when debugging
         ValidateComplexInvariants();
     }
-    
+
     // Manual breakpoint for debugging specific conditions
     if (someRareCondition) {
         DebugBreak();
@@ -85,22 +85,22 @@ Two new intrinsic functions are added:
 void DebugBreak();
 ```
 
-Triggers a breakpoint if a graphics debugger is attached. If no debugger is
-attached or the runtime does not support this operation, it is treated as a
+Triggers a breakpoint if debugging is enabled in the runtime. If debugging is
+not enabled or the runtime does not support this operation, it is treated as a
 no-op. Execution continues after the breakpoint.
 
-#### `dx::IsDebuggerPresent()`
+#### `dx::IsDebuggingEnabled()`
 
 ```hlsl
-bool dx::IsDebuggerPresent();
+bool dx::IsDebuggingEnabled();
 ```
 
-Returns `true` if a graphics debugger is currently attached to the process,
-`false` otherwise. This allows shader authors to conditionally execute expensive
-debug validation code only when a debugger is present:
+Returns `true` if debugging is enabled on the current thread, `false` otherwise.
+This allows shader authors to conditionally execute expensive debug validation
+code only when debugging is enabled:
 
 ```hlsl
-if (dx::IsDebuggerPresent()) {
+if (dx::IsDebuggingEnabled()) {
     // Expensive bounds checking, validation, etc.
     for (uint i = 0; i < arraySize; ++i) {
         if (data[i] < 0.0f || data[i] > 1.0f) {
@@ -126,19 +126,19 @@ declare void @dx.op.debugBreak(
 ```
 
 Triggers a debugger breakpoint. Must be treated as `convergent` to prevent code
-motion. Should not be marked `readonly` or `readnone`. If no debugger is
-attached, this is a no-op.
+motion. Should not be marked `readonly` or `readnone`. If debugging is not
+enabled, this is a no-op.
 
-#### `dx.op.isDebuggerPresent`
+#### `dx.op.IsDebuggingEnabled`
 
 ```llvm
-declare i1 @dx.op.isDebuggerPresent(
+declare i1 @dx.op.IsDebuggingEnabled(
   immarg i32             ; opcode
 ) readonly
 ```
 
-Returns `true` (1) if a debugger is attached, `false` (0) otherwise. Marked
-`readonly` as it only queries state. 
+Returns `true` (1) if debugging is enabled, `false` (0) otherwise. Marked
+`readonly` as it only queries state.
 
 ### Convergence Requirements
 
@@ -162,10 +162,10 @@ per-pipeline basis.
 
 Behavioral changes may include:
 
-- Breaking regardless of a debugger being attached
+- Breaking regardless of a debugging enabled
 - Disabling debug break instructions entirely
 
-It is expected that the driver compiler will alter behavior during lowering 
+It is expected that the driver compiler will alter behavior during lowering
 based on information provided by the runtime at pipeline creation.
 
 ### SPIR-V Lowering
@@ -182,7 +182,7 @@ Uses the existing `NonSemantic.DebugBreak` instruction:
 While this instruction is not widely supported by Vulkan debuggers, it is
 supported by NVIDIA's NSight and can be safely ignored by Vulkan runtimes.
 
-No SPIR-V lowering is defined for `dx::IsDebuggerPresent()`.
+No SPIR-V lowering is defined for `dx::IsDebuggingEnabled()`.
 
 ## Testing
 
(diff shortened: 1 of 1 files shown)


damyanp (approved):
@joecitizen can you file an issue to make the corresponding changes to the implementation in DXC please?

proposals/0039-debugbreak.md
```diff
@@ -36,8 +36,8 @@ Conditional breakpoints can be a powerful tool for shader authors to narrow down
 and identify these complex rare-occurring problems.
 
 Additionally, shader authors need a way to conditionally enable expensive debug
-checks only when a debugger is attached, avoiding runtime overhead in production
-scenarios.
+checks only when debugging is enabled in the runtime, avoiding runtime overhead
```
t-tye, line 39: Suggest s/in the runtime/at runtime/ . Do not require that the checking has to be done in a runtime, as it may be the device driver that does the checking. Both these cases are happening at runtime.

proposals/0039-debugbreak.md
```diff
@@ -85,22 +85,22 @@ Two new intrinsic functions are added:
 void DebugBreak();
 ```
 
-Triggers a breakpoint if a graphics debugger is attached. If no debugger is
-attached or the runtime does not support this operation, it is treated as a
+Triggers a breakpoint if debugging is enabled in the runtime. If debugging is
+not enabled or the runtime does not support this operation, it is treated as a
```
t-tye, line 89: Suggest s/or the runtime does not support this operation/or the device does not support this operation/ . Do not require that the checking has to be done in a runtime, as it may be the device driver that does the checking. Both these cases are providing support for the specific device executing the code.

t-tye (commented):
I think the name change from debugger to debugging makes a lot of sense. The facilities are about controlling information intended for debugging. The control of when that information is produced could be due to a debugger being attached, or it could be controlled by the driver or runtime inspecting options the user has supplied when executing the application.

proposals/0039-debugbreak.md
```diff
@@ -59,11 +59,11 @@ bool dx::IsDebuggerPresent(); // Query if a debugger is attached (DirectX only)
 [numthreads(8,1,1)]
 void main(uint GI : SV_GroupIndex) {
     // Conditional expensive debug checks
-  if (dx::IsDebuggerPresent()) {
+  if (dx::IsDebuggingEnabled()) {
```
t-tye, line 62: The if statement has 2 spaces of indentation but should have 4 spaces to match the rest of the function body.

proposals/0039-debugbreak.md
```diff
@@ -126,19 +126,19 @@ declare void @dx.op.debugBreak(
 ```
 
 Triggers a debugger breakpoint. Must be treated as `convergent` to prevent code
-motion. Should not be marked `readonly` or `readnone`. If no debugger is
-attached, this is a no-op.
+motion. Should not be marked `readonly` or `readnone`. If debugging is not
+enabled, this is a no-op.
```
t-tye: I suggest `IsDebuggingEnabled`/`DebugBreak` should not be marked convergent. There is no communication between the dynamic execution instances of a convergent version of `IsDebuggingEnabled`/`DebugBreak` — as long as each lane executes a `IsDebuggingEnabled`/`DebugBreak`, it does not matter if they are reported by the same instruction. As long as each lane that is active correctly executes the code to report a DebugBreak that is all that is required. Marking it convergent would limit code motion optimization unnecessarily. For example, the compiler may decide to clone basic blocks when optimizing complex IF conditions, resulting in the DebugBreak code being replicated. Each lane may execute a different path through the optimized code, but will still report the DebugBreak, and the active lane mask indicates which lane(s) are actually executing it. So instead of all lanes executing a single instruction to report the DebugBreak, some lanes may execute one instruction, and other lanes may report at a different cloned version of the instruction. Collectively, all the lanes are reporting. My view is that diagnostic code should not inhibit optimization as far as possible, provided it does execute under the correct conditions.

However, what I think is needed instead is to treat `IsDebuggingEnabled`/`DebugBreak` as "volatile read+write" in the sense that languages like C define it. It indicates communication with something outside the abstract machine model of the language, which is exactly what the debugger is here. A debugger can attach at any time, so each execution of the `IsDebuggingEnabled`/`DebugBreak` needs to happen using the current machine state (which could have been modified by the debugger).

This prevents code motion: each instance of `IsDebuggingEnabled`/`DebugBreak` in the source program needs to execute, so it cannot be moved out of loops (which would change the number of times it executes). Similarly, CSE is prevented. Generally, volatile and read + write semantics also inhibit other code motion so that `IsDebuggingEnabled`/`DebugBreak` will tend to remain where they are relative to other instructions that have observable effects, such as other load and stores. This is similar to how the time() builtin needs to be handled as the value it returns is also volatile and there is a desire to keep it in the same place that the source code implies.

Suggest replacing "Must be treated as convergent" with a description of volatile semantics. Possible change to this paragraph for `IsDebuggingEnabled`/`DebugBreak`:

> **Must** be treated as a `volatile` memory read and write to prevent code motion and to ensure the current runtime debugging state value is always used. The runtime debugging state can change through the execution of the application due to attaching a debugger and changing the setting. Should not be marked `readonly` or `readnone` since it is reading the current state of the debugging state.

I would delete the "If debugging is not enabled, this is a no-op." since the purpose of it is to determine is debugging is enabled, and it is a function that must always return a boolean value. Instead, perhaps:

> If compilation requests that debugging code is eliminated, then **must** be treated as unconditionally returning `FALSE`. The compiler is then permitted to remove all code guarded by `IsDebuggingEnabled` as dead code.

proposals/0039-debugbreak.md
```diff
@@ -126,19 +126,19 @@ declare void @dx.op.debugBreak(
...
-declare i1 @dx.op.isDebuggerPresent(
+declare i1 @dx.op.IsDebuggingEnabled(
   immarg i32             ; opcode
 ) readonly
 ```
 
-Returns `true` (1) if a debugger is attached, `false` (0) otherwise. Marked
-`readonly` as it only queries state. 
+Returns `true` (1) if debugging is enabled, `false` (0) otherwise. Marked
+`readonly` as it only queries state.
```
t-tye, line 139: As mentioned above, it is proposed that `IsDebuggingEnabled` should also have `volatile` read and write semantics rather than being marked `readonly`. This would ensure the current debugging status is queried each time the operation is executed. This is important to support the ability to attach a debugger to an already-executing shader — a common approach for debugging shaders that do not terminate due to synchronization errors. In these cases, TDR detection is often disabled and a debugger is attached to the running hung shader. Making the semantics include write to unknown memory address will help prevent the compiler optimization from reordering the calls with other source statements so it will stay where the user expected from a source language perspective.

The operation should also be marked as expected to return `FALSE` (e.g., the equivalent of `__builtin_expect(expr, 0)` in Clang). This will give the compiler the information to allow it to move any conditionally-executed debugging code out of the inline path as "cold" code to reduce instruction cache pressure and improve branch prediction. The compiler may choose to limit optimizations in the cold code to minimize the impact on the hot code. For example, limiting scheduling so that register pressure in the cold path does not exceed that in the hot path which could cause occupancy and other performance issues.

The desire is that debugging code can be left in production code and have negligible performance impact.

proposals/0039-debugbreak.md
t-tye: See the discussion above about `convergent` and `volatile`.

proposals/0039-debugbreak.md
```diff
@@ -162,10 +162,10 @@ per-pipeline basis.
 
 Behavioral changes may include:
 
-- Breaking regardless of a debugger being attached
+- Breaking regardless of a debugging enabled
```
t-tye: Grammar: suggest "Breaking regardless of debugging being enabled" or "Breaking unconditionally".

proposals/0039-debugbreak.md
```diff
@@ -162,10 +162,10 @@ per-pipeline basis.
...
 - Disabling debug break instructions entirely
 
-It is expected that the driver compiler will alter behavior during lowering 
+It is expected that the driver compiler will alter behavior during lowering
```
t-tye, line 158: I am not sure this section belongs in the HLSL compiler specification. Runtime finalization is not the only compilation model — shaders may be compiled offline, in which case it is not known at compilation time whether debugging will be enabled when the generated code is executed. The runtime could provided ways for the user to control when debugging is enabled independently of the compilation process. An advantage is that the developer does not have to wait for a re-finalization each time they change the setting: a single code object can execute in either mode.

If the intent is to allow the compiler to eliminate debugging code, that is reasonable but suggest it is specified differently: compile out all `DebugBreak` operations and treat `IsDebuggingEnabled` as the constant ``FALSE``, then perform dead code elimination. This is distinct from the compiler needing to know about runtime behavior during lowering.

proposals/0039-debugbreak.md
```diff
@@ -182,7 +182,7 @@ Uses the existing `NonSemantic.DebugBreak` instruction:
 While this instruction is not widely supported by Vulkan debuggers, it is
 supported by NVIDIA's NSight and can be safely ignored by Vulkan runtimes.
 
-No SPIR-V lowering is defined for `dx::IsDebuggerPresent()`.
+No SPIR-V lowering is defined for `dx::IsDebuggingEnabled()`.
```
t-tye, line 175: Is there a plan to pursue adding an equivalent to `IsDebuggingEnabled` for SPIR-V? Could SPIR-V's Specialization Constants be used for this?

proposals/0039-debugbreak.md
```diff
@@ -199,13 +199,13 @@ No SPIR-V lowering is defined for `dx::IsDebuggerPresent()`.
...
-- Test `dx::IsDebuggerPresent()` returns correct value based on debugger state
+- Test `DebugBreak()` triggers breakpoint when debugging is enabled
+- Test `DebugBreak()` is no-op when debugging is not enabled
+- Test `dx::IsDebuggingEnabled()` returns correct value based on runtime state
 
 ## Open Questions
 
 * Consider introducing the `convergent` attribute to DXIL.
   * This should be "cheap" and would potentially address pre-existing bugs.
   * This would preserve the requirement that these operations not be moved
```
t-tye: Adding `convergent` to DXIL may be a good thing for other primitives that cause communication between dynamic instances of an operation across the lanes of a wave, but `IsDebuggingEnabled`/`DebugBreak` are not such operations.

proposals/0039-debugbreak.md
llvm-beanz: > As mentioned above, it is proposed that IsDebuggingEnabled should also have volatile read and write semantics rather than being marked readonly.

`volatile` does not apply to functions in LLVM IR, it can only apply to load and store instructions. A call to a function function marked `readonly` cannot be legally eliminated by a compiler optimization, so it will provide the expected behaviors.

proposals/0039-debugbreak.md
llvm-beanz: This comment is nitpicking beyond the scope of this spec. To a user enabling debug break will be something they set on the runtime, and it is up to the "runtime implementation" (which includes the driver implementation) to determine how that is implemented.

proposals/0039-debugbreak.md
llvm-beanz: `volatile` is not a relevant concept to DXIL (it only applies to load and store instructions in LLVM). The behavior you're describing is achieved with the `readonly` attribute.

proposals/0039-debugbreak.md
llvm-beanz: This language doesn't bind the implementation to any specific behavior but is intended to allow a backend compiler to optimize how it sees fit. If you choose not to do it, fine.

proposals/0039-debugbreak.md
llvm-beanz: Eventually for SM7, but I suspect your question is more about Vulkan support. We're not currently planning to pursue that.

alsepkow (approved):

damyanp:
@llvm-beanz is this ready to merge?

proposals/0039-debugbreak.md
t-tye: @llvm-beanz, I did some poking around the langref and the problem of ``readonly`` is that it will allow the compiler to do optimizations that are inconsistent with the function accessing a volatile variable. As you point out, there is no volatile attribute available on functions, so the closest you seem to be able to get is to declare as ``memory(inaccessiblemem: readwrite)``, and then ``nomerge`` prevents other optimizations.

Every call to the function can use LLVM's ``expect`` intrinsic to improve code layout so the debugging code is seen to be cold.

## Function Declaration

```llvm
declare i1 @IsDebuggingEnabled() nomerge memory(inaccessiblemem: readwrite)
```

## Call Site Usage

```llvm
%raw = call i1 @IsDebuggingEnabled()
%val = call i1 @llvm.expect.i1(i1 %raw, i1 false)
br i1 %val, label %rare_true, label %common_false
```

## Summary of what each piece does

| Annotation | Purpose |
|---|---|
| `memory(inaccessiblemem: readwrite)` | Prevents CSE, dead elimination, and reordering between calls |
| `nomerge` | Prevents duplication and merging across control flow |
| `llvm.expect(... , i1 false)` | Hints that `false` is the common result — improves branch layout without removing the true path |

PR #875, opened 2026-05-19
[0035] Fix various VectorAccumlator details

### LINKED ISSUE #868: [0035] dx.op.vectorAccumulateToDescriptor needs an alignment parameter
https://github.com/microsoft/hlsl-specs/blob/c2f44e38a31b92d7f110f8bbbb80dba36bb5af3b/proposals/0035-linalg-matrix.md?plain=1#L1779-L1783

For consistency with `dx.op.linAlgMatrixAccumulateToDescriptor`, I think `dx.op.vectorAccumulateToDescriptor` needs an alignment parameter. At least I think we should include it until we possibly rule out the need for one in the final resolution of #839. If we are to include it in the preview, it will need to be updated ASAP in the [implementation PR](https://github.com/microsoft/DirectXShaderCompiler/pull/8448) or shortly after that PR merges.

For the high-level API, we should add a default alignment value, like we did for Matrix. It should probably be `64` bytes to match the proposed alignment requirement in #839.

First location:
https://github.com/microsoft/hlsl-specs/blob/c2f44e38a31b92d7f110f8bbbb80dba36bb5af3b/proposals/0035-linalg-matrix.md?plain=1#L287-L290

Second location:
https://github.com/microsoft/hlsl-specs/blob/c2f44e38a31b92d7f110f8bbbb80dba36bb5af3b/proposals/0035-linalg-matrix.md?plain=1#L2267-L2270

**Additional context**
See comment thread here: https://github.com/microsoft/DirectXShaderCompiler/pull/8448#discussion_r3230590681

V-FEXrt:
Fixes the following issues with the VectorAccumulate spec that was discovered during implementation.

Specifically
- ~~Parameter order in the header was inconsistent~~ This was reverted based on internal discussion
- DXIL op name was inconsistent
- DXIL op needed an `align` parameter

Fixes https://github.com/microsoft/hlsl-specs/issues/868

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -286,8 +286,8 @@ OuterProduct(vector<InputElTy, M> VecA, vector<InputElTy, N> VecB);
 
 template <typename InputElTy, SIZE_TYPE M>
 typename hlsl::enable_if<hlsl::is_arithmetic<InputElTy>::value, void>::type
-InterlockedAccumulate(vector<InputElTy, M> Vec, RWByteAddressBuffer Res,
-                      uint StartOffset);
+InterlockedAccumulate(RWByteAddressBuffer Res, vector<InputElTy, M> Vec,
+                      uint StartOffset, uint Align = 64);
 
 } // namespace linalg
 } // namespace dx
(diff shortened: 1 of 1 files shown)


V-FEXrt:
@llvm-beanz I'm trying to also add you to the reviewer list but GitHub decided that feature doesn't work today

tex3d (approved):
Looks good to me.

I was going to leave a nit for the indent-only changes on unrelated ops, but they aren't a big deal and seem to make things more consistent with some of the other ops at least.

llvm-beanz (approved):

xiaolin-ji:
When input vector is half and buffer align to 2, it might be a problem to hw if it doesn't support half buffer accumulate but b32 accumulate, can we specify the base and start offset must align to 64-byte or 4-byte? like https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md#arguments-2

V-FEXrt:
@xiaolin-ji yep that was intended just didn't make it into the spec. I filed an issue here https://github.com/microsoft/hlsl-specs/issues/876

edit: oh lol this _is_ a spec PR. Sorry I have too many PRs bouncing around right now. I suppose I can just make the change here

edit 2: Actually, I think I'm going to keep the PR as is. Right now this PR forces the alignment parameter to be 64 anyways and we still had some on going discussions about alignment details. With the issue filed we'll made sure to address this before retail but I don't want to rush to get language into the spec for the next preview build we are making

tex3d (approved):

jenatali (approved):

PR #879, opened 2026-06-05
[0035] Update wave-scope matrix groupshared operations

### LINKED ISSUE #866: [0035] Loading Matrix from groupshared memory without conversion
[linalg](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

The proposal states that loading/storing with groupshared memory will perform an automatic data conversion if the groupshared element type does not match the matrix element type.

The tint compiler (for WebGPU) will sometimes decompose groupshared memory into a generic type (e.g. 32-bit or 16-bit integers). A major motivating factor in that is an upcoming feature, [buffer_view](https://github.com/gpuweb/gpuweb/blob/main/proposals/buffer-view.md). That feature allows data type reinterpretation. Our concern is if we do not encode the same data type as the matrix element type an incorrect conversion would occur.

Would it be possible to have a way to skip conversions in the final release (say via an extra parameter)?

**Additional context**
SPIR-V's cooperative matrix does not care about the variable data type when loading/storing. MSL can do reinterpret casts. HLSL ByteAddressBuffer already handles reinterpretation so just the groupshared is problematic.

llvm-beanz:
This updates the load, store, and interlocked accumlation operations on groupshared memory to align data conversion handling and support accumulating into packed data types.

Fixes #866

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -1544,10 +1544,13 @@ declare %dx.types.LinAlgMatrix<mangling> @dx.op.linAlgMatrixLoadFromMemory.[MatT
   )
 ```
 
-Populates a matrix with data from a `groupshared` array. If the output matrix's
-component type has a native HLSL representation the input groupshared array must
-match that type, otherwise the input groupshared array must be `i32`. Validation
-will verify this requirement.
+Populates a matrix with data from a `groupshared` array. If the groupshared
+memory is any type other than `i32` the matrix's component type must match the
+scalar ype of the groupshared array. If the groupshared memory is `i32` the
+matrix may be of any type. Validation will verify this requirement.
+
+No data conversions are applied during this operation, the data in the
+groupshared array is written in the format of the input matrix element.
 
 ```llvm
 declare i32 @dx.op.linAlgMatrixLength.[MatTy](
@@ -1625,10 +1628,14 @@ declare void @dx.op.linAlgMatrixStoreToMemory.[MatTy].[Ty](
   )
 ```
 
-Store a matrix to groupshared memory. If the input matrix's
-component type has a native HLSL representation the groupshared array must
-match that type, otherwise the input groupshared array must be `i32`. Validation
-will verify this requirement.
+Store a matrix to groupshared memory. If the groupshared memory is any type
+other than `i32` the matrix's component type must match the scalar ype of the
+groupshared array. If the groupshared memory is `i32` the matrix may be of any
+type. Validation will verify this requirement.
+
+No data conversions are applied during this operation, the data in the
+groupshared array is assumed to be in the format of the destination matrix
+element.
 
 The validator will ensure that the group shared target memory is large enough
 for the write.
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -101,8 +101,9 @@ class Matrix {
   Cast();
 
   template <typename T>
-  static typename hlsl::enable_if<hlsl::is_arithmetic<T>::value, Matrix>::type
-  Splat(T Val);
+  [[nodiscard]] static
+      typename hlsl::enable_if<hlsl::is_arithmetic<T>::value, Matrix>::type
+      Splat(T Val);
 
   static Matrix Load(ByteAddressBuffer Res, uint StartOffset, uint Stride,
                      MatrixLayoutEnum Layout, uint Align = 128);
@@ -111,10 +112,13 @@ class Matrix {
                      MatrixLayoutEnum Layout, uint Align = 128);
 
   template <typename T, SIZE_TYPE Size>
-  static typename hlsl::enable_if<M * N / ElementsPerScalar <= Size,
-                                  Matrix>::type
-  Load(/*groupshared*/ T Arr[Size], uint StartIdx, uint Stride,
-       MatrixLayoutEnum Layout);
+  static
+      typename hlsl::enable_if<(hlsl::is_same<T, ElementType>::value) &&
+                                   (__detail::ScalarCountFromPackedComponents<
+                                        ComponentTy, M * N>::Value <= Size),
+                               Matrix>::type
+      Load(groupshared T Arr[Size], uint StartIdx, uint Stride,
+           MatrixLayoutEnum Layout);
 
   template <ComponentEnum LocalComp = ComponentTy>
   typename hlsl::enable_if<LocalComp == ComponentTy && IsNativeScalar,
@@ -140,9 +144,11 @@ class Matrix {
              MatrixLayoutEnum Layout, uint Align = 128);
 
   template <typename T, SIZE_TYPE Size>
-  typename hlsl::enable_if<M * N / ElementsPerScalar <= Size,
+  typename hlsl::enable_if<(hlsl::is_same<T, ElementType>::value) &&
+                               (__detail::ScalarCountFromPackedComponents<
+                                    ComponentTy, M * N>::Value <= Size),
                            void>::type
-  Store(/*groupshared*/ T Arr[Size], uint StartIdx, uint Stride,
+  Store(groupshared T Arr[Size], uint StartIdx, uint Stride,
         MatrixLayoutEnum Layout);
 
   // Accumulate methods
@@ -150,17 +156,34 @@ class Matrix {
   typename hlsl::enable_if<Use == MatrixUse::Accumulator && UseLocal == Use,
                            void>::type
   InterlockedAccumulate(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
-                        MatrixLayoutEnum Layout,
-                        uint Align = 128);
+                        MatrixLayoutEnum Layout, uint Align = 128);
 
   template <typename T, MatrixUseEnum UseLocal = Use,
             MatrixScopeEnum ScopeLocal = Scope, SIZE_TYPE Size>
   typename hlsl::enable_if<
-      UseLocal == Use && (M * N / ElementsPerScalar <= Size) &&
+      hlsl::is_arithmetic<T>::value && Use == MatrixUse::Accumulator &&
+          UseLocal == Use &&
+          (__detail::ScalarCountFromPackedComponents<ComponentTy,
+                                                     M * N>::Value <= Size) &&
           Scope == MatrixScope::Wave && ScopeLocal == Scope,
       void>::type
-  InterlockedAccumulate(/*groupshared*/ T Arr[Size], uint StartIdx, uint Stride,
+  InterlockedAccumulate(groupshared T Arr[Size], uint StartIdx, uint Stride,
                         MatrixLayoutEnum Layout);
+
+#ifdef __hlsl_dx_compiler
+  template <ComponentEnum TargetCompTy = ComponentTy,
+            MatrixUseEnum UseLocal = Use, MatrixScopeEnum ScopeLocal = Scope,
+            SIZE_TYPE Size>
+  typename hlsl::enable_if<
+      !__detail::ComponentTypeTraits<TargetCompTy>::IsNativeScalar &&
+          Use == MatrixUse::Accumulator && UseLocal == Use &&
+          (__detail::ScalarCountFromPackedComponents<TargetCompTy,
+                                                     M * N>::Value <= Size) &&
+          Scope == MatrixScope::Wave && ScopeLocal == Scope,
+      void>::type
+  InterlockedAccumulate(groupshared uint8_t4_packed Arr[Size], uint StartIdx,
+                        uint Stride, MatrixLayoutEnum Layout);
+#endif
 
   template <ComponentEnum CompTy, MatrixUseEnum UseLocal = Use>
   typename hlsl::enable_if<Use == MatrixUse::Accumulator && UseLocal == Use,
@@ -190,8 +213,7 @@ class Matrix<ComponentTy, M, N, Use, MatrixScope::Thread> {
   template <MatrixLayoutEnum Layout, MatrixUseEnum UseLocal = Use>
   static typename hlsl::enable_if<Use == MatrixUse::A && UseLocal == Use,
                                   Matrix>::type
-  Load(ByteAddressBuffer Res, uint StartOffset, uint Stride,
-       uint Align = 128);
+  Load(ByteAddressBuffer Res, uint StartOffset, uint Stride, uint Align = 128);
 
   template <MatrixUseEnum UseLocal = Use>
   typename hlsl::enable_if<Use == MatrixUse::Accumulator && UseLocal == Use,
@@ -298,6 +320,11 @@ InterlockedAccumulate(vector<InputElTy, M> Vec, RWByteAddressBuffer Res,
 ```c++
 RWByteAddressBuffer B : register(u0);
 
+groupshared float GSMat[128];
+#ifdef __hlsl_dx_compiler
+groupshared uint8_t4_packed GSMatPacked[32];
+#endif
+
 void WaveMatrixExample() {
   using namespace dx::linalg;
   using MatrixATy =
@@ -327,6 +354,20 @@ void WaveMatrixExample() {
 
   MatrixAccumTy Accum = Multiply(MatA, MatB);
   MatrixAccum32Ty Accum32 = Multiply<ComponentType::F32>(MatA, MatB);
+#ifdef __hlsl_dx_compiler
+  MatrixAccum32Ty M =
+      MatrixAccum32Ty::Load(GSMat, 0, 8, MatrixLayout::RowMajor);
+
+  // The next line is an error because the groupshared array isn't the right
+  // type for the matrix component. MatrixAccumTy M2 =
+  // MatrixAccumTy::Load(GSMat, 0, 8, MatrixLayout::RowMajor);
+
+  M.Store(GSMat, 0, 8, MatrixLayout::RowMajor);
+  M.InterlockedAccumulate(GSMat, 0, 8, MatrixLayout::RowMajor);
+
+  M.InterlockedAccumulate<ComponentType::I8>(GSMatPacked, 0, 2,
+                                             MatrixLayout::RowMajor);
+#endif
 }
 ```
 
@@ -727,6 +768,8 @@ __MATRIX_SCALAR_COMPONENT_MAPPING(ComponentType::U64, uint64_t)
 __MATRIX_SCALAR_COMPONENT_MAPPING(ComponentType::F64, double)
 
 template <ComponentEnum DstTy, ComponentEnum SrcTy, int SrcN> struct DstN {
+  // Make sure to round up in case SrcN isn't an even multiple of the number of
+  // elements per scalar
   static const int Value =
       (SrcN * ComponentTypeTraits<SrcTy>::ElementsPerScalar +
        ComponentTypeTraits<DstTy>::ElementsPerScalar - 1) /
@@ -741,6 +784,14 @@ template <SIZE_TYPE MVal, SIZE_TYPE NVal, bool Transposed> struct DimMN {
 template <SIZE_TYPE MVal, SIZE_TYPE NVal> struct DimMN<MVal, NVal, true> {
   static const SIZE_TYPE M = NVal;
   static const SIZE_TYPE N = MVal;
+};
+
+template <ComponentEnum CompTy, SIZE_TYPE PackedComponentCount>
+struct ScalarCountFromPackedComponents {
+  static const SIZE_TYPE ElementsPerScalar =
+      ComponentTypeTraits<CompTy>::ElementsPerScalar;
+  static const SIZE_TYPE Value =
+      (PackedComponentCount + ElementsPerScalar - 1) / ElementsPerScalar;
 };
 
 } // namespace __detail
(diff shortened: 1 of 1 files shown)


llvm-beanz:
Tagging a few people from hardware vendors for feedback:
@anteru, @pow2clk, @anupamachandra & @mbedy-amd.

proposals/0035-linalg-matrix.md
```diff
@@ -1406,9 +1485,10 @@ declare %dx.types.LinAlgMatrix<mangling> @dx.op.linAlgMatrixLoadFromMemory.[MatT
   )
 ```
 
-Populates a matrix with data from a `groupshared` array. Data conversions
-between opaque matrices and groupshared memory are defined in the
-[Conversions](#data-conversion-rules) section below.
+Populates a matrix with data from a `groupshared` array. If the output matrix's
+compnent type has a native HLSL representation the input groupshared array must
```
```suggestion
component type has a native HLSL representation the input groupshared array must
```

proposals/0035-linalg-matrix.md
```diff
@@ -1486,9 +1566,10 @@ declare void @dx.op.linAlgMatrixStoreToMemory.[MatTy].[Ty](
   )
 ```
 
-Store a matrix to groupshared memory. Data conversions between opaque matrices
-and groupshared memory are defined in the [Conversions](#data-conversion-rules) section
-below.
+Store a matrix to groupshared memory. If the input matrix's
+compnent type has a native HLSL representation the groupshared array must
```
```suggestion
component type has a native HLSL representation the groupshared array must
```

proposals/0035-linalg-matrix.md
```diff
@@ -849,8 +902,8 @@ can only be read from `ByteAddressBuffer` objects. Wave scope matrices can be
 read from `[RW]ByteAddressBuffer` objects or `groupshared` arrays. When read
 from `[RW]ByteAddressBuffer` objects the data is assumed to already be in the
 expected target data format. When read from `groupshared` memory, the data may
-be in any arithmetic or packed data type. If the type mismatches the target data
-type of the matrix a data conversion is applied on load.
+be in any arithmetic or packed data type. The type of the array must match the
+type of the matrix being loaded into.
```
alan-baker: The way I'm reading this is if I want to load a packed uint8_t matrix the array must also be packed uint8_t. Is that correct?

Our concern with WebGPU is someone declaring a buffer (opaque type) in workgroup and using with both packed 8bit and 16-bit granularity. We aren't sure it will be guaranteed that equivalent matrices would be supported to use the `Cast()` method.

Also in the dxil section it says non-natively supported types must use i32 for the array. Would that be possible here?

llvm-beanz: > The way I'm reading this is if I want to load a packed uint8_t matrix the array must also be packed uint8_t. Is that correct?

DXIL doesn't have 8-bit integer types, so uint8_t isn't a native type, so it would need to be uint8_t4_packed, which becomes an i32.

> Our concern with WebGPU is someone declaring a buffer (opaque type) in workgroup and using with both packed 8bit and 16-bit granularity. We aren't sure it will be guaranteed that equivalent matrices would be supported to use the `Cast()` method.

In general DXIL doesn't allow casting groupshared pointers, so we don't really have a reasonable way to reinterpret the data from a single allocation to mean different things. The one oddity of this is non-native types that we treat as i32 (uint8_t4_packed in the API), where we carry both the data type and the interpretation so the accumulation operations know how to operate on the data.

> Also in the dxil section it says non-natively supported types must use i32 for the array. Would that be possible here?

That is what the `uint8_t4_packed` type is, but it only works for a subset of types.

jrprice: As Alan says, it looks like we would still not be able to support LinAlg in WebGPU due to this requirement. In addition to the opaque buffer type that we will soon have in WGSL, we also decompose groupshared buffers into flat arrays of (e.g.) u32 or u16 in order to work around bugs in D3D drivers. This means that we won't always have to ability to force the type to be a `uint8_t4_packed`, or to use `f32` for an `f32` matrix etc.

> In general DXIL doesn't allow casting groupshared pointers

Maybe naive question, but does DXIL really need to support casting? Could the DXIL representation for these load/store operations not just allow the type mismatch, and specify that the operations are performed without conversion? This is essentially the SPIR-V approach - you can use any scalar/vector type in the src/dst pointer, and it's only used for addressing calculations.

We also considered trying to polyfill the matrix load/store operations ourselves by manually loading the data and then constructing the matrix objects using `Set()`, but a) I don't know how efficient this will be and b) the 8-bit matrix types don't support `Get()/Set()` as far as I can tell.

proposals/0035-linalg-matrix.md
```diff
@@ -101,8 +101,9 @@ class Matrix {
   Cast();
 
   template <typename T>
-  static typename hlsl::enable_if<hlsl::is_arithmetic<T>::value, Matrix>::type
-  Splat(T Val);
+  [[nodiscard]] static
+      typename hlsl::enable_if<hlsl::is_arithmetic<T>::value, Matrix>::type
+      Splat(T Val);
```
pow2clk, line 116: While a welcome change, this isn't mentioned in the description

proposals/0035-linalg-matrix.md
```diff
@@ -140,28 +144,47 @@ class Matrix {
...
+          UseLocal == Use &&
+          (__detail::ScalarCountFromPackedComponents<ComponentTy,
+                                                     M * N>::Value <= Size) &&
           Scope == MatrixScope::Wave && ScopeLocal == Scope,
       void>::type
-  InterlockedAccumulate(/*groupshared*/ T Arr[Size], uint StartIdx, uint Stride,
+  InterlockedAccumulate(groupshared T Arr[Size], uint StartIdx, uint Stride,
                         MatrixLayoutEnum Layout);
 
+#ifdef __hlsl_dx_compiler
```
pow2clk, line 183: why the restriction?

proposals/0035-linalg-matrix.md
```diff
@@ -327,6 +354,20 @@ void WaveMatrixExample() {
 
   MatrixAccumTy Accum = Multiply(MatA, MatB);
   MatrixAccum32Ty Accum32 = Multiply<ComponentType::F32>(MatA, MatB);
+#ifdef __hlsl_dx_compiler
+  MatrixAccum32Ty M =
+      MatrixAccum32Ty::Load(GSMat, 0, 8, MatrixLayout::RowMajor);
+
+  // The next line is an error because the groupshared array isn't the right
+  // type for the matrix component. MatrixAccumTy M2 =
```
pow2clk: clang-format did you dirty here and moved part of the "next line" onto the description of it. Is this a known bug or an example of how the header is expected to error on this usage?

proposals/0035-linalg-matrix.md
```diff
@@ -1026,6 +1096,15 @@ available for matrices with `MatrixUse::Accumulator` use. The
...
 `groupshared` overload is only available for `Wave` scope matrices.
 
+When accumulating to groupshared memory, a data conversion to the destination
+type following the [conversion rules](#data-conversion-rules) may occur if the
+matrix component type does not match the groupshared memory target. If the
+accumulation is into a `uint8_t4_packed` array, the API requires specifying the
+interpretation type of the `groupshared` memory.
+
+> User beware: if you pass the wrong interpretation type to an accumulation to
+> `uint8_t4_packed` groupshared memory you'll likely get crazy results!!!
```
pow2clk: There are at least two reasons why I don't think this is an appropriate adjective
```suggestion
> `uint8_t4_packed` groupshared memory you'll likely get undesired results!!!
```

proposals/0035-linalg-matrix.md
```diff
@@ -1891,7 +1988,7 @@ template <typename T> struct enable_if<true, T> {
 // Put this in a dxil constants header.
 namespace dxil {
 
-// This enum is _exactly_ the DXIL constants.
+// This enum must _exactly_ match the DXIL constants.
```
pow2clk, line 2121: Corresponding change is not made above circa line 596

proposals/0035-linalg-matrix.md
```diff
@@ -1922,6 +2019,7 @@ enum class ComponentType : uint32_t {
 
   LastEntry
 };
+
```
pow2clk, line 2152: It's minor, but again, the change is not mirrored above

proposals/0035-linalg-matrix.md
```diff
@@ -1954,6 +2052,9 @@ struct ComponentType {
     __COMPONENT_TYPE(F64),
   };
 };
+
+#undef __COMPONENT_TYPE
+
```
pow2clk, line 2187: line 660

proposals/0035-linalg-matrix.md
llvm-beanz: Treating the data in a groupshared array as two (or more) types of different sizes is currently illegal in DXIL. You can kinda work around that by just always treating the array as 32-bit integers and doing load->mask->convert chains, but the fundamental limitation exists.

There isn't a good way to work around that limitation through the linalg spec as specified.

I think we should probably have a conversation with the hardware vendors about whether or not those restrictions are still required for DXIL, or if we should remove them. Removing them has some general benefits, but would make some transformations that DXC performs on groupshared allocations potentially illegal (which is also maybe fine).

proposals/0035-linalg-matrix.md
```diff
@@ -1993,11 +2094,19 @@ template <ComponentEnum CompTy> struct ComponentTypeTraits {
   static const uint ElementsPerScalar = 4;
 };
 
+template <typename T> struct TypeTraits {
+  static const ComponentEnum CompType =
+      (ComponentEnum)dxil::ComponentType::Invalid;
+};
+
```
pow2clk, line 2231: line 749

proposals/0035-linalg-matrix.md
```diff
@@ -1993,11 +2094,19 @@ template <ComponentEnum CompTy> struct ComponentTypeTraits {
...
 #define __MATRIX_SCALAR_COMPONENT_MAPPING(enum_val, type)                      \
   template <> struct ComponentTypeTraits<enum_val> {                           \
     using Type = type;                                                         \
     static const bool IsNativeScalar = true;                                   \
     static const uint ElementsPerScalar = 1;                                   \
+  };                                                                           \
+  template <> struct TypeTraits<type> {                                        \
+    static const ComponentEnum CompType = enum_val;                            \
```
pow2clk, line 2239: 755

proposals/0035-linalg-matrix.md
```diff
@@ -2051,8 +2170,11 @@ InterpretedVector<T, N, DT> MakeInterpretedVector(vector<T, N> Vec) {
 }
 
 template <ComponentEnum DestTy, ComponentEnum OriginTy, typename T, int N>
-InterpretedVector<typename __detail::ComponentTypeTraits<DestTy>::Type,
-                  __detail::DstN<DestTy, OriginTy, N>::Value, DestTy>
+typename hlsl::enable_if<
+    DestTy != OriginTy,
+    InterpretedVector<typename __detail::ComponentTypeTraits<DestTy>::Type,
+                      __detail::DstN<DestTy, OriginTy, N>::Value,
+                      DestTy> >::type
```
pow2clk, line 2307: Prototypes at lines 78 and 820 not updated

proposals/0035-linalg-matrix.md
```diff
@@ -2061,6 +2183,13 @@ Convert(vector<T, N> Vec) {
   return MakeInterpretedVector<DestTy>(Result);
 }
 
+template <ComponentEnum DestTy, ComponentEnum OriginTy, typename T, int N>
+typename hlsl::enable_if<DestTy == OriginTy,
+                         InterpretedVector<T, N, DestTy> >::type
+Convert(vector<T, N> Vec) {
+  return MakeInterpretedVector<DestTy>(Vec);
+}
+
```
pow2clk, line 2322: Not included at line 85, nor is the function documented along with the other Convert() variant

proposals/0035-linalg-matrix.md
```diff
@@ -2216,18 +2364,17 @@ Multiply(Matrix<MatrixDT, M, K, MatrixUse::A, MatrixScope::Thread> MatrixA,
...
-          SIZE_TYPE M, SIZE_TYPE K, SIZE_TYPE VecK,
-          ComponentEnum MatrixDT>
+          SIZE_TYPE M, SIZE_TYPE K, SIZE_TYPE VecK, ComponentEnum MatrixDT>
 typename hlsl::enable_if<
     InterpretedVector<InputElTy, VecK, InputInterp>::Size == K,
     vector<OutputElTy, M> >::type
-    Multiply(Matrix<MatrixDT, M, K, MatrixUse::A, MatrixScope::Thread> MatrixA,
-             InterpretedVector<InputElTy, VecK, InputInterp> InterpVec);
+Multiply(Matrix<MatrixDT, M, K, MatrixUse::A, MatrixScope::Thread> MatrixA,
+         InterpretedVector<InputElTy, VecK, InputInterp> InterpVec);
```
pow2clk, line 2501: Minor, but still indented at line 265

proposals/0035-linalg-matrix.md
```diff
@@ -2216,18 +2364,17 @@ Multiply(Matrix<MatrixDT, M, K, MatrixUse::A, MatrixScope::Thread> MatrixA,
...
 
 template <typename OutputElTy, typename InputElTy, typename BiasElTy,
           SIZE_TYPE M, SIZE_TYPE K, ComponentEnum MatrixDT>
 typename hlsl::enable_if<hlsl::is_arithmetic<InputElTy>::value &&
-                         hlsl::is_arithmetic<BiasElTy>::value,
+                             hlsl::is_arithmetic<BiasElTy>::value,
```
pow2clk, line 2506: minor, but not indented at line 271

proposals/0035-linalg-matrix.md
alan-baker: If it helps SPV_KHR_cooperative_matrix does not require the matrix component type to match the memory type. So any vendor supporting that extension should already be able to support this.

proposals/0035-linalg-matrix.md
```diff
@@ -2236,8 +2383,8 @@ template <typename OutputElTy, typename InputElTy, ComponentEnum InputInterp,
           typename BiasElTy, SIZE_TYPE M, SIZE_TYPE K, SIZE_TYPE VecK,
           ComponentEnum MatrixDT>
 typename hlsl::enable_if<
-    InterpretedVector<InputElTy, VecK, InputInterp>::Size == K &&
-    hlsl::is_arithmetic<BiasElTy>::value,
+    VecK == __detail::ScalarCountFromPackedComponents<InputInterp, K>::Value &&
```
pow2clk, line 2515: I've been thinking about how much simpler some of this would be if we represented the size of InterpretedVectors as the number of elements like we do with matrices and VectorRefs rather than the number of scalars, but that's a topic for another day.

Meanwhile, this change is not propagated to the matching code on line 280 and 1267

proposals/0035-linalg-matrix.md
```diff
@@ -2261,8 +2408,9 @@ MultiplyAdd(Matrix<MatrixDT, M, K, MatrixUse::A, MatrixScope::Thread> MatrixA,
             VectorRef<BiasElTy, M> BiasRef);
 
 template <ComponentEnum OutTy, typename InputElTy, SIZE_TYPE M, SIZE_TYPE N>
-typename hlsl::enable_if<hlsl::is_arithmetic<InputElTy>::value,
-                         Matrix<OutTy, M, N, MatrixUse::Accumulator, MatrixScope::Thread> >::type
+typename hlsl::enable_if<
+    hlsl::is_arithmetic<InputElTy>::value,
+    Matrix<OutTy, M, N, MatrixUse::Accumulator, MatrixScope::Thread> >::type
```
pow2clk, line 2542: Lines 305 and 1242

proposals/0035-linalg-matrix.md
```diff
@@ -2304,6 +2457,20 @@ void WaveMatrixExample() {
 
   MatrixAccumTy Accum = Multiply(MatA, MatB);
   MatrixAccum32Ty Accum32 = Multiply<ComponentType::F32>(MatA, MatB);
+#ifdef __hlsl_dx_compiler
+  MatrixAccum32Ty M =
+      MatrixAccum32Ty::Load(GSMat, 0, 8, MatrixLayout::RowMajor);
+
+  // The next line is an error because the groupshared array isn't the right
+  // type for the matrix component. MatrixAccumTy M2 =
```
pow2clk: clang-format again

pow2clk (commented):
Most of my comments relate to changes made to the header that weren't mirrored to the other 1 or 2 places where the code is duplicated.

We discussed transforming this spec from containing so many of the requirements in header code to simplifying the code to leave that to the actual header while better explaining restrictions in the text. Is that no longer the plan? This seems to be moving in the opposite direction.

proposals/0035-linalg-matrix.md
llvm-beanz: Because we haven't implemented the packed types in Clang yet, so this is needed to keep the code compiling with Clang.

proposals/0035-linalg-matrix.md
llvm-beanz: It is an expected error based on the current state of this spec, although there is a request (in other comments) to change that requirement.

llvm-beanz:
A bunch of the code snippets across the file were out-of-sync with the appendix. I've re-sync'd all of that now.

llvm-beanz:
> We discussed transforming this spec from containing so many of the requirements in header code to simplifying the code to leave that to the actual header while better explaining restrictions in the text. Is that no longer the plan? This seems to be moving in the opposite direction.

Where we are without this change is a bit of a worst-of-all-worlds. The header in the annex was so far out of date from the implementation that it was no longer useful for crafting interface changes, and the code snippets across the file were in even worse state.

I did not fully sync this to the implementation, but I did pull in some of the differences in the implementation required to help me craft API changes that were needed for this PR.

proposals/0035-linalg-matrix.md
mjbedy: This looks like it needs updating to match the new DXIL sections?

proposals/0035-linalg-matrix.md
```diff
@@ -1406,9 +1544,13 @@ declare %dx.types.LinAlgMatrix<mangling> @dx.op.linAlgMatrixLoadFromMemory.[MatT
   )
 ```
 
-Populates a matrix with data from a `groupshared` array. Data conversions
-between opaque matrices and groupshared memory are defined in the
-[Conversions](#data-conversion-rules) section below.
+Populates a matrix with data from a `groupshared` array. If the groupshared
+memory is any type other than `i32` the matrix's component type must match the
+scalar ype of the groupshared array. If the groupshared memory is `i32` the
```
mjbedy:
```suggestion
scalar type of the groupshared array. If the groupshared memory is `i32` the
```

proposals/0035-linalg-matrix.md
```diff
@@ -1486,9 +1628,14 @@ declare void @dx.op.linAlgMatrixStoreToMemory.[MatTy].[Ty](
   )
 ```
 
-Store a matrix to groupshared memory. Data conversions between opaque matrices
-and groupshared memory are defined in the [Conversions](#data-conversion-rules) section
-below.
+Store a matrix to groupshared memory. If the groupshared memory is any type
+other than `i32` the matrix's component type must match the scalar ype of the
```
mjbedy:
```suggestion
other than `i32` the matrix's component type must match the scalar type of the
```

proposals/0035-linalg-matrix.md
```diff
@@ -1486,9 +1628,14 @@ declare void @dx.op.linAlgMatrixStoreToMemory.[MatTy].[Ty](
   )
```
mjbedy, line 1649: Are the Offset and Stride intended to be in terms of the element type of the groupshared array? (e.g., if you are storing f16 elements to i32 array, the Offset and Strides will be in terms of 4 byte elements.)

mjbedy (commented):
This seems like a good general direction. I'm going to follow up to confirm a couple things, but left some comments in the meantime.

proposals/0035-linalg-matrix.md
```diff
@@ -1653,6 +1800,7 @@ declare void @dx.op.linAlgMatrixAccumulateToMemory.[MatTy].[Ty](
   immarg i32,                         ; opcode
   %dx.types.LinAlgMatrix<mangling>,   ; matrix
   [Ty] addrspace(3)*,                 ; groupshared Ty[M * N]
+  immarg i32,                         ; target data type
```
V-FEXrt, line 1833: I can't add a comment directly to the right spot because there is no diff there but can you update the sections that start with

`For the Load operations on groupshared arrays:` and `For the Store operations on groupshared arrays:` to comment on what `Stride` and `Offset` are expected to be when target data type and gs array data type disagree?

V-FEXrt (approved):
There are a number of changes in here that will trigger updating/filing new issues against the DXC implementation

llvm-beanz (commented):
Issues that need to be filed on DXC:

- [x] Header updates on Load, Store, and InterlockedAccumulate methods operating on groupshared arrays and associated traits (https://github.com/microsoft/DirectXShaderCompiler/issues/8634)
- [x] Validation rules for load and store groupshared array types (https://github.com/microsoft/DirectXShaderCompiler/issues/8635)
- [x] Validation rules for load, store, and interlocked accumulate offset and stride (if constant) (this is also related to changes in #891) (https://github.com/microsoft/DirectXShaderCompiler/issues/8636)
- [x] New parameter for dx.op.linAlgMatrixAccumulateToMemory (https://github.com/microsoft/DirectXShaderCompiler/issues/8637)

proposals/0035-linalg-matrix.md
llvm-beanz: Before this PR merged, this language was updated to:

> The type of the array must match the type of the matrix being loaded into, or it must be a `uint8_t4_packed` which may be used to represent matrix data at rest in any format.

I believe this addresses the WebGPU request. Please let me know if there is something else you need.

PR #882, opened 2026-06-08
Update VK SDK Release Spec

joaosaffran:
This patch proposes a set of steps required to be performed to release DXC into the VK SDK.

### DIFF
--- a/proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
+++ b/proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
@@ -119,7 +119,8 @@ opts in.
 4. **Offload tests.** A separate job runs the LLVM offload-test-suite against the
    published artifact, with DXC as the only compiler (the in-tree clang compiler is
    disabled). The Vulkan tests (`check-hlsl-vk`) on lavapipe and the Direct3D 12
-   tests (`check-hlsl-warp-d3d12`) on WARP. Because it consumes the published artifact
+   tests (`check-hlsl-warp-d3d12`) on WARP (developers still use the dxc which ships
+   in the Vulkan SDK to compile dx shaders). Because it consumes the published artifact
    rather than the build tree, this job can also be run on its own against any earlier candidate.
 
 ### Release manifest
@@ -150,12 +151,16 @@ candidate was built against and the tests results:
 These steps are performed by whoever is currently responsible for monitoring the
 llvm-build, and may be repeated as needed:
 
-1. Update the SPIRV-Headers and SPIRV-Tools submodules to the commits specified by
+1. File an issue in DXC repository using the `vk_release_checklist.md` issue template,
+   assign this to yourself. This should be handed off at the end of rotation. 
+3. Update the SPIRV-Headers and SPIRV-Tools submodules to the commits specified by
    LunarG.
-2. Create the `release/vulkan/<version>` branch, which triggers the pipeline.
-3. Check whether the resulting candidate is validated (see
+4. Create the `release/vulkan/<version>` branch, which triggers the pipeline.
+5. Check whether the resulting candidate is validated (see
    [Release Candidate readiness](#release-candidate-readiness)).
-4. Email the validated DXC commit back to LunarG.
+6. Tag the validated commit, `vulkan-sdk-<version>-rc`, as a release candidate
+7. Report that tag to LunarG. This can be done through email
+   or by updating Khronos GitLab release issue with the validated commit.
 
 ### Release Candidate readiness
 
--- a/proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
+++ b/proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
@@ -42,7 +42,9 @@ For an SDK release, the pipeline also publishes the candidate as an artifact and
 the LLVM [offload-test-suite](https://github.com/llvm/offload-test-suite) against it,
 using lavapipe as the SPIRV driver and WARP as the DXIL driver. DXC from the Vulkan SDK
 is also used to compile shader targeting DirectX, so its DXIL output is worth testing too. 
-LunarG is handed the validated DXC commit, recorded in the release candidate manifest.
+Each validated candidate handed to LunarG is tagged as a release candidate (its
+results recorded in the manifest); the latest is promoted to a release once the Vulkan
+SDK ships.
 
 The automation is split into three workflows, each shown below starting from its triggers:
 
@@ -146,6 +148,20 @@ candidate was built against and the tests results:
 }
 ```
 
+### Release tagging
+
+A release is prepared on a branch and shipped to LunarG as a series of release
+candidate tags. The branch carries the Vulkan SDK's major and minor version; each tag
+adds the patch version and a release-candidate number. For SDK version `1.4.360`, for
+example, the release branch is `release/vulkan/1.4.360`, the candidates sent to LunarG
+are tagged `1.4.360.0rc1`, `1.4.360.0rc2`, and so on, and once the Vulkan SDK is
+published the latest candidate is promoted to a release tag with no `rc` suffix:
+`1.4.360.0`.
+
+A single SDK release might involves more than one handoff: LunarG may surface an
+issue, which is fixed on the branch and re-validated before the next candidate tag is
+cut. Only the final, shipped candidate becomes the release.
+
 ### Release Steps
 
 These steps are performed by whoever is currently responsible for monitoring the
@@ -153,14 +169,19 @@ llvm-build, and may be repeated as needed:
 
 1. File an issue in DXC repository using the `vk_release_checklist.md` issue template,
    assign this to yourself. This should be handed off at the end of rotation. 
-3. Update the SPIRV-Headers and SPIRV-Tools submodules to the commits specified by
+2. Update the SPIRV-Headers and SPIRV-Tools submodules to the commits specified by
    LunarG.
-4. Create the `release/vulkan/<version>` branch, which triggers the pipeline.
-5. Check whether the resulting candidate is validated (see
+3. Create the `release/vulkan/<version>` branch — named for the SDK major and minor
+   version, e.g. `release/vulkan/1.4.360` — which triggers the pipeline.
+4. Check whether the resulting candidate is validated (see
    [Release Candidate readiness](#release-candidate-readiness)).
-6. Tag the validated commit, `vulkan-sdk-<version>-rc`, as a release candidate
-7. Report that tag to LunarG. This can be done through email
-   or by updating Khronos GitLab release issue with the validated commit.
+5. Tag the validated commit with the next release-candidate tag (e.g. `1.4.360.0rc1`)
+   and report that tag to LunarG — by email, or by updating the Khronos GitLab release
+   issue with the validated commit.
+6. If LunarG reports a problem, fix it on the release branch and re-validate, then cut
+   the next release-candidate tag (`...rc2`, and so on) and report it.
+7. Once the Vulkan SDK is published, promote the latest release-candidate tag to a
+   release tag with no `rc` suffix, e.g. `1.4.360.0`.
 
 ### Release Candidate readiness
 
--- a/proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
+++ b/proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
@@ -34,15 +34,15 @@ releases and can be ingested into Godbolt.
 
 The SPIRV-Headers and SPIRV-Tools submodules are the single source of truth for the
 SPIRV revisions a candidate is built against. Once a week a workflow bumps them to the
-latest upstream commit and opens a pull request to merge that bump into `main`. The
+latest upstream commit and opens a pull request to merge that update into `main`. The
 pull request runs the release-candidate pipeline, building DXC against
-the bumped submodules and running the SPIRV tests.
+the updated submodules and running the SPIRV tests.
 
 For an SDK release, the pipeline also publishes the candidate as an artifact and runs
 the LLVM [offload-test-suite](https://github.com/llvm/offload-test-suite) against it,
 using lavapipe as the SPIRV driver and WARP as the DXIL driver. DXC from the Vulkan SDK
-is also used for DirectX, so its DXIL output is worth testing too. The SDK builders are
-handed the validated DXC commit, recorded in the manifest, not the artifact.
+is also used to compile shader targeting DirectX, so its DXIL output is worth testing too. 
+LunarG is handed the validated DXC commit, recorded in the release candidate manifest.
 
 The automation is split into three workflows, each shown below starting from its triggers:
 
@@ -107,17 +107,16 @@ opts in.
    tests enabled. The build also produces the tools the tests need, including
    `spirv-val`.
 
-2. **Test.** Every SPIRV test in the DXC repo runs here, across all of its harnesses:
-   the googletest unit tests, the lit CodeGenSPIRV tests, and the TAEF tests; the
-   binary's output is then checked with `spirv-val`. Each result is recorded
-   in the manifest, including failed tests. The logs also contain further information
-   where each file is located and what was exected. This stage is non-blocking —
-   a release candidate is published even if some tests fail.
+2. **Test.** Googletest unit tests, the lit CodeGenSPIRV tests, and the TAEF tests are
+   all run in this stage; the binary's output is then checked with `spirv-val`. Each
+   result is recorded in the manifest, including failed tests. The logs also contain
+   further information where each file is located and what was exected. This stage is
+   non-blocking — a release candidate is published even if some tests fail.
 
-4. **Publish.** The DXC binaries (`dxc`, `dxv`, and `dxcompiler.dll`), the manifest,
+3. **Publish.** The DXC binaries (`dxc`, `dxv`, and `dxcompiler.dll`), the manifest,
    and the test reports are uploaded as a single artifact named `dxc_rc_<version>`.
 
-5. **Offload tests.** A separate job runs the LLVM offload-test-suite against the
+4. **Offload tests.** A separate job runs the LLVM offload-test-suite against the
    published artifact, with DXC as the only compiler (the in-tree clang compiler is
    disabled). The Vulkan tests (`check-hlsl-vk`) on lavapipe and the Direct3D 12
    tests (`check-hlsl-warp-d3d12`) on WARP. Because it consumes the published artifact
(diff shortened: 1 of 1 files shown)


proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,141 @@ releases and can be ingested into Godbolt.
...
-
+The SPIRV-Headers and SPIRV-Tools submodules are the single source of truth for the
+SPIRV revisions a candidate is built against. Once a week a workflow bumps them to the
+latest upstream commit and opens a pull request to merge that bump into `main`. The
+pull request runs the release-candidate pipeline, building DXC against
+the bumped submodules and running the SPIRV tests.
+
+For an SDK release, the pipeline also publishes the candidate as an artifact and runs
+the LLVM [offload-test-suite](https://github.com/llvm/offload-test-suite) against it,
+using lavapipe as the SPIRV driver and WARP as the DXIL driver. DXC from the Vulkan SDK
```
bob80905, line 43: Can you explain why we're opting to use WARP / lavapipe, instead of changing the release candidate pipeline to run on a GPU? Is the only reason because it's easier?

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,141 @@ releases and can be ingested into Godbolt.
...
+is also used for DirectX, so its DXIL output is worth testing too. The SDK builders are
```
alsepkow: 'DirectX' what?

alsepkow: What are 'the SDK builders' ?

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,141 @@ releases and can be ingested into Godbolt.
...
+handed the validated DXC commit, recorded in the manifest, not the artifact.
```
alsepkow: which manifest?

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,141 @@ releases and can be ingested into Godbolt.
...
-reviewing.
-
-## Acknowledgments (Optional)
-
-Take a moment to acknowledge the contributions of people other than the author
-and sponsor.
-
-
+The SPIRV-Headers and SPIRV-Tools submodules are the single source of truth for the
+SPIRV revisions a candidate is built against. Once a week a workflow bumps them to the
```
alsepkow, line 36: nit: I think it would be clearer to say 'updates' instead of 'bumps'. Applies to other uses of bump/s in this doc

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,141 @@ releases and can be ingested into Godbolt.
...
+also publishes the artifact and runs the offload tests.
+
+```mermaid
+flowchart TD
+    push(["Trigger: push to a<br/>vk-update or release/vulkan branch"]) --> build
+    md2(["Trigger: manual dispatch"]) --> build
+    build["Build DXC against the<br/>submodule pointers, plus spirv-val"] --> test
+    test["Run every SPIRV test:<br/>googletest, lit, TAEF, spirv-val"] --> gate{"Release candidate?<br/>release/vulkan branch<br/>or manual opt-in"}
+    gate -->|no| prchecks(["Validation only — checks surface<br/>on the vk-update PR, merge when green"])
+    gate -->|yes| publish["Publish dxc_rc artifact + manifest"]
```
bob80905, line 73: How exactly do we trigger the publishing of this artifact? Is it done manually, or will there be some automatic check to see if the branch name is prefixed with release ?

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,141 @@ releases and can be ingested into Godbolt.
...
+The build and test stages run on every push; the publish and offload stages run only 
+for a release candidate — a `release/vulkan/<version>` branch, or a manual run that 
+opts in.
+
+1. **Build.** DXC is checked out with its submodules, so it builds against exactly
+   the SPIRV the pointers name, configured with SPIRV code generation and the SPIRV
+   tests enabled. The build also produces the tools the tests need, including
+   `spirv-val`.
+
+2. **Test.** Every SPIRV test in the DXC repo runs here, across all of its harnesses:
```
alsepkow: How do we know we run every SPIRV test?

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,141 @@ releases and can be ingested into Godbolt.
...
+   where each file is located and what was exected. This stage is non-blocking —
+   a release candidate is published even if some tests fail.
+
+4. **Publish.** The DXC binaries (`dxc`, `dxv`, and `dxcompiler.dll`), the manifest,
+   and the test reports are uploaded as a single artifact named `dxc_rc_<version>`.
+
+5. **Offload tests.** A separate job runs the LLVM offload-test-suite against the
+   published artifact, with DXC as the only compiler (the in-tree clang compiler is
+   disabled). The Vulkan tests (`check-hlsl-vk`) on lavapipe and the Direct3D 12
+   tests (`check-hlsl-warp-d3d12`) on WARP. Because it consumes the published artifact
```
bob80905: Can you defend the reasoning for why we even need to run Direct3D 12 tests, if this is for vulkan SDK releases? Is it just extra coverage for the sake of coverage, or is there some need?

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,141 @@ releases and can be ingested into Godbolt.
...
+   rather than the build tree, this job can also be run on its own against any earlier candidate.
+
+### Release manifest
+
+The manifest records the DXC commit, the SPIRV-Headers and SPIRV-Tools commits the
+candidate was built against, the tests results, and a single `validated` flag
```
alsepkow: Do we have a precedent of calling this 'validated'? That name seems odd to me. I'd interpret that more as a 'did we run tests against this release?'.  And even that seems implied by the results in test_suites?

So, the name seems odd and I question if this is even giving us any information that we don't already have?

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,141 @@ releases and can be ingested into Godbolt.
...
+}
+```
+
+### Release Steps
+
+These steps are performed by whoever is currently responsible for monitoring the
+llvm-build, and may be repeated as needed:
+
+1. Update the SPIRV-Headers and SPIRV-Tools submodules to the commits specified by
+   LunarG.
```
alsepkow, line 172: Where do they specify these commits?

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,141 @@ releases and can be ingested into Godbolt.
...
+
+### Release manifest
+
+The manifest records the DXC commit, the SPIRV-Headers and SPIRV-Tools commits the
+candidate was built against, the tests results, and a single `validated` flag
+that is true only when every test passed:
+
+```json
+{
+  "dxc_commit": "<sha>",
```
bob80905, line 135: Wouldn't we want to document software renderer versions here as well? (Since some tests might pass on WARP in 1.0.18 and not 1.0.19).
And for lavapipe, maybe the commit sha that is being used should be included as well.

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,141 @@ releases and can be ingested into Godbolt.
...
+
+These steps are performed by whoever is currently responsible for monitoring the
+llvm-build, and may be repeated as needed:
+
+1. Update the SPIRV-Headers and SPIRV-Tools submodules to the commits specified by
+   LunarG.
+2. Create the `release/vulkan/<version>` branch, which triggers the pipeline.
+3. Check whether the resulting candidate is validated (see
+   [Release Candidate readiness](#release-candidate-readiness)).
+4. Report the validated DXC commit to LunarG.
```
alsepkow: What is the reporting mechanism?

bob80905 (commented):
Reviewed with @alsepkow, couple comments.

alsepkow (commented):
Left some comments with @bob80905

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
joaosaffran: They get in touch with people, I believe, through email.

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
joaosaffran: That was a request from @llvm-beanz. 

Running the `offload-tests` on a GPU would require one of 2 things:
- Create a new GPU lab only to run the dxc tests, then we can replicate all `offload-tests` infrastructure in DXC.
- Send the release candidate to the `offload-tests` repo and trigger an action inside the `offload-tests` repository to run the tests using the release candidate and send the results back to DXC repo.

After I spoke with Chris about this, he suggested using Warp and Lavapipe.

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
joaosaffran: It is automatically done by the pipeline. If you trigger it by creating a release branch, it will automatically publish it, if you trigger it in the UI there is a flag you can set to publish it

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
joaosaffran: Developers use the DXC within the Vulkan SDK to generate both vk and dx shaders.

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
alsepkow: I wonder if it would be good for us to clarify that process here? Or maybe thats not something to openly advertise - I'm not sure.

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
alsepkow: I think it would be good to capture a summary of the root of the reasoning in this spec.

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
alsepkow: I think it would be good to add a quick blurb on that. "Because developers still use the dxc which ships in the Vulkan SDK to compile dx shaders...."

bob80905 (approved):
The only unresolved comment I have is including an explanation for why we should use lavapipe/warp instead of actually offloading to a GPU for testing, but I don't think that's critical to include in the spec.

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
llvm-beanz: The point of testing the release branch is to validate that DXC generates correct DXIL and SPIRV, we can do that based on WARP and lava pipe which allows us to use VMs which we can spawn on demand instead of adding additional load to our already over-burdened test lab.

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,139 @@ releases and can be ingested into Godbolt.
...
+
+These steps are performed by whoever is currently responsible for monitoring the
+llvm-build, and may be repeated as needed:
+
+1. Update the SPIRV-Headers and SPIRV-Tools submodules to the commits specified by
+   LunarG.
+2. Create the `release/vulkan/<version>` branch, which triggers the pipeline.
+3. Check whether the resulting candidate is validated (see
+   [Release Candidate readiness](#release-candidate-readiness)).
+4. Email the validated DXC commit back to LunarG.
```
llvm-beanz: There's an issue with a checklist for the release in the Khronos GitLab. We can report the hash there as well.

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,139 @@ releases and can be ingested into Godbolt.
...
+    { "name": "spirv-val", "passed": 6, "failed": 0 }
+  ]
+}
+```
+
+### Release Steps
+
+These steps are performed by whoever is currently responsible for monitoring the
+llvm-build, and may be repeated as needed:
+
```
llvm-beanz, line 168: I think the first step should be to file an issue with a release schedule and checklist. That also gives us something that we can hand off so that if this starts under one person's time on rotation it can pass to the next person as needed.

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
llvm-beanz: At the point where we're communicating a commit, it should probably be tagged at least as a release candidate, and the tag is what we should be communicating.

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,144 @@ releases and can be ingested into Godbolt.
...
+1. File an issue in DXC repository using the `vk_release_checklist.md` issue template,
+   assign this to yourself. This should be handed off at the end of rotation. 
+3. Update the SPIRV-Headers and SPIRV-Tools submodules to the commits specified by
+   LunarG.
+4. Create the `release/vulkan/<version>` branch, which triggers the pipeline.
+5. Check whether the resulting candidate is validated (see
+   [Release Candidate readiness](#release-candidate-readiness)).
+6. Tag the validated commit, `vulkan-sdk-<version>-rc`, as a release candidate
```
llvm-beanz: At what point do we tag a final release vs a release candidate?

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,144 @@ releases and can be ingested into Godbolt.
...
+
+The following must be true and validated for a release candidate to be considered
+ready for the Vulkan SDK.
+
+* It builds against the SPIRV-Headers and SPIRV-Tools commits the submodules are
+  pinned to.
+* Every SPIRV test harness passes — the googletest, lit, and TAEF tests — and the
+  SPIRV the binary emits validates under `spirv-val`.
+* The shaders the candidate compiles execute on the offload-test-suite under both
+  software renderers: the SPIRV on lavapipe (Vulkan) and the DXIL on WARP (Direct3D 12).
```
llvm-beanz, line 195: A question that should maybe wait until @damyanp can weigh in: should we publish releases aligned with the Vulkan SDK to NuGet and on GitHub?

It seems to me that having an "official" DXC release for every Vulkan SDK and (when required) Agility SDK might be a desirable outcome.

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
joaosaffran: I think we could tag the validated commit as an actual release. At this point, the tests passed and everything looks right. So there is no point to tag it twice; we can just branch to run the tests. That way, main contributions don't interfere, and once it is correct, we can tag it. Thoughts? @llvm-beanz

llvm-beanz: That precludes the ability for us to take a bug fix. What if SPIRV-Tools or SPIRV-Headers has a bug and comes to us with an updated hash?

joaosaffran: Based on our offline discussion, I added a section mentioning the tagging strategy we came up with.

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
joaosaffran: I filed an issue to discuss this with him once he is back from vacation: https://github.com/microsoft/hlsl-specs/issues/884

proposals/infra/INF-0007-DXC-Vulkan-SDK-Release-Strategy.md
```diff
@@ -30,40 +32,165 @@ releases and can be ingested into Godbolt.
...
+
+### Release tagging
+
+A release is prepared on a branch and shipped to LunarG as a series of release
+candidate tags. The branch carries the Vulkan SDK's major and minor version; each tag
+adds the patch version and a release-candidate number. For SDK version `1.4.360`, for
+example, the release branch is `release/vulkan/1.4.360`, the candidates sent to LunarG
+are tagged `1.4.360.0rc1`, `1.4.360.0rc2`, and so on, and once the Vulkan SDK is
+published the latest candidate is promoted to a release tag with no `rc` suffix:
+`1.4.360.0`.
```
llvm-beanz: The Vulkan SDK tags should probably be tagged in a way that marks them as Vulkan SDK versions otherwise it would be easy to confuse them with other DXC tags. Our existing DXC tagging scheme is `v<major>.<minor>.YYMM.<patch>`

SPIRV-Tools uses tags of the format `v<year>.<count>[.rc<count>]` for its normal releases and `vulkan-sdk-<major>.<minor>.<sdkversion>.<patch>` for the Vulkan SDK tags.

If we're tagging a version that is the Vulkan SDK version I think the tag should start with `vulkan-sdk`.

If we instead unify the releases and are tagging based on a DXC release version, that probably has some wider changes to the branch management and tagging that we'd want to consider. We can discuss that with @damyanp as part of #884.

llvm-beanz (approved):

PR #891, opened 2026-06-30
[0035] Add alignment requirements

### LINKED ISSUE #723: [0035] Specify Alignment requirements for Matrix (and Bias?) memory.
(https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

The current spec doesn't mention alignments for baseaddress for the Matrix memory and offset at load. Same for Bias Vectors. These were in place in the original [Coopvector specification](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md) for efficient vectorization of load/store operations. Can these  be added to the spec?

**Additional context**
While alignment guarantees are useful in general and there are several proposals/issues in flight to handle these generally in HLSL, the lack of these guarantees in operations that are accelerated through special Matrix-Multiplication hardware causes significant performance cliffs.

### LINKED ISSUE #839: [0035] Alignment guarantee for Matrix Base, Offset, Stride and Size

### LINKED ISSUE #876: [0035] State offset/align params in VecAccumulate align to 64b

Explicitly state that base address and align of VectorAccumulate op must align to 64 bytes

**Additional context**
This was also part of the coopvec spec 
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md#arguments-2

llvm-beanz:
This adds the alignment requirements for Load/Store and Accumulate operations.

Resolves #723, Resolves #839, Resolves #876

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -864,8 +864,14 @@ Applications need to query the driver to determine if a matrix transpose is
 supported.
 
 For the `Load` operations on `[RW]ByteAddressBuffers`:
-  - the `Stride` argument is the row or column stride in bytes.
+  - the `Stride` argument is the row or column stride in bytes, and must be a
+    multiple of 16.
   - the `Offset` argument is the number of bytes to skip before loading.
+
+For overloads operating on device memory (`RWByteAddressBuffer`), the address of
+the first element of the matrix (base address of the resource + the offset) must
+be 128-byte aligned. The `Stride` argument must refer to 16-byte aligned byte
+offsets.
 
 For the `Load` operations on `groupshared` arrays:
   - an element is a type matching the element type of the `groupshared` array.
@@ -975,8 +981,14 @@ is converted to the target arithmetic or packed data type if the data types do
 not match.
 
 For the `Store` operations on `RWByteAddressBuffers`:
-  - the `Stride` argument is the row or column stride in bytes.
+  - the `Stride` argument is the row or column stride in bytes, and must be a
+    multiple of 16.
   - the `Offset` argument is the number of bytes to skip before storing.
+
+For overloads operating on device memory (`RWByteAddressBuffer`), the address of
+the first element of the matrix (base address of the resource + the offset) must
+be 128-byte aligned. The `Stride` argument must refer to 16-byte aligned byte
+offsets.
 
 For the `Store` operations on `groupshared` arrays:
   - an element is a type matching the element type of the `groupshared` array.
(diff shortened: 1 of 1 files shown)


V-FEXrt (approved):
LGTM

I'm not sure how much of these changes need to be directly enforced but for the ones that do we should file header/validation issues against the DXC implementation to ensure they aren't missesd

proposals/0035-linalg-matrix.md
```diff
@@ -864,9 +864,15 @@ Applications need to query the driver to determine if a matrix transpose is
...
   - the `Offset` argument is the number of bytes to skip before loading.
 
+For overloads operating on device memory (`RWByteAddressBuffer`), the address of
+the first element of the matrix (base address of the resource + the offset) must
+be 128-byte aligned. The `Stride` argument must refer to 16-byte aligned byte
+offsets.
+
 For the `Load` operations on `groupshared` arrays:
   - an element is a type matching the element type of the `groupshared` array.
   - the `Stride` argument is the row or column stride in elements.
```
anupamachandra, line 987: The stride and offset of groupshared load/store should be multiples of 16B as well.

PR #892, opened 2026-07-01
[0035] Update vector InterlockedAccumulate

### LINKED ISSUE #869: [0035] vector InterlockedAccumulate parameter ordering is inconsistent
https://github.com/microsoft/hlsl-specs/blob/c2f44e38a31b92d7f110f8bbbb80dba36bb5af3b/proposals/0035-linalg-matrix.md?plain=1#L287-L290

Parameter ordering for Interlocked operations typically starts with a parameter for the location into which you are accumulating. This new operation reverses the parameter ordering, starting with the vector to accumulate, followed by the target buffer.

I think it would make more sense to reverse the order of these parameters.

The earlier we do this, the less pain, since this operation is brand new and we haven't released a preview that includes it yet.

For the DXIL op (`dx.op.vectorAccumulateToDescriptor`), I think it would be more consistent with existing atomic operations to place the resource and offset before the vector.

While we're at it, I also think the DXIL op naming should be made more consistent with other atomic operations in DXIL as well, such as: `AtomicVectorAccumulate`. Though this could be considered an independent issue.

llvm-beanz:
This updates the vector interlocked accumulate to correct parameter ordering, and make alignment an immediate (and template) argument.

Resolves #869

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -284,10 +284,10 @@ typename hlsl::enable_if<hlsl::is_arithmetic<InputElTy>::value,
                          Matrix<OutTy, M, N, MatrixUse::Accumulator, MatrixScope::Thread> >::type
 OuterProduct(vector<InputElTy, M> VecA, vector<InputElTy, N> VecB);
 
-template <typename InputElTy, SIZE_TYPE M>
+template <uint Align = 64, typename InputElTy, SIZE_TYPE M>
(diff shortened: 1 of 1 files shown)


V-FEXrt (approved):
Will need DXC issues filed to track the implementation changes needed!

llvm-beanz:
Filed: https://github.com/microsoft/DirectXShaderCompiler/issues/8639

V-FEXrt:
Another random question! Does Matrix InterlockedAccumulate (with non-template param Align = 128) need the same treatment here?

PR #899, opened 2026-07-23
[0035] Make Align a template argument

llvm-beanz:
These need to be compile-time immediate values and they have default values. This doesn't change any example usage, but it does allow compile-time enforcement that it is an immediate.


V-FEXrt (approved):

bob80905 (approved):
LGTM, template argument order also looks consistent

V-FEXrt:
I had an extra moment so I already filed the implementation issue for this change!

PR #900, opened 2026-07-27
[0035] Correct an error from a recent PR

llvm-beanz:
The Matrix::Store function has void return type


V-FEXrt (approved):

bob80905 (approved):

V-FEXrt:
This is already fixed in the implementation! No new issue needed

PR #901, opened 2026-07-27
[0035] Remove redundant bias interpretation type

### LINKED ISSUE #870: [0035] Remove now redundnat bias interpretation parameter from matvecmuladd
https://github.com/microsoft/hlsl-specs/blob/c2f44e38a31b92d7f110f8bbbb80dba36bb5af3b/proposals/0035-linalg-matrix.md?plain=1#L1608

On account of the convert operation added to MultiplyAdd that makes the bias vector always match the output type, the bias interp enum parameter will always match the output type, making it redundant and adding some validation overhead to the runtime and potentially compiler

For the preview, it makes sense to assume or enforce that the enum match the output type. Post preview, the dxil op should be modified to remove the parameter.

llvm-beanz:
Since we always load and convert the bias to the output type this argument is redundant.

Fixes #870

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -1801,8 +1801,7 @@ declare <[NUMo] x [TYo]> @dx.op.linAlgMatVecMulAdd.v[NUMo][TYo].[MatTy].v[NUMi][
   immarg i1,                          ; is output signed
   <[NUMi] x [TYi]>,                   ; input vector
   immarg i32,                         ; input interpretation type (DXIL::ComponentType)
-  <[NUMo] x [TYb]>,                   ; bias vector
-  immarg i32                          ; bias interpretation type (DXIL::ComponentType)
+  <[NUMo] x [TYb]>                   ; bias vector
   )
 ```
 
@@ -1812,6 +1811,7 @@ of `Thread` scope with a bias vector added to the result.
 Validation will enforce that:
 * The input vector length matches the `K` matrix dimension
 * The bias vector length matches the `M` matrix dimension
+* The bias vector's element type and length must match the output vector type.
 * The matrix A is an `A` matrix of `Thread` scope
 * The input and bias interpretation type must be one of the valid linalg
   component types specified in the list in the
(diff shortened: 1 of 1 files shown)


V-FEXrt (approved):

V-FEXrt:
I had an extra second, so I already filed the implementation issue for this PR!

proposals/0035-linalg-matrix.md
```diff
@@ -1812,6 +1811,7 @@ of `Thread` scope with a bias vector added to the result.
 Validation will enforce that:
 * The input vector length matches the `K` matrix dimension
 * The bias vector length matches the `M` matrix dimension
+* The bias vector's element type and length must match the output vector type.
```
alsepkow: Is it weird to mention length again in this statement when we already say it must match `M` matrix dimension directly above? So, this could just be 'The bias vector's element type must match the output vector type' ?

```suggestion
* The bias vector and output vector length matches the `M` matrix dimension
* The bias vector's element type matches the output vector type.
```

llvm-beanz: I made a small edit to the bullet above and this to make it more clear.

alsepkow (approved):

proposals/0035-linalg-matrix.md
```diff
@@ -1811,7 +1810,8 @@ of `Thread` scope with a bias vector added to the result.
 
 Validation will enforce that:
 * The input vector length matches the `K` matrix dimension
-* The bias vector length matches the `M` matrix dimension
+* The bias vector and output vector length matches the `M` matrix dimension
+* The bias vector's element type matches the output vector type.
 * The matrix A is an `A` matrix of `Thread` scope
 * The input and bias interpretation type must be one of the valid linalg
```
V-FEXrt, line 1816: Sorry for the late comment! I just noticed this

```suggestion
* The input must be one of the valid linalg
```

proposals/0035-linalg-matrix.md
```diff
@@ -1811,7 +1810,8 @@ of `Thread` scope with a bias vector added to the result.
 
 Validation will enforce that:
 * The input vector length matches the `K` matrix dimension
-* The bias vector length matches the `M` matrix dimension
+* The bias vector and output vector length matches the `M` matrix dimension
+* The bias vector's element type matches the output vector type.
```
V-FEXrt, line 1814: and this one:)
```suggestion
* The bias vector's element type matches the output vector's element type.
```

PR #902, opened 2026-07-27
[0035] Add scope validation rules

### LINKED ISSUE #896: [0035] Explicitly state scope restrictions for builtins

Builtins should explicitly list scope level restrictions that must be validated. As an example, `CopyConvert` is only valid to use with a `Wave` or `ThreadGroup` matrix so that should be called out as a required validation

llvm-beanz:
This adds the validation rules for each DXIL opeation based on matrix scope. Also updated formatting on a few rules to be consistent and added some missing validations for resource handles.

Fixes #896

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -1524,6 +1524,9 @@ Fills a matrix with a scalar value. The scalar's type does not need to match the
 matrix component's type, a type conversion is applied following the rules
 documented in the [Conversions](#data-conversion-rules) section.
 
+Validation shall enforce that:
+* The output matrix scope must be `Wave` or `ThreadGroup`
+
 ```llvm
 declare %dx.types.LinAlgMatrix<mangling> @dx.op.linAlgCopyConvertMatrix.[MatTy1].[MatTy2](
   immarg i32,                         ; opcode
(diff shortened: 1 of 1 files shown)


proposals/0035-linalg-matrix.md
```suggestion
the form `RET = LHS + RHS`.
```

proposals/0035-linalg-matrix.md
```diff
@@ -1740,8 +1759,8 @@ Validation rules will enforce that:
 * Argument RHS is an `A` or `B` matrix
 * Argument LHS is an `Accumulator` matrix
```
```suggestion
* RHS is an `A` or `B` matrix
* RET and LHS are `Accumulator` matrices
```

proposals/0035-linalg-matrix.md
```diff
@@ -1740,8 +1759,8 @@ Validation rules will enforce that:
...
 * Type of LHS is the same as the return type
```
```suggestion
* Type of LHS and RET are the same
```

proposals/0035-linalg-matrix.md
```diff
@@ -1943,6 +1965,9 @@ Accumulation occurs in the type of the input vector. This operation must observe
 
 The base alignment of the vector must be at least 64-bytes.
 
+Validation rules will enforce that:
+* The resource handle must be an UAV RWByteAddressBuffer
```
V-FEXrt, line 1969: Not asking for a spec change: I'm assuming in cases like this where nothing is said "all matrix scopes" are acceptable?

V-FEXrt (approved):
Couple small suggestions that have been bouncing around in my head for a bit and a question

proposals/0035-linalg-matrix.md
llvm-beanz: Yes, that's the intent because there is nothing to validate.

V-FEXrt (approved):

V-FEXrt:
Instead of filing new issues for this I'm going to go update the existing validation issues with the new requirements

inbelic (approved):

PR #924, opened 2026-09-01
[0035] Relax alignment requirements on wave and threadgroup

### LINKED ISSUE #922: [0035] Clarifications on alignment requirements for Load/Store
[linalg](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

Would it be possible to expand on the reasoning behind the alignment requirements for Matrix::Load and Matrix::Store? The proposal says:
> For overloads operating on device memory (RWByteAddressBuffer), the address of the first element of the matrix (base address of the resource + the offset) must be 128-byte aligned. The Stride argument must refer to 16-byte aligned byte offsets.

My understanding was that without [0051](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0051-byte-address-buffer-alignment.md), ByteAddressBuffer base alignments cannot be guaranteed to be large enough.

Also, it reads like the alignment requirement is to the start HLSL matrix. 128-bytes would make it difficult to read consecutive tiles in many cases. As an example, if an implementation exposes a 16x16 f32 geometry, it would only be 64 bytes from the start of one tile to the next. I am misunderstanding this requirement?

I would appreciate some clarification around the exact requirements.

llvm-beanz:
This relaxes the alignment requirements on wave and threadgroup scope matrices to 16-byte alignment.

In theory we might be able to go as low as 4 in some cases, but I kinda suspect that 16 is small enough and will give better performance.

Resolves #922

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -1008,7 +1008,7 @@ the first element of the matrix (base address of the resource + the offset) must
 be 128-byte aligned for `Thread` scope matrices. For `Wave` and `ThreadGroup`
 scope matrices alignment must be at least 4, but will aim to default to a wider
 alignment for better performance. The `Stride` argument must be a multiple of 16
-bytes.
+bytes for `Thread` scope matrices and 4 bytes for `Wave` and `ThreadGroup`.
 
 For the `Load` operations on `groupshared` arrays:
   - an element is a type matching the element type of the `groupshared` array.
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -115,11 +115,11 @@ class Matrix {
       typename hlsl::enable_if<hlsl::is_arithmetic<T>::value, Matrix>::type
       Splat(T Val);
 
-  template <uint Align = 16>
+  template <uint Align = __detail::DefaultAlign<ComponentTy, M, N>::Value>
   static Matrix Load(ByteAddressBuffer Res, uint StartOffset, uint Stride,
                      MatrixLayoutEnum Layout);
 
-  template <uint Align = 16>
+  template <uint Align = __detail::DefaultAlign<ComponentTy, M, N>::Value>
   static Matrix Load(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
                      MatrixLayoutEnum Layout);
 
@@ -153,7 +153,7 @@ class Matrix {
                            void>::type
   Set(uint Index, ElementType Value);
 
-  template <uint Align = 16>
+  template <uint Align = __detail::DefaultAlign<ComponentTy, M, N>::Value>
   void Store(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
              MatrixLayoutEnum Layout);
 
@@ -168,7 +168,8 @@ class Matrix {
         MatrixLayoutEnum Layout);
 
   // Accumulate methods
-  template <uint Align = 16, MatrixUseEnum UseLocal = Use>
+  template <uint Align = __detail::DefaultAlign<ComponentTy, M, N>::Value,
+            MatrixUseEnum UseLocal = Use>
   typename hlsl::enable_if<Use == MatrixUse::Accumulator && UseLocal == Use,
                            void>::type
   InterlockedAccumulate(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
@@ -855,6 +856,15 @@ struct ScalarCountFromPackedComponents {
       (PackedComponentCount + ElementsPerScalar - 1) / ElementsPerScalar;
 };
 
+template <ComponentEnum ElementType, SIZE_TYPE M, SIZE_TYPE N>
+struct DefaultAlign {
+  static const SIZE_TYPE MinDim = M < N ? M : N;
+  static const uint ElementAlign =
+      ScalarCountFromPackedComponents<ElementType, MinDim>::Value;
+  static const uint MinElementAlign = ElementAlign < 4 ? 4 : ElementAlign;
+  static const uint Value = MinElementAlign < 16 ? MinElementAlign : 16;
+};
+
 } // namespace __detail
 ```
 
@@ -941,7 +951,7 @@ static Matrix Matrix::Load(ByteAddressBuffer Res, uint StartOffset, uint Stride,
                            MatrixLayoutEnum Layout);
 
 // Not available on Thread scope matrices.
-template <uint Align = 16>
+template <uint Align>
 static Matrix Matrix::Load(RWByteAddressBuffer Res, uint StartOffset,
                            uint Stride, MatrixLayoutEnum Layout);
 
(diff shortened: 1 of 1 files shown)


llvm-beanz:
Tagging some folks for review: @anupamachandra & @mjbedy.

proposals/0035-linalg-matrix.md
```diff
@@ -992,10 +992,11 @@ For the `Load` operations on `[RW]ByteAddressBuffers`:
     multiple of 16.
   - the `Offset` argument is the number of bytes to skip before loading.
 
-For overloads operating on device memory (`RWByteAddressBuffer`), the address of
+For overloads operating on device memory (`[RW]ByteAddressBuffer`), the address of
 the first element of the matrix (base address of the resource + the offset) must
-be 128-byte aligned. The `Stride` argument must refer to 16-byte aligned byte
-offsets.
+be 128-byte aligned for `Thread` scope matrices. For `Wave` and `ThreadGroup`
+scope matrices alignment must be at least 16. The `Stride` argument must
```
anupamachandra: Rewording Suggestion:  For `Wave` and `ThreadGroup`
scope matrices alignment must be at least 16 bytes. The `Stride` argument must be a multiple of 16 bytes.

proposals/0035-linalg-matrix.md
```diff
@@ -1118,7 +1119,7 @@ For the `Store` operations on `RWByteAddressBuffers`:
 
 For overloads operating on device memory (`RWByteAddressBuffer`), the address of
 the first element of the matrix (base address of the resource + the offset) must
-be 128-byte aligned. The `Stride` argument must refer to 16-byte aligned byte
+be 16-byte aligned. The `Stride` argument must refer to 16-byte aligned byte
```
anupamachandra: Suggestion: Here and in other place should where stride and offsets requirements are called out explicitly can we just use "must be a multiple of X bytes", just for uniformity.

anupamachandra:
Minor suggestions. Other that that, LGTM!

anupamachandra (approved):

mjbedy:
I'm checking into the alignments we might prefer on the Wave and ThreadGroup scope matrix. I had missed that they had previously inherited the 128 byte alignment of the Thread scope in the spec.

proposals/0035-linalg-matrix.md
```diff
@@ -936,12 +936,12 @@ Matrix::Splat(WaveReadLaneFirst(Val));
 #### Matrix::Load
 
 ```c++
-template <uint Align = 128>
+template <uint Align> // defaults 128 on thread-scope, 16 on other scopes.
```
V-FEXrt, line 949: how is this actually enforced/defaulted? Can we have two overloads enabled/disabled on scope?

llvm-beanz: Because the thread-scope uses a specialization, we put the alignment as 128 on the overload in the thread-scope specialization and 16 everywhere else.

V-FEXrt: I see, we are already doing that split overload and I just missed it

V-FEXrt (approved):

jrprice:
I'm concerned that 16-bytes will still be too strict for devices that support 8-bit element types with matrices where one dimension is <16.

For example, I have an NVIDIA GPU that supports subgroup-scope cooperative matrices where `MNK = 16x8x32` with an 8-bit integer element type via the Vulkan driver. I haven't yet tested via the D3D12 driver, but I would expect to see the same support for WaveScope with LinAlg. Loading the RHS input matrix for this geometry would not be able to satisfy the 16-byte requirement when the layout is row-major.

Is there a reason that the alignment/stride requirement couldn't be tied to the actual byte-size of a row/col of the matrix, instead of a hardcoded number like 16? Or something like `min(row_byte_size, 16)` to allow for these sub-16 byte alignment scenarios?

llvm-beanz:
> For example, I have an NVIDIA GPU that supports subgroup-scope cooperative matrices where `MNK = 16x8x32` with an 8-bit integer element type... Loading the RHS input matrix for this geometry would not be able to satisfy the 16-byte requirement when the layout is row-major.

This is not strictly true, this would require padding by setting the row stride, which the API does support.
 
> Is there a reason that the alignment/stride requirement couldn't be tied to the actual byte-size of a row/col of the matrix, instead of a hardcoded number like 16? Or something like `min(row_byte_size, 16)` to allow for these sub-16 byte alignment scenarios?

We've been discussing this tradeoff. We can probably support smaller alignment for 8-bit (and potentially 16-bit) types, but there is an associated performance degradation. A 16x8x32 matrix padded to the shape of a 16x16x32 matrix may seem wasteful, but the performance benefit of loading it with vectorized loads would be potentially large.

The question that came up in discussion that we didn't really have a good answer for is _how common_ are matrices (or tiles) that small really in practice. Is it worth optimizing around or not?

I don't know that I have a firm understanding of the answers to that question.

proposals/0035-linalg-matrix.md
```diff
@@ -992,10 +1002,12 @@ For the `Load` operations on `[RW]ByteAddressBuffers`:
...
 
-For overloads operating on device memory (`RWByteAddressBuffer`), the address of
+For overloads operating on device memory (`[RW]ByteAddressBuffer`), the address of
 the first element of the matrix (base address of the resource + the offset) must
-be 128-byte aligned. The `Stride` argument must refer to 16-byte aligned byte
-offsets.
+be 128-byte aligned for `Thread` scope matrices. For `Wave` and `ThreadGroup`
+scope matrices alignment must be at least 4, but will aim to default to a wider
+alignment for better performance. The `Stride` argument must be a multiple of 16
+bytes.
```
mjbedy: Is there a motivation for Stride having a larger alignment constraint for Wave and ThreadGroup scope than the base alignment, or should it get the same alignment value as the base for these cases?

proposals/0035-linalg-matrix.md
```diff
@@ -992,10 +1002,12 @@ For the `Load` operations on `[RW]ByteAddressBuffers`:
...
   - the `Offset` argument is the number of bytes to skip before loading.
 
-For overloads operating on device memory (`RWByteAddressBuffer`), the address of
+For overloads operating on device memory (`[RW]ByteAddressBuffer`), the address of
 the first element of the matrix (base address of the resource + the offset) must
-be 128-byte aligned. The `Stride` argument must refer to 16-byte aligned byte
-offsets.
+be 128-byte aligned for `Thread` scope matrices. For `Wave` and `ThreadGroup`
+scope matrices alignment must be at least 4, but will aim to default to a wider
+alignment for better performance. The `Stride` argument must be a multiple of 16
```
mjbedy, line 1010: It might help clarity to be specific here; default alignment value will be min(4, max(16, bytes_in_row/col)) or something like that.

mjbedy (commented):
We've also been debating which alignment constraints make sense. We would want at least 4 byte alignment for [RW]ByteAddressBuffer accesses. 

From the header/language level, as I understand the template math it is implementing max(4, min(16, row/col_bytes)) is that correct? The effect of that is to default the value in the alignment field in the DXIL to something other than 4.

llvm-beanz:
> We've also been debating which alignment constraints make sense. We would want at least 4 byte alignment for [RW]ByteAddressBuffer accesses.
> 
> From the header/language level, as I understand the template math it is implementing max(4, min(16, row/col_bytes)) is that correct? The effect of that is to default the value in the alignment field in the DXIL to something other than 4.

Yea, the idea is to try and choose 16 if it makes sense, but fall back to 4 (or 8) if 16 is too restrictive.

In terms of the stride. I think I was too focused on the alignment and not thinking through the stride. The stride requirement should match the alignment. I'll update that.

proposals/0035-linalg-matrix.md
V-FEXrt: +1 everywhere else stride seems to have been lowered to 4 but its still 16 here? was that intentional?

V-FEXrt (approved):

V-FEXrt (approved):

PR #927, opened 2026-09-03
[0035] Fix accumulation to groupshared

### LINKED ISSUE #925: [0035] Discrepancy in Interlockedaccumulate to Groupshared : (1) Clarify Type requirement for MatrixTy and GSMem type  (2) MatrixScope (Wave + TG?)
[LinAlg Matrix](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0035-linalg-matrix.md)

In the spec, interlockedaccumulate comes in 2 versions:
From the header:
```
template <typename T, MatrixUseEnum UseLocal = Use,
            MatrixScopeEnum ScopeLocal = Scope, SIZE_TYPE Size>
  typename hlsl::enable_if<
      hlsl::is_arithmetic_vector<T>::value && Use == MatrixUse::Accumulator &&
          UseLocal == Use && Scope == MatrixScope::Wave && ScopeLocal == Scope,
      void>::type
  InterlockedAccumulate(groupshared T Arr[Size], uint StartIdx, uint Stride,
                        MatrixLayoutEnum Layout);

#ifdef __hlsl_dx_compiler
  template <ComponentEnum TargetCompTy = ComponentTy, typename T,
            MatrixUseEnum UseLocal = Use, MatrixScopeEnum ScopeLocal = Scope,
            SIZE_TYPE Size>
  typename hlsl::enable_if<
      hlsl::is_same<typename hlsl::strip_vector_type<T>::type,
                    uint8_t4_packed>::value &&
          Use == MatrixUse::Accumulator && UseLocal == Use &&
          Scope == MatrixScope::Wave && ScopeLocal == Scope,
      void>::type
  InterlockedAccumulate(groupshared T Arr[Size], uint StartIdx, uint Stride,
                        MatrixLayoutEnum Layout);
```
and the requirement for the fist overload is that when MatrixTy if different from the shared memort type, a conversion is inserted, though it is not clear where the conversion should be performed (DXC or the driver implementation). This can be simplified to have semantics similar to to load/stores where the memory and matrix component types match or the memory be declared as an integral/packed type and a reinterpretation type be passed explicitly.

Tangetially, the spec in incosistent in the MatrixScope allowed for this operation, in the header it is marked as Wave scope only adn elsewhere it is Wave and Threadgroup scope.

llvm-beanz:
This corrects some errors I recently introduced to the spec and removes some undesirable complexitly.

When accumulating to groupshared memory we should effectively always treat the groupshared memory as if it is a blob of bytes of the matrix element type, and no data conversions are applied.

Fixes #925

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -1901,8 +1901,8 @@ match the scalar type or vector component type of the groupshared array. If the
 groupshared memory is `i32` or vector of _N_ by `i32` (`< N x i32 >`) the matrix
 may be of any type. Validation will verify this requirement.
 
-Matrix accumullation occurs in the data type of the matrix element with data in
-groupshared memory being treated as-if it were bitcast to the element type.
+Matrix accumulation occurs in the data type of the matrix element with data in
+groupshared memory being treated as if it were bitcast to the element type.
 
 For the Accumulate operation on groupshared arrays the `Offset` and `Stride`
 parameters are the number of scalar elements of the scalar element type of the
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -174,24 +174,21 @@ class Matrix {
   InterlockedAccumulate(RWByteAddressBuffer Res, uint StartOffset, uint Stride,
                         MatrixLayoutEnum Layout);
 
-  template <typename T, MatrixUseEnum UseLocal = Use,
-            MatrixScopeEnum ScopeLocal = Scope, SIZE_TYPE Size>
+  template <typename T, MatrixUseEnum UseLocal = Use, SIZE_TYPE Size>
   typename hlsl::enable_if<
       hlsl::is_arithmetic_vector<T>::value && Use == MatrixUse::Accumulator &&
-          UseLocal == Use && Scope == MatrixScope::Wave && ScopeLocal == Scope,
+          UseLocal == Use,
       void>::type
   InterlockedAccumulate(groupshared T Arr[Size], uint StartIdx, uint Stride,
                         MatrixLayoutEnum Layout);
 
 #ifdef __hlsl_dx_compiler
   template <ComponentEnum TargetCompTy = ComponentTy, typename T,
-            MatrixUseEnum UseLocal = Use, MatrixScopeEnum ScopeLocal = Scope,
-            SIZE_TYPE Size>
+            MatrixUseEnum UseLocal = Use, SIZE_TYPE Size>
   typename hlsl::enable_if<
       hlsl::is_same<typename hlsl::strip_vector_type<T>::type,
                     uint8_t4_packed>::value &&
-          Use == MatrixUse::Accumulator && UseLocal == Use &&
-          Scope == MatrixScope::Wave && ScopeLocal == Scope,
+          Use == MatrixUse::Accumulator && UseLocal == Use
       void>::type
   InterlockedAccumulate(groupshared T Arr[Size], uint StartIdx, uint Stride,
                         MatrixLayoutEnum Layout);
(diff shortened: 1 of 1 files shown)


proposals/0035-linalg-matrix.md
V-FEXrt: Validation rules need to be updated

proposals/0035-linalg-matrix.md
```diff
@@ -1920,6 +1901,9 @@ match the scalar type or vector component type of the groupshared array. If the
 groupshared memory is `i32` or vector of _N_ by `i32` (`< N x i32 >`) the matrix
 may be of any type. Validation will verify this requirement.
 
+Matrix accumullation occurs in the data type of the matrix element with data in
```
damyanp:
```suggestion
Matrix accumulation occurs in the data type of the matrix element with data in
```

proposals/0035-linalg-matrix.md
V-FEXrt, line 1925: > only available for
matrices with `MatrixUse::Accumulator` and `Wave` or `ThreadGroup` scope.

Is this still true now?

proposals/0035-linalg-matrix.md
```diff
@@ -1920,6 +1901,9 @@ match the scalar type or vector component type of the groupshared array. If the
...
+groupshared memory being treated as-if it were bitcast to the element type.
```
damyanp:
```suggestion
groupshared memory being treated as if it were bitcast to the element type.
```

damyanp (approved):

proposals/0035-linalg-matrix.md
llvm-beanz: Yes, did I screw that up somewhere? It should still only be usable with Wave and ThreadGroup accumulator matrices.

V-FEXrt: Does line 192ish not remove the scope restriction for accumulate?

llvm-beanz: It removes the scope restriction that was locking to just wave scope. The overload is in the specialization for Wave and ThreadGroup, so it still excludes Thread, the template restriction wasn't necessary for that.

```suggestion
`groupshared` overload is only available for `Wave` and `ThreadGroup` scope matrices.
```

V-FEXrt: fyi that suggestion looks wrong so I don't think you can directly accept it?

```suggestion
Offset * element byte width must be 128-byte aligned from the base offset, and stride * element byte width
```

V-FEXrt:
Sorry for drip feeding comments, I wanted to do a deeper read and kept finding small things. That should be everything!

proposals/0035-linalg-matrix.md
```diff
@@ -174,24 +174,21 @@ class Matrix {
...
   typename hlsl::enable_if<
       hlsl::is_arithmetic_vector<T>::value && Use == MatrixUse::Accumulator &&
-          UseLocal == Use && Scope == MatrixScope::Wave && ScopeLocal == Scope,
+          UseLocal == Use,
       void>::type
   InterlockedAccumulate(groupshared T Arr[Size], uint StartIdx, uint Stride,
                         MatrixLayoutEnum Layout);
 
 #ifdef __hlsl_dx_compiler
   template <ComponentEnum TargetCompTy = ComponentTy, typename T,
```
V-FEXrt: Comment from copilot on another PR that belongs over here!

> `TargetCompTy` is now accepted but never used, so calls such as InterlockedAccumulate<ComponentType::I32>(...) compile while producing exactly the same F32 operation as the default call. Since the DXIL target-type operand was removed, keeping this argument silently misrepresents the operation's element interpretation; remove the template parameter and update the affected call sites/tests so obsolete explicit target types are rejected rather than ignored.

llvm-beanz: I'll remove that. Thanks for the catch!

proposals/0035-linalg-matrix.md
llvm-beanz: This sentence got rewritten in PR #924.

V-FEXrt (approved):

PR #929, opened 2026-09-10
[SM6.10] Fix Default Align Value Calculation

V-FEXrt:
The default alignment had a small error in how the alignment was calculated. The count of scalar elements was being used directly instead of being scaled to total number of bytes

### DIFF
--- a/proposals/0035-linalg-matrix.md
+++ b/proposals/0035-linalg-matrix.md
@@ -862,11 +862,13 @@ struct ScalarCountFromPackedComponents {
 
 template <ComponentEnum ElementType, SIZE_TYPE M, SIZE_TYPE N>
 struct DefaultAlign {
-  static const SIZE_TYPE MinDim = M < N ? M : N;
-  static const uint ElementAlign =
-      ScalarCountFromPackedComponents<ElementType, MinDim>::Value;
-  static const uint MinElementAlign = ElementAlign < 4 ? 4 : ElementAlign;
-  static const uint Value = MinElementAlign < 16 ? MinElementAlign : 16;
+  enum {
+    MinDim = M < N ? M : N,
+    ScalarCount = ScalarCountFromPackedComponents<ElementType, MinDim>::Value,
+    ByteAlign = ScalarCount * 4,
+    MinByteAlign = ByteAlign < 4 ? 4 : ByteAlign,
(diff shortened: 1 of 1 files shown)


llvm-beanz (approved):

proposals/0035-linalg-matrix.md
```diff
@@ -862,11 +862,13 @@ struct ScalarCountFromPackedComponents {
 
 template <ComponentEnum ElementType, SIZE_TYPE M, SIZE_TYPE N>
 struct DefaultAlign {
-  static const SIZE_TYPE MinDim = M < N ? M : N;
-  static const uint ElementAlign =
-      ScalarCountFromPackedComponents<ElementType, MinDim>::Value;
-  static const uint MinElementAlign = ElementAlign < 4 ? 4 : ElementAlign;
-  static const uint Value = MinElementAlign < 16 ? MinElementAlign : 16;
+  enum {
```
bogner, line 865: I guess the switch to `enum` from `static const` variables is meant as a pre-constexpr hack? I'd kind of prefer `static const` so that it's trivial to update these to `static constexpr` if/when HLSL learns the keyword, and given how these will be used I don't think it matters (See https://godbolt.org/z/vcsMf6dr1). Since it's basically just a style point I'll let you make the call though.

bogner (approved):

proposals/0035-linalg-matrix.md
V-FEXrt: I personally don't have a strong opinion on this, and my gut feeling is that @llvm-beanz doesn't either.

Assuming that's the case then I can switch it back here but I'm going to (briefly) punt on switching it back in the implementation PR because I won't have time for CI to complete before I have to sign off and I'd really like to get `main`'s implementation back in line with the spec ASAP

V-FEXrt:
Ope, I clicked merge on the wrong PR! I'll put up another PR in a couple mins to revert the `enum` bit

V-FEXrt:
Testing locally and doing the enum change is (surprisingly) a bigger lift. I have to run for now so I'll have to leave the summary on why until later but for now I'm going to keep the `enum` implementation


Edit: Slightly longer details. DXC falls over without the enum when doing the template instantiation so we need to keep it

